PR #2, opened 2018-12-08
Quirks for CentraLite 3300-S

roblandry:


zhaquirks/centralite/3300S.py
```diff
@@ -0,0 +1,88 @@
...
+                DIAGNOSTICS_CLUSTER_ID
+            ],
+            'output_clusters': [
+                Identify.cluster_id
+            ],
+        },
+    }
+
+    replacement = {
+        'manufacturer': 'CentraLite',
```
dmulcahey: I had originally defined these incorrectly. manufacturer and model have to be defined on each endpoint.

dmulcahey:
Thanks for the contribution!

Issue #19, opened 2018-12-20
Conflict with Centralite 3325-s and 3326-l


dmulcahey:
Both devices have the same exact signature. In this case everything works but we need to handle this better before we have devices that match but function differently.

PR #20, opened 2018-12-26
Quirks for Smartthings MotionV4

roblandry:


roblandry:
It actually looks like this matches the signatures for the CentraLite 3325-S and 3326-L. Being a different manufacturer im not sure if this will conflict like #19

zhaquirks/smartthings/motionv4.py
```diff
@@ -0,0 +1,57 @@
...
+    PollControl, Ota, BinaryInput
+from zigpy.zcl.clusters.security import IasZone
+from zhaquirks.centralite import PowerConfigurationCluster
+from zigpy.zcl.clusters.measurement import TemperatureMeasurement
+from zigpy.quirks import CustomDevice
+
+
+class SmartThingsMotionV4(CustomDevice):
+
+    def __init__(self, *args, **kwargs):
```
dmulcahey: unneeded can remove the entire function

zhaquirks/smartthings/motionv4.py
```diff
@@ -0,0 +1,57 @@
...
+        super().__init__(*args, **kwargs)
+
+    signature = {
+        #  <SimpleDescriptor endpoint=1 profile=260 device_type=1026
+        #  device_version=0
+        #  input_clusters=[0, 1, 3, 15, 1026, 1280, 32]
+        #  output_clusters=[25]>
+        1: {
+            'profile_id': zha.PROFILE_ID,
```
dmulcahey: we match on man / model now so add them here

zhaquirks/smartthings/motionv4.py
roblandry: Done

zhaquirks/smartthings/motionv4.py
roblandry: Done

PR #21, opened 2018-12-26
Quirks for Smartthings MultiV4

roblandry:


roblandry:
I am assuming this one will be left open until the accelerometer cluster can be properly implemented, but sending this one also due to battery reports working.

zhaquirks/smartthings/multiv4.py
```diff
@@ -0,0 +1,59 @@
...
+    PollControl, Ota, BinaryInput
+from zigpy.zcl.clusters.security import IasZone
+from zhaquirks.centralite import PowerConfigurationCluster
+from zigpy.zcl.clusters.measurement import TemperatureMeasurement
+from zigpy.quirks import CustomDevice
+
+
+class SmartThingsMultiV4(CustomDevice):
+
+    def __init__(self, *args, **kwargs):
```
dmulcahey: unneeded can remove the entire function

zhaquirks/smartthings/multiv4.py
```diff
@@ -0,0 +1,59 @@
...
+        super().__init__(*args, **kwargs)
+
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=1026
+        # device_version=0
+        # input_clusters=[0, 1, 3, 15, 32, 1026, 1280, 64514]
+        # output_clusters=[25]>
+        1: {
```
dmulcahey: we match on man / model now so add them here

zhaquirks/smartthings/__init__.py
```diff
@@ -0,0 +1,19 @@
+import zigpy.types as t
```
dmulcahey: drop this file from the PR and import this from zigpy quirks

dmulcahey: scratch that... move it to Centralite and we have to change it too... https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/b2c942bd2220e7f5367465c49b8460f9edf70763/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy#L236

zhaquirks/smartthings/multiv4.py
roblandry: Done

zhaquirks/smartthings/multiv4.py
roblandry: Done

PR #30, opened 2019-02-05
allow "flip" actions to provide activated_face

dlashua:


dmulcahey:
Not at a machine to test this at the moment... do the face #’s correspond to the existing ones on the other 2 actions that contain activated face?

dlashua:
Yes, the face numbers match with the face numbers presented by "knock" and "slide". 

Reading the data field directly, the faces are numbered 0-5. The "+1" on the end of that line makes it line up with the numbering you're already using.

There's additional data available when flip_degrees=90. Namely, which side it was flipped FROM (in addition to which side it was flipped TO). However, I didn't find this information to be valuable since, with this in place, I'd already have a means of knowing the previous side. However, I could certainly add code for it if you find it useful.

The sensor doesn't provide this extra data when flip_degress=180, presumably because the previous side will always be the same based on the new side (if it is 3 now it was 6 before, if it is 2 now it was 5 before, etc, etc). So, the previous side on a 180 degree flip could also be presented if desired.

dmulcahey:
Completely up to you. Thanks for the contribution!

PR #42, opened 2019-03-11
Add manufacturer/model to centralite signatures

Socalix:
CentraLite devices share the same profile/input/output signature, which makes HA pick the wrong quirk to use.
Since HA v0.89 I also noticed that the device name for my CentraLite 3320-L has change to 3315-S due to wrong quirk.

With this change, the quirk for 3315 is skipped and 3320 is picked up as expected:
```
2019-03-10 22:11:19 DEBUG (MainThread) [zigpy.quirks] Considering <class 'zhaquirks.centralite.3315S.CentraLite3315S'>
2019-03-10 22:11:19 DEBUG (MainThread) [zigpy.quirks] Fail because model mismatch on at least one endpoint
2019-03-10 22:11:19 DEBUG (MainThread) [zigpy.quirks] Considering <class 'zhaquirks.centralite.3320.CentraLite3320'>
2019-03-10 22:11:19 DEBUG (MainThread) [zigpy.quirks] Found custom device replacement for 00:0d:6f:00:11:22:33:44: <class 'zhaquirks.centralite.3320.CentraLite3320'>
```

### DIFF
--- a/zhaquirks/centralite/3130.py
+++ b/zhaquirks/centralite/3130.py
@@ -19,8 +19,6 @@ class CentraLite3130(CustomDevice):
         #  input_clusters=[0, 1, 3, 32, 1026, 2821]
         #  output_clusters=[3, 6, 8, 25]>
         1: {
-            'manufacturer': 'CentraLite',
-            'model': '3130',
             'profile_id': zha.PROFILE_ID,
             'device_type': zha.DeviceType.LEVEL_CONTROL_SWITCH,
             'input_clusters': [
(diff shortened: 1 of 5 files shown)


dmulcahey:
I have to look at this a bit. If the devices are exactly the same the solution is to remove 1 quirk all together and to remove the manufacturer and model completely from the remaining one. I’ll peek at this tomorrow.

Socalix:
Thanks for explaining! I didn't realize that. In that case, only 3320 and 3315 share the same signature for both endpoint 1 and 2. The quirk files are identical except for the manufacturer/model, so maybe it's OK to combine into one.

I updated the PR to have only these 2 files now. (Earlier I went ahead and added model to all others that didn't have it thinking it'll potentially help with other devices.)

dmulcahey:
used #47 for this instead. Thanks for the contribution though!!!

PR #53, opened 2019-03-23
Initial support for LUMI lumi.plug

roblandry:
Possibly needs cleanup. 
Currently working with power reports on endpoints 2 and 3.
Removed Manufacturer and model on signature matches due to "unable to match on at least one endpoint"?


dmulcahey:
closed in favor of #434

PR #55, opened 2019-03-27
Add 2nd form of the Samjin Button

dmulcahey:
This will need to be cleaned up when we have a better way to support multiple signatures.


ryanwinter:
I was expecting you to just put the two signatures in the same file and share the button pressing code between.

One question I did have here, is that support of the double press and hold. Is this a common function across different zha devices? Should there be a standard definition to hold all these event types rather than being redefined for each device?

PR #82, opened 2019-05-17
Add support for xbee3 and xbee binary sensors and switches

prairiesnpr:
This adds support for using the xbee3 and xbee (S2/S2C) as binary inputs or switches. To use the xbee as a switch, the coordinator must be an xbee since it relies on the remote_at_command in zigpy_xbee.
I have not tested with a non-xbee coordinator.

### DIFF
--- a/zhaquirks/xbee/__init__.py
+++ b/zhaquirks/xbee/__init__.py
@@ -1,0 +1,1 @@
+"""Module for xbee devices as remote sensors/switches."""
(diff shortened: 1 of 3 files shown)


Shulyaka:
Do both xbee and xbee3 have identical signatures?

prairiesnpr:
No, the signatures are different. The device type varies between them.

Shulyaka:
Yes, you are right. Sorry for not noticing it myself

PR #91, opened 2019-06-07
Add support for innr RB 285 C

jurriaan:
So the color is adjustable and the light is dimmable.

And add innr devices to the README

### DIFF
--- a/zhaquirks/innr/rb285c.py
+++ b/zhaquirks/innr/rb285c.py
@@ -1,0 +1,82 @@
+"""Innr RB 285 C device."""
+from zigpy.profiles import zha
+from zigpy.profiles.zha import DeviceType
+from zigpy.quirks import CustomDevice
+from zigpy.zcl.clusters.general import (
+    Basic, GreenPowerProxy, Groups, Identify, LevelControl, OnOff, Ota, Scenes)
+from zigpy.zcl.clusters.lighting import Color
+from zigpy.zcl.clusters.lightlink import LightLink
+
+
+class RB285C(CustomDevice):
+    """Innr RB 285 C device."""
+
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=269
+        # device_version=1
+        # input_clusters=[0, 3, 4, 5, 6, 8, 768, 4096]
+        # output_clusters=[25]>
+        1: {
+            'manufacturer': 'innr',
+            'model': 'RB 285 C',
+            'profile_id': zha.PROFILE_ID,
+            'device_type': 269,
+            'input_clusters': [
+                Basic.cluster_id,
+                Identify.cluster_id,
+                Groups.cluster_id,
+                Scenes.cluster_id,
+                OnOff.cluster_id,
+                LevelControl.cluster_id,
+                Color.cluster_id,
+                LightLink.cluster_id
+            ],
+            'output_clusters': [
+                Ota.cluster_id
+            ],
+        },
+        # <SimpleDescriptor endpoint=242 profile=41440 device_type=97
+        # device_version=0
+        # input_clusters=[]
+        # output_clusters=[33]>
+        242: {
+            'profile_id': 41440,
+            'device_type': 97,
+            'input_clusters': [
+            ],
+            'output_clusters': [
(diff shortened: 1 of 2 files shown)


zhaquirks/innr/rb285c.py
```diff
@@ -0,0 +1,82 @@
...
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=269
+        # device_version=1
+        # input_clusters=[0, 3, 4, 5, 6, 8, 768, 4096]
+        # output_clusters=[25]>
+        1: {
+            'manufacturer': 'innr',
+            'model': 'RB 285 C',
+            'profile_id': zha.PROFILE_ID,
+            'device_type': 269,
```
dmulcahey: Please use a constant for this.

zhaquirks/innr/rb285c.py
```diff
@@ -0,0 +1,82 @@
...
+                Ota.cluster_id
+            ],
+        },
+        # <SimpleDescriptor endpoint=242 profile=41440 device_type=97
+        # device_version=0
+        # input_clusters=[]
+        # output_clusters=[33]>
+        242: {
+            'profile_id': 41440,
+            'device_type': 97,
```
dmulcahey: ditto

zhaquirks/innr/rb285c.py
```diff
@@ -0,0 +1,82 @@
...
+            'output_clusters': [
+                Ota.cluster_id
+            ],
+        },
+        # <SimpleDescriptor endpoint=242 profile=41440 device_type=97
+        # device_version=0
+        # input_clusters=[]
+        # output_clusters=[33]>
+        242: {
+            'profile_id': 41440,
```
dmulcahey: ditto

zhaquirks/innr/rb285c.py
```diff
@@ -0,0 +1,82 @@
...
+                    LevelControl.cluster_id,
+                    Color.cluster_id,
+                    LightLink.cluster_id
+                ],
+                'output_clusters': [
+                    Ota.cluster_id
+                ],
+            },
+            242: {
+                'profile_id': 41440,
```
dmulcahey: ditto

zhaquirks/innr/rb285c.py
```diff
@@ -0,0 +1,82 @@
...
+                'device_type': 97,
```
dmulcahey: ditto

zhaquirks/innr/rb285c.py
jurriaan: Sorry, should've read the contributer documentation about this, just copied the existing innr quirk. 

I found out that this device type is an 'Extended Colour Light', so probably we need to add support to home assistant/zigpy for these types so we don't need quirks. What do you think?

dmulcahey: The device types are defined in Zigpy here: https://github.com/zigpy/zigpy/blob/a607f6284cd9dffbe7030ba265f9c7d3a02d95d1/zigpy/profiles/zha.py#L7 and here: https://github.com/zigpy/zigpy/blob/a607f6284cd9dffbe7030ba265f9c7d3a02d95d1/zigpy/profiles/zll.py#L7

and then that info is linked to domains in ZHA/HA here: 
https://github.com/home-assistant/home-assistant/blob/b71baef7c8effa036c17e4955bdf829580b1401a/homeassistant/components/zha/core/registries.py#L112

so you could PR the changes there instead and avoid the quirks all together.

jurriaan:
Thanks for the pointers on how to fix this @dmulcahey!

https://github.com/zigpy/zigpy/pull/176 has been merged, and I just opened https://github.com/home-assistant/home-assistant/pull/24429. 

When both are merged I think the quirks for the innr RS 228 T can be removed from this repository, as that bulb also does device type reassignment, but then for the newly added `zha.DeviceType.COLOR_TEMPERATURE_LIGHT`.

dmulcahey:
sounds good. Thanks for the contributions!!!

jurriaan:
:) Will close this PR since this doesn't make sense anymore.

PR #97, opened 2019-06-16
Misc. updates

dmulcahey:
This PR does the following:
- initial support for Ikea 5 button remote
- initial support for Ikea motion sensor
- add mfg attr to hue remote
- add initial handler for adurolight remote
- fix aqara vibration sensor after lock support added to ZHA
- remove model and manufacturer from all replacement objects
- clean up SmartThings tagV4 device
- misc. cleanup


dmulcahey:
> Deconz doesn't have `zigpy` device for coordinator yet. Wrap it in `try`/`except KeyError` and just ignore it on failure?

what will it take to add it? not supporting devices on certain sticks seems like an easy way to make users riot :)

dmulcahey:
@Adminiuga can you see if you can read the custom scene commands from the Ikea 5 button remote? I have tried a few times and I'm not making any progress.

Adminiuga:
IIRC schema on Scenes clusters needs an overhaul. Could you read it from any other device?

dmulcahey:
These are custom commands for this device only

PR #99, opened 2019-06-22
Move and convert manufacturer and model to tuples

dmulcahey:
This PR:
- adds endpoints key to signature and moves endpoints to it
- moves model and manufacturer from endpoint to signature and converts to tuples


Adminiuga:
can you add `Centralite 3320-L` window/door open/close sensor?
Here's the sig
```
<SimpleDescriptor endpoint=1 profile=260 device_type=1026 device_version=0 input_clusters=[0, 1, 3, 32, 1026, 1280, 2821] output_clusters=[25]>
Manufacturer: CentraLite
Model: 3320-L
<SimpleDescriptor endpoint=2 profile=49887 device_type=12 device_version=0 input_clusters=[0, 1, 3, 2821, 64527] output_clusters=[3]>
```

Before it was matched as 3315S, with tuples it is not matched any longer.

dmulcahey:
Will do

PR #112, opened 2019-07-11
Return status from custom Ikea bind method

### LINKED ISSUE #111: Ikea group support bind method doesn't return status as expected
https://github.com/dmulcahey/zha-device-handlers/blob/b5b383939944ff541ee38a94c7f4d6cf3edc611f/zhaquirks/ikea/__init__.py#L25

https://github.com/home-assistant/home-assistant/blob/a30c37017b7782473294d7999e85d7a369a0539a/homeassistant/components/zha/core/helpers.py#L56

reported by @Adminiuga 

we should return the status in [ ] so the bind helper in HA is happy.

dmulcahey:
fixes #111 

@Adminiuga better?


Adminiuga:
Yes.

Issue #109, opened 2019-07-08
Fix Keen and Ikea battery pct reporting


dmulcahey:
https://github.com/ebaauw/deconz-rest-plugin/blob/7a9b939efef6c35a3a093b0483972aa44aa2b724/de_web_plugin.cpp#L5130

See lines 5168-5170

PR #113, opened 2019-07-12
Correct non-standard battery pct remaining for Ikea and Keen Home

dmulcahey:
Fixes: #109 

@Adminiuga thoughts?


Adminiuga:
Looks fine by me. Just out of interest, what is the reported battery count for these devices? Just wondering if all devices with double batteries double bpr as well

dmulcahey:
remote has 1, motion has 2 IIRC and the vents have 4

PR #114, opened 2019-07-16
Fix xiaomi aqara to not report bogus temperature values

frog32:
Xiaomi Aqara Weather sensors sometimes report temperature of -100°C. This PR caps temperature reports at the documented temperature range for this sensor.

https://xiaomi-mi.com/sockets-and-sensors/aqara-temperature-and-humidity-sensor/

So far I could test that the sensor still works but I do not find a way to provoke this bogus report from one of my sensors.


dmulcahey:
Hey, I fixed the last lint issue and I moved the impl of the cluster up a level and added it to the Xiaomi Mija sensor too. If you're happy w/ the changes I think this is good to go. Thanks for the contribution!

frog32:
looks good to me

PR #117, opened 2019-07-20
Rename movionV4 to iasV4 and add moistureV4 model info

### LINKED ISSUE #116: Smartthings moisture sensor v1 not returning battery (0.96)
I'm sure this was working in the past, however now I'm seeing 'unknown' battery for this device since upgrading to 0.96.

When I query the 'battery_voltage' attribute, I get 26 which indicates 100%.

dmulcahey:
Fixes #116


zhaquirks/smartthings/iasv4.py
```diff
@@ -9,7 +9,7 @@
 from zhaquirks.centralite import PowerConfigurationCluster
 
 
-class SmartThingsMotionV4(CustomDevice):
+class SmartThingsIASV4(CustomDevice):
     """SmartThingsMotionV4."""
```
ryanwinter: Should be IASV4

PR #127, opened 2019-08-10
Add Centralite 3157100 thermostat quirks

presslab-us:
This thermostat does not report battery_percentage_remaining,
like most other Centralite devices.
Use the existing code to convert from battery_voltage.


presslab-us:
There is a pylint error `Module name "3157100" doesn't conform to snake_case naming style (invalid-name)`, but it's named similar to the other files in that folder.

dmulcahey:
Update this: https://github.com/dmulcahey/zha-device-handlers/blob/52f480ee595d69f4629c35333436bbf73c0a3170/pylintrc#L53

presslab-us:
I also found this which seems to work too.  Which would you prefer? https://github.com/dmulcahey/zha-device-handlers/blob/52f480ee595d69f4629c35333436bbf73c0a3170/zhaquirks/centralite/3321S.py#L2

dmulcahey:
Either one, I’ll eventually take a pass through all of this to clean this up.

dmulcahey:
Thanks!

PR #129, opened 2019-08-10
Modify Centralite 3305S quirk to support single endpoint devices

presslab-us:
My PEQ motion sensor (Centralite 3305) only has one endpoint,
so it doesn't match the existing signature.  Add another object
with only one endpoint so it will match.


presslab-us:
This fixes https://github.com/dmulcahey/zha-device-handlers/issues/128

dmulcahey:
Can you explicitly set models info for your device please.

dmulcahey:
Thanks!

PR #130, opened 2019-08-10
Add Centralite 3460-L smart button quirk

presslab-us:
This button does not report battery_percentage_remaining,
like most other Centralite devices.
Use the existing code to convert from battery_voltage.


dmulcahey:
Thanks!

PR #131, opened 2019-08-10
Add Yale Real Living lock quirk

presslab-us:
This lock reports battery_percentage_remaining as half of what it should.
Supports the YRD210 and YRL220 as these are the two locks I tested with;
they both use the same Zigbee hardware module.


dmulcahey:
Thanks!

ss4sgoku:
hi i m new in HA i want to add my yrd210 and yrl210 locks to it, what USB Adapter or  Pi shield you use to work with ZHA?

presslab-us:
See here for the radio modules supported:
https://www.home-assistant.io/components/zha/
Here is some more discussion:
https://community.home-assistant.io/t/zha-lock-i-have-a-yale/116992/12

PR #134, opened 2019-08-17
fix xiaomi aqara temperature sensor and also limit humidity

frog32:
Last commit I mixed up if statement and since I can not really force this behavior I did not catch it. Turns out the same issue happens with humidity where the sensor sometimes reports a relative humidity of 654.4.


frog32:
To finally test those changes I have manually backported them into my hassio production system (docker exect and then vi to change the actual files) and can confirm that it is having the desired effect. No more unrealistic values from a sensor that was lately giving me wrong readings multiple times a day.

dmulcahey:
Thanks!

PR #138, opened 2019-08-20
Add quirk for Osram BR RGBW and update Tunable White quirk

tube0013:
Adds new quirk, and updates a19twhite to add model so it doesn't match the BRRGBW bulb

Resolves https://github.com/dmulcahey/zha-device-handlers/issues/136


dmulcahey:
is this still necessary?

tube0013:
I think it would be good to differentiate the bulb models if this is the best way to go about it I'll leave up to you.

PR #143, opened 2019-08-31
Add secondary Waxman LeakSmart class without PollControl cluster for sensors that don't support it.

brg468:
This adds a secondary LeakSmart class in the quirk to support sensor models without the PollControl cluster.


brg468:
Tested to work with my LeakSmart sensors.

dmulcahey:
Thanks!

PR #147, opened 2019-09-08
Update xiaomi quirk with zigpy API changes

Adminiuga:
This is to match https://github.com/zigpy/zigpy/pull/212 change in zigpy.


Adminiuga:
@prairiesnpr can you take a look at this PR and `zigpy-homeassistant==0.8.0` ? just want to make sure with changes in zigpy we're not breaking XBee quirks.

PR #150, opened 2019-09-16
Add 2019 refresh of Samjin Multi

Thalagyrt:
Opening here instead of https://github.com/zigpy/zigpy/pull/228 per @Adminiuga's request.

I am not sure precisely what the `IasZone` cluster is used for here as I've not delved into Zigbee much. Seems like open/close perhaps? It seems to work properly without an additional handler in zigpy#228 above, so I figured there wasn't a need to refactor the `IASCluster` out of `button.py` as it seems that's used to make an open/close sensor appear as a button or some such? Might be worth a refactor there eventually as it seems that `IASCluster` is used for both buttons.

Does this all look correct? Here's the log output for this device refresh:

```
[0x00ea:1] Discovered endpoint information: <SimpleDescriptor endpoint=1 profile=260 device_type=1026 device_version=0 input_clusters=[0, 1, 3, 32, 1026, 1280, 2821, 64514] output_clusters=[3, 25]>
[0x00ea:1] Manufacturer: Samjin
[0x00ea:1] Model: multi
```

Seems it's just an added diagnostics cluster in the newer Samjin devices, looking at the button quirk. Note that there is another Samjin multi implementation in the `zigpy` repo in the file modified by the above linked PR for the 2018 version of this device that does not have the diagnostics cluster, perhaps it should be moved over here into `multi.py` which is currently empty other than a comment? I feel that should probably be a separate PR if so.

Cheers!
James

### DIFF
--- a/zhaquirks/samjin/multi2.py
+++ b/zhaquirks/samjin/multi2.py
@@ -1,0 +1,64 @@
+"""Samjin Multi 2019 Refresh Quirk."""
+from zigpy.quirks import CustomDevice
+from zigpy.quirks.smartthings import SmartThingsAccelCluster
+from zigpy.zcl.clusters.general import (
+    Basic, Identify, Ota, PollControl, PowerConfiguration)
+from zigpy.zcl.clusters.measurement import TemperatureMeasurement
+from zigpy.zcl.clusters.security import IasZone
+
+DIAGNOSTICS_CLUSTER_ID = 0x0B05  # decimal = 2821
+
+
+class SmartthingsMultiPurposeSensor2019(CustomDevice):
+    """Samjin multi device."""
+
+    signature = {
+        'models_info': [
+            ('Samjin', 'multi')
+        ],
+        'endpoints': {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=1026
+            # device_version=0 input_clusters=[0, 1, 3, 32, 1026, 1280,
+            # 2821, 64514]
+            # output_clusters=[3, 25]>
+            1: {
+                'profile_id': 0x0104,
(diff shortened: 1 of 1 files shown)


dmulcahey:
@Adminiuga in cases like this where there is a single additional cluster... do you think we should use this format for the additional classes: https://github.com/dmulcahey/zha-device-handlers/blob/ccaedc7e526eb231b9e11d4ca3687c1e05043928/zhaquirks/lutron/lzl4bwhl01remote.py#L87

Thalagyrt:
Sorry for the linter failures here, I don't touch Python much and don't know the usual style guides. :)

dmulcahey:
> Sorry for the linter failures here, I don't touch Python much and don't know the usual style guides. :)

no need to apologize!

Thalagyrt:
As an outsider, that suggestion seems way cleaner, but we'd probably want to port the existing Samjin multi quirk over to this repo for that, correct?

Adminiuga:
> in cases like this where there is a single additional cluster... do you think we should use this format for the additional classes

Hrm, we could. The downside of this approach is the replacement attribute which is less inclusive, eg in the case of multi sensor it would omit the Diagnostic cluster. ATM not using that cluster for anything, but it could provide some insight for troubleshooting.  In other words i'm open to ideas how to handle these slight variations better.

dmulcahey:
I'll play with some options soon and we can pick one. I'll merge this for now as is.

Thalagyrt:
Thanks David + Alexei! I've already got my branch running in my homeassistant install, so I'll keep an eye out for updates when this goes through the release process. If I see anything weird I'll let you know.

Issue #176, opened 2019-10-15
Sylvania/Osram Lightify Dimmer Switch battery always at 100%


rpatterson:
I have several Sylvania/Osram Lightify Dimmer Switches that I moved over from a SmartThings hub v2 to Home Assistant Docker v0.99.3 with a HUSBZB-1 USB stick.  Under ST, I could see battery percentages but under HA all their batteries show 100% all the time.  I've tried removing and repairing some of them but they still show the same 100% battery level after re-pairing.  It's recognized as using the `zhaquirks.centralite.3130.CentraLite3130` quirk.

Here's a log snipped from a pairing session if it helps:

```
2019-10-15 00:23:11 DEBUG (MainThread) [homeassistant.components.zha.api] Reconfiguring node with ieee_address: 84:18:26:00:00:e9:05:a4
2019-10-15 00:23:11 DEBUG (MainThread) [homeassistant.components.zha.core.device] [0xfc69](LIGHTIFY Dimming Switch): started configuration
2019-10-15 00:23:11 DEBUG (MainThread) [homeassistant.components.zha.core.channels] [0xfc69:ZDO](LIGHTIFY Dimming Switch): channel: 'async_configure' stage succeeded
2019-10-15 00:23:11 DEBUG (MainThread) [bellows.zigbee.application] Extending timeout for 84:18:26:00:00:e9:05:a4/0xfc69
2019-10-15 00:23:11 DEBUG (MainThread) [bellows.ezsp] Send command setExtendedTimeout: (84:18:26:00:00:e9:05:a4, True)
2019-10-15 00:23:11 DEBUG (MainThread) [bellows.zigbee.application] Extending timeout for 84:18:26:00:00:e9:05:a4/0xfc69
2019-10-15 00:23:11 DEBUG (MainThread) [bellows.ezsp] Send command setExtendedTimeout: (84:18:26:00:00:e9:05:a4, True)
2019-10-15 00:23:11 DEBUG (MainThread) [bellows.zigbee.application] Extending timeout for 84:18:26:00:00:e9:05:a4/0xfc69
2019-10-15 00:23:11 DEBUG (MainThread) [bellows.ezsp] Send command setExtendedTimeout: (84:18:26:00:00:e9:05:a4, True)
2019-10-15 00:23:11 DEBUG (bellows.thread_0) [bellows.uart] Sending: b'772d21d6f02ffcb259b252a1abfbde7e'
[...]

PR #178, opened 2019-10-17
Fix Centralite/Lightify button/switch battery level (#176)

rpatterson:
See #176 for history and where the number used come from.

This updates the battery percentages in the UI for me locally if I use the ZHA configuration UI to get the `battery_voltage` from the individual device.  So far at least other devices haven't shown the update, I'm not sure how ZHA refreshes such information from individual devices normally.


rpatterson:
All the batteries did eventually update over time to show correct values.

PR #182, opened 2019-10-21
make aqara leak sensor be of moisture type

abmantis:


dmulcahey:
What about adding an init function and doing this there that way it isn’t updated every time an attr is updated?

abmantis:
I was inspired by: https://github.com/dmulcahey/zha-device-handlers/commit/d75709d82f8e95f5e7f8d7ed8be0cc054b495bc4

But I'm ok changing it.

dmulcahey:
I'll merge this for now since there are no documented standards for things like this.

PR #195, opened 2019-11-20
Created a Contributors.md file with list of contributors

Hedda:
Created a Contributors.md file as per https://github.com/firstcontributions/first-contributions/blob/master/README.md


dmulcahey:
We should probably add a way to automate the handling of this file

Issue #199, opened 2019-11-22
Smartthings battery levels reporting incorrectly


ryanwinter:
Samsung devices shares it's battery range with centralite. Since this change, my devices are reporting low levels, with a bunch now reporting 0% battery, and they are still working.

https://github.com/dmulcahey/zha-device-handlers/commit/6cba0ff725db1e5d66b577ddfb283d2f7fd73313

Do the Samsung devices need their own battery logic to have them report correctly?

dmulcahey:
I’m not surprised. I’m not sold on the new ranges myself... can you monitor them so we can track the min side voltage where they stop working? That’s what I’m doing myself.

Adminiuga:
I think for devices that we know are using CR2 or CR123 battery we could use a lower voltage threshold.

ryanwinter:
The max voltage is also incorrect, I replaced batteries in a few devices recently, and now they are reporting 30%. 

I just checked and these devices are reporting 24 volts, which would have reported 90% on the old scale which seems correct.

For min voltage, I just replaced most of my devices as they had reached 1%. They were still working at that level (16 volts).

So basically I have seen devices working in the 16 - 26 volts range which seems to align well with the original range.

dmulcahey:
What kind of batteries? 100 is usually 3.0v for Lithium batteries at a minimum. Most start between 3.4 and 3.2ish. 2.4v seems really low for a new battery.

ryanwinter:
They are CR2032 so yes they are 3V. Maybe the reported voltage is not actual. I'll pull one down and measure it when I get a chance.

dmulcahey:
I have plenty of those in use and they generally start at 3.2 when new... I have 8 of the sylvania dimmers that use these

ryanwinter:
Referring to the smartthings community repo for this device, it uses the old scale:
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy

ryanwinter:
[...]

PR #204, opened 2019-11-30
[WIP] Revert centralite battery map, add new 3.1v map

roblandry:
To open discussion...

Reverts centralite battery map to last know working map. Uses [2.8-1.5v battery logic](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy#L294-L297) for centralite/smartthings.

Incorporates logic from: [Smartthings Zigbee Button](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/2ec2ca46f05e82a190d44890b9f7fe72dc4c3e85/devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy) for #176.

Reverts: 6cba0ff
Fixes: #199 
Potential fix: #176

### DIFF
--- a/zhaquirks/centralite/3130.py
+++ b/zhaquirks/centralite/3130.py
@@ -11,7 +11,7 @@ from zigpy.zcl.clusters.general import (
 )
 from zigpy.zcl.clusters.measurement import TemperatureMeasurement
 
-from zhaquirks.centralite import CENTRALITE, PowerConfigurationCluster
+from zhaquirks.centralite import CENTRALITE, PowerConfigurationCluster31
 from zhaquirks.const import (
     COMMAND,
     COMMAND_MOVE,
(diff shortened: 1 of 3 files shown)


Adminiuga:
IMO we just should have a linear volts to % function and each manufacturer/model just need to define the min/Max volts.

roblandry:
@Adminiuga i started the logic found in st for battery logic. #205 But, i would imagine there was a reason ST used the map for some devices... possibly the way they report voltage? I am not stuck on using a map, i honestly just wanted something to get my battery values back as the previous commit made like the bottom 50% of my batteries None. 
Both of these PRs are mostly about opening discussion and us getting to a solid fix.

dmulcahey:
> IMO we just should have a linear volts to % function and each manufacturer/model just need to define the min/Max volts.

I like this idea the more that I think about it.

roblandry:
If I can get some ideas how to pass values from the quirks to __init__ I'll get it done.

Adminiuga:
I just would define a child class overriding parent class attributes for
min/max

roblandry:
Closing in favor of #208

PR #206, opened 2019-12-01
Changed the way the args were being parsed to prevent duplicate trigg…

martinlong1978:
Changed the way the args were being parsed to prevent duplicate triggers.

Due to the way that the args are matched in HA, the current
implementation meant that the left and right button were each firing
triggers for both themself, and the other. So left would fire
automations triggered on left and right, and vice versa.

I've changed the "press" and "hold" command parsing to start with a
16 bit int, the same as release, this prevents the zeros being
generated for both, which was spooking the way that HA matches these.

### DIFF
--- a/zhaquirks/ikea/fivebtnremote.py
+++ b/zhaquirks/ikea/fivebtnremote.py
@@ -56,8 +56,8 @@ class ScenesCluster(EventableCluster, Scenes):
         super().__init__(*args, **kwargs)
         self.server_commands.update(
             {
-                0x0007: ("press", (t.int8s, t.int8s, t.int8s, t.int8s), False),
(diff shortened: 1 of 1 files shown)


dmulcahey:
@martinlong1978 can you log into travisci.org with your GH acct? Your PR won't build and it is flagged like this:

image

martinlong1978:
Thanks,

I've signed in, but my account is flagged for abuse. I've emailed them as
the error suggested, but guess I have to wait for business hours for them
to reset it.

Cheers

Martin

martinlong1978:
The abuse flag has cleared, however, I'm not sure how to get this PR to build.

PR #217, opened 2019-12-20
Move zigpy quirks to zhaquirks

prairiesnpr:
Remove refrences to zigpy quirks,
move zigpy quirks to zha quirks,
rework moved quirks to remove magic numbers

See https://github.com/zigpy/zigpy/pull/270


Adminiuga:
We should keep the tests for quirks were applicable, eg for KoF quirks.

dmulcahey:
I agree. We definitely have to move existing tests as well.

prairiesnpr:
> I agree. We definitely have to move existing tests as well.

Agreed, the KoF test is now included.

PR #221, opened 2019-12-22
Added additional attributes to Mija smoke detector.

prairiesnpr:


zhaquirks/xiaomi/mija/smoke.py
```diff
@@ -11,13 +24,31 @@
...
+    """Xiaomi smoke IAS cluster implementation."""
+
+    cluster_id = IasZone.cluster_id
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        super().__init__(*args, **kwargs)
+        self.attributes = super().attributes.copy()
+        self.attributes.update(
+            {0xFFF1: ("set_options", t.uint32_t), 0xFFF0: ("get_status", t.uint32_t)}
```
Adminiuga: I think if you update attributes that way, you won't get `self._attridx` dict updated. Which is done in https://github.com/zigpy/zigpy/blob/b3e08d0d82a685b39461ca3d8997c4aebf5d6b72/zigpy/zcl/__init__.py#L23

are we allowed to do:
```
class XiaomiSmokeIASCluster(CustomCluster, IasZone):
    """Xiaomi smoke IAS cluster implementation."""

    attributes = super().attributes.copy()
    attributes.update(....)
```
?

Adminiuga: after talking to David the
```
class XiaomiSmokeIASCluster(CustomCluster, IasZone):
    """Xiaomi smoke IAS cluster implementation."""

    attributes = IasZone.attributes.copy()
    attributes.update(....)
```
Seems to be a better approach

prairiesnpr: See if this is acceptable.

https://github.com/dmulcahey/zha-device-handlers/pull/224

PR #227, opened 2019-12-28
Fix LocalDataCluster

Shulyaka:
Fixed attribute read for LocalDataCluster. 
The zigpy.zcl.Cluster.read_attributes() function seems to require a more complex data structure in the response than just an array of values.

### DIFF
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -43,7 +43,7 @@ class LocalDataCluster(CustomCluster):
         ]
         for record in records:
             record.value.value = self._attr_cache.get(record.attrid)
-            if record.value.value:
+            if record.value.value is not None:
                 record.status = foundation.Status.SUCCESS
         return [records]
 
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -4,7 +4,6 @@ import importlib
 import pkgutil
 
 from zigpy.quirks import CustomCluster
-import zigpy.types as types
 from zigpy.util import ListenableMixin
 from zigpy.zcl import foundation
 from zigpy.zcl.clusters.general import PowerConfiguration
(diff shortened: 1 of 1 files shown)


dmulcahey:
```
lint run-test: commands[0] | flake8
237./zhaquirks/__init__.py:7:1: F401 'zigpy.types' imported but unused
```

zhaquirks/__init__.py
```diff
@@ -35,9 +36,17 @@ class LocalDataCluster(CustomCluster):
 
     async def read_attributes_raw(self, attributes, manufacturer=None):
         """Prevent remote reads."""
-        attributes = [types.uint16_t(a) for a in attributes]
-        values = [self._attr_cache.get(attr) for attr in attributes]
-        return values
+        records = [
+            foundation.ReadAttributeRecord(
+                attr, foundation.Status.UNSUPPORTED_ATTRIBUTE, foundation.TypeValue()
```
dmulcahey: Why are we doing this?

dmulcahey: Oh, are you establishing this as the default in case there is nothing in the cache?

Shulyaka: Yes, exactly.

Shulyaka:
> ```
> lint run-test: commands[0] | flake8
> 237./zhaquirks/__init__.py:7:1: F401 'zigpy.types' imported but unused
> ```

Will fix in an hour

zhaquirks/__init__.py
```diff
@@ -43,7 +43,7 @@ class LocalDataCluster(CustomCluster):
         ]
         for record in records:
             record.value.value = self._attr_cache.get(record.attrid)
-            if record.value.value:
+            if record.value.value is not None:
```
dmulcahey: good catch!

Shulyaka: Thanks... I wish I had noticed that before filing the PR :)

PR #229, opened 2019-12-29
LocalDataCluster: prevent remote attribute writes

Shulyaka:
If I understand the idea of the LocalDataCluster, we should as well prevent remote attribute writes


dmulcahey:
Thx. Unrelated, are you on discord?

Shulyaka:
> Thx. Unrelated, are you on discord?

Yes. I believe you can find me as Shulyaka.

PR #230, opened 2019-12-30
XBee PWM support

Shulyaka:
PWM output support for XBee3 by adding of an AnalogOutput cluster.

Important points:
- Requires zigpy-xbee>=0.8.0 to work
- HA support for AnalogOutput clusters is not ready yet, but one can still use the functionality of this PR with `zha.set_zigbee_cluster_attribute` service for `present_value` attribute. And here is a proof-of-concept code for the full HA support: home-assistant/home-assistant#30211

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ If you are looking to make your first code contribution to this project then we 
 - Some functionality requires a coordinator device to be XBee as well
 - GPIO pins are exposed to Home Assistant as switches
 - Analog inputs are exposed as sensors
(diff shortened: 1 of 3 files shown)


dmulcahey:
@prairiesnpr @Adminiuga deferring to you guys here for your thoughts.

Shulyaka:
I think it would be important to point out that this PR could be merged independently of home-assistant/home-assistant#30211. This PR implements the AnalogOutput cluster as per ZCL specification (except the attribute update report, for which we would need some kind of polling, and I don't know how to do it yet)

PR #241, opened 2020-01-15
XBee: Use LocalDataCluster where appropriate

### LINKED ISSUE #194: Slow startup with xbee
When an xbee or xbee3 device is added to HA, it causes startup time to significantly increase (by about 1 minute). Probably something is waiting for a timeout.
I plan to also look at it myself later.

Shulyaka:
Reduce the number of remote attribute read requests by using LocalDataCluster where appropriate.

Currently we do not handle remote attribute read requests for custom clusters and they are forwarded to the XBee device. Because the device is not aware of our custom clusters and endpoints, it does not respond on them and we have to wait for a timeout. This slows down the startup significantly, among other side effects.

The fix uses LocalDataCluster which uses cache to respond to the attribute read requests and prevents unwanted remote reads.

Closes #194

### DIFF
--- a/zhaquirks/xbee/__init__.py
+++ b/zhaquirks/xbee/__init__.py
@@ -19,7 +19,7 @@ the xbee stays alive in Home Assistant.
 import logging
 import struct
 
-from zigpy.quirks import CustomCluster, CustomDevice
+from zigpy.quirks import CustomDevice
 import zigpy.types as t
 from zigpy.zcl import foundation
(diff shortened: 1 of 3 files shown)


dmulcahey:
@prairiesnpr wanna take a peek please?

prairiesnpr:
@dmulcahey  Looks good, tested, no issues noted.

Shulyaka:
Thank you for the testing!

PR #244, opened 2020-01-16
XBee: refactor deserialization

Shulyaka:
Hi! I think I have made the code for deserialization simpler and less prone to break due to future zigpy changes.

The data from xbee does not contain a header. The old code first tries to interpret first 3 bytes of data as a zcl header, which removes them from the data, then adds the parsed header fields in front of data again and fixes the misinterpreted header fields. Instead, the new code builds a new header, inserts it in front of the data, and allows zigpy to parse it as usual.

Also fixed IO sample report parse: according to the documentation, if digital pin mask contains no pins, the digital samples block will be omitted. This will happen if no pin is configured as digital input or output.

@prairiesnpr may I ask you to look into it as well?

### DIFF
--- a/zhaquirks/xbee/__init__.py
+++ b/zhaquirks/xbee/__init__.py
@@ -17,7 +17,6 @@ the xbee stays alive in Home Assistant.
 """
 
 import logging
-import struct
 
 from zigpy.quirks import CustomDevice
 import zigpy.types as t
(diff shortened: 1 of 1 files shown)


prairiesnpr:
I'm good with this, tested and agree with the methodology.

Shulyaka:
Thank you! I will probably submit a couple more of xbee PR...

Issue #212, opened 2019-12-11
Hue button - logs for device automation


tube0013:
Here are the join jogs and zha_events for the Phillips Hue Button that recently came out.

Joining:
https://paste.ubuntu.com/p/MwpyQ4JD7s/

Events
1st single tap:
```
{
    "event_type": "zha_event",
    "data": {
        "unique_id": "00:17:88:01:06:54:aa:fe:1:0x0006",
        "device_ieee": "00:17:88:01:06:54:aa:fe",
        "endpoint_id": 1,
        "cluster_id": 6,
        "command": "on",
        "args": []
    },
    "origin": "LOCAL",
    "time_fired": "2019-12-10T23:36:56.921176+00:00",
    "context": {
        "id": "254a844669a5414895127a013ef23b6c",
        "parent_id": null,
        "user_id": null
    }
}
```
2nd single tap:

```
{
    "event_type": "zha_event",
    "data": {
        "unique_id": "00:17:88:01:06:54:aa:fe:1:0x0006",
        "device_ieee": "00:17:88:01:06:54:aa:fe",
        "endpoint_id": 1,
        "cluster_id": 6,
        "command": "off_with_effect",
        "args": [
            0,
            0
        ]
    },
    "origin": "LOCAL",
    "time_fired": "2019-12-10T23:36:59.457089+00:00",
    "context": {
        "id": "8a2884ad568446828d93f5b7cd5ce559",
        "parent_id": null,
        "user_id": null
    }
```

Here is a sequence: On > Held for about 9 seconds > Off.
https://paste.ubuntu.com/p/kBbx5dPbxJ/

the Step events seem to fire on average about ever 0.8 seconds.

Here is a sequence of holding after an Off for about 20 seconds... it does not send a stop event when held after in the "off" state:
https://paste.ubuntu.com/p/F75QmMXsDz/

majkrzak:
I'll dive into it tomorrow. 

P.S. some link
https://community.home-assistant.io/t/does-anyone-know-if-this-will-work/154624/24

majkrzak:
```python
    signature = {
        #  <SimpleDescriptor endpoint=1 profile=260 device_type=2096
        #  device_version=1
        #  input_clusters=[0, 1, 3, 64512, 4096]
        #  output_clusters=[25, 0, 3, 4, 6, 8, 5, 4096]>
        ENDPOINTS: {
            1: {
                PROFILE_ID: zha.PROFILE_ID,
[...]

PR #261, opened 2020-02-02
Add initial support for Philips Hue Smart Button (ROM001)

majkrzak:
Philips Hue Smart Button while connected to Home Assistant reports only battery status. As far as I understand how ZigBee works, this mapping should let it operate as a On/Off switch.

Based on data from #212

### DIFF
--- a/zhaquirks/philips/rom001.py
+++ b/zhaquirks/philips/rom001.py
@@ -13,13 +13,7 @@ from zigpy.zcl.clusters.general import (
 )
 from zigpy.zcl.clusters.lightlink import LightLink
 
-from ..const import (
-    DEVICE_TYPE,
-    ENDPOINTS,
-    INPUT_CLUSTERS,
-    OUTPUT_CLUSTERS,
-    PROFILE_ID,
-)
+from ..const import DEVICE_TYPE, ENDPOINTS, INPUT_CLUSTERS, OUTPUT_CLUSTERS, PROFILE_ID
 
 DEVICE_SPECIFIC_UNKNOWN = 64512
 
@@ -52,7 +46,7 @@ class PhilipsROM001(CustomDevice):
                     LevelControl.cluster_id,
                     Scenes.cluster_id,
                     LightLink.cluster_id,
-                ]
+                ],
             }
         }
     }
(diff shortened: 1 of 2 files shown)


dmulcahey:
This isn't correct. These are remotes and should not have any entities in HA.

dmulcahey:
The replacement signature shouldn't change the device type and it shouldn't unnecessarily remove a bunch of clusters. All this quirk should do is add `device_automation_triggers`

zhaquirks/philips/rom001.py
```diff
@@ -0,0 +1,67 @@
...
+                ],
+            }
+        }
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
```
dmulcahey: Don't change the device type like this. There aren't entities for this device because there shouldn't be any. It is a scene controller (remote) that emits events (as outlined in the issue).

zhaquirks/philips/rom001.py
```diff
@@ -0,0 +1,67 @@
...
+                INPUT_CLUSTERS: [
```
dmulcahey: Include all of the clusters that were removed from this list.

zhaquirks/philips/rom001.py
```diff
@@ -0,0 +1,67 @@
...
+                    Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
+                    Identify.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [OnOff.cluster_id],
```
dmulcahey: Include all of the clusters that were removed from this list.

zhaquirks/philips/rom001.py
```diff
@@ -0,0 +1,95 @@
...
+                    OnOff.cluster_id,
+                    LevelControl.cluster_id,
+                    Scenes.cluster_id,
+                    LightLink.cluster_id,
+                ],
+            }
+        }
+    }
+
+    replacement = {
```
majkrzak: Do I even need to specify replacement if I'm not touching it?

dmulcahey:
looks good for the initial implementation. Just waiting on someone to test it. Thanks for contributing!

majkrzak:
I was finally able to test it :sweat_smile: Hopefully main on/off feature works, but something is still wrong with battery status :disappointed:

NikoGrano:
As far based on my testing, many Phillips accessories has problems with giving correct battery level... Its like, its either 100% or or empty.

PR #274, opened 2020-02-19
Fix IKEA pairing issues

dmulcahey:


bwynants:
tested and confirmed to work now

PR #296, opened 2020-03-09
Limit Xiaomi PressureMeasurementCluster values

chmielowiec:
According to docs (https://www.aqara.com/us/temperature_humidity_sensor.html) pressure range is 30 kPa – 110 kPa.

On current HA dev Xiaomi sensor returned values out of specs range (-1 or 10020-10040). After that fix everything went back to normal.

**Before:**
image
**After:**
image


dmulcahey:
Thanks!

PR #304, opened 2020-03-17
Add support for the older rwl020 remote

bs42:
This adds support for a slightly older Philips hue remote that doesn't have the exact same support profile.  It is a simpler interface that does not support scenes.

### DIFF
--- a/zhaquirks/philips/rwl020.py
+++ b/zhaquirks/philips/rwl020.py
@@ -126,11 +126,6 @@ class PhilipsRWL020(CustomDevice):
             ENDPOINT_ID: 1,
             ARGS: [0, 56, 9],
         },
-        (LONG_RELEASE, "Dim Up/Down"): {
-            COMMAND: COMMAND_STOP,
-            ENDPOINT_ID: 1,
-            CLUSTER_ID: 8,
-        },
         (SHORT_PRESS, DIM_DOWN): {
             COMMAND: COMMAND_STEP,
             CLUSTER_ID: 8,
--- a/zhaquirks/philips/rwl020.py
+++ b/zhaquirks/philips/rwl020.py
@@ -1,0 +1,139 @@
+"""Phillips RWL020 device."""
+from zigpy.profiles import zha, zll
+from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
+from zigpy.zcl.clusters.general import (
+    Basic,
+    BinaryInput,
+    Groups,
+    Identify,
+    LevelControl,
+    OnOff,
+    Ota,
+    PowerConfiguration,
+)
+
+from ..const import (
+    ARGS,
+    CLUSTER_ID,
+    COMMAND,
+    COMMAND_OFF_WITH_EFFECT,
+    COMMAND_ON,
+    COMMAND_STEP,
+    DEVICE_TYPE,
+    DIM_DOWN,
+    DIM_UP,
+    ENDPOINT_ID,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    LONG_PRESS,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+    SHORT_PRESS,
+    TURN_OFF,
+    TURN_ON,
+)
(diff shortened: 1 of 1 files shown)


dmulcahey:
do all of these actions work with this? I don't have one of these to test with.

bs42:
Yes, I have 2 of these remotes and have been using them with this quirk for a while.  All actions working.

bs42:
A minor correction that I never noticed, it was incorrectly listing the OFF button as a "continuously pressed" button when it should have been just a short press.  I've corrected that.

bs42:
There is an unhandled "stop" event that happens when releasing the dimmer buttons after holding them so that it triggers continuous mode.  Should an event be added for that?

bs42:
I went ahead and added the "stop" event handler.  There is not a unique event for each button so I gave it a description that I believe would be clearest.

zhaquirks/philips/rwl020.py
```diff
@@ -0,0 +1,146 @@
...
+            ENDPOINT_ID: 1,
+            ARGS: [0, 30, 9],
+        },
+        (LONG_PRESS, DIM_UP): {
+            COMMAND: COMMAND_STEP,
+            CLUSTER_ID: 8,
+            ENDPOINT_ID: 1,
+            ARGS: [0, 56, 9],
+        },
+        (LONG_RELEASE, "Dim Up/Down"): {
```
dmulcahey: These are std keys in ZHA for the translation enging... just remove the release action

bs42: Without the (LONG_RELEASE, "Dim Up/Down"): { } block the action is not exposed to the automation engine and would have to be manually handled through zha_events.

Adding it makes it so that the automation frontend can easily assign an event to it.

dmulcahey: I wrote the feature support in ZHA. I know exactly what it does :)

dmulcahey: Doing that makes it so that the value you have supplied can’t be translated to other languages in HA. All of the keys in () actually get translated into the set language in HA. They correspond to strings in here: https://github.com/home-assistant/core/blob/d3f9408650da467ce0dbde0656837d269054566e/homeassistant/components/zha/strings.json#L25

dmulcahey: If both buttons (dim up and dim down) have an event the correct solution is to add 2 actions and use the appropriate type / sub_type combo for each

bs42: I thought you meant to remove the block entirely.

None of the existing strings directly apply to the release action since there is no specific event for each button's release.  There is a generic "step" that has the same args regardless of which button is released.

What value would you prefer to have placed into the second argument of the (LONG_RELEASE, ... ) block?

dmulcahey: I don’t follow... you should have this block for dim up and dim down. 1 for each button and there should be a difference in the event generated?

bs42: I agree, there *should* be a different event for the release, but alas, there is not.  Only a single "stop" event ( I misspoke in the previous comment, "stop" not "step").  I watched zha_event multiple times to confirm.  This event is the only one that fires regardless of which dim button was released:

{
    "event_type": "zha_event",
    "data": {
        "unique_id": "00:17:88:01:10:33:6e:f3:1:0x0008",
        "device_ieee": "00:17:88:01:10:33:6e:f3",
        "endpoint_id": 1,
        "cluster_id": 8,
        "command": "stop",
        "args": []
    },
    "origin": "LOCAL",
    "time_fired": "2020-03-26T16:08:00.362954+00:00",
    "context": {
        "id": "516d7dbc5a85425fb9bd7ef317ae55ee",
        "parent_id": null,
        "user_id": null
    }
}

dmulcahey:
In that case remove the block entirely because the action is not clear. People should use the events if they want to catch this. @adminiuga what do you think?

dmulcahey:
can you push an empty commit to this to see if travis will build it?

PR #330, opened 2020-04-08
Add Legrand dimmer without neutral

dmulcahey:
Fixes: #299

PR #333, opened 2020-04-15
Add leading whitespaces to Legrand Dimmer switch w/o neutral manufacturer and model strings

pdecat:
Just a hack for #330, so submitted as a draft PR for now.


dmulcahey:
change incorporated into #330

PR #341, opened 2020-04-19
Add lumi.sensor_motion.aq2 illuminance reports

### LINKED ISSUE #340: [Device Support Request] Aqara motion sensor illuminance updates regardless of detected motion
Aqara motion sensors on the Aqara Hub report an updated illuminance level at least once every hour, regardless of motion detection. On ZHA they only report illuminance updates when motion is detected, making them not as useful as before.

More regular illuminance updates, without the need for motion before an update is sent.

**Device signature**
Is that needed for this issue? If so, I will add it!

**Additional context**
image  
This line shows at least 8 hours without an update, since I didn't walk by the location off the sensor. The weather / sun brightness today was very different per hour; bright sun mixed with darker clouds, so a lot of brightness changes that usually would have been reported way more often.

I've found this[ similar issue from zigbee2mqtt ](https://github.com/Koenkk/zigbee2mqtt/issues/827)that talks about the same thing, with a solution that solved this in their case.

It is also possbile that this is no timed-update, but a change in lux that exceeds a certain tolerance (for example, updating every 30lx)

chmielowiec:
Fixes #340

### DIFF
--- a/zhaquirks/xiaomi/__init__.py
+++ b/zhaquirks/xiaomi/__init__.py
@@ -2,19 +2,21 @@
 import asyncio
 import binascii
 import logging
+import math
 
+import zigpy.zcl.foundation as foundation
 from zigpy import types as t
 from zigpy.quirks import CustomCluster, CustomDevice
 from zigpy.zcl.clusters.general import AnalogInput, Basic, PowerConfiguration
 from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
 from zigpy.zcl.clusters.measurement import (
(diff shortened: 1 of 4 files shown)


gentoorax:
@dmulcahey David can I just check, does this change just cause lux to be reported but we are still limited by it only being reported when there is movement, or does it resolve this issue as well? I'm guessing we are limited by the device only reporting lux when motion detected?

chmielowiec:
This device should report illuminance periodically every ~50min and every motion detected.

PR #360, opened 2020-05-23
Add third variant of Opple 6 button remote

TheTimeWalker:
This adds support for another variant for the Aqara Opple 6 button remote which has different input and output clusters for endpoints. There's no real indication from the outside that it's some different version of it. It has the same model number and the reported model id is exactly the same as the other versions.

```
<Optional endpoint=1 profile=260 device_type=261 device_version=1 input_clusters=[0, 3, 1] output_clusters=[3, 6, 8, 768]>
<Optional endpoint=2 profile=260 device_type=259 device_version=1 input_clusters=[3] output_clusters=[6, 3]>
<Optional endpoint=3 profile=260 device_type=259 device_version=1 input_clusters=[18, 3] output_clusters=[6]>
<Optional endpoint=4 profile=260 device_type=259 device_version=1 input_clusters=[18, 3] output_clusters=[6]>
<Optional endpoint=5 profile=260 device_type=259 device_version=1 input_clusters=[18, 3] output_clusters=[6]>
<Optional endpoint=6 profile=260 device_type=259 device_version=1 input_clusters=[18, 3] output_clusters=[6]>
```

### DIFF
--- a/zhaquirks/xiaomi/aqara/opple_remote.py
+++ b/zhaquirks/xiaomi/aqara/opple_remote.py
@@ -792,3 +792,131 @@ class RemoteB686OPCN01V2(XiaomiCustomDevice):
     }
 
     device_automation_triggers = RemoteB686OPCN01.device_automation_triggers
+
+
+class RemoteB686OPCN01V3(XiaomiCustomDevice):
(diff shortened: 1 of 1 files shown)


dmulcahey:
Thanks!

Issue #356, opened 2020-05-18
[Device Support Request] Eurotronic Spirit Zigbee - Battery-operated Thermostat Radiator Valve (TRV)


Hedda:
Request ZHA quirk for Eurotronic Spirit Zigbee which is a popular TRV (Thermostat Radiator Valve)

https://eurotronic.org/produkte/zigbee-heizkoerperthermostat/spirit-zigbee/

Looks like this official PDF documentation might contain all or most Zigbee attributes?

https://eurotronic.org/wp-content/uploads/2019/11/Spirit_ZigBee_BAL_web_EN_November-2019.pdf

Dependent on ZHA climate support making it into to HA https://github.com/home-assistant/core/pull/35682

There is a quirk discussion about it for deCONZ here which their gateway now have supported for:

https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1098

FYI, Eurotronic also makes a Z-Wave version of this TRV (Thermostat Radiator Valve):

https://eurotronic.org/produkte/z-wave-heizkoerperthermostat/spirit-z-wave-plus/

https://eurotronic.org/wp-content/uploads/2018/08/Spirit_Z-Wave_BAL_web_DE_view_V5.pdf

That same Z-Wave version is also being sold as OEM and rebranded to Aeotec Radiator Thermostat

https://aeotec.com/z-wave-home-automation/thermostat-radiator-valve-trv/

PS: I already have the Aeotec Z-Wave version of this TVR myself and that works very well for me.

Hedda:
FYI, Adminiuga has now submitted a pull request for "ZHA Climate" (beta) with thermostats support.

It will probably not make it into Home Assistant 0.110 but it might hopefully make it into 0.111 +

Dependent on ZHA climate support making it into to HA https://github.com/home-assistant/core/pull/35682 by @Adminiuga

docgalaxyblock:
@Hedda I purchased one of these without checking if they are supported... ^^
I logged the pairing process here (HA 110):
https://pastebin.com/PCbqqP1M

Maybee it is usefull or not

Greetings from Germany ^^

Adminiuga:
you should try the dev build of HA

docgalaxyblock:
ok will setup a VM for testing

docgalaxyblock:
I try it at a dev VM and it works fine.
I think a quirk needs to be created.

docgalaxyblock:
[...]

PR #373, opened 2020-06-11
Add quirk for Eurotronic Spirit Zigbee

vigonotion:
This quirk adds support for the thermostat "Eurotronic Spirit Zigbee".
Manual: https://eurotronic.org/wp-content/uploads/2019/11/Spirit_ZigBee_BAL_web_DE_Okt.-2019.pdf (German, but the last pages containing the Zigbee information are in English)

Related issue: #356

The thermostat works without the quirk, but not well. This quirk adds support for:

- OFF: Ability to turn off the thermostat using system mode
- Bidirectional temperature: Temperature set on the thermostat will be reported back into HA

After joining, HA needs to be restarted for the thermostat to be working properly. It seems that some attributes are not read on first join (including system_mode). Other than that, the quirk should work well.

### DIFF
--- a/zhaquirks/eurotronic/__init__.py
+++ b/zhaquirks/eurotronic/__init__.py
@@ -83,7 +83,6 @@ class ThermostatCluster(CustomCluster, Thermostat):
 
     async def read_attributes_raw(self, attributes, manufacturer=None):
         """Override wrong attribute reports from the thermostat."""
-
         success = []
         error = []
 
@@ -144,7 +143,6 @@ class ThermostatCluster(CustomCluster, Thermostat):
 
     def write_attributes(self, attributes, manufacturer=None):
         """Override wrong writes to thermostat attributes."""
-
         if "system_mode" in attributes:
 
             host_flags = self._attr_cache.get(HOST_FLAGS_ATTR, 1)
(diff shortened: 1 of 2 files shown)


Adminiuga:
Wow. Thanks. I wouldn't be able to do it without access to the device.

zhaquirks/eurotronic/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+                CURRENT_TEMP_SETPOINT_ATTR: (
+                    "current_temperature_setpoint",
+                    types.int16s,
+                ),
+                HOST_FLAGS_ATTR: ("host_flags", types.uint24_t),
+            }
+        )
+        self.attridx = {
+            attrname: attrid for attrid, (attrname, datatype) in self.attributes.items()
+        }
```
Adminiuga, line 66:
```suggestion
    attributes = {
        TRV_MODE_ATTR: ("trv_mode", types.enum8),
        SET_VALVE_POS_ATTR: ("set_valve_position", types.uint8_t),
        ERRORS_ATTR: ("errors", types.uint8_t),
        CURRENT_TEMP_SETPOINT_ATTR: ("current_temperature_setpoint", types.int16s),
        HOST_FLAGS_ATTR: ("host_flags", types.uint24_t),
    }
    attributes.update(Thermostat.attributes)
```

You could do this as well, then there's no need for custom init or updating attridx

vigonotion: Thanks, I really struggled with that one, this looks so much cleaner!

PR #374, opened 2020-06-11
Add quirk for Quirky+GE TAPT Smart Switch

coolacid:
This is my initial attempt for a quirks file for the Quirky+GE TAPT Smart Switch. 

This implements: 

- CustomCluster allowing the configuration mode of the switch. 
- Automation triggers when switch is in a compatible smart switch mode


zhaquirks/quirky/tapt.py
```diff
@@ -0,0 +1,187 @@
...
+        0x0000: ("Switch Mode", t.uint8_t),
+        # Value, Description
+        # 0 - Standard On/Off Switch, No Smart Status, Relay Status Endpoint 2
+        # 1 - Standard On/Off Switch, Smart Status on Endpoint 1, Relay Status on Endpoint 2
+        # 2 - Smart Switch, Mains always On, Smart Status on Endpoint 1
+        # 3 - Upper Toggles Load, Smart Status on Endpoint 1, Relay Status on Endpoint 2
+        # 4 - Lower Toggles Load, Smart Status on Endpoint 1, Relay Status on Endpoint 2
+    }
+
+    async def read_attributes_raw(self, attributes, manufacturer=None):
```
dmulcahey, line 67: Why not just replace none in the signatures with the mfg code you want? @Adminiuga thoughts?

zhaquirks/quirky/tapt.py
```diff
@@ -0,0 +1,187 @@
...
+        """Set the default manufacturer for Switch Mode Setting."""
+        if manufacturer == TAPT_NODE_MFG_CODE:
+            manufacturer = TAPT_MFG_CODE
+        return await super().read_attributes_raw(attributes, manufacturer)
+
+    async def write_attributes(self, attributes, manufacturer=None):
```
dmulcahey, line 73: Same as above

zhaquirks/quirky/tapt.py
Adminiuga: I think we do that in HA channels, not in the clusters. But I was thinking lately to do it in zigpy instead. Then you could override the mfg code in the quirks

@coolacid what is the mfg code reported by the node descriptor?

coolacid: When read_attributes_raw or write_attributes fires while using the Get/Set attributes in the HA UI, it uses the mfg code from the node descriptor, it doesn't appear to ever be none. 

I went up and down the classes trying to find a place I could just override the mfg code as a simple setting for the custom cluster, however I couldn't find any, thus resorting to overloading these two functions - This allowed me to Get/Set the attributes without having to specify the mfg code in the UI every time - To me, that's a quality of life bit for people that want to make use of this setting. 

I actually define, and check for the mfg code from the node descriptor, just in case somehow this custom cluster gets added to something else - but as far as I can tell, this is the only device under the quirky brand - TAPT_NODE_MFG_CODE = 0xBBAA  # 48042 base 10

dmulcahey: @Adminiuga what are your thoughts on how this was done? I saw some mfg specific PR’s to zigpy iirc. Were they related to this sort of thing?

dmulcahey:
@coolacid there may be a new way to handle this in Zigpy coming very soon. Just waiting a bit to see if that is the case. If not I'll merge this in time for the next beta and we can clean it up when we have a better solution.

coolacid:
I went thru the changes you're referring too and couldn't immediately see how to implement using the new way. If you want to merge then, then ping me on Discord after the new release, and more details, I'm happy to submit a second PR.

dmulcahey:
@coolacid wanna peek at this: https://github.com/zigpy/zha-device-handlers/pull/400 all you should have to do is define manufacturer_attributes now.

dmulcahey:
I should add that this needs the newest zigpy if you're gonna test it. You'll have to update the zha manifest in HA or pip install it and use hassocks --skip-pip to start HA.

coolacid:
Unfortunately, my deploy is all in docker, so updating to latest isn't easy. 

I looked at that PR, but it wasn't clear how to actually define the overload.

zhaquirks/quirky/tapt.py
```diff
@@ -0,0 +1,187 @@
...
+
+class TaptConfig(CustomCluster):
+    """TAPT Configuration Cluster."""
+
+    cluster_id = TAPT_CLUSTER
+    name = "TAPTCluster"
+    ep_attribute = "tapt_cluster"
+    server_commands = {}
+    client_commands = {}
+    attributes = {
```
dmulcahey, line 57: change this to: `manufacturer_attributes` and then fully remove: `read_attributes_raw` and `write_attributes` below and that should be enough.

dmulcahey:
> Unfortunately, my deploy is all in docker, so updating to latest isn't easy.
> 
> I looked at that PR, but it wasn't clear how to actually define the overload.

https://github.com/zigpy/zha-device-handlers#testing-quirks-in-development-in-docker-based-install

coolacid:
Testing the quarks is easier then your suggestion - I just volume mount the quark so it's always available between restarts. But, upgrading zigpy inside docker is not something I wanted to do in my "prod" deployment while people were in the house actively using things :)

I'll update to latest docker image, I think that has latest and try the new way of doing things.

dmulcahey:
👍 yeah this is all in HA now so should be easy to test if you upgrade HA.

coolacid:
Upgraded to latest docker container, zigpy shows version 0.22.2

Changing `attributes` to `manufacturer_attributes` and removing the manufacturer "patch" didn't work as expected. Attempted to read an attribute, and it resulting in `None` as opposed to the expected int value of `2`. 

I also don't see anywhere where I can override the Manufacturer Code. Remember, the patching is because the node_descriptor  defines the manufacturer_code as `48042` however, the `Switch Mode` attributes require code `21640`

dmulcahey:
Copy this block into the quirk: https://github.com/zigpy/zha-device-handlers/blob/408b14accfcb4dd4c447fcf24f63cf9847b790a5/zhaquirks/xiaomi/aqara/opple_remote.py#L203 and change `0x115F` to your manufacturer id

Adminiuga:
by default  we use the manufacturer_id indicated by the device node descriptor. If for some reason manufacturer_id is in the node descriptor is different from the one which works, then you need to override it in your `CustomDevice` class definition the `manufacturer_id_override = None`

Adminiuga:
in other words override https://github.com/zigpy/zigpy/blob/64e4f731d34e3ed0aa75cb1146cb79ea1314c91a/zigpy/device.py#L34 attribute, so you don't have to mess with node descriptors :)

dmulcahey:
Or that :)

Adminiuga:
yeah, the manufacturer id override was added recently. Opple quirk was created much earlier than that and had to override the node descriptor

zhaquirks/quirky/tapt.py
Adminiuga: fixed in https://github.com/zigpy/zigpy/pull/429 see PR description for details

coolacid:
I def see those changes in the zigpy version in the docker container. 

I added `manufacturer_id_override = TAPT_MFG_CODE` to the `TAPTSwitch(CustomDevice)` class.  However, at the read_attributes_raw call (hooking there because I know I can see what's about to happen) manufacturer is still coming back as the node descriptor value. 

Forcing the ID by replacing it in the Node Descriptor did work - I'm going to leave it this way for a bit to make sure it didn't effect the other functions of the switch, then dive back into the `manufacturer_id_override`

Adminiuga:
btw, for `manufacturer_id_overrride` to work, your attributes should be defined in `manufacturer_attributes` dict as per https://github.com/zigpy/zha-device-handlers/pull/374#discussion_r482917671

dmulcahey:
Please rebase to pick up the CI change. We swapped from travis to GH Actions. Sorry for the inconvenience.

dmulcahey:
@coolacid I made the suggested changes to the PR. Can you try this and let us know if this works?

HAelnino80:
If you need test, I have a Tapt, but i don't know how to test a quirks.

PR #375, opened 2020-06-12
Fix plaid battery sensor

Adminiuga:
Fix plaid battery sensor.

PS:  and that's why you write the tests, kids :) Next time

### DIFF
--- a/zhaquirks/plaid/soil.py
+++ b/zhaquirks/plaid/soil.py
@@ -25,7 +25,7 @@ class PowerConfigurationClusterMains(PowerConfigurationCluster):
     def _update_attribute(self, attrid, value):
         super()._update_attribute(attrid, value)
         if attrid == self.MAINS_VOLTAGE_ATTR:
-            super()._update_attribute(self.self.BATTERY_VOLTAGE_ATTR, value)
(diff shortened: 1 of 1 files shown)


Adminiuga:
Fixes https://github.com/zigpy/zha-device-handlers/issues/366

zhaquirks/plaid/soil.py
```diff
@@ -25,7 +25,7 @@ class PowerConfigurationClusterMains(PowerConfigurationCluster):
     def _update_attribute(self, attrid, value):
         super()._update_attribute(attrid, value)
         if attrid == self.MAINS_VOLTAGE_ATTR:
-            super()._update_attribute(self.self.BATTERY_VOLTAGE_ATTR, value)
+            super()._update_attribute(self.BATTERY_VOLTAGE_ATTR, round(value / 100))
```
dmulcahey, line 28: Doesn’t the rounding already happen in HA?

dmulcahey: Nvm... missed what was rounded

Adminiuga:
meh, it  would need an approach similar to Eurotronic TRV valve...

Issue #385, opened 2020-06-25
[Device Support Request] - Lumi Weather Sensor needs an alternate quirk


dmulcahey:
**Is your feature request related to a problem? Please describe.**

The Aqara weather sensor has no quirk and no battery info as a result. Info here: https://github.com/jcallaghan/home-assistant-config/issues/167#issuecomment-649669395

**Describe the solution you'd like**
Add an alternate quirk for this device

**Device signature**

```
{
  "node_descriptor": "<Optional byte1=2 byte2=64 mac_capability_flags=128 manufacturer_code=4151 maximum_buffer_size=127 maximum_incoming_transfer_size=100 server_mask=0 maximum_outgoing_transfer_size=100 descriptor_capability_field=0>",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0302",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0402",
        "0x0403",
        "0x0405",
        "0xffff"
      ],
      "out_clusters": [
        "0x0000",
        "0x0004",
        "0xffff"
      ]
    }
  },
  "manufacturer": "LUMI",
  "model": "lumi.weather",
  "class": "zigpy.device.Device"
}
```

**Additional context**

Adminiuga:
Fixed by #386

Issue #368, opened 2020-06-03
Aqara Temps - Wrong type detected


martynstar:
Hi there

It seems that something has slightly changed in the hardware on the temp sensor.
Previously this was detected and shows - Temperature, Pressure, Humidity and Battery life.
Zigbee information
{
  "node_descriptor": "<Optional byte1=2 byte2=64 mac_capability_flags=128 manufacturer_code=4151 maximum_buffer_size=127 maximum_incoming_transfer_size=100 server_mask=0 maximum_outgoing_transfer_size=100 descriptor_capability_field=0>",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x5f01",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0402",
        "0x0403",
        "0x0405",
        "0xffff"
      ],
      "out_clusters": [
        "0x0000",
        "0x0004",
        "0xffff"
      ]
    }
  },
  "manufacturer": "LUMI",
  "model": "lumi.weather",
  "class": "zhaquirks.xiaomi.aqara.weather.Weather"
}

Now the device i purchased recently is showing only the Temperature, Pressure and Humidity.
Zigbee info.

{
  "node_descriptor": "<Optional byte1=2 byte2=64 mac_capability_flags=128 manufacturer_code=4151 maximum_buffer_size=127 maximum_incoming_transfer_size=100 server_mask=0 maximum_outgoing_transfer_size=100 descriptor_capability_field=0>",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0302",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0402",
        "0x0403",
        "0x0405",
        "0xffff"
      ],
      "out_clusters": [
        "0x0000",
        "0x0004",
        "0xffff"
      ]
    }
  },
  "manufacturer": "LUMI",
  "model": "lumi.weather",
  "class": "zigpy.device.Device"
}

If we could get this added to the quirks it would be great! what more info can i grab or do we need some debug logs?

Thanks

The slight hardware change, everything else looks the same on the kit but maybe a firmware?

caylor93:
Is there a timeline for this update?

Adminiuga:
if you can make changes to your environment, then try manually updating `zhaquirks/xiaomi/aqara/weather.py` file with changes from https://github.com/zigpy/zha-device-handlers/pull/386/files

[...]

PR #386, opened 2020-06-26
New signature for Aqara temp/humid/pressure sensor

Adminiuga:
Fixes:
- https://github.com/zigpy/zha-device-handlers/issues/368
- https://github.com/zigpy/zha-device-handlers/issues/385


dmulcahey:
wait... where is replacement?

Adminiuga:
Inherited from the class above. I could add it explicitly if needed

dmulcahey:
wow completely missed that... my bad.... Good idea!

Issue #86, opened 2019-05-28
Hue Dimmer Long Press Events


andrewpc:
Ive successfully joined a hue dimmer (RWL021) to ZHA. The basic events are present, these being single tap on/off, single tap dim up/down and hold dim up/down step. These can be mapped through to HA for automations.

Reading into the Zigbee2MQTT support for this dimmer, it appears that this device is also capable of providing discrete long press and multi-press events for all 4 buttons. To enable this, from reading the code and the related issue, it appears that when initialized, the dimmer needs an attribute written to it. 

The relevant code reference is here: https://github.com/Koenkk/zigbee-shepherd-converters/blob/69befbcce5d744cf5f81e1cdf6f59cd5bcb5917e/devices.js#L909

The Issue link is here: https://github.com/Koenkk/zigbee2mqtt/issues/585

Is this assumption correct, and how difficult is this to add to ZHA? Im happy to try and have a go at this, but want to get some validation on whats needed and any ZHA limitations that might prevent this?

dmulcahey:
This should be fairly trivial. You may be able to write the attribute via the config panel already. That said, I do see a bind for a mfg specific cluster there. Not sure what that’s about. I don’t have one of these... do you have a link to the exact model? I can pick one up to mess with this a bit.

David F. Mulcahey

jonathanweinberg:
Not the OP, but this should be the model: https://www.amazon.com/Philips-Dimmer-Switch-Installation-Free-Exclusively/dp/B076MGKTGS

As far as I know, there's only one real model of the switch, different SKU's for Europe / US + Style ( I/O vs On/Off printed on the remote).

Found this: https://www.reddit.com/r/Hue/comments/7ia4f5/is_there_a_second_version_of_the_hue_dimmer_switch/

andrewpc:
Yes, thanks @jonathanweinberg Confirming thats the one I am using.

I did try sending the configuration using the config panel, but either didnt get the values write or got some warnings about the values being invalid? I wasnt sure if there was some kind of schema limiting what needed to be sent or I had converted them incorrectly.

andrewpc:
Just an update, I did re-try setting the attribute using the service, but I believe it failed because the attribute is manufacturer specific, and therefore didnt match the schema. If I can get around that, by setting some other flag, I am happy to try again? It might be that it should be done in code?

dmulcahey:
@andrewpc I picked one up. Should be here tomorrow.

[...]

PR #388, opened 2020-06-27
[BREAKING CHANGE] Phillips Hue remote manufacturer specific cluster handling

dmulcahey:
This PR adds support for 4 actions per button: press, hold, short release and long release. It needs https://github.com/home-assistant/core/pull/37172 in order to function. 

Fixes: #86 

Things still needed:

- [x] device trigger support
- [x] deduplicate code between remote versions

### DIFF
--- a/zhaquirks/philips/__init__.py
+++ b/zhaquirks/philips/__init__.py
@@ -61,11 +61,11 @@ class PhilipsOnOffCluster(CustomCluster, OnOff):
     attributes.update({0x4003: ("power_on_state", PowerOnState)})
 
 
-class PhillipsBasicCluster(CustomCluster, Basic):
-    """Phillips Basic cluster."""
+class PhilipsBasicCluster(CustomCluster, Basic):
+    """Philips Basic cluster."""
 
     attributes = Basic.attributes.copy()
-    attributes.update({0x0031: ("phillips", t.bitmap16)})
+    attributes.update({0x0031: ("philips", t.bitmap16)})
 
     attr_config = {0x0031: 0x000B}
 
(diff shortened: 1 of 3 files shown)


dmulcahey:
@Adminiuga I changed the triggers to make these what we use since they are more robust. Should we mark this as a breaking change in HA since if anyone was using the old ones they will no longer work?

Adminiuga:
Yes, should be marked as breaking then. Would it break all device triggers or only hue ones

dmulcahey:
Just these 2 remotes

Adminiuga:
Maybe still should mark it as breaking but with limited scope for these two remotes.

dmulcahey:
@Adminiuga think if I merge this and cut a release that we can pull this and the corresponding HA change into the beta? Or should we wait?

Adminiuga:
Since this is a breaking change, maybe should wait a bit

zhaquirks/philips/__init__.py
```diff
@@ -19,3 +59,57 @@ class PhilipsOnOffCluster(CustomCluster, OnOff):
 
     attributes = OnOff.attributes.copy()
     attributes.update({0x4003: ("power_on_state", PowerOnState)})
+
+
+class PhilipsBasicCluster(CustomCluster, Basic):
+    """Philips Basic cluster."""
+
+    attributes = Basic.attributes.copy()
+    attributes.update({0x0031: ("philips", t.bitmap16)})
```
Adminiuga, line 68: @dmulcahey where this is coming from? isn't this attr supposed to be on `0x0406` Occupancy cluster id?

dmulcahey: https://github.com/Koenkk/zigbee-herdsman-converters/blob/69befbcce5d744cf5f81e1cdf6f59cd5bcb5917e/devices.js#L931

dmulcahey: It makes the notification on the manufacturer cluster work from what I could see. It’s like a mode switch similar to what the Opple and Lightify x4 remotes do

Adminiuga: ah, ok. Just didn't see this in https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Philips-Hue-sniffing#hue-motion-sensor and thought it was the same attribute as sensitivity on occupancy cluster

PR #396, opened 2020-07-05
Added power on state for Philips Hue LCB001, LCT011, LCT024, LST002

jamarzka:
Also fixed what I think is a copy/paste error in LCT016


dmulcahey:
Thanks for the contribution!

PR #400, opened 2020-07-07
Refactor "manufacturer specific" attributes and commands declaration

Adminiuga:
Sync with https://github.com/zigpy/zigpy/pull/429 changes. Refactor manufacturer specific attribute, client and server commands.

Allow Zigpy Device class to override manufacturer code, normally supplied by the node_descriptor.
Implement the following new attributes for the CustomCluster class:

`manufacturer_attributes` -- a dict of manufacturer specific attributes, in addition to ZCL standard attributes
`manufacturer_client_commands` -- a dict of manufacturer specific commands, in addition to ZCL standard client commands
`manufacturer_server_commands` -- a dict of manufacturer specific commands, in addition to ZCL standard server commands

`CustomDevice` allow definition of the `manufacturer_id_override: Optional[int]` and this manufacturer code would be used automatically for operations related to any manufacturer specific commands or cluster. If not specified, then `manufacturer_id` from the node descriptor is being used.

Requesting manufacturer specific commands or attributes would automatically provide manufacturer_id for the ZCL frame, unless overridden by the manufacturer parameter for the request.

e.g. for quirks now you can just do:
```
class ThermostatCluster(CustomCluster, Thermostat):
    """Thermostat cluster."""

    manufacturer_attributes = {
        TRV_MODE_ATTR: ("trv_mode", types.enum8),
        SET_VALVE_POS_ATTR: ("set_valve_position", types.uint8_t),
        ERRORS_ATTR: ("errors", types.uint8_t),
        CURRENT_TEMP_SETPOINT_ATTR: ("current_temperature_setpoint", types.int16s),
        HOST_FLAGS_ATTR: ("host_flags", types.uint24_t),
    }
```
And this would extend standard ZCL attributes with new manufacturer specific attributes for this cluster

### DIFF
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -41,7 +41,7 @@ class LocalDataCluster(CustomCluster):
         """Prevent unbind."""
         return (foundation.Status.SUCCESS,)
 
-    async def _configure_reporting(self, *args, **kwargs):
+    async def _configure_reporting(self, *args, **kwargs):  # pylint: disable=W0221
         """Prevent remote configure reporting."""
         return foundation.ConfigureReportingResponse.deserialize(b"\x00")[0]
 
--- a/zhaquirks/xiaomi/aqara/vibration_aq1.py
+++ b/zhaquirks/xiaomi/aqara/vibration_aq1.py
@@ -61,6 +61,8 @@ _LOGGER = logging.getLogger(__name__)
 class VibrationAQ1(XiaomiCustomDevice):
     """Xiaomi aqara smart motion sensor device."""
 
+    manufacturer_id_override = 0x115F
+
     def __init__(self, *args, **kwargs):
         """Init."""
         self.motion_bus = Bus()
--- a/zhaquirks/xiaomi/aqara/vibration_aq1.py
+++ b/zhaquirks/xiaomi/aqara/vibration_aq1.py
@@ -70,15 +70,13 @@ class VibrationAQ1(XiaomiCustomDevice):
         """Vibration cluster."""
 
         cluster_id = BasicCluster.cluster_id
-        attributes = Basic.attributes.copy()
-        attributes.update({0xFF0D: ("sensitivity", types.uint8_t)})
+        manufacturer_attributes = {0xFF0D: ("sensitivity", types.uint8_t)}
 
     class MultistateInputCluster(CustomCluster, MultistateInput):
         """Multistate input cluster."""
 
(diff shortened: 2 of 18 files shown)


zhaquirks/xiaomi/aqara/vibration_aq1.py
```diff
@@ -70,15 +70,13 @@ class VibrationBasicCluster(BasicCluster):
         """Vibration cluster."""
 
         cluster_id = BasicCluster.cluster_id
-        attributes = Basic.attributes.copy()
-        attributes.update({0xFF0D: ("sensitivity", types.uint8_t)})
+        manufacturer_attributes = {0xFF0D: ("sensitivity", types.uint8_t)}
```
dmulcahey, line 73: this guy needs a different mfg code. I think it is 4447 but let me double check

dmulcahey: Yeah it is 4447

Adminiuga: Is it the same for all xiaomi or just for vibration sensor?

Adminiuga: In other words should it be overriden just for vibration sensor or all xiaomi? I think just for vibration for starters

dmulcahey: > Is it the same for all xiaomi or just for vibration sensor?

Just this one. It reports something different in the node descriptor

Adminiuga: Added override in https://github.com/zigpy/zha-device-handlers/pull/400/commits/b6373fa7a1a551bf412606404200cb873e1edc10

zhaquirks/__init__.py
```diff
@@ -41,7 +41,7 @@ async def unbind(self):
         """Prevent unbind."""
         return (foundation.Status.SUCCESS,)
 
-    async def _configure_reporting(self, *args, **kwargs):
+    async def _configure_reporting(self, *args, **kwargs):  # pylint: disable=W0221
```
dmulcahey, line 44: Just curious, why not make the signature match?

Adminiuga: I was thinking about it, but particularly this implementation doesn't really care about the signature or parameters and in the future if the signature in zigpy has to change for any reason, it's just one less dependency to be a breaking change.
But could be a matching signature,

PR #402, opened 2020-07-10
Add sercomm flood sensor quirk

xdissent:
Adds a quirk for Sercomm SZ-WTD02N, sold as [Securifi Flood Sensor](https://www.securifi.com/securifi-flood-sensor)


dmulcahey:
Thanks!

PR #419, opened 2020-07-20
Support for WXKG07LM

adchevrier:
Add support for the Xiaomi Aqara D1 double key wireless wall switch.

It is detected as a "lumi.remote.b286acn02" when paired to Home Assistant.

I tested all actions (left/right/both single, double and long press) and they all map correctly.


Adminiuga:
can you run ` zhaquirks/xiaomi/aqara/remote_b286acn01.py ` through black to make lint happy?

adchevrier:
Done.

I was wondering if the file should be renamed? There are 3 models covered by this quirk so maybe it should be named in a more generic way?

Btw Xiaomi also released a new 2020 single key wireless switch, I don't have one yet but I'm guessing they'll have changed the model number in the same way.

PR #422, opened 2020-07-27
add quirks for Sinope light and dimmer

claudegel:
This is my first quirks for zigbee devices. Hope it's ok for code formating etc.


claudegel:
Need to fix some  check error

PR #424, opened 2020-07-27
Add quirks to support Sinope light and dimmer

claudegel:
First pull request for Sinope light and dimmer support in zha/zigpy


Adminiuga:
You don't need lines 38 through 53. Just use cluster id for manufacturer specific cluster feeding n signature/replacement

claudegel:
Oups need formating

claudegel:
If I remove line 38 to 53 I also need to remove line `SinopeTechnologiesLightingCluster,` and
 ` SinopeTechnologiesManufacturerCluster,` in replacement.
How do I force an update of my PR to add those changes

claudegel:
I've closed that PR but open a new one #425. So how can I add changes to PR 425. Do I need to close and open a new one ?

Adminiuga:
1. Just use cluster IDs for those. You are not doing anything with the clusters themselves, you are just overriding the endpoint's device type.

2. Just push changes to the same branch for which you be opened PR

To fix formatting:
pip3 install pre-commit
pre-commit install
pip3 install black
black <your py file>

claudegel:
I've install and run black to reformat then removed some unused code but I'm still stuck with
`The command "travis_wait 40 tox --develop" exited with 1 `

claudegel:
About the cluster_id, I think I need to have it everywhere because I don't change any cluster. So for example I need to put 
`Basic.cluster_id,` in both signature and replacement. If I put only `Basic,` I'll need to have a class extenting Basic
```
class Basic(Basic):
    """ Changing something in Basic cluster """
    things_to_be_changed
```

Adminiuga:
In your case just put Basic.cluster_id in the signature and replacement. You are not changing any clusters, are you?

Adminiuga:
You put a cluster class instead of a cluster id in the replacement, only when you want to use your CustomCkuster class because you are doing something different there, like a new command or new attribute

claudegel:
For now I'm not changing anything in the clusters. That's the next step
I've set the DeviceType in signature back to it's original one ON_OFF_LIGHT_SWITCH and add the new DeviceType ON_OFF_LIGHT in replacement. Then reformat with black

claudegel:
I've updated my PR but still getting the same error
`ERROR: InvocationError for command /home/travis/build/zigpy/zha-device-handlers/script/check_format (exited with code 1)`
Black formating just changed the OUTPUT_CLUSTERS: format, each cluster_id have been splitted on separate line

PR #429, opened 2020-07-31
Starting to flesh out the building example

walthowd:
Starting to flesh out the building example and some more spelling fixes


dmulcahey:
mind rebasing this?sorry about that... I completely missed the fact that this was open!

dmulcahey:
Thanks!!!

Issue #426, opened 2020-07-28
[Documentation] Documentation to help get devices added


jlongman:
Hey there, thanks in advance for what you do.

While trying to do #94 I had a hard time with two things:

- how to do a new device
- how to test the new files once added

The second is maybe simplest, the docs talk about using pypi as a source, (I've never done this but) do I push my code to pypi to validate whether what I'm doing works?  How would I do this?  (If it matters I'm in a hassio environment).

For the first, well I was trying to write this custom handling for the orvibo motion detector that works in zigbee2mqtt. I took an aqara device that had the same requirement to reset the state on a timer and copied a bunch off of it.  But it wasn't clear whether the MOTION AND IAS_ZONE was a choice/a feature to make integration or merely reflecting the devices zigbee reporting.

Additionally the aqara had a signature and a replacement which was referencing `cluster_id`s and `Clusters` _with_ `cluster_id`s so it was confusing. e.g.

```
    signature = {
# skip the SimpleDescriptor
        MODELS_INFO: [(LUMI, "lumi.sensor_motion.aq2")],
        ENDPOINTS: {
            1: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.OCCUPANCY_SENSOR,
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    XIAOMI_CLUSTER_ID,
                    OccupancySensing.cluster_id, 
# etc
    }

    replacement = {
        SKIP_CONFIGURATION: True,
        ENDPOINTS: {
            1: {
                INPUT_CLUSTERS: [
                    BasicCluster,
                    PowerConfigurationCluster,
                    Identify.cluster_id,
# etc
```

(Is that the point?  the first is used for matching the device and the second is what we present?)

Anyways, I think I made my point is that how to build one isn't clear.  Maybe there's some mandatory fields.  And I get that this probably isn't trivial and even harder to explain, but are there resources somewhere I'm missing maybe?

jlongman:
These are what I came up for what it's worth.  (The illuminance appears to be something I'm not sure is supported in the orvibo but maybe is in the aqara).

# orvibo/__init__.py
```
import asyncio
import logging
import math

from zigpy.quirks import CustomCluster, CustomDevice
from zigpy.zcl.clusters.measurement import (
    IlluminanceMeasurement,
)
from zigpy.zcl.clusters.security import IasZone

from .. import Bus, LocalDataCluster
[...]

PR #431, opened 2020-08-03
Update contributing guide

dmulcahey:
Fixes: #426

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ ZHA device handlers and it's provided Quirks allow Zigpy, ZHA and Home Assistant
 
 ## What is a device in human terms?
 
-A device is a physical object that you want to join to a Zigbee network: a light bulb, a switch, a sensor etc. The host application, in this case Zigpy, needs to understand how to interact with the device so there are standards that define how the application and devices can communicate. The device's functionality is described by several descriptors while he device itself contain endpoints and endpoints contain clusters. Zigpy needs to understand all these elements in order to correctly work with the device.
+A device is a physical object that you want to join to a Zigbee network: a light bulb, a switch, a sensor etc. The host application, in this case Zigpy, needs to understand how to interact with the device so there are standards that define how the application and devices can communicate. The device's functionality is described by several descriptors while the device itself contains endpoints and endpoints contain clusters. Zigpy needs to understand all these elements in order to correctly work with the device.
 
 ### Endpoints
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ A device is a physical object that you want to join to a Zigbee network: a light
 
 ### Endpoints
 
-Endpoints are essentially groupings of functionality. For example, a typical Zigbee light bulb will have a single endpoint for the light. A multi-gang wall switch may have an endpoint for each invidual switch so they can all be controlled separately. Each endpoint has several functions represented by clusters.
+Endpoints are essentially groupings of functionality. For example, a typical Zigbee light bulb will have a single endpoint for the light. A multi-gang wall switch may have an endpoint for each individual switch so they can all be controlled separately. Each endpoint has several functions represented by clusters.
 
 ### Clusters
 
@@ -38,25 +38,25 @@ A node descriptor explains some basic device attributes to Zigpy. The manufactur
 
 #### What is a Simple Descriptor and why do we care?
 
-A simple descriptor is a description of a Zigbee device endpoint and is responsible for explaining the endpoint's functionality. It contains a profile id, the device type, and collections of clusters. The profile id tells the application what set of Zigbee rules to use. The device type tells the application what logical type of device this is ex: on off light, color light, etc. The clusters explain to the application what types of functionality exist on the endpoint. Here is an example:
+A simple descriptor is a description of a Zigbee device endpoint and is responsible for explaining the endpoint's functionality. It contains a profile id, the device type, and collections of clusters. The profile id tells the application what set of Zigbee rules to use. The most common profile will be 260 (0x0104) for the Home Automation profile. The device type tells the application what logical type of device this is ex: on off light, color light, etc. The clusters explain to the application what types of functionality exist on the endpoint. Here is an example:
 `<SimpleDescriptor endpoint=1 profile=260 device_type=1026 device_version=0 input_clusters=[0, 1, 3, 32, 1026, 1280, 2821] output_clusters=[25]>`
 
 ## What the heck is a quirk?
 
 In human terms you can think of a quirk like google translate. I know it's a weird comparison but lets dig in a bit. You may only speak one language but there is an interesting article written in another language that you really want to read. Google translate takes the original article and displays it in a format (language) that you understand. A quirk is a file that translates device functionality from the format that the manufacturer chose to implement it in to a format that Zigpy and in turn ZHA understand. The main purpose of a quirk is to serve as a translator. A quirk is comprised of several parts:
 
-- signature
-- replacement
-- device_automation_triggers
+- Signature -- To identify and apply the correct quirk
+- Replacement -- To allow Zigpy and ZHA to correctly work with the device
+- device_automation_triggers - To let the Home Assistant Device Automation engine and users interact with the device
 
-## signature
+### Signature
 
 The signature on a quirk identifies the device as the manufacturer implemented it. You can think of it as a fingerprint or the dna of the device. The signature is what we use to identify the device. If any part of the signature doesn't match what the device returns during discovery the quirk will not match and as a result it will not be applied. The signature is made up of several parts:
 
 - `models_info`
 - `endpoints`
 
-Models info tells the application what devices the quirk should apply to. Endpoints are the simple descriptors that we spoke about earlier exactly as they are on the device. `endpoints` is a dict where the key is the id of the endpoint and the value is an object with the following properties: `profile_id`, `device_type`, `input_clusters` and `output_clusters`. Creating the signature element is generally just a job of transcribing what the device gives us. Here is an example:
+Models info tells the application which devices should use this particular quirk. Endpoints are the simple descriptors that we spoke about earlier exactly as they are on the device. `endpoints` is a dict where the key is the id of the endpoint and the value is an object with the following properties: `profile_id`, `device_type`, `input_clusters` and `output_clusters`. Creating the signature element is generally just a job of transcribing what the device gives us. Here is an example:
 
 ```python
 signature = {
@@ -87,7 +87,7 @@ signature = {
 }
 ```
 
-## replacement
+### Replacement
 
 The replacement on a quirk is what we want the device to be. Remember, we said that quirks were like Google translate... you can think of the replacement like the output from Google translate. The replacement dict is what will actually be used by Zigpy and ZHA to interact with the device. The structure of `replacement` is the same as signature with 2 key differences: `models_info` is generally omitted and there is an extra element `skip_configuration` that instructs the application to skip configuration if necessary. Some manufacturers have not implemented the specifications correctly and the devices come pre-configured and therefore the configuration calls fail (non Zigbee 3.0 Xiaomi devices for instance) Here is an example:
 
@@ -116,11 +116,11 @@ replacement = {
 }
 ```
 
-## device_automation_triggers
+### device_automation_triggers
 
 Device automation triggers are essentially representations of the events that the devices fire in HA. They allow users to use actions in the UI instead of using the raw events.
 
-# Great, all of the definitions are out of the way. Let's break down an example!
+# Building a quirk
 
 Now that we got that out of the way we can focus on the task at hand: make our devices work the way they should with Zigpy and ZHA. Because the device doesn't work correctly out of the box we have to write a quirk for it. First lets look at what the quirk looks like when complete:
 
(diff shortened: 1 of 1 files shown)


dmulcahey:
@Adminiuga, @abmantis wanna poke at this when you have some free time?

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        self._current_state = {}
+        super().__init__(*args, **kwargs)
+
+    def _update_attribute(self, attrid, value):
+        super()._update_attribute(attrid, value)
+        if value is not None and value >= 0:
+            self.endpoint.device.power_bus.listener_event(POWER_REPORTED, value)
```
jlongman, line 270: Are the `voltage` and `consumption` omitted for the sake of brevity?  Are they derivative of the power?

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+            DEVICE_TYPE: zha.DeviceType.OCCUPANCY_SENSOR,
+            INPUT_CLUSTERS: [BinaryInput.cluster_id],
+            OUTPUT_CLUSTERS: [BinaryInput.cluster_id, Groups.cluster_id],
+        },
+```
+
+You can see that we have replaced `ElectricalMeasurement.cluster_id` from endpoint 1 in the `signature` dict with the name of our cluster that we created: `ElectricalMeasurementCluster` and on endpoint 2 we replaced `AnalogInput.cluster_id` with the implementation we created for that: `AnalogInputCluster`. This instructs Zigpy to use these `CustomCluster` derivatives instead of the normal cluster definitions for these clusters and this is why this part of the quirk is called `replacement`.
+
+Now lets put this all together. If you examine the device definition above you will see that we have defined our custom device, we defined the `signature` dict where we transcribed the `SimpleDescriptor` output we obtained when the device joined the network and we defined the `replacement` dict where we swapped the cluster ids for the culsters that we wanted to replace with the `CustomCluster` implementations that we created.
+
```
jlongman, line 365: While working on my (non-working, ongoing) attempt on the orvibo motion detector, I noticed that Xiaomi had added in the `replacement` either an extra (occupancy or motion) cluster - this kind of makes sense as one is derivative of the other and adding the cluster makes the component more viable.

Does this imply the actual `signature` supported value must publish to the unsupported `replacement` value?

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+- Capture the SimpleDescriptor log entries for each endpoint on the device. These can be found in the HA logs after joining a device and they look like this: `<SimpleDescriptor endpoint=1 profile=260 device_type=1026 device_version=0 input_clusters=[0, 1, 3, 32, 1026, 1280, 2821] output_clusters=[25]>`. This information can also be obtained from the zigbee.db if you want to take the time to query the tables and reconstitute the log entry. I find it easier to just remove and rejoin the device. ZHA entity ids are stable for the most part so it _shouldn't_ disrupt anything you have configured. These need to match what the device reports EXACTLY or zigpy will not match them when a device joins and the handler will not be used for the device. You can also obtain this information from the device screen in HA for the device. The `Zigbee Device Signature` button will launch a dialog that contains all of the information necessary to create quirks.
```
jlongman, line 370: Log settings to get these might be helpful to add here, ie how to see the `SimpleDescriptor`.  Note, I IIRC I didn't see the `SimpleDescriptor` string but maybe I had the wrong log level or because I was using a cc2531?

The docker/local install methodology of testing might also be helpful to add to this document.

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
-# Initial Contribution Guidelines - WIP
+# Primer
```
jlongman, line 1: Generally this is great and thank you.  As someone wrestling with trying to get the Orvibo motion detector, I realized right away that my replacement cluster didn't include my replacement code.

Generally I found sometimes the document described the what and not the why or how to use it.  I tried to give a few cases below.

Extra/later steps might include the a discussion on the timers required by some quirks and maybe a troubleshooting/debugging section.  (additional notes below).

CONTRIBUTING.md
jlongman: my descriptor had a slightly different format:
```

#   <Optional endpoint=1 profile=260 device_type=1026 device_version=1
#       input_clusters=[0, 3, 1280, 65535, 1]
#       output_clusters=[0, 4, 3, 5, 1]
#       >
#  {
#   "node_descriptor": "<NodeDescriptor
#       byte1=2
#       byte2=64
#       mac_capability_flags=128
#       manufacturer_code=4151
#       maximum_buffer_size=127
#       maximum_incoming_transfer_size=100
#       server_mask=0
#       maximum_outgoing_transfer_size=100
#       descriptor_capability_field=0
#       >",
#   "endpoints": {
#     "1": {
#       "profile_id": 260,
#       "device_type": "0x0402",
#       "in_clusters": [
#         "0x0000",
#         "0x0001",
#         "0x0003",
#         "0x0500",
#         "0xffff"
#       ],
#       "out_clusters": [
#         "0x0000",
#         "0x0001",
#         "0x0003",
#         "0x0004",
#         "0x0005"
#       ]
#     }
#   },
#   "manufacturer": "ORVIBO",
#   "model": "895a2d80097f4ae2b2d40500d5e03dcc",
#   "class": "zigpy.device.Device"
# }
```

jlongman: and a reminder to turn off the logs too!

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+}
+```
+
+### Replacement
+
+The replacement on a quirk is what we want the device to be. Remember, we said that quirks were like Google translate... you can think of the replacement like the output from Google translate. The replacement dict is what will actually be used by Zigpy and ZHA to interact with the device. The structure of `replacement` is the same as signature with 2 key differences: `models_info` is generally omitted and there is an extra element `skip_configuration` that instructs the application to skip configuration if necessary. Some manufacturers have not implemented the specifications correctly and the devices come pre-configured and therefore the configuration calls fail (non Zigbee 3.0 Xiaomi devices for instance) Here is an example:
+
+```python
+replacement = {
+    SKIP_CONFIGURATION: True,
```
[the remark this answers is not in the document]
Adminiuga: `SKIP_CONFIGURATION = True` means the ZHA wont bind and configure ZCL attribute reporting for supported (by ZHA) clusters. This is a workaround for Xioami legacy devices and is not supposed to be used with any other compliant devices. Legacy xiaomi devices are hardcoded to report attributes to NWK 0x0000 address -- coordinator and  because those devices are too "sleepy" often binding/attribute report configuration was timing out for those, so this hack was introduced. 
If `SKIP_CONFIGURATION = False` or just not present on the custom cluster, then ZHA will bind and configure attribute reporting for that cluster.

CONTRIBUTING.md
Adminiuga: Simple descriptor is logged by the `zigpy` but only if the device is not present in the database -- very new join. On rejoins the  full discovery is skipped, so you might not see it. And it also could be burried by all other logging noise. 
In zigpy logs, you should see `...new device joined...` and after that be on lookout for SimpleDescriptor messages

CONTRIBUTING.md
Adminiuga: I'm not quite following what are you asking. Perhaps references to specific examples/code would help.

Consider `signature` as a must have match, for the `replacement` to kick in. The content of replacement doesn't have to match the signature, it is entirely up to quirk author and the way device is functioning, but in 99% of the use cases, the replacement is a copy of signature with some attributes/clusters replaced by the quirk.
e.g https://github.com/zigpy/zha-device-handlers/blob/e09f7f81171de15c84825f5e0b3f3b987616f829/zhaquirks/sinope/light.py#L67 just changes the device type, so ZHA discovers that device a "light" and not as a "switch", leaving everything else as is. 

In other cases, replacement may contain extra "virtual" cluster not present on the original devices, like ElectricalMeasurement cluster for Xiaomi devices. Xiaomi reports power usage on a analog cluster (IIRC), but ZHA doesn't know about power metering on analog cluster. So quirks gets data from analog cluster and presents it through a virtual "ElectricalMeasurement" cluster which ZHA knows how to represent.

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+### Clusters
+
+Clusters are objects that contain the information (attributes and commands) for individual functions. There is the ability to turn the switch on and off, maybe there is energy monitoring, maybe there is the ability to add each switch to an individual group or a scene, etc.
+
+### Descriptors
+
+For the purposes of Zigpy and Quirks we will focus on two descriptors:
+
+- Node Descriptor: desc goes here
+- Simple Descriptor: desc goes here
```
abmantis, line 32:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+
+Endpoints are essentially groupings of functionality. For example, a typical Zigbee light bulb will have a single endpoint for the light. A multi-gang wall switch may have an endpoint for each individual switch so they can all be controlled separately. Each endpoint has several functions represented by clusters.
+
+### Clusters
+
+Clusters are objects that contain the information (attributes and commands) for individual functions. There is the ability to turn the switch on and off, maybe there is energy monitoring, maybe there is the ability to add each switch to an individual group or a scene, etc.
+
+### Descriptors
+
+For the purposes of Zigpy and Quirks we will focus on two descriptors:
```
abmantis, line 29:
```suggestion
For the purposes of Zigpy and Quirks we will focus on two descriptors: **Node Descriptor** and **Simple Descriptor**.
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+
+- Node Descriptor: desc goes here
+- Simple Descriptor: desc goes here
+
+#### Node Descriptors
```
abmantis, line 34:
```suggestion
#### Node Descriptor
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+
+A node descriptor explains some basic device attributes to Zigpy. The manufacturer code and the power type are the ones that we generally care about. In most cases you won't have to worry about this but it is good to know why it is there in case you come across it while looking at an existing quirk. Here is an example:
+`<Optional byte1=2 byte2=64 mac_capability_flags=128 manufacturer_code=4174 maximum_buffer_size=82 maximum_incoming_transfer_size=82 server_mask=0 maximum_outgoing_transfer_size=82 descriptor_capability_field=0>`
+
+#### What is a Simple Descriptor and why do we care?
```
abmantis, line 39: We don't really explain why do we care in this section (and I think we shouldn't, since that comes later).
```suggestion
#### Simple Descriptor
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+`<SimpleDescriptor endpoint=1 profile=260 device_type=1026 device_version=0 input_clusters=[0, 1, 3, 32, 1026, 1280, 2821] output_clusters=[25]>`
+
+## What the heck is a quirk?
+
+In human terms you can think of a quirk like google translate. I know it's a weird comparison but lets dig in a bit. You may only speak one language but there is an interesting article written in another language that you really want to read. Google translate takes the original article and displays it in a format (language) that you understand. A quirk is a file that translates device functionality from the format that the manufacturer chose to implement it in to a format that Zigpy and in turn ZHA understand. The main purpose of a quirk is to serve as a translator. A quirk is comprised of several parts:
+
+- Signature -- To identify and apply the correct quirk
```
abmantis, line 48:
```suggestion
- Signature - To identify and apply the correct quirk
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+- Replacement -- To allow Zigpy and ZHA to correctly work with the device
```
abmantis, line 49:
```suggestion
- Replacement - To allow Zigpy and ZHA to correctly work with the device
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+            ],
+            OUTPUT_CLUSTERS: [Ota.cluster_id, Time.cluster_id],
+        },
+    },
+}
+```
+
+### Replacement
+
+The replacement on a quirk is what we want the device to be. Remember, we said that quirks were like Google translate... you can think of the replacement like the output from Google translate. The replacement dict is what will actually be used by Zigpy and ZHA to interact with the device. The structure of `replacement` is the same as signature with 2 key differences: `models_info` is generally omitted and there is an extra element `skip_configuration` that instructs the application to skip configuration if necessary. Some manufacturers have not implemented the specifications correctly and the devices come pre-configured and therefore the configuration calls fail (non Zigbee 3.0 Xiaomi devices for instance) Here is an example:
```
abmantis, line 92:
```suggestion
The replacement on a quirk is what we want the device to be. Remember, we said that quirks were like Google translate... you can think of the replacement like the output from Google translate. The replacement dict is what will actually be used by Zigpy and ZHA to interact with the device. The structure of `replacement` is the same as signature with 2 key differences: `models_info` is generally omitted and there is an extra element `skip_configuration` that instructs the application to skip configuration if necessary. Some manufacturers have not implemented the specifications correctly and the devices come pre-configured and therefore the configuration calls fail (non Zigbee 3.0 Xiaomi devices for instance). Usually, you should not add `skip_configuration`.

Here is an example:
```

CONTRIBUTING.md
```diff
@@ -1,12 +1,379 @@
...
+
+[Zigbee Cluster Library](https://zigbeealliance.org/wp-content/uploads/2019/12/07-5123-06-zigbee-cluster-library-specification.pdf)
+
+[Zigbee Base Device Specification](https://zigbeealliance.org/wp-content/uploads/zip/zigbee-base-device-behavior-bdb-v1-0.zip)
+
+[Zigbee Primer](https://docs.smartthings.com/en/latest/device-type-developers-guide/zigbee-primer.html)
+
+## What is a device in human terms?
+
+A device is a physical object that you want to join to a Zigbee network: a light bulb, a switch, a sensor etc. The host application, in this case Zigpy, needs to understand how to interact with the device so there are standards that define how the application and devices can communicate. The device's functionality is described by several **descriptors** while the device itself contains **endpoints** and **endpoints** contain **clusters**. Zigpy needs to understand all these elements in order to correctly work with the device.
```
Adminiuga, line 17:
```suggestion
A device is a physical object that you want to join to a Zigbee network: a light bulb, a switch, a sensor etc. The host application, in this case Zigpy, needs to understand how to interact with the device so there are standards that define how the application and devices can communicate. The device's functionality is described by several **descriptors** while the device itself contains **endpoints** and **endpoints** contain **clusters**. There are two types of clusters an endpoint contains:
- **in_clusters** - are "Server" clusters in ZCL terms. These clusters control the device, e.g. a smart plug or light bulb would have an `on_off` server cluster. **in_clusters** are also the ones which also send attribute reports and/or you can read an attribute from a **in_cluster**.
- **out_clusters** - are "Client" clusters. These clusters control some other device, as "Client" cluster sends commands to "Server" cluster. For example an On/Off remote would have an `on_off` client cluster and will generate cluster commands and send those to some other device.
Zigpy needs to understand all these elements in order to correctly work with the device.
```

PR #432, opened 2020-08-04
Add quirk for Xiaomi Aqara plug_mitw01

delikat:
Adds a quirk to support the Taiwan version of the Aqara smart plug (`plug_mitw01`), based on the quirk for `plug_maus01`.


dmulcahey:
I think we can just add the model to the models info for the US plug and this will work. Can you try that locally and report back please.

delikat:
@dmulcahey Great idea. I made the change and confirmed it's working locally.

PR #436, opened 2020-08-11
Sort README.md list of device manufacturers in alphabetical order

Hedda:
Sort README.md existing list of device manufacturers supported via quirks in alphabetical order.

I wanted to also sort the devices themself by name under each manufacturer as well, however that can wait until if and when this is approved.

Yes, I have CDO. It's like OCD, but all the letters are in alphabetical order, AS THEY SHOULD BE. ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ If you are looking to make your first code contribution to this project then we 
 - https://github.com/firstcontributions/first-contributions/blob/master/README.md
 - https://github.com/firstcontributions/first-contributions/blob/master/github-desktop-tutorial.md
 
-# Devices currently supported via quirks in ZHA Device Handlers:
+# Zigbee devices not following specifications which there already are quirks for:
 
 ### CentraLite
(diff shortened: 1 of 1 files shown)


dmulcahey:
I have been thinking about completely removing this from the readme... @Adminiuga what do you think?

Adminiuga:
I think we should have a list, but it shouldn't be worded as "supported", cause I think some people think if there isnt a quirk, then the device won't work with ZHA.
We just should have an automatically generated list of vendors and models and corresponding quirks,.

Hedda:
OK so I changed the phrasing a little regarding "supported hardware". Probably should be clarified further but maybe good for now?

dmulcahey:
I think we are gonna close this in favor of something dymnamically generated as I agree with @Adminiuga comment above: "We just should have an automatically generated list of vendors and models and corresponding quirks,."

Hedda:
OK then submitted https://github.com/zigpy/zha-device-handlers/pull/470 for that one line that mentions "supported hardware".

Issue #327, opened 2020-04-08
[SUGGESTION] Rename the zha-device-handlers library and repository to "zigpy-device-handlers" for clarity and consistency


Hedda:
Please consider renaming the zha-device-handlers library and its repository to "zigpy-device-handlers" (or something else with "zigpy" in it and without "zha" in it) for better clarity and more consistency.

- https://github.com/zigpy/zha-device-handlers

As I understand it, even if zha-device-handlers library is today only used by the ZHA integration component for Home Assistant there is nothing really stopping other developers/projects using zigpy from also utilizing this zha-device-handlers library too even if it is not for the ZHA component for Home Assistant?

- https://github.com/zigpy

Note! This is indirectly related to the request of renaming bellows library and repository -> https://github.com/zigpy/bellows/issues/215 and also a request of renaming zigpy module name -> https://github.com/zigpy/zigpy/issues/351

These are only suggestions and not a must or showstoppers in any way, but IMHO 'name branding' is important even for hobby open-source organizations that are only made up by volunteers like the zigpy organization. As this project is now like the other zigpy projects part of a collection of projects which are a part of an organization shared by many members and is upstream for multiple sub-projects, using the same brand name can help everyone feel that they belong to the same team who are working for the same or similar goals.

There are now already all these libraries for zigpy where only "zha-device-handlers" and "bellows" do not follow the name-standard for the library and repository names (or 'brand' if you will):

- zigpy https://github.com/zigpy/zigpy/
  - **bellows https://github.com/zigpy/bellows**
  - zigpy-cc https://github.com/zigpy/zigpy-cc
  - zigpy-deconz https://github.com/zigpy/zigpy-deconz
  - zigpy-xbee https://github.com/zigpy/zigpy-xbee
  - zigpy-zigate https://github.com/zigpy/zigpy-zigate
  - zigpy-znp https://github.com/zha-ng/zigpy-znp
  - zigpy-deconz-parser https://github.com/zha-ng/zigpy-deconz-parser
  - **zha-device-handlers https://github.com/zigpy/zha-device-handlers**

If zha-device-handlers library and repo was renamed to "zigpy-device-handlers" then also PyPI project cand and probably should be renamed as well; 

- "zigpy-device-handlers" instead of "zha-quirks" https://pypi.org/project/zha-quirks/

[...]

PR #453, opened 2020-08-18
Update README.md based on feedback of other application integration 

Hedda:
@dmulcahey  Update README.md based on some general feedback from @pipiche38 regarding lack of documentation about integration into other projects/applications that are not ZHA in Home Assistant, for more info see: https://github.com/zigpy/zigpy/issues/452

> One more comment, I wanted to use the Quirks library but this looks ZHA specific, isn't ?

I think this is also related to issue https://github.com/zigpy/zha-device-handlers/issues/327 which is a suggestion to rename the zha-device-handlers library and repository to "zigpy-device-handlers" for clarity and consistency as it is not dependant on ZHA.

Again, the main point with both that and this is just to try to clarify that the zha-quirks library can be used as a standalone library without ZHA in HA as there is no dependence on Home Assistant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,164 +1,464 @@
-[![Build Status](https://travis-ci.org/zigpy/zha-device-handlers.svg?branch=master)](https://travis-ci.org/zigpy/zha-device-handlers)
-
 # ZHA Device Handlers For Home Assistant
 
-ZHA Device Handlers are custom quirks implementations for [Zigpy](https://github.com/zigpy/zigpy), the library that provides the [Zigbee](http://www.zigbee.org) support for the [ZHA](https://www.home-assistant.io/components/zha/) component in [Home Assistant](https://www.home-assistant.io). 
+![CI](https://github.com/zigpy/zha-device-handlers/workflows/CI/badge.svg?branch=dev)
+[![Coverage Status](https://coveralls.io/repos/github/zigpy/zha-device-handlers/badge.svg)](https://coveralls.io/github/zigpy/zha-device-handlers)
 
-ZHA device handlers bridge the functionality gap created when manufacturers deviate from the ZCL specification, handling deviations and exceptions by parsing custom messages to and from Zigbee devices. Zigbee devices that deviate from or do not fully conform to the standard specifications set by the Zigbee Alliance may require the development of custom ZHA Device Handlers (ZHA custom quirks handler implementation) to for all their functions to work properly with the ZHA component in Home Assistant. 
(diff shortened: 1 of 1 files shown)


Hedda:
Please also see the somewhat related https://github.com/zigpy/zha-device-handlers/pull/468

README.md
```diff
@@ -8,6 +8,8 @@ ZHA device handlers bridge the functionality gap created when manufacturers devi
...
 
+Note! The ZHA Device Handlers library is a standalone library and not dependent [Home Assistant](https://www.home-assistant.io), however a reference integration implementation of the ZHA Device Handlers library is available in [Home Assistant](https://www.home-assistant.io/) (a Python based open source home automation software) as part of its [ZHA integration component](https://www.home-assistant.io/integrations/zha/). The source code for that reference integration implementation is avalable in Home Assistant's "core" code under the "home-assistant/core/tree/dev/homeassistant/components/zha" directory.
```
dmulcahey, line 11: This isn't really true. there is no reference implementation in ZHA. All ZHA does is import the library... I think this is misleading. @Adminiuga thoughts?

pipiche38: I think that you should refered to 'The Device Handlers library' and not the 'ZHA Device handlers'

Adminiuga: Reference implementation using zha device handlers library?

Hedda: By "reference implementation" I no not mean to imply that there is any kind of dependency on Home Assistant. I only meant that developers can look at Home Assistant's ZHA code as a reference in order to see how zigpy and zha-quirks can be used.

That terminology wording is used as a standard for these types of scenarios, see  https://en.wikipedia.org/wiki/Reference_implementation

"In the software development process, a reference implementation (or, less frequently, sample implementation or model implementation) is the standard from which all other implementations and corresponding customizations are derived. A reference implementation often accompanies a technical standard, and demonstrates what should be considered the "correct" behaviour of any other implementation of it."

dmulcahey: I know what a reference implementation is... this is not one. it's a dependency pure and simple. Again, HA only depends on and imports this library. That's it.

dmulcahey:
Please rebase to pick up the CI change. We swapped from travis to GH Actions. Sorry for the inconvenience.

README.md
Hedda: Removed the phrasing "reference implementation".

Hedda:
@dmulcahey What do you think about resolving this pull request merge conflict with https://github.com/zigpy/zha-device-handlers/issues/577 so there would be no need to rebase?

Hedda:
PR replaced with https://github.com/zigpy/zha-device-handlers/pull/578

PR #463, opened 2020-08-23
ORVIBO switch, dimmer and motion sensor support

udoewich:
Initial support for ORVIBO in-wall switch and dimmer, as well as battery operated motion sensor.


zhaquirks/orvibo/switch.py
```diff
@@ -0,0 +1,62 @@
...
+                ],
+                OUTPUT_CLUSTERS: [Basic.cluster_id],
+            }
+        },
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
```
dmulcahey, line 52: why are you changing the device type?

zhaquirks/orvibo/dimmer.py
```diff
@@ -0,0 +1,67 @@
...
+                ],
+                OUTPUT_CLUSTERS: [Basic.cluster_id],
+            }
+        },
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                DEVICE_TYPE: zha.DeviceType.DIMMABLE_LIGHT,
```
dmulcahey, line 56: This one makes some sense because HA doesn't support dimming on switches. @Adminiuga I am ok w/ this if you are.

zhaquirks/orvibo/dimmer.py
```diff
@@ -0,0 +1,67 @@
...
+
+class T10D1ZW(CustomDevice):
+    """T10D1ZW in-wall dimmer"""
+
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=1
+        # device_version=0
+        # input_clusters=[0, 4, 5, 6, 8, 768]
+        # output_clusters=[0]>
+        MODELS_INFO: [(ORVIBO, "abb71ca5fe1846f185cfbda554046cce")],
```
dmulcahey, line 35: what a model #

zhaquirks/orvibo/switch.py
dmulcahey: The dimmable one above makes sense. This one shouldn't be changed. The device is an in wall switch not a light. Please remove this. You can override the platform in HA if you want it to be a light. If this was the only change you can remove the entire quirk. @Adminiuga do you feel any different about this?

udoewich: Strictly not necessary, but I use it as a in-wall light switch for lights. I just re-used what I had done for deconz/Phoscon.  If it's an issue, we can drop this quirk.

zhaquirks/orvibo/dimmer.py
udoewich: Yes, this was the main reason for me to look at adding a quirk, since it didn't allow setting a light level without it.

dmulcahey: Yep this one is OK w/ me because it corrects missing device functionality.

zhaquirks/orvibo/switch.py
dmulcahey: I would remove this as other users may not want this. This can be overridden w/ configuration for the device on a per user basis in configuration.yaml in HA.

udoewich: Sure, no problem. I'll make the change.

udoewich: One more thing though, before I delete this: I noticed that if I use the physical button to turn the light on/off the state is not updated in HA. Is this something that has to be added via the quirk? Turning the switch on/off via HA works fine.  Same behavior with the dimmer: changes with the physical button are not picked up by HA...

dmulcahey: Ugh, this could be 2 things: they didn’t pair correctly and they aren’t reporting changes as a result or they just flat out don’t report at all... and may need to be polled.

udoewich: Ok, I'll try to investigate. For now, I'll remove the quirk.

zhaquirks/orvibo/dimmer.py
Adminiuga: looks good to me.  Although I was expecting `zha.DeviceType.ON_OFF_LIGHT,` to be mapped to light and since it has the level cluster I was expecting it to be a level controllable light.

udoewich:
Is it possible to change the model name in the replacement block?  ORVIBO seems to use some kind of hash instead of the actual model designations for all their devices (at least the ones I've come across so far).  That makes it a bit unwieldy and not very intuitive when shown in HA...

dmulcahey:
just a lint failure left: ./zhaquirks/orvibo/dimmer.py:26:1: D400 First line should end with a period

udoewich:
@dmulcahey not sure what the issue is now; not clear to me from the error message.  I've linted the code with black in VSC before I did the pull request and it showed no issues...

dmulcahey:
```
lint run-test: commands[0] | flake8
240lint run-test: commands[1] | pydocstyle zhaquirks tests
241lint run-test: commands[2] | /home/travis/build/zigpy/zha-device-handlers/script/check_format
242would reformat /home/travis/build/zigpy/zha-device-handlers/zhaquirks/orvibo/dimmer.py
243All done! 💥 💔 💥
2441 file would be reformatted, 151 files would be left unchanged.
245ERROR: InvocationError for command /home/travis/build/zigpy/zha-device-handlers/script/check_format (exited with code 1)
```

Something in dimmer.py isn’t formatted correctly

udoewich:
Sorry, no idea what the issue is.  I tried all kinds of ways to lint and format this locally, but am not getting the error the CI script shows.  Without more info I don't know how to resolve this.

dmulcahey:
I’ll reformat it for you in the morning

PR #467, opened 2020-08-31
Refactored Philips ZHA lights to ZHAExtendedColorLight, new bulb support

TheJulianJES:
Refactored Philips ZHA lights to ZHAExtendedColorLight, added 5th gen Philips bulb support, added latest gen Bloom support.

Apparently, all new Philips devices mostly use the ZHA profile instead of ZLL. So, I think it makes sense to refactor the current "LCA003", "LCB001" lights into ZHAExtendedColorLight along with the 5th gen E27 Hue bulbs (LCA001).

The new Philips Hue Bloom (in black and white color: "929002376001", "929002375901" -- only tested the black model) do not have the "64514" input cluster. They are separated out into the "ZHAExtendedColorLight2" class. Both classes are in the same file though.

I chose the "ZHAExtendedColorLight2" name as I couldn't really think of another (useful) name. A similar name is also used with "IkeaTradfriRemote" and "IkeaTradfriRemote2". I'm open to changing this.

(In the next couple of days I'll also get access to a Philips Hue Go 2nd Gen. It's likely that this model would also be a ZHA extended color light. Will add support then. Edit: Philips Hue Go 2nd Gen lights are now supported.)


dmulcahey:
Thanks!

PR #485, opened 2020-09-12
Add support for the Osram Smart+ GU10 AC05347

Goldwing1973:
My first attempt of a quirk so i can use the AC05347 lights in ZHA.

PR #491, opened 2020-09-17
OSRAM SMART+ CLASSIC A60 TW Integration

makuser:


Adminiuga:
Can you provide a brief description of the PR, exactly what's it doing and what it fixes. Then it's good to go

makuser:
> Can you provide a brief description of the PR, exactly what's it doing and what it fixes. Then it's good to go

Well, it does what the title said and adds support for the "Osram SMART+ CLASSIC A60" bulb. There's really not much more to say other then that. The bulb was not supported previously due to a different signature then other existing devices, and I needed support for devices that I own, so I added it.?

Adminiuga:
Exactly how it wasn't supported prior to this?

makuser:
I can not tell if you are joking or just having a bad day, but if you accept #485, then you should also accept this one, as it is more or less the same, simply for a device with a different signature.

While the "base function, turning the light on and off and tuning the light color" works, this device model is not fully supported and forgets its last state after power-cycling.

Osram has a special command that needs to be called, for lights to keep their state after power-cycling. This quirk and virtually every other quirk for Osram devices in this repo does exactly that, plus sometimes some more. This is implemented in the class `OsramLightCluster`, you can read up on it if you like.

You accepted #485 a few minutes ago, which does just that, simply for a different model. You did not ask for more information there and @Goldwing1973 did not really provide any more on his/her own other than "My first attempt of a quirk so i can use the AC05347 lights in ZHA." which is more or less redundant. I mean sure, one could argue over this, of course, but the main information is "new quirk for AC05347" and that was already presented in the title.

I only gave the title, and that contains all you and everyone else needs to know about this PR.

It adds support for the SMART+ CLASSIC A60 bulb, which exists for example with an E27 socket.

Accept the PR or not, I will not argue over 'how exactly it was not supported prior to this', because sure, the base function already worked as almost all clusters were initialized properly, the vendor specific one however was not.

So if HA pulls the next version of the zhaquirks in one of the upcoming updates and my implementation of that device that I did directly in the installed library on my machine gets deleted by the new version, the `save_defaults` function is going to stop working/will be missing. When that happens the WAF drops because the light does no longer behave as it should and while under the right circumstances I might join an argument with someone on the web, I will definitely not do that with my better half, so then I will switch to zigbee2mqtt for this installation, because the device is implemented there, because @Koenkk apparently did not screw with the contributor that implemented the device there, instead of simply saying
> THANK YOU

or if you like to keep it even shorter like you did in #485:

> LGTM

petegoes:
@makuser why did you close the PR?

Issue #401, opened 2020-07-08
[Device Support Request] Hue Dimmer Switch RWL021 click count


tigattack:
**Is your feature request related to a problem? Please describe.**
ZHA doesn't report the click count for Hue Dimmer Switch button presses.

**Describe the solution you'd like**
I would like the click count to be reported as part of the event data sent when a button is pressed on the Hue Dimmer Switch.   
e.g.
```
{
    "event_type": "zha_event",
    "data": {
        "device_ieee": "ieee_here",
        "unique_id": "unique_id_here",
        "endpoint_id": 1,
        "cluster_id": 6,
        "command": "on",
        "click_count": "1",
# Like this ^
        "args": []
    },
    "origin": "LOCAL",
    "time_fired": "2020-07-08T14:27:48.530416+00:00",
    "context": {
        "id": "id_here",
        "parent_id": null,
        "user_id": null
    }
}
```
Or similar.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
```
{
  "node_descriptor": "<NodeDescriptor byte1=2 byte2=64 mac_capability_flags=128 manufacturer_code=4107 maximum_buffer_size=71 maximum_incoming_transfer_size=45 server_mask=0 maximum_outgoing_transfer_size=45 descriptor_capability_field=0>",
  "endpoints": {
    "1": {
      "profile_id": 49246,
      "device_type": "0x0830",
      "in_clusters": [
        "0x0000"
      ],
      "out_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x000c",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x000f",
        "0xfc00"
      ],
      "out_clusters": [
        "0x0019"
      ]
    }
  },
  "manufacturer": "Philips",
  "model": "RWL021",
  "class": "zhaquirks.philips.rwl021.PhilipsRWL021"
}
```

**Additional context**
zigbee2mqtt appears to support this and it's mentioned that `click_count` is supported by this device here: https://zigbee.blakadder.com/Philips_RWL021.html

dmulcahey:
Click count is done in software not by the device. We may support this in the future but currently we do not.

[...]

PR #501, opened 2020-09-23
[RWL021 & RWL020] Add Multiple Click Support

evanreichard:
Software implementation for multiple press support (Double / Triple / Quadruple / Quintuple) for the Philips Hue Dimmer Switch. (RWL021 & RWL020)

This capability uses a 500ms timeout (which is reset every time we press). Currently this is statically defined, as I'm not sure theres any way to change configuration on the ZHA side. Downside is that this adds a 500ms delay to any "press" `press_type`. However, in practice I didn't notice it.

Also possible to add a `click_count` as requested here: https://github.com/zigpy/zha-device-handlers/issues/401, but this should solve the "why" to that request, so it's likely not necessary.

### DIFF
--- a/zhaquirks/philips/__init__.py
+++ b/zhaquirks/philips/__init__.py
@@ -181,7 +181,7 @@ class PhilipsRemoteCluster(CustomCluster):
             # Override PRESS_TYPE
             event_args[PRESS_TYPE] = press_type
 
-            action = "{}_{}".format(button, press_type)
+            action = f"{button}_{press_type}"
             self.listener_event(ZHA_SEND_EVENT, action, event_args)
 
         # Derive Multiple Presses
--- a/zhaquirks/philips/__init__.py
+++ b/zhaquirks/philips/__init__.py
@@ -165,8 +165,7 @@ class PhilipsRemoteCluster(CustomCluster):
 
         def send_press_event(click_count):
             _LOGGER.debug(
-                "PhilipsRemoteCluster - send_press_event click_count: [%s]",
-                click_count,
+                "PhilipsRemoteCluster - send_press_event click_count: [%s]", click_count
             )
             if click_count == 1:
                 press_type = "press"
--- a/zhaquirks/philips/__init__.py
+++ b/zhaquirks/philips/__init__.py
@@ -1,5 +1,8 @@
 """Module for Philips quirks implementations."""
 import logging
+import time
+
+from threading import Timer
 
 from zigpy.quirks import CustomCluster
 import zigpy.types as t
@@ -12,6 +15,10 @@ from ..const import (
     COMMAND_ID,
     DIM_DOWN,
     DIM_UP,
+    DOUBLE_PRESS,
+    TRIPLE_PRESS,
+    QUADRUPLE_PRESS,
+    QUINTUPLE_PRESS,
     LONG_PRESS,
     LONG_RELEASE,
     PRESS_TYPE,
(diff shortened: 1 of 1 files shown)


dmulcahey:
First off thanks for the contribution!

We can't accept this as is. All of the code needs to be safe to use in the event loop so you will need to convert the Timer usage into asyncio equivalents. You will need something similar to this: https://stackoverflow.com/a/45430833

zhaquirks/philips/__init__.py
```diff
@@ -1,5 +1,8 @@
 """Module for Philips quirks implementations."""
 import logging
+import time
+
+from threading import Timer
```
dmulcahey, line 5: We can't use this in asyncio

dmulcahey (changes requested):
See comments above.

evanreichard:
@dmulcahey Thanks for pointing me in the right direction! I've switched over to `asyncio`. In the process I created a generalized `ButtonPressQueue` class that could potentially be used in future software based multi-press implementations.

Let me know if there's anything else you'd like tweaked.

zhaquirks/philips/__init__.py
```diff
@@ -108,5 +162,31 @@ def handle_cluster_request(self, tsn, command_id, args):
...
+            event_args[PRESS_TYPE] = press_type
+
+            action = "{}_{}".format(button, press_type)
+            self.listener_event(ZHA_SEND_EVENT, action, event_args)
+
+        # Derive Multiple Presses
+        if press_type == "press":
+            self.button_press_queue.press(send_press_event)
+        else:
+            action = "{}_{}".format(button, press_type)
```
Adminiuga, line 191:
```suggestion
            action = f"{button}_{press_type}"
```

zhaquirks/philips/__init__.py
```diff
@@ -108,5 +162,31 @@ def handle_cluster_request(self, tsn, command_id, args):
...
+                press_type = "triple_press"
+            elif click_count == 4:
+                press_type = "quadruple_press"
+            elif click_count == 5:
+                press_type = "quintuple_press"
+
+            # Override PRESS_TYPE
+            event_args[PRESS_TYPE] = press_type
+
+            action = "{}_{}".format(button, press_type)
```
Adminiuga, line 184:
```suggestion
            action = f"{button}_{press_type}"
```

zhaquirks/philips/__init__.py
```diff
@@ -75,6 +98,34 @@ async def bind(self):
...
+        self._callback(self._click_counter)
+
+    def press(self, callback):
+        """Process a button press."""
+        self._callback = callback
+        now_ms = time.time() * 1000
+        if now_ms - self._ms_last_click > self._ms_threshold:
+            self._click_counter = 1
+        else:
+            self._task.cancel()
```
Adminiuga, line 123: Just to confirm, there's no chance we could get here and task be already done?

Adminiuga:
Overall LGTM

@dmulcahey may be should make this part of global clusters, cause people were asking for multiple presses support for xiaomis too? Or may be should implement it at ZHA for basic client channel 🤔

zhaquirks/philips/__init__.py
evanreichard: No I don't believe so.

dmulcahey (approved):

dmulcahey:
I think ZHA may be challenging because some devices support this natively and adding this to those may be problematic. I see a potential issue here too (or maybe a breaking change) these remotes send "short release" events. I am assuming that in the multi press scenarios those are now sent multiple times. This may be confusing if someone is using them in events already. It is something we can track in the beta and adjust if necessary. 

@evanreichard thanks for the contribution!!

Adminiuga:
Fixes https://github.com/zigpy/zha-device-handlers/issues/401

PR #513, opened 2020-10-06
Fix XBee manufacturer and model

Shulyaka:
1. Added model and manufacturer on device level for XBee
2. Removed model and manufacturer on endpoint level since I couldn't find where it is used.

This is required for home-assistant/core#41311 to actually work.


Shulyaka:
Should we change manufacturer to Digi by the way?

Adminiuga (approved):

PR #517, opened 2020-10-07
Add Danfoss specific configuration and attributes according to manufacturer

zoic21:
Add Danfoss specific configuration and attributes according to manufacturer

### DIFF
--- a/zhaquirks/danfoss/__init__.py
+++ b/zhaquirks/danfoss/__init__.py
@@ -1,0 +1,2 @@
+"""Module for Sinope quirks implementations."""
+DANFOSS = "Danfoss"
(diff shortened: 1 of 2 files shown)


Adminiuga:
Looks good, nothing appears to be missing on the 1st glance.

Have you tried it in your environment and were able to write the 0x4000 attribute with the quirk applied?

zoic21:
I try it's ok for open windows detection. For other no error on write but do nothing and for read attributes I got "Status.UNSUPPORTED_ATTRIBUTE". Manufacturer R&D search the root cause because in their labs it's ok (maybe it's because I have wrong firmware...). I should have an answer tomorrow morning

Adminiuga (approved):
lgtm

Issue #399, opened 2020-07-07
[Device Support Request] Philips Hue Power Cycle Attribute


kurtwarwick-new:
Hi 

It would be great to have the Philips Hue Power Cycle attribute added. This allows for the bulb to remain powered on, off or set to previous settings when powered on.

The vendor attribute is `0x4003` and falls part of the `0x0006` (on_off) cluster. The possible values (enum8) are 

```
Off = 0x00
On = 0x01
Previous = 0xff
```

The device signature for the ambient color bulb is as follows:

```
{
  "node_descriptor": "<NodeDescriptor byte1=1 byte2=64 mac_capability_flags=142 manufacturer_code=4107 maximum_buffer_size=71 maximum_incoming_transfer_size=45 server_mask=0 maximum_outgoing_transfer_size=45 descriptor_capability_field=0>",
  "endpoints": {
    "11": {
      "profile_id": 49246,
      "device_type": "0x0210",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0300",
        "0x1000",
        "0xfc01"
      ],
      "out_clusters": [
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [
        "0x0021"
      ],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "Philips",
  "model": "LCT015",
  "class": "zigpy.device.Device"
}
```

The device signature for the dimmable white bulb is as follows:

```
{
  "node_descriptor": "<NodeDescriptor byte1=1 byte2=64 mac_capability_flags=142 manufacturer_code=4107 maximum_buffer_size=71 maximum_incoming_transfer_size=45 server_mask=0 maximum_outgoing_transfer_size=45 descriptor_capability_field=0>",
  "endpoints": {
    "11": {
      "profile_id": 49246,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x1000"
      ],
      "out_clusters": [
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [
        "0x0021"
      ],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
[...]

PR #520, opened 2020-10-08
Philips power on color temperature and brightness

Shulyaka:
Add attributes for power on color temperature and power on brightness configuration.
Attribute numbers are taken from this comment: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1055#issuecomment-559993258
Tested on Philips LCT003 (`ZLLExtendedColorLight`). `ZHAExtendedColorLight` and `ZHAExtendedColorLight2` are not tested but they should be the same for what I know.
Apparently, to set the power on color itself one could use the Color (0x0300) cluster, current_x (0x0003) and current_y (0x0004) attributes with manufacturer code 0x100b (4107).

### DIFF
--- a/zhaquirks/philips/__init__.py
+++ b/zhaquirks/philips/__init__.py
@@ -3,7 +3,8 @@ import logging
 
 from zigpy.quirks import CustomCluster
 import zigpy.types as t
-from zigpy.zcl.clusters.general import Basic, OnOff
+from zigpy.zcl.clusters.general import Basic, LevelControl, OnOff
+from zigpy.zcl.clusters.lighting import Color
 
 from ..const import (
     ARGS,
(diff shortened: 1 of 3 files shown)


Adminiuga:
There were some changes in zigpy and we are no longer copying attr dict, but using manufacturer_attributes dict instead. See some other quirk for an example. On the phone and can't provide direct links right now

Shulyaka:
Ok, will do that. I just copied the approach from the existing code for PhilipsOnOffCluster.

Adminiuga:
ah, effing philips, could be because https://github.com/zigpy/zha-device-handlers/issues/399#issuecomment-678631540
can you confirm?

Shulyaka:
Looks like that...
I've tested it now, it doesn't work with manufacturer_attributes.

Adminiuga:
Thanks for testing.

TheAlphaLaw:
@Shulyaka So is power on to previous state not possible? It's writable in deCONZ for my hue lights, but doesn't function.

Adminiuga:
1. is it writtable in ZHA?
2. Device signature as reported by ZHA
3. Debug log of you writing this attribute

TheAlphaLaw:
It's writable in deCONZ, can get more info later today.

Shulyaka:
As per https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1055#issuecomment-559993258, it should be cluster OnOff 0x0006, attribute PowerOn OnOff 0x4003, value Previous (0xff)

TheAlphaLaw:
Correct, but when you write "Previous" it writes the data but does not change the behavior of the light. The light will reset temp/brightness on power loss.

Adminiuga:
you write `255`, not `Previous`

TheAlphaLaw:
255 would turn the light on 100% brightness on reset. I'm looking for previous temp/brightness on reset.

Adminiuga:
you have to do it for all clusters. And we have to assume you do it correctly. For the light that is on the network and is not powered off.

PR #530, opened 2020-10-18
Support for Tuya-based NEO Siren (closes #444)

### LINKED ISSUE #444: [Device Support Request] Tuya Siren
Could we please have support for this device added please?

image

{
  "node_descriptor": "<NodeDescriptor byte1=2 byte2=64 mac_capability_flags=128 manufacturer_code=0 maximum_buffer_size=82 maximum_incoming_transfer_size=255 server_mask=11264 maximum_outgoing_transfer_size=255 descriptor_capability_field=0>",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0003"
      ],
      "out_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TYST11_d0yu2xgi",
  "model": "0yu2xgi",
}

FrnchFrgg:
Also implement a base class to map non-standard tuya data to attributes.
Custom tuya clusters can derive from that class and any defined
attribute on the cluster will be automatically mapped.

Standard ZHA clusters will be implemented next, by listening on attributes changes and modifying them on the Tuya cluster.

### DIFF
--- a/Contributors.md
+++ b/Contributors.md
@@ -21,3 +21,4 @@
 - [Gleb Sinyavskiy](https://github.com/zhulik)
 - [Michael Thingnes](https://github.com/thimic)
 - [Piotr Mis](https://github.com/piomis)
+- [Julien Rivaud](https://github.com/FrnchFrgg)
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ If you are looking to make your first code contribution to this project then we 
 
 ### Tuya-based
 - [TS0601 switch](https://zigbee.blakadder.com/Lerlink_X701A.html): Tuya-based 1-gang switches with neutral (e.g. Lerlink, Lonsonho)
+- [Neo Siren](https://zigbee.blakadder.com/Neo_NAS-AB02B0.html): Tuya-based alarm siren with temperature and humidity sensors
 
 # Configuration:
 
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -51,6 +51,75 @@ class TuyaManufCluster(CustomCluster):
     }
 
 
+class TuyaManufClusterAttributes(TuyaManufCluster):
+    """Manufacturer specific cluster for Tuya converting attributes <-> commands."""
+
+    def handle_cluster_request(self, tsn: int, command_id: int, args: Tuple) -> None:
+        """Handle cluster request."""
+        if command_id not in (0x0001, 0x0002):
+            return super().handle_cluster_request(tsn, command_id, args)
+
+        tuya_cmd = args[0].command_id
+        tuya_value = args[0].data[1:]  # first uint8_t is length
+
+        _LOGGER.debug(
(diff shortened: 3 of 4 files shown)


FrnchFrgg:
This is preliminary work on issue #444. I am going to make Tox happy before removing the WIP token even if I do not like when things do not line up.

FrnchFrgg:
This MR makes the tuya siren available as an On/Off device, a Temperature sensor and an humidity sensor. Other attributes (melody, sound duration, automatic alarms, etc.) are available as attributes on the Tuya cluster, and are persistent.
A followup might implement IAS warning device clusters, but that would entail saving the current melody, setting it to a specific "alarm-like" melody, probably changing the maximum duration, then activating the alarm, and reverting back the settings to what they were before, which might be overkill for a zhaquirks implementation.

meremortals70:
Julien

Thanks for all your work on this?  When is it likely to be included in a GA release?  I just upgraded to 0.117.0 and re-paired the siren with no change.

FrnchFrgg:
I have no idea about that. The work is not yet included in zha-device-handlers, so we will have to wait for the review and merge, then wait for a release of zha-device-handlers, then for the update of zhaquirks to that release in home assistant. During that time, you can use the portainer method described in the README of this project to just update the contents of the tuya subdirectory of zhaquirks. It will not survive an update of HA, but you can do it after each update, until the changes are included directly.

BTW, it seems that pairing the valve again is not needed. In my case at least, HA picked up the changes directly. I suppose that re-pairing is needed when there are changes to the binding methods, which my implementation didn't need.

meremortals70:
Awesome thanks for the explanation, helps a lot to understand the process.

Adminiuga (approved):
LGTM

@FrnchFrgg can confirm you tested this and it works?

FrnchFrgg:
@Adminiuga Yes, I have the device, I tested it and it works. It exposes a zha humidity sensor, a zha temperature sensor, and a zha on/off switch. The other features of the device are accessed by using the exposed attributes on the tuya-specific manufacturer.

Note that my particular device starts after a reset with automatic triggering of the alarm based on temperature and humidity ranges (all settable through attributes), which is less than useful since the reported temperature is quite imprecise (between 1.5°C and 2°C below the real temperature as measured by other devices).

An user would typically have to disable that automatic triggering by setting the corresponding `enable_*` attributes to `0`. The setting sticks even across loss of power (without the energy cells) and is only reverted with a full reset of the device which is the only way to put the device into pair mode.

As I found that deconz seems to know of a calibration offset parameter for other tuya devices, I want to try if my siren reacts to that command, and advertise the attribute accordingly. But that would be better as a followup PR I think, since my leisure time is running out. Another followup would be sent if/when I figure out how to ask the device for the current value of a parameter (currently I only know the last value sent by the device on its own or in response to a `set` command).

PR #539, opened 2020-10-25
Add support for Terncy Awareness switch and Terncy Knob

ilarrain:
Awareness is a button, 2 motion sensors, illuminance and temperature sensor. https://www.terncy.com/en-us/awarenessswitch/

Knob is a button and rotation knob. https://www.terncy.com/en-us/smartdial/

### DIFF
--- a/zhaquirks/terncy/pp01.py
+++ b/zhaquirks/terncy/pp01.py
@@ -1,0 +1,103 @@
+"""Device handler for Terncy awareness switch."""
+from zigpy.profiles import zha
+from zigpy.quirks import CustomDevice
+from zigpy.zcl.clusters.general import (
+    Basic,
+    Identify,
+    Ota,
+    PollControl,
+    PowerConfiguration,
+)
+from zigpy.zcl.clusters.measurement import (
+    TemperatureMeasurement,
+    IlluminanceMeasurement,
+    OccupancySensing,
+)
+
+from zhaquirks import DoublingPowerConfigurationCluster
+
+from . import (
+    IlluminanceMeasurementCluster,
+    TemperatureMeasurementCluster,
+    TerncyRawCluster,
+    OccupancyCluster,
+    MotionClusterLeft,
+    MotionClusterRight,
+    BUTTON_TRIGGERS,
+)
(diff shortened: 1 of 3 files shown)


tube0013:
I have both these devices, and just tested this quirk, it works great with both - thanks!

ilarrain:
Great! It took some debugging and research due to my limited python (programming in general, tbh), Zigbee, and home-assistant knowledge.

tube0013:
In the events I see a counter for for each type of button click, do you know when this gets reset? seems to just keep going up.

ilarrain:
> In the events I see a counter for for each type of button click, do you know when this gets reset? seems to just keep going up.

It doesn't seem to reset. It is reported directly by the button, and counts the times the button have been activated that particular number of times. The button registers from 1 to 9 clicks, but I had to limit it to 5, because it wasn't displayed correctly in the UI.

zhaquirks/terncy/pp01.py
```diff
@@ -0,0 +1,103 @@
...
+        self.motion_right_bus = Bus()
+        self.occupancy_bus = Bus()
+        super().__init__(*args, **kwargs)
+
+    signature = {
+        #  <SimpleDescriptor endpoint=1 profile=260 device_type=496
+        #  device_version=0
+        #  input_clusters=[0, 1, 3, 32, 1024, 1026, 1030, 64716]
+        #  output_clusters=[25]>
+        MODELS_INFO: [("Xiaoyan", "TERNCY-PP01"), (None, "TERNCY-PP01")],
```
Adminiuga, line 56: @ilarrain any reason why you are not defining manufacturer for the second tuple in `MODELS_INFO`???

ilarrain: Yes, the reason is that I have 3 Terncy devices, 2 Awareness Switch and 1 knob, the first ones correctly report the manufacturer as Xiaoyan, and the latter one (the knob) just reports None polling the attribute on the Basic cluster (and zha reports as unk_manufacturer). I assumed "TERNCY-XXX" is sufficiently unique.

PR #543, opened 2020-10-25
Add model info to Osram A19TunableWhite quirk (Fixes #536)

### LINKED ISSUE #536: [BUG] Mismatched Quirk: A19TunableWhite without model info (and Osram RGBW Outdoor Flex)
**Describe the bug**
Someone mentioned on the Hass Discord in #zigbee that a OSRAM Lightify Outdoor Flex RGBW always got a wrong quirk: A19TunableWhite. This quirk removes the colorwheel (since #488) (which is needed for a RGBW led strip).
This happens because the quirk (https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/osram/a19twhite.py) never included any "MODELS_INFO" and the signature matches exactly between those two devices.
It's also possible that further OSRAM devices also match and thus don't show the color wheel.
A solution to this seems like to add the proper model (and manufacturer info) to the quirk.
(Or is there a reason this quirk doesn't include any model info at all? If not, perhaps a test that every quirk has at least one model in MODELS_INFO would be a good idea?)

**To Reproduce**
1. Bind a OSRAM Lightify Outdoor FLEX RGBW to ZHA
2. Try to change the color using Hass
3. See color wheel missing

**Expected behavior**
The OSRAM A19TunableWhite quirk doesn't apply for the RGBW light.

**Additional context**
(The user on Discord with this issue wanted to submit an issue, but it seems like that didn't happen after multiple days.)

TheJulianJES:
Fixes #536 by also including MODELS_INFO for the Osram A19TunableWhite quirk (based on the device signature given here: https://github.com/zigpy/zha-device-handlers/issues/536#issuecomment-716198388)

### DIFF
--- a/zhaquirks/osram/a19twhite.py
+++ b/zhaquirks/osram/a19twhite.py
@@ -13,8 +13,15 @@ from zigpy.zcl.clusters.general import (
 from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
 from zigpy.zcl.clusters.lighting import Color
 
-from . import OsramLightCluster
-from ..const import DEVICE_TYPE, ENDPOINTS, INPUT_CLUSTERS, OUTPUT_CLUSTERS, PROFILE_ID
+from . import OsramLightCluster, OSRAM
+from ..const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+    MODELS_INFO,
+)
 
 
 class OsramColorCluster(CustomCluster, Color):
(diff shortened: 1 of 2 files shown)


zhaquirks/osram/a19twhite.py
```diff
@@ -30,6 +37,7 @@ class A19TunableWhite(CustomDevice):
         # <SimpleDescriptor endpoint=3 profile=260 device_type=258
         # device_version=2 input_clusters=[0, 3, 4, 5, 6, 8, 768, 64527]
         # output_clusters=[25]>
+        MODELS_INFO: [(OSRAM, "LIGHTIFY A19 Tunable White")],
```
Adminiuga, line 40:
```suggestion
        MODELS_INFO: [(OSRAM, "LIGHTIFY A19 Tunable White"), (OSRAM, "LIGHTIFY RT Tunable White")],
```

Adminiuga: the OSRAM recessed lighting has exactly the same signature, just different model. Lets add this model too and maybe rename the quirk class to slightly more neutral, eg just `OsramTunableWhite`? Don't really bother me, just recessed lights are not A19 types :)

Adminiuga (approved):
LGTM, once RT Tunable white added to the model

Issue #519, opened 2020-10-07
[Device Support Request] Iris IL07_1 (GreatStar iMagic 1117-S) Motion Sensor not reporting battery status


coyoteT:
**Is your feature request related to a problem? Please describe.**
I have a few Iris IL07_01 Motion/Temp/Humidity sensors and none of them report the battery level.

**Describe the solution you'd like**
I would like to be able to see what percentage the batteries have remaining.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
````
{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4667, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0402",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0020",
        "0x0402",
        "0x0405",
        "0x0500",
        "0x0b05",
        "0xfc01",
        "0xfc02"
      ],
      "out_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "iMagic by GreatStar",
  "model": "1117-S",
  "class": "zigpy.device.Device"
}
````

**Additional context**
When going into manage cluster -> power configuration, none of the battery_percentage_remaining options report a number, they all display "none"

PR #549, opened 2020-10-29
Added support for imagic by greatstar closes (#519)

mjumnito:
Support for imagic by greatstar open/close remotes


mjumnito:
created by mistake

PR #551, opened 2020-10-30
Add Quirk for Ecolink 4655BC0-R for Battery

tube0013:
Add CustomPowerConfigurationCluster.

Use MIN_VOLTS = 2.1 and MAX-VOLTS = 3.0 until we know better.


tube0013:
I'm not sure why it is failing, when I run black locally it is happy and has no changes.

Adminiuga:
Could be different black versions. Need to upgrade to newer black 0.20.x

PR #554, opened 2020-10-31
Add support for GL-C-007 2ID version

jedi7:
GL-C-007 2ID have issue when on destination endpoint 15 is used source endpoint other than 1.

Quirk is probably not perfect, but at least both lights are working (on/off, dimmable, rgb selection)
Issue: CCT light part have RGB selection and not color temperature selection.

### DIFF
--- a/zhaquirks/gledopto/glc007_2id.py
+++ b/zhaquirks/gledopto/glc007_2id.py
@@ -1,7 +1,5 @@
 """GLEDOPTO GL-C-007 2ID device.."""
-import binascii
 import logging
-import math
 from typing import Optional, Union
 
 import zigpy.zcl.foundation as foundation
@@ -60,6 +58,7 @@ class BasicCluster(Basic, CustomCluster):
             expect_reply=expect_reply,
         )
 
+
 class OnOffCluster(OnOff, CustomCluster):
     """GLEDOPTO GL-C-007 2ID on/off cluster."""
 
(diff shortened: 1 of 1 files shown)


Adminiuga:
What is the problem that it is fixing? 
What do you mean by "GL-C-007 2ID have issue when on destination endpoint 15 is used source endpoint other than 1." ?
What radio are you using?
Can you provide logs?
Cause I don't understand how this fixes the issue, as it effectively kills any ZCL cluster general command like Read Attributes, Write Attributes etc

jedi7:
It fix problem with second light (on endpoint 15), see logs: https://pastebin.com/rA4x5uBp
The issue is following: when I want to communicate with the light on endpoint 15, then I must use source endpoint 1. Unless it fails (see pastebin)
When I use the source endpoint 1, then it is successful, see: https://pastebin.com/rT44yfku

Radio: TI CC2531 USB CDC

If you have another suggestion how to solve this problem, then I will be happy.

Adminiuga:
Ah, ok. Yes, there's a problem, but the problem is with `zigpy_cc` library.  I'd recommend switching to `zigpy_znp` library for TI radios and give it a go. Pretty sure that would fix the problem without requiring this quirk.

jedi7:
You are right. I did not believed, but tried to force the znp in HA and now it works.
Many thanks!

PR #556, opened 2020-11-01
Hue White Dimmable - Expose PhilipsLevelControlCluster

glassbase:
In order for PhilipsOnOff power restore to work, you also need to be able to set Level Control cluster 0x0008, PowerOn Level attribute 0x4000 to value 255 (0xff).

I edited the quirk to add the Level Control cluster. When I read/get to attribute ox4000, it is set as 254 as default. This causes the bulb to restore to full brightness after power restore. If set as 255, it restores to previous brightness. 

Findings based on this https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1055#issuecomment-559993258

### DIFF
--- a/zhaquirks/philips/zlldimmablelight.py
+++ b/zhaquirks/philips/zlldimmablelight.py
@@ -22,8 +22,12 @@ from zhaquirks.const import (
     PROFILE_ID,
     MODELS_INFO,
 )
-from zhaquirks.philips import PHILIPS, PhilipsOnOffCluster
 
+from zhaquirks.philips import (
+    PHILIPS,
+    PhilipsLevelControlCluster,
+    PhilipsOnOffCluster
+)
 
 class ZLLDimmableLight(CustomDevice):
     """Philips ZigBee LightLink dimmable bulb device."""
@@ -73,7 +77,7 @@ class ZLLDimmableLight(CustomDevice):
                     Groups.cluster_id,
                     Scenes.cluster_id,
                     PhilipsOnOffCluster,
-                    LevelControl.cluster_id,
+                    PhilipsLevelControlCluster,
                     LightLink.cluster_id,
                 ],
                 OUTPUT_CLUSTERS: [Ota.cluster_id],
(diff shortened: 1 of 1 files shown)


zhaquirks/philips/zlldimmablelight.py
```diff
@@ -22,8 +22,12 @@
     PROFILE_ID,
     MODELS_INFO,
 )
-from zhaquirks.philips import PHILIPS, PhilipsOnOffCluster
 
+from zhaquirks.philips import (
+    PHILIPS,
+    PhilipsLevelControlCluster,
+    PhilipsOnOffCluster
+)
```
Adminiuga, line 31:
```suggestion


```

zhaquirks/philips/zlldimmablelight.py
```diff
@@ -22,7 +22,12 @@
     PROFILE_ID,
     MODELS_INFO,
 )
-from zhaquirks.philips import PHILIPS, PhilipsOnOffCluster
+
+from zhaquirks.philips import (
+    PHILIPS,
+    PhilipsLevelControlCluster,
+    PhilipsOnOffCluster
+)
```
TheJulianJES, line 30: ``from zhaquirks.philips import PHILIPS, PhilipsLevelControlCluster, PhilipsOnOffCluster``

glassbase: I put them on separate lines the same as zllextendedvolorlight.py
https://github.com/zigpy/zha-device-handlers/blob/6682592d6b3935e495503ab20177312b8dd6b55a/zhaquirks/philips/zllextendedcolorlight.py#L26-L31

TheJulianJES: PhilipsColorCluster not being needed here makes it only 3 (not 4) imports. Hence the shorter form should be used. That’s also why the lint fails.

Adminiuga: Lint also tells you what it wants to reformat, e.g. https://travis-ci.org/github/zigpy/zha-device-handlers/jobs/740724957#L240

But just use black and  install pre-commit
```
pip3 install -U pre-commit
pre-commit install
```

Adminiuga (approved):

PR #568, opened 2020-11-07
Switch CI to GitHub Actions

dmulcahey:
This PR adds a CI workflow leveraging GH Actions so we can migrate from TravisCI to GH Actions. It also updates the precommit configuration.


requirements_test_all.txt
```diff
@@ -3,19 +3,19 @@
...
-codecov==2.0.15
-coveralls==1.2.0
-flake8-docstrings==1.3.1
-flake8==3.7.8
-mock-open==1.3.1
-mypy==0.720
-pre-commit==1.18.3
-pydocstyle==4.0.1
-pylint==2.3.1
+black==20.8b1
```
Adminiuga, line 6: I don't think black, I sort, flake8 dependencies could be omitted here, as pre-commit would take care to install the correct one. And allows you to keep those up-to-date in a single place -- pre-commit config

PR #575, opened 2020-11-09
Update thermostat.py

zoic21:
Add manufacturer code.

How can we auto-set manufacturer code on manufacturer specific attribute ?

### DIFF
--- a/zhaquirks/danfoss/thermostat.py
+++ b/zhaquirks/danfoss/thermostat.py
@@ -27,6 +27,8 @@ from ..const import (
     PROFILE_ID,
 )
 
+MANUFACTURER = 0x1246
+
 
 class DanfossThermostatCluster(CustomCluster, Thermostat):
     """Danfoss custom cluster."""


Adminiuga:
is it different from one reported by NodeDescriptor?

zoic21:
Indeed it is the same. But in the zigbee documentation they say it's a code (a kind of protection therefore). So I think it may be different from that of the node descriptor, right?

In fact, I don't really understand the meaning of this code if it is in the information that the node transmits.

Adminiuga:
zigpy would use manufacturer_code from the node descriptor for all commands related to manufacturer_specific attributes

zoic21:
Ok I understand, when it's in manufacturer_specific I don't need to specify that it's a manufacturer specific attribute. Good to known.

Adminiuga:
in other words device should have a quirk and for any commands and attributes defined as being `manufacturer_server_commands`, `manufacturer_client_commands` or `manufacturer_attributes` like in https://github.com/jeedom/zha-device-handlers/blob/4d54767698fd3324efd18d437afb1dec0c2ef250/zhaquirks/danfoss/thermostat.py#L36 zigpy would use manufacturer code automatically. 

If the manufactuer_code does not match the one in the node descriptor, then quirk could force it by overriding the `manufacturer_id_override` attribute in the custom device class https://github.com/zigpy/zigpy/blob/e83733847e848c9a52ef76e4a3026add640c72ec/zigpy/device.py#L35

And application can always supply `manufacturer_id` code for most of the requests

PR #580, opened 2020-11-10
Terncy: Add command schemas and filter repeated click events

ilarrain:
Fixes command data deserialization and repeated click events generated by some devices (for single action).

### DIFF
--- a/zhaquirks/terncy/__init__.py
+++ b/zhaquirks/terncy/__init__.py
@@ -1,7 +1,9 @@
 """Module for Terncy quirks."""
+from collections import deque
 import math
 
 from zigpy.quirks import CustomCluster
+import zigpy.types as t
 from zigpy.zcl.clusters.measurement import (
     IlluminanceMeasurement,
     TemperatureMeasurement,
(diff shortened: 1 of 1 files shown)


zhaquirks/terncy/__init__.py
```diff
@@ -120,14 +122,26 @@ class TerncyRawCluster(CustomCluster):
...
+        super().__init__(*args, **kwargs)
+        self._last_clicks = deque(maxlen=10)
 
     def handle_cluster_request(self, tsn, command_id, args):
         """Handle a cluster command received on this cluster."""
-        args = list(args)
         if command_id == 0:  # click event
             count = args[0]
             state = args[1]
+            if (state, count) in self._last_clicks:
```
dmulcahey, line 141: so unless this device sends an insane amount of traffic I struggle to see how this doesn't swallow interactions. Tell me, what happens in this scenario... using the exact same action ex: single press:

- press the button once
- wait like 10 seconds 
- press the button again

The way I read this the first press will insert this into the queue and send the event. If I understand what is happening it looks like the 2nd single press would be swallowed until 10 events (that are not the same) have been sent by the device.

ilarrain: The count attribute is a rolling counter (1 to 254) of the times that particular action has been called. Each number of clicks has its own counter (from single to nonuple clicks). The first single click gets counter=1, the second single click counter=2, the first triple click counter=1, the third single click counter=3, and so on.
The thing is, this devices tend to repeat the previous actions, especially when newer interactions get called, but the counter identifies correctly the action being repeated. For instance, If you single click 2 times, and then double click, you might get about triple the amount of events reported (single, count 1; single, count 2; single, count 1; double, count 1; single, count 1; ....), so the automation get called multiple times, and in no logical order (except the first event for that particular click and counter).
So, I implemented a filter.
I couldn't just expect the counter to increase, as swapping batteries, HA reboots or signal losses would mean the counters would became unsynchronized.
So, after multiple attempts, I implemented _last_clicks as a deque that stores the last 10 tuples (clicks, count) of actions reported, in order to dismiss an action if it gets repeated.
The chance of the filter swallowing events is pretty slim, and I've been using the integration (with the pull request applied) for 3 weeks without issues (no repeated, nor missed actions that I have noticed).

Sorry for the long explanation. I hope it makes sense.

dmulcahey: That makes sense. No need to apologize! I appreciate the long explanation.

Adminiuga (approved):

tube0013:
@ilarrain I was able to pull a FW image for the Smart Dial from sniffing the traffic from the terncy hub. and worked with @walthowd to get into the form of an OTA file. If you're dial is not upgraded to the 26 file would you be willing to test it? mine got updated in the process of getting the FW.

Sorry for posting in the closed PR. you can ping me on discord, I'm tube there.

PR #609, opened 2020-11-29
Set a name on automatically created groups

dmulcahey:
Fixes: https://github.com/home-assistant/core/issues/39572


Adminiuga (approved):
Lgtm

MattWestb:
Sorry for reporting little late but i was messing my network completely then changing coordinator and updating the firmware and must rejoining all devices (Zigbee 3 :-(( )

The old IKEA motion sensor is ZLL and have one random groupe that is changed after device reset.

So ZHA was doing one (wrong 3 then i was messing 2 times) grupes with un friendly names like this:
ZHAGROUPE01

I think somthing is not working like it was intended to do with the fix :-(

Its no large problem for my, i stopping HA and ending the name in the DB if its necessary.

PR #615, opened 2020-12-04
Add quirk for Bosch isw-zdl1-wp11g motion sensor

nasa931:
This implements a quirk to support the Bosch isw-zdl1-wp11g motion sensor.
Add battery_percent_remaining attribute to power_configuration cluster.

This was modeled directly from the existing quirk for the Bosch ISW-ZPR1-WP13 model - only the MAX_VOLTS and MIN_VOLTS values are different because the WP11G model holds more batteries and operates on a higher voltage than the WP13 model.

### DIFF
--- a/zhaquirks/bosch/isw-zdl1-wp11g.py
+++ b/zhaquirks/bosch/isw-zdl1-wp11g.py
@@ -1,0 +1,71 @@
+"""Device handler for Bosch ISWZDL1WP11G motion sensor."""
+from zigpy.profiles import zha
+from zigpy.quirks import CustomDevice
+from zigpy.zcl.clusters.general import Basic, Identify, Ota, PollControl
+from zigpy.zcl.clusters.homeautomation import Diagnostic
+from zigpy.zcl.clusters.measurement import TemperatureMeasurement
+from zigpy.zcl.clusters.security import IasZone
+
+from zhaquirks import PowerConfigurationCluster
+
+from . import BOSCH
+from ..const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+
+
+class BoschWP11GPowerConfiguration(PowerConfigurationCluster):
+    """Bosch power configuration cluster for ISWZDL1WP11G motion sensor."""
+
+    MAX_VOLTS = 6.0
+    MIN_VOLTS = 3.0
+
+
+class ISWZDL1WP11G(CustomDevice):
+    """Custom device representing Bosch ISWZDL1WP11G motion sensor."""
+
+    signature = {
+        #  <SimpleDescriptor endpoint=1 profile=260 device_type=1026
+        #  device_version=0
+        #  input_clusters=[0, 1, 3, 32, 1026, 1280, 2821]
+        #  output_clusters=[25]>
+        MODELS_INFO: [(BOSCH, "ISW-ZDL1-WP11G")],
+        ENDPOINTS: {
+            5: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.IAS_ZONE,
+                INPUT_CLUSTERS: [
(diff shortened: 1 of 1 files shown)


zhaquirks/bosch/isw-zdl1-wp11g.py
```diff
@@ -0,0 +1,71 @@
...
+
+    MAX_VOLTS = 6.0
+    MIN_VOLTS = 3.0
+
+
+class ISWZDL1WP11G(CustomDevice):
+    """Custom device representing Bosch ISWZDL1WP11G motion sensor."""
+
+    signature = {
+        #  <SimpleDescriptor endpoint=1 profile=260 device_type=1026
```
dmulcahey, line 33: The endpoint id in the signature and here in the comment do not match. Which one is right?

dmulcahey (changes requested):
Just 1 small thing. Thanks for the contribution!

Adminiuga:
Post the original device signature (the one without the quirk applied) here for posterity.

zhaquirks/bosch/isw-zdl1-wp11g.py
nasa931: verified in Home Assistant it is 5 and working properly as configured.  will correct comment.  Thank you.

Adminiuga:
what do you get for battery_voltage attribute if you read it? e.g. the other bosch also has 4 batteries but reports the battery voltage as 3v, cause it looks like it is using 2in series + 2 in parallel battery setup.

nasa931:
Original Device Descriptor:
{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=8, manufacturer_code=0, maximum_buffer_size=108, maximum_incoming_transfer_size=0, server_mask=0, maximum_outgoing_transfer_size=0, descriptor_capability_field=0)",
  "endpoints": {
    "5": {
      "profile_id": 260,
      "device_type": "0x0402",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0020",
        "0x0402",
        "0x0500",
        "0x0b05"
      ],
      "out_clusters": [
        "0x0019"
      ]
    }
  },
  "manufacturer": "Bosch",
  "model": "ISW-ZDL1-WP11G",
}

nasa931:
@Adminiuga - there are 6 AA batteries in the WP11G but the arrangement is interesting, it appears to be 4 batteries in series for the first set (1.5 x 4 = 6 volts) and a second set with 2 other batteries powering other circuitry.
"battery_quantity" returns "4"
"battery_voltage" returns "60" (on a brand new set of batteries)
"battery_2_quantity" returns "2"
"battery_2_voltage" returns "0"

For what it's worth, the batteryMap developed for SmartThings also represents these two different models as follows, which matches what we're considering to deploy here:

if (model == "ISW-ZDL1-WP11G") {	
    	batteryMap = [60: 100, 59: 100, 58: 100, 57: 100, 56: 100, 55: 100,            
    				  54: 100, 53: 100, 52: 100, 51: 100, 50: 90, 49: 90,
                      48: 90, 47: 90, 46: 70, 45: 70, 44: 70, 43: 70, 42: 50, 
                      41: 50, 40: 50, 39: 50, 38: 30, 37: 30, 36: 30, 35: 30,
                      34: 15, 33: 15, 32: 1, 31: 1, 30: 0]                          
    	minVolts = 30        
        maxVolts = 60			           
    } else if(model == "ISW-ZPR1-WP13") {
    	batteryMap = [30: 100, 29: 100, 28: 100, 27: 100, 26: 100, 25: 90, 24: 90, 23: 70,
							  22: 70, 21: 50, 20: 50, 19: 30, 18: 30, 17: 15, 16: 1, 15: 0]
			minVolts = 15
			maxVolts = 30         
    }

that was found at https://community.smartthings.com/t/bosch-isw-zdl1-wp11g-tritech-motion-detectors-work-with-smartthings/119131

Adminiuga:
Interesting. I guess if there's no voltage for the 2nd set then we could ignore it, and `max_volts=6.0` is correct setting.

dmulcahey (approved):

PR #619, opened 2020-12-05
Add 2nd variant of konke button and device automation triggers

tube0013:
2nd model of the button had different signature. (thanks @walthowd ) Also added device automation triggers.

### DIFF
--- a/zhaquirks/konke/button.py
+++ b/zhaquirks/konke/button.py
@@ -13,17 +13,21 @@ from zigpy.zcl.clusters.general import (
 
 from .. import CustomCluster, CustomDevice, PowerConfigurationCluster
 from ..const import (
+    COMMAND,
     COMMAND_DOUBLE,
     COMMAND_HOLD,
     COMMAND_ID,
     COMMAND_SINGLE,
     DEVICE_TYPE,
+    DOUBLE_PRESS,
     ENDPOINTS,
     INPUT_CLUSTERS,
+    LONG_PRESS,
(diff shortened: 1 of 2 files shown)


zhaquirks/konke/button.py
```diff
@@ -112,3 +123,79 @@ def handle_message(self, profile, cluster, src_ep, dst_ep, message):
...
     }
+
+    device_automation_triggers = {
+        (DOUBLE_PRESS, DOUBLE_PRESS): {COMMAND: COMMAND_DOUBLE},
+        (SHORT_PRESS, SHORT_PRESS): {COMMAND: COMMAND_SINGLE},
+        (LONG_PRESS, LONG_PRESS): {COMMAND: COMMAND_HOLD},
+    }
+
+
+class KonkeButtonRemote2(CustomDevice):
```
dmulcahey, line 134: Let’s make a KonkeButton class that only has handle_message and extends CustomDevice. Then have both quirk device classes extend this so we don’t have 2 copies of handle_message. Ping me on discord if this isn’t clear.

dmulcahey (changes requested):

zhaquirks/konke/button.py
tube0013: I may have gone too far, this is tested and worked for me, and sort of follows what was done in the Tuya remotes.

dmulcahey (approved):
Great work @tube0013! Thanks!

PR #626, opened 2020-12-11
Fix Signiture Match

tube0013:
Fixes the quirk matching .

### DIFF
--- a/zhaquirks/konke/button.py
+++ b/zhaquirks/konke/button.py
@@ -11,7 +11,7 @@ from zigpy.zcl.clusters.general import (
     Scenes,
 )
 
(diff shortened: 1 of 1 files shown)


dmulcahey:
Why not use the const from init?

tube0013:
Do I have to import it? When I tried just KONKE no quotes it through an error on ha start up

dmulcahey:
Yes like this: https://github.com/zigpy/zha-device-handlers/blob/580adf4a8f66339f4da4ecd4c0e4589bada06225/zhaquirks/konke/magnet.py#L8

dmulcahey (approved):

PR #611, opened 2020-12-03
Add quirks for ts0041, ts0043 and ts0044 (Tuya 1, 3 and 4 button Remotes)

tube0013:
Add Quirk for Tuya 3 button remote
Make TuyaSmartRemote Class a bit more generic

Solves #595

Tested and work with my LoraTap 3 button remote.

Issue #505, opened 2020-09-26
[Device Support Request] Zemismart TS0041


oxamafia:
Hi,

I bought a wireless one-gang zigbee switch from Zemismart (TS0041/_TZ3400_keyjqthh).

The device is recognized by ZHA but nothing happens when I push the button. The switch stays in off mode in Home Assistant.

I would love to have the same things as Xiaomi Wireless Button for one/double/triple press.

There is no ZHA EVENT as well.

The debug says (when I press the button) :
2020-09-26 16:27:25 WARNING (MainThread) [zigpy.zcl] [0x281e:1:0x0006] Unknown cluster-specific command 253

and ZHA device signature is :
{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0006"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZ3400_keyjqthh",
  "model": "TS0041",
  "class": "zigpy.device.Device"
}

Can you please help me? Do you need anything else?

I also want to thank you for all the great job that you do. I was on deconz at first and then I switched to ZHA, that is wayyyyy better for many reasons.

Max

oxamafia:
Nobody can help me? :(

meremortals70:
This might help 

From the looks you have to watch the event stream and use an automation

https://www.zigbee2mqtt.io/devices/TS0041.html

avee87:
This should be fixed by #627

oxamafia:
Hello
Thank you for this hard work! You’re giving my Christmas gift in advance! So thank you
Actually I updated HA with the lastest version but the quirk doesn’t seem to apply to my device. 
i don’t understand why?
Is there a way to force the quirk?
I have tried to remove the device (and restart HA), then relink it... like many times... and it still doesn’t take the quirk 🙁

Informations sur l’appareil
TS0041
par _TZ3400_keyjqthh
dresden elektronik ConBee II

avee87:
Hm, that is weird - signature seems to be correct...
[...]

PR #627, opened 2020-12-11
Add quirk for Tuya wireless switch by Zemismart

avee87:
This is very similar to https://github.com/zigpy/zha-device-handlers/pull/611 but these switches have a slightly different signature.
Adding a separate quirk for them.

Tested that button presses generate correct zha events in Home Assistant on single, double and long press.

### DIFF
--- a/zhaquirks/tuya/ts0041_zemismart.py
+++ b/zhaquirks/tuya/ts0041_zemismart.py
@@ -1,0 +1,61 @@
+"""Tuya 1 Button Remote."""
+
+from zigpy.profiles import zha
+from zigpy.zcl.clusters.general import Basic, OnOff, Ota, PowerConfiguration, Time
+
+from . import TuyaSmartRemote, TuyaSmartRemoteOnOffCluster
+from ..const import (
+    BUTTON_1,
+    COMMAND,
+    DEVICE_TYPE,
+    DOUBLE_PRESS,
+    ENDPOINT_ID,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    LONG_PRESS,
+    MODELS_INFO,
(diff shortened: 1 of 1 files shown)


avee87:
Looks like signature is similar to the one from #505 although my device has "_TZ3000_tk3s5tyg" instead of "_TZ3400_keyjqthh"

avee87:
I actually think that "_TZ3400_keyjqthh" was misconfigured in #611 - i.e., it should use this new quirk but I don't have that device to test.

zhaquirks/tuya/ts0041_zemismart.py
```diff
@@ -0,0 +1,61 @@
...
+    SHORT_PRESS,
+)
+
+
+class TuyaZemismartSmartRemote0041(TuyaSmartRemote):
+    """Tuya Zemismart 1-button remote device."""
+
+    signature = {
+        # SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=0, device_version=1, input_clusters=[0, 1, 6], output_clusters=[25, 10])
+        MODELS_INFO: [("_TZ3000_tk3s5tyg", "TS0041")],
```
tube0013, line 28: can add the other one in here 
```
MODELS_INFO: [("_TZ3000_tk3s5tyg", "TS0041") ,("_TZ3400_keyjqthh", "TS0041")]

tube0013:
> I actually think that "_TZ3400_keyjqthh" was misconfigured in #611 - i.e., it should use this new quirk but I don't have that device to test.

you are right, I missed that difference. I'll put in a PR to remove it from the other 1 button quirk. signature for _TZ3400_keyjqthh to confirm is in #505

dmulcahey (approved):

PR #638, opened 2020-12-14
Revolt NX-4911 Bump

MattWestb:
Was missing then doing copy and pace then fighting with black so wrong model was left in the new record.
Fixed and holding my finger crossed.


MattWestb:
Sorry @dmulcahey Was messing up the device ID but have fixing it.
And thanks for all enchantments in 2020.12.0 !!

dmulcahey (approved):

PR #656, opened 2020-12-20
Fix develco __init_.py typo (missing underscore), add missing cluster_id, fixed imports

TheJulianJES:
Rename develco ``__init_.py`` to ``__init__.py`` which previously missed one underscore.
Also see https://github.com/zigpy/zha-device-handlers/pull/656#issuecomment-748692607


TheJulianJES:
It seems like that only now the tests are actually "checking" this file.
The issue occurs because ``general.Time`` should be ``general.Time.cluster_id``.
However, I'm wondering why this is the only/one of the only files where the imports where done like this. Was/Is there any reason to having it done like this?

TheJulianJES:
I've now gone ahead and changed the imports to be more like all other classes.
If this isn't desired, I can just revert the last commit.

TheJulianJES:
I've now gone ahead and changed the imports to be more like all other classes.
If this isn't desired, I can just revert the last two commits.

Since @Adminiuga made the original quirk. Are you okay with these changes (and does the quirk still work)?
(The basic issue was that ``__init__.py`` was missing one underscore. After adding it, the tests were failing (which I guess didn't run correctly before) because of the ``Time`` cluster not using the ``cluster_id``. I've also gone ahead and changed how the imports are done to be more in line with other files of the project while I'm at it.)

Adminiuga:
This is fine by me. I've done that way because it was easier to import all "general" clusters than pick individual ones.

Issue #646, opened 2020-12-18
Xiaomi Sensor actual first timeout is 90sec


tyjtyj:
https://github.com/zigpy/zha-device-handlers/blob/288913ecb1604c7f20c161d526fea9dc481fcfdc/zhaquirks/xiaomi/__init__.py#L343

I understand migateway setting is 120sec for returning from on to off is 120sec but xiaomi motion sensor actual timeout is 90. 
After first motion, the next motion detection is set every 30s. Below is example of the timeline

```
0s motion detected
60s motion redetect activated
60-89s any motion here will extend the motion to 120s
if no motion 
90s motion switch to off if no motion detected. 
if motion
90s motion remain on till 120s, redetection start at 90s again.
We can see this behaving correctly in z2m setting motion off to 90s.
```

Edit to explain further the behavior

if motion is at detect at 80s
eg.
0s motion detected
60s motion detection reactivated:
80s motion detected
110s motion redetect activated
if no motion for 30s, sensor should be off at 140s 
if motion here, sensor remain on
120s motion detected
if no motion for 30s, sensor should be off at 180s

dmulcahey:
the Z2M code looks like a straight 90 second timeout...


I also found this: https://community.smartthings.com/t/xiaomi-motion-sensor-aqara-how-to-test/170752/11

I'll change this to 65 seconds and do some testing in a few hrs when my sensor that I had to re-pair is no longer in test mode.

Issue #523, opened 2020-10-11
[BUG] Mijia/Xiaomi motion sensor with hack reports false motion


krue81:
**Describe the bug**
I have 3 Mijia/Xiaomi Motion Sensors (lumi.sensor_motion) working in HA via a Sonoff Zigbee Bridge. I did the hack for reporting motion every 5 seconds (https://community.smartthings.com/t/making-xiaomi-motion-sensor-a-super-motion-sensor/139806), but after that, I always get a fake/ghost trigger after the real motion trigger. Sometimes it happens 3 o 4 seconds after the real trigger, sometimes it is 1 minute later, but the ghost trigger happens ALWAYS after a real trigger.

**To Reproduce**
Hack the motion sensor with a bridge between TP4 and the switch.
![image](https://user-images.githubusercontent.com/70699930/95669306-19e4b500-0b45-11eb-8c66-930178924c3c.png)

Use a Sonoff Zigbee Bridge flashed with Tasmota and connect it to Home Assistant with ZHA

Add/Pair the motion sensor to HA

The motion sensor reports every 120/180 seconds... with the hack and a Node Red reset flow I can read the sensor every 5 seconds: Use an API node with POST to the path states/binary_sensor.sensor_name and data {"state":"off"}

After the sensor state is reset, It will always report a false/ghost trigger after a real trigger.

**Expected behavior**
A single motion trigger after motion is detected.

Adminiuga:
Debug logs?

krue81:
> Debug logs?

Sorry, here they are (the first motion is the real one, the second one, more than 60s later, was false (no one was in the room):
This screenshot is from node red, for the reset function mentioned earlier:
![image](https://user-images.githubusercontent.com/70699930/95679704-39f29380-0b9a-11eb-9500-51a998bccc05.png)


`
[...]

PR #658, opened 2020-12-21
Small cleanups in several classes

dmulcahey:
This PR:

- cleans up IASZone implementations that use constant attributes to set zone type
- changes the Xiaomi motion cool down to 70 seconds
- cleans up the Aqara vibration sensor motion class
- changes the Aqara vibration sensor cool down to 70 seconds
- adds an event to publish the reported vibration strength

~~I still have to test the motion cool down after the device vacates test mode...~~ I can confirm that the device does in fact trigger about every 60s like this says:  https://community.smartthings.com/t/xiaomi-motion-sensor-aqara-how-to-test/170752/11

Fixes: #646
Fixes: #523

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -62,7 +62,7 @@ async def test_motion(zigpy_device_from_quirk, quirk):
         tuya_cluster.handle_message(hdr, args)
 
     assert len(motion_listener.cluster_commands) == 1
-    assert len(motion_listener.attribute_updates) == 0
+    assert len(motion_listener.attribute_updates) == 1
     assert motion_listener.cluster_commands[0][1] == ZONE_STATE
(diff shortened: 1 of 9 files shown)


Adminiuga:
Do you want to incorporate https://github.com/zigpy/zha-device-handlers/issues/523#issuecomment-744755416 here as well?

Adminiuga (approved):
Looks good!

Adminiuga:
welp!
Guess those tests did not work correctly in the 1st place. give me a few i'll make a fix

dmulcahey:
> welp!
> Guess those tests did not work correctly in the 1st place. give me a few i'll make a fix

was just setting the attire update to 1 incorrect? we weren't updating the attractions before and now we are... so I figured making them 1 now made sense... Am I missing something?

PR #243, opened 2020-01-15
Add support for IKEA Tradfri Light bulbs

sandhose:
I noticed the IKEA Trafri light bulbs (I think I have [this one](https://www.ikea.com/gb/en/p/tradfri-led-bulb-e27-1000-lumen-wireless-dimmable-white-spectrum-opal-white-60408483/)) reported the ZHA profile (260) while having a ZLL device type (COLOR_TEMPERATURE_LIGHT / 0x0220). I added a quirk that changes the profile id to the ZLL one (49246)

Not sure if it is the right thing to do, but with this patch the light bulb shows up correctly with brightness and color temperature control in Home Assistant

PR #666, opened 2020-12-22
Add support for TRADFRI GU10 WS 400lm

pgrondek:
I had a problem with GU10 bulb, after paring i could only turn it on or off, this quirk add support for brightness control and color temperature

### DIFF
--- a/zhaquirks/ikea/tradfrigu10bulb.py
+++ b/zhaquirks/ikea/tradfrigu10bulb.py
@@ -1,0 +1,80 @@
+"""Tradfri GU10 bulb Quirk."""
+from zigpy.profiles import zha, zll
+from zigpy.quirks import CustomDevice
+from zigpy.zcl.clusters.general import (
+    Basic,
+    Groups,
+    Identify,
+    LevelControl,
+    OnOff,
+    Ota,
+    PollControl,
+    Scenes,
+)
+from zigpy.zcl.clusters.homeautomation import Diagnostic
+from zigpy.zcl.clusters.lighting import Color
+from zigpy.zcl.clusters.lightlink import LightLink
+
+from . import IKEA
+
+
+class TradfriGU10Bulb(CustomDevice):
+    """TRADFRI GU10 WS 400lm."""
+
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=544
+        # device_version=2
+        # input_clusters=[0, 3, 4, 5, 6, 8, 768, 2821, 4096]
+        # output_clusters=[5, 25, 32, 4096]>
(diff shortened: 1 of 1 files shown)


dmulcahey:
Took me a sec to see what this was doing... @Adminiuga another case of IKEA reporting a ZLL device type on a ZHA profile 🤦🏻‍♂️

Adminiuga:
What firmware version the bulb is reporting?

I think the problem with that, the entity discover uses clusters and not the device type, so it matches to the switch domain. 
Should we add a work around and lookup zll device ID if zha lookup fails?

dmulcahey:
> What firmware version the bulb is reporting?
> 
> I think the problem with that, the entity discover uses clusters and not the device type, so it matches to the switch domain.
> Should we add a work around and lookup zll device ID if zha lookup fails?

Up to you... I’m ok with this as a solution

Adminiuga:
That's fine, but if the firmware update fixes it, maybe should also recommend the firmware update. Although Ikea OTA updates have been a separate problem lately

pgrondek:
> What firmware version the bulb is reporting?
0x23050631 

I've enabled ota updates so I don't know if bulbs has been updated. I have also 4 bulbs connected to smartthings that reports version 0x11115720

TheJulianJES:
Just wondering, what is the model number of the IKEA bulb that you have? (There are multiple "TRADFRI bulb GU10 WS 400lm" -> I've got the older LED1737R5 model and they work.)
Is it the LED1739R5 you're having issues with or a completely different model? (Model number should be written on the box/perhaps even the light itself?)

pgrondek:
> Just wondering, what is the model number of the IKEA bulb that you have? (There are multiple "TRADFRI bulb GU10 WS 400lm" -> I've got the older LED1737R5 model and they work.)
> Is it the LED1739R5 you're having issues with or a completely different model? (Model number should be written on the box/perhaps even the light itself?)

I think I have even older model. On bulb I found model: LED1537R6, I thrown out box about 2 years ago.

TheJulianJES:
I guess a fix is really needed for some (older) IKEA bulbs then. Adminiuga recommended this workaround which would probably fix all IKEA lights misreporting a ZLL device type on the ZHA profile. Individual quirks wouldn't be needed when doing it this way. #243 could also be fixed by doing this and there wouldn't be any issues with the special character in the model id for some of the bulbs.
> I think the problem with that, the entity discover uses clusters and not the device type, so it matches to the switch domain.
> Should we add a work around and lookup zll device ID if zha lookup fails?

dmulcahey:
going to merge this for now... we can reassess if necessary.

Issue #643, opened 2020-12-15
[BUG] Non-RGB Lidl Bulbs still have xy_color, hs_color and rgb_color


bfayers:
**Describe the bug**
I have three of the Lidl E27 bulbs (Manufacturer: `_TZ3000_49qchf10` Model: `TS0502A`) still present the xy,hs and rgb color attributes in Home Assistant despite not being RGB bulbs - there is also a colour wheel shown in hass.

**To Reproduce**
1. Pair a Non-RGB Lidl Bulb
2. Go to the entity
3. See a colour wheel which is both non-functional and redundant.

**Expected behavior**
The xy,hs and rgb color attributes should not exist and the colour wheel should not be present.

**Additional context**
Device Signature:
```
{
  "node_descriptor": "NodeDescriptor(byte1=1, byte2=64, mac_capability_flags=142, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x010c",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0300",
        "0x1000"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZ3000_49qchf10",
  "model": "TS0502A",
  "class": "zigpy.device.Device"
}
```

TheJulianJES:
I'll take a look at this later on. Can you check by going to "Manage Clusters" -> select Color in the first dropdown. Then select color_capabilities in the second dropdown and click "Get Zigbee attribute". Please post the output from the text field here.

bfayers:
Sure! That comes back as `None`

TheJulianJES:
> > Can you check by going to "Manage Clusters" -> select Color in the first dropdown. Then select color_capabilities in the second dropdown and click "Get Zigbee attribute".
> 
> 
> 
> Sure! That comes back as `None`
> 
> 

If you have any other Zigbee lights: do they also come back as "None"?

bfayers:
[...]

PR #672, opened 2020-12-27
Add Lidl E14 model for CCT quirk

TheJulianJES:
https://github.com/zigpy/zha-device-handlers/issues/643#issuecomment-751415102

(Successfully tested by @mwsimmonite)


mwsimmonite:
I have tested this and it works in my config, correcting the E14 bulbs as the previous fix did the E27 ones

Issue #557, opened 2020-11-01
[Device Support Request]  Ikea Tradfri Wireless Dimmer Model ICTC-G-1


MattWestb:
**Is your feature request related to a problem? Please describe.**
https://github.com/home-assistant/core/issues/42672#issue-733729408

**Describe the solution you'd like**
Battery is not working OK without quirk (have some one pointed out).
Binding to groups is working and dimming lights but not the HA part for automations then its not implanted without quirk.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**

logs while pairing:

[0x0000:zdo] ZDO request ZDOCmd.Mgmt_Permit_Joining_req: [60, <Bool.false: 0>]
Tries remaining: 1
[0x8a5f] Extending timeout for 0x25 request
[0x8a5f] Delivery error for seq # 0x25, on endpoint id 0 cluster 0x0004: message send failure
[0x8a5f:1] Failed to discover endpoint_id 1
Device not found for removal: 00:0b:57:ff:fe:97:d2:cd
Device 0xdf77 (00:0b:57:ff:fe:97:d2:cd) joined the network
[0xdf77] Requesting 'Node Descriptor'
Tries remaining: 2
[0xdf77] Extending timeout for 0x27 request
Device 0xdf77 (00:0b:57:ff:fe:97:d2:cd) joined the network
Skip initialization for existing device 00:0b:57:ff:fe:97:d2:cd
[0xdf77:zdo] ZDO request ZDOCmd.Device_annce: [0xDF77, 00:0b:57:ff:fe:97:d2:cd, 128]
[0xdf77] Node Descriptor: NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4476, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=10752, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)
[0xdf77] Discovering endpoints
Tries remaining: 3
[0xdf77] Extending timeout for 0x29 request
[0xdf77] Discovered endpoints: [1]
[0xdf77:1] Discovering endpoint information
Tries remaining: 3
[0xdf77] Extending timeout for 0x2b request
[...]

PR #685, opened 2020-12-30
Support (old) IKEA Tradfri Wireless Dimmer ICTC-G-1

TheJulianJES:
Similar to the IKEA SYMFONISK "dimmer", but doesn't support clicks.
- adds doubling of the battery percentage: fixes https://github.com/zigpy/zha-device-handlers/issues/557, https://github.com/home-assistant/core/issues/42672
- also adds basic device automation triggers for moving the dial left and right

The support is for an IKEA wireless dimmer now with the updated ZB3 firmware. I didn't include the previous quirk for the ZLL firmware.
Based upon https://github.com/zigpy/zha-device-handlers/issues/557 and tested by @MattWestb

### DIFF
--- a/zhaquirks/ikea/dimmer.py
+++ b/zhaquirks/ikea/dimmer.py
@@ -13,7 +13,7 @@ from zigpy.zcl.clusters.general import (
 )
 from zigpy.zcl.clusters.lightlink import LightLink
 
-from . import IKEA
+from . import IKEA, ROTATED
 from .. import DoublingPowerConfigurationCluster
 from ..const import (
     ARGS,
@@ -30,8 +30,6 @@ from ..const import (
     PROFILE_ID,
     RIGHT,
 )
-
-ROTATED = "device_rotated"
 
 
 class IkeaDimmer(CustomDevice):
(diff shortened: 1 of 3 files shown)


zhaquirks/ikea/dimmer.py
```diff
@@ -0,0 +1,107 @@
...
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    LEFT,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+    RIGHT,
+)
+
+ROTATED = "device_rotated"
```
TheJulianJES, line 34: This is required for the ``device_automation_triggers`` and is also present in the ``symfonisk.py``.
Perhaps this should be moved into the ``__init.py__`` in the IKEA "folder" and then be imported in both quirks?

Adminiuga: either move it into `__init__.py` or you could do `from .sympfonisk import ROTATED`

Adminiuga (approved):
LGTM

TheJulianJES:
I moved ``device_rotated`` to the IKEA ``__init__.py`` file now as that seems *cleaner*.
It does look like that Black reformatted some parts of the file though. One second

MattWestb:
Was updating the 3 changed files after doing the last chopping round this year and the quirk is loading OK and battery is reported as expected !!
HA is firing ZHA events from cluster 0x0008 commands so its looks doing wath is expected :-))

Thanks for great coding and one great end and happy new year with greetings from [Miss Sophie and James](https://en.wikipedia.org/wiki/Dinner_for_One) ;-))

PR #714, opened 2021-01-10
Better Moes HY369 (tuya) implementation

xonestonex:
I've implemented almost all of the device functionality based on the device point descriptions of Zigbee2Mqtt.
The manufacturer specific stuff is also mapped to the individual clusters as manufacturer specific attributes, this should make it easier to provide more functionality in home assistant in a later step.

Expect update delays until [https://github.com/zigpy/zigpy/issues/592](url) gets fixed.

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -1,5 +1,6 @@
 """Tuya devices."""
 import logging
+import datetime
 from typing import Optional, Tuple, Union
 
 from zigpy.quirks import CustomCluster, CustomDevice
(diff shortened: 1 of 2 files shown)


sylvaing:
Hi @xonestonex, why you closed this ? I didn't test this but I think that is a good idea to improve this TRV support.

xonestonex:
I've decided to write some tests, to improve the test coverage. Will create a new PR then again.

sylvaing:
nice !

MattWestb:
Hi Hope you is getting the tests done and the PR because the TRV need little more coding for getting then full functioning. 
The first part was not so easy for the dev then hi was not having any TRV for testing and not easy testing the code and getting feedback.
The Moes is one with nearly all things tuya have implanting so if its in place the rest is more or less adjusting the commands that the device is using.

xonestonex:
I've create a new pull request.
https://github.com/zigpy/zha-device-handlers/pull/726

PR #715, opened 2021-01-10
Decoupled mode for xiaomi.aqara.CtrlNeutral

### LINKED ISSUE #713: [Device Support Request] Decoupled mode for ctrl_neutral (QBKG04LM/QBKG03LM/QBKG21LM/QBKG22LM)
Similar to #365, but for `ctrl_neutral` implementation. I think that once implemented it can be easily adapted.

In my case I want to implement QBKG04LM decoupled mode.
It can be done easily, following @albalaing indications (https://github.com/zigpy/zha-device-handlers/issues/365#issuecomment-636510316).

The problem is that without the click button events it is not useful, so it is necessary to also implement the generation of ZHA events.

Before implement anything you can see that device generate 4 messages:
```
2021-01-09 21:46:29 DEBUG (MainThread) [zigpy.device] Ignoring message (b'18500a0000100000001001') on cluster 6: unknown endpoint or cluster id: 4
2021-01-09 21:46:31 DEBUG (MainThread) [zigpy.device] Ignoring message (b'18510a00001000') on cluster 6: unknown endpoint or cluster id: 4
2021-01-09 21:46:32 DEBUG (MainThread) [zigpy.device] Ignoring message (b'18520a00001001') on cluster 6: unknown endpoint or cluster id: 4
2021-01-09 21:46:32 DEBUG (MainThread) [zigpy.device] Ignoring message (b'18530a00001002') on cluster 6: unknown endpoint or cluster id: 4
```

This messages are (in order):
* click
* hold
* release
* double_click

You can see that the click message corresponds in some way to the hold+release messages (00001000+00001001)

I think I'm halfway done (the decoupled part is working as expected), but I'm not satisfied with the event generation part. 

When a `replacement` for endpoint 4 (and 5) is implemented you can get the `on_off` attribute reported. But IMO this approach has the following problems:
* add a new entity in HA for click button events
* it is not possible to differentiate the `click` event from `hold` and `release`
* the `double_click` event cannot be handled (parse error)

Want I think that would be implemented is:
* no HA button entity
* full events control, not only hold and release

Any help in this regard is welcome.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
The device signature is known and correct.

**Additional context**
Although I am not sure if it can be done this way, I have tried without success to implement a replacement that transforms events from the `OnOff` cluster to the `MultistateInput` cluster.

My current implementation is as follows (only the relevant part):
```
    class BasicClusterDecoupled(BasicCluster):
        """Adds attributes for decoupled mode"""
            """Init."""
            self.attributes = BasicCluster.attributes.copy()
            self.attributes.update({ 0xFF22: ("left_decoupled_mode", t.uint8_t)})
            self.attributes.update({ 0xFF23: ("right_decoupled_mode", t.uint8_t)})


    class CustomOnOffCluster(OnOffCluster):
        """Fire ZHA events for on off cluster."""

        cluster_id = OnOff.cluster_id

            """Init."""
            self._current_state = {}

            _LOGGER.info("%s: %s", attrid, value)
            if attrid == 0:
                self._current_state = PRESS_TYPES.get(value)
                event_args = {
                    PRESS_TYPE: self._current_state,
                    VALUE: value,
                }
                self.listener_event(ZHA_SEND_EVENT, COMMAND_CLICK, event_args)
            super()._update_attribute(attrid, self._current_state)


    replacement = {
        SKIP_CONFIGURATION: True,
        ENDPOINTS: {
            .../...
            4: {                                                                   
                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,                         
                INPUT_CLUSTERS: [                                                  
                    MultistateInput.cluster_id,                                        
                    CustomOnOffCluster,                                        
                ],                                                                 
                OUTPUT_CLUSTERS: [],                                               
            },  
            5: {                                                                   
                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,                         
                INPUT_CLUSTERS: [                                                  
                    MultistateInput.cluster_id,                                        
                    CustomOnOffCluster,                                        
                ],                                                                 
                OUTPUT_CLUSTERS: [],                                               
            },  
        },
    }
```

javicalle:
Some Aqara wall switches can work in decoupled mode, allowing the rocker to be independent from the relay.

The current implementation for these switches only exposes the relays, so for this change to be functional it is also necessary to implement the quirks of the buttons endpoints.

This PR implements the "decoupled mode" command and these quirks in `xiaomi.aqara.CtrlNeutral`.
This class supports devices: QBKG04LM/QBKG03LM/QBKG21LM/QBKG22LM.
Decoupled mode would be configurable from UI (thanks @albalaing):
image

Value:
* 254 (decoupled)
* 18 (coupled)

Manufacturer code: 
* 4447 (important)

All the tests I am doing with the single rocker model (QBKG04LM). So I cannot guarantee the correct implementation for the 2 rockets models.
IMO is probably preferable to separate the implementations for 1 rocker model from the 2 rockers, but I think it will be best addressed in a second iteration when we have test data.

Some things to check:
- [ ] do not generate the HA entity
- [x] `double_click` event handling (related #719)
- [x] events format, more homogeneous with the rest of the events
- [x] implement `device_automation_triggers` (must have)
- [ ] check class nomenclature
- [ ] implement tests?

resolves #713

### DIFF
--- a/zhaquirks/xiaomi/aqara/ctrl_neutral.py
+++ b/zhaquirks/xiaomi/aqara/ctrl_neutral.py
@@ -24,15 +24,19 @@ from .. import (
     XiaomiCustomDevice,
     XiaomiPowerConfiguration,
 )
+from ... import EventableCluster
 from ...const import (
     COMMAND_CLICK,
     DEVICE_TYPE,
+    DOUBLE_PRESS,
     ENDPOINTS,
     INPUT_CLUSTERS,
     MODELS_INFO,
     OUTPUT_CLUSTERS,
     PRESS_TYPE,
     PROFILE_ID,
+    SHORT_PRESS,
+    SHORT_RELEASE,
     SKIP_CONFIGURATION,
     VALUE,
     ZHA_SEND_EVENT,
@@ -59,13 +63,50 @@ class CtrlNeutral(XiaomiCustomDevice):
     """Aqara single and double key switch device."""
 
     class BasicClusterDecoupled(BasicCluster):
-        """Adds attributes for decoupled mode"""
+        """Adds attributes for decoupled mode."""
+
+        # Known Options for 'decoupled_mode_<button>':
+        # * 254 (decoupled)
+        # * 18 (relay controlled)
+        manufacturer_attributes = {
+            0xFF22: ("decoupled_mode_left", t.uint8_t),
+            0xFF23: ("decoupled_mode_right", t.uint8_t),
+        }
+
+    class WallSwitchOnOffCluster(OnOff, EventableCluster):
+        """WallSwitchOnOffCluster: fire events corresponding to press type."""
+
+        press_type = {
+            0x00: SHORT_PRESS,
+            0x01: SHORT_RELEASE,
+            0x02: DOUBLE_PRESS,
+        }
+        name = "WallSwitch_cluster"
+        ep_attribute = "WallSwitch_cluster"
+
         def __init__(self, *args, **kwargs):
             """Init."""
-            self.attributes = BasicCluster.attributes.copy()
-            self.attributes.update({0xFF22: ("decoupled_mode_left", t.uint8_t)})
-            self.attributes.update({0xFF23: ("decoupled_mode_right", t.uint8_t)})
+            self.last_tsn = -1
             super().__init__(*args, **kwargs)
+
+        manufacturer_server_commands = {
+            0xFD: ("press_type", (t.uint8_t,), False),
+        }
+
+        def handle_cluster_request(self, tsn, command_id, args):
+            """Handle press_types command."""
+            if tsn == self.last_tsn:
+                _LOGGER.debug("WallSwitch: ignoring duplicate frame")
+                return
+
+            self.last_tsn = tsn
+            _LOGGER.error("handle_cluster_request: %s, %s", command_id, args)
+            super().handle_cluster_request(tsn, command_id, args)
+            if command_id == 0xFD:
+                press_type = args[0]
+                self.listener_event(
+                    ZHA_SEND_EVENT, self.press_type.get(press_type, "unknown"), []
+                )
 
     class CustomOnOffCluster(OnOffCluster):
         """Fire ZHA events for on off cluster."""
@@ -225,7 +266,7 @@ class CtrlNeutral(XiaomiCustomDevice):
                 DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
                 INPUT_CLUSTERS: [
                     MultistateInput.cluster_id,
-                    CustomOnOffCluster,
+                    WallSwitchOnOffCluster,
                 ],
                 OUTPUT_CLUSTERS: [],
             },
@@ -233,7 +274,7 @@ class CtrlNeutral(XiaomiCustomDevice):
                 DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
                 INPUT_CLUSTERS: [
                     MultistateInput.cluster_id,
-                    CustomOnOffCluster,
+                    WallSwitchOnOffCluster,
                 ],
                 OUTPUT_CLUSTERS: [],
             },
--- a/zhaquirks/xiaomi/aqara/ctrl_neutral.py
+++ b/zhaquirks/xiaomi/aqara/ctrl_neutral.py
@@ -25,8 +25,8 @@ from .. import (
     XiaomiPowerConfiguration,
 )
 from ...const import (
-    DEVICE_TYPE,
     COMMAND_CLICK,
+    DEVICE_TYPE,
     ENDPOINTS,
     INPUT_CLUSTERS,
     MODELS_INFO,
@@ -58,16 +58,14 @@ _LOGGER = logging.getLogger(__name__)
 class CtrlNeutral(XiaomiCustomDevice):
     """Aqara single and double key switch device."""
 
-
     class BasicClusterDecoupled(BasicCluster):
         """Adds attributes for decoupled mode"""
         def __init__(self, *args, **kwargs):
             """Init."""
             self.attributes = BasicCluster.attributes.copy()
-            self.attributes.update({ 0xFF22: ("decoupled_mode_left", t.uint8_t)})
-            self.attributes.update({ 0xFF23: ("decoupled_mode_right", t.uint8_t)})
+            self.attributes.update({0xFF22: ("decoupled_mode_left", t.uint8_t)})
+            self.attributes.update({0xFF23: ("decoupled_mode_right", t.uint8_t)})
             super().__init__(*args, **kwargs)
-
 
     class CustomOnOffCluster(OnOffCluster):
         """Fire ZHA events for on off cluster."""
@@ -89,7 +87,6 @@ class CtrlNeutral(XiaomiCustomDevice):
                 }
                 self.listener_event(ZHA_SEND_EVENT, COMMAND_CLICK, event_args)
             super()._update_attribute(attrid, self._current_state)
-
 
     signature = {
         MODELS_INFO: [
--- a/zhaquirks/xiaomi/aqara/ctrl_neutral.py
+++ b/zhaquirks/xiaomi/aqara/ctrl_neutral.py
@@ -26,20 +26,13 @@ from .. import (
 )
 from ... import EventableCluster
 from ...const import (
-    COMMAND_CLICK,
     DEVICE_TYPE,
-    DOUBLE_PRESS,
     ENDPOINTS,
     INPUT_CLUSTERS,
     MODELS_INFO,
     OUTPUT_CLUSTERS,
-    PRESS_TYPE,
     PROFILE_ID,
-    SHORT_PRESS,
-    SHORT_RELEASE,
     SKIP_CONFIGURATION,
-    VALUE,
-    ZHA_SEND_EVENT,
 )
 
 DOUBLE = "double"
@@ -75,59 +68,6 @@ class CtrlNeutral(XiaomiCustomDevice):
 
     class WallSwitchOnOffCluster(OnOff, EventableCluster):
         """WallSwitchOnOffCluster: fire events corresponding to press type."""
-
-        press_type = {
-            0x00: SHORT_PRESS,
-            0x01: SHORT_RELEASE,
-            0x02: DOUBLE_PRESS,
-        }
-        name = "WallSwitch_cluster"
-        ep_attribute = "WallSwitch_cluster"
-
-        def __init__(self, *args, **kwargs):
-            """Init."""
-            self.last_tsn = -1
-            super().__init__(*args, **kwargs)
-
-        manufacturer_server_commands = {
-            0xFD: ("press_type", (t.uint8_t,), False),
-        }
-
-        def handle_cluster_request(self, tsn, command_id, args):
-            """Handle press_types command."""
-            if tsn == self.last_tsn:
-                _LOGGER.debug("WallSwitch: ignoring duplicate frame")
-                return
-
-            self.last_tsn = tsn
-            _LOGGER.error("handle_cluster_request: %s, %s", command_id, args)
-            super().handle_cluster_request(tsn, command_id, args)
-            if command_id == 0xFD:
-                press_type = args[0]
-                self.listener_event(
-                    ZHA_SEND_EVENT, self.press_type.get(press_type, "unknown"), []
-                )
-
-    class CustomOnOffCluster(OnOffCluster):
-        """Fire ZHA events for on off cluster."""
-
-        cluster_id = OnOff.cluster_id
-
-        def __init__(self, *args, **kwargs):
-            """Init."""
-            self._current_state = {}
-            super().__init__(*args, **kwargs)
-
-        def _update_attribute(self, attrid, value):
-            _LOGGER.info("%s: %s", attrid, value)
-            if attrid == 0:
-                self._current_state = PRESS_TYPES.get(value)
-                event_args = {
-                    PRESS_TYPE: self._current_state,
-                    VALUE: value,
-                }
-                self.listener_event(ZHA_SEND_EVENT, COMMAND_CLICK, event_args)
-            super()._update_attribute(attrid, self._current_state)
 
     signature = {
         MODELS_INFO: [
--- a/zhaquirks/xiaomi/aqara/ctrl_neutral.py
+++ b/zhaquirks/xiaomi/aqara/ctrl_neutral.py
@@ -64,8 +64,8 @@ class CtrlNeutral(XiaomiCustomDevice):
         def __init__(self, *args, **kwargs):
             """Init."""
             self.attributes = BasicCluster.attributes.copy()
-            self.attributes.update({ 0xFF22: ("left_decoupled_mode", t.uint8_t)})
-            self.attributes.update({ 0xFF23: ("right_decoupled_mode", t.uint8_t)})
+            self.attributes.update({ 0xFF22: ("decoupled_mode_left", t.uint8_t)})
+            self.attributes.update({ 0xFF23: ("decoupled_mode_right", t.uint8_t)})
             super().__init__(*args, **kwargs)
 
 
--- a/zhaquirks/xiaomi/aqara/ctrl_neutral.py
+++ b/zhaquirks/xiaomi/aqara/ctrl_neutral.py
@@ -1,6 +1,7 @@
 """Xiaomi aqara single key wall switch devices."""
 import logging
 
+from zigpy import types as t
 from zigpy.profiles import zha
 from zigpy.zcl.clusters.general import (
     AnalogInput,
@@ -23,15 +24,19 @@ from .. import (
     XiaomiCustomDevice,
     XiaomiPowerConfiguration,
 )
-from ...const import (
-    DEVICE_TYPE,
-    ENDPOINTS,
-    INPUT_CLUSTERS,
-    MODELS_INFO,
-    OUTPUT_CLUSTERS,
-    PROFILE_ID,
-    SKIP_CONFIGURATION,
-)
+from ...const import (                                                         
+    DEVICE_TYPE,                                                               
+    COMMAND_CLICK,                                                             
+    ENDPOINTS,                                                                 
+    INPUT_CLUSTERS,                                                                
+    MODELS_INFO,                                                               
+    OUTPUT_CLUSTERS,                                                           
+    PRESS_TYPE,                                                                      
+    PROFILE_ID,                                                                
+    SKIP_CONFIGURATION,                                                        
+    VALUE,                                                                           
+    ZHA_SEND_EVENT,                                                            
+) 
 
 DOUBLE = "double"
 HOLD = "long press"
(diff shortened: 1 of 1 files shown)


dmulcahey:
You should define these as manufacturer_attributes.  Search the repo to find examples.

dmulcahey:
define them like this: https://github.com/zigpy/zha-device-handlers/blob/56cf0cfadd63795d2e360dd593e3676a2da0f441/zhaquirks/xiaomi/mija/smoke.py#L57

dmulcahey:
For events, you can override the onoff cluster and send them similar to what is done here: https://github.com/zigpy/zha-device-handlers/blob/56cf0cfadd63795d2e360dd593e3676a2da0f441/zhaquirks/xiaomi/mija/sensor_switch.py#L63

dmulcahey:
Just make your onoff implementation extend this: https://github.com/zigpy/zha-device-handlers/blob/b97c196c16dea43f0ea56dc822b64d487c76c15b/zhaquirks/__init__.py#L101

and that should do it

javicalle:
Thank you.
It's just what I need to move forward.

javicalle:
Second round.

I have implemented an EventableCluster (inspired in `TuyaSmartRemoteOnOffCluster`):
```
    class WallSwitchOnOffCluster(OnOff, EventableCluster):
        """WallSwitchOnOffCluster: fire events corresponding to press type."""

        press_type = {
            0x00: SHORT_PRESS,
            0x01: SHORT_RELEASE,
            0x02: DOUBLE_PRESS,
        }
        name = "WallSwitch_cluster"
        ep_attribute = "WallSwitch_cluster"

            """Init."""
            self.last_tsn = -1

        manufacturer_server_commands = {
            0xFD: ("press_type", (t.uint8_t,), False),
        }

        def handle_cluster_request(self, tsn, command_id, args):
            """Handle press_types command."""
            if tsn == self.last_tsn:
                _LOGGER.debug("WallSwitch: ignoring duplicate frame")
                return

            self.last_tsn = tsn
            _LOGGER.error("handle_cluster_request: %s, %s", command_id, args)
            super().handle_cluster_request(tsn, command_id, args)
            if command_id == 0xFD:
                press_type = args[0]
                self.listener_event(
                    ZHA_SEND_EVENT, self.press_type.get(press_type, "unknown"), []
                )
```

But I can't get any trace from my implementation.

As in the previous attempts, the traces I get are (`click`):
```
2021-01-10 20:27:21 DEBUG (MainThread) [zigpy.zcl] [0x46bb:4:0x0006] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=False disable_default_response=True> manufacturer=None tsn=146 command_id=Command.Report_Attributes>
2021-01-10 20:27:21 DEBUG (MainThread) [zigpy.zcl] [0x46bb:4:0x0006] ZCL request 0x000a: [[Attribute(attrid=0, value=<TypeValue type=Bool, value=Bool.false>), Attribute(attrid=0, value=<TypeValue type=Bool, value=Bool.true>)]]
2021-01-10 20:27:21 DEBUG (MainThread) [zigpy.zcl] [0x46bb:4:0x0006] Attribute report received: on_off=0, on_off=1
```

And for `double_click`:
```
2021-01-10 20:28:15 DEBUG (MainThread) [zigpy.zcl] [0x46bb:4:0x0006] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=False disable_default_response=True> manufacturer=None tsn=149 command_id=Command.Report_Attributes>
2021-01-10 20:28:15 ERROR (MainThread) [zigpy.device] Failed to parse message (b'18950a00001002') on cluster 6, because 2 is not a valid Bool
```

Here I am lost.
I interpret that the event is captured by the OnOff cluster ('2 is not a valid Bool') before the new implementation get the message, but I can't dig more to understand what I'm doing wrong.
Also I don't understand why I'm getting these `on_off` attributes.

javicalle:
Maybe I need to correctly set the value of `manufacturer_server_commands` or maybe those magic numbers (`0xFD`) are more important than I think, but I really don't know where to get valid values from.

When I activate the traces in debug mode the output is very verbose (I have several sensors at home) and I cannot see anything.

javicalle:
Ummmm, I have found this implementation:
https://github.com/zigpy/zha-device-handlers/blob/3775e13d4ab5d67d558cc64204d1723d8240e9cd/zhaquirks/konke/__init__.py#L63

I'm going to see where it takes me.

dmulcahey:
You don’t need any custom logic... just define an onoff cluster and extend eventable. No logic needed in the class

zhaquirks/xiaomi/aqara/ctrl_neutral.py
```diff
@@ -53,6 +62,73 @@
...
+            0x01: SHORT_RELEASE,
+            0x02: DOUBLE_PRESS,
+        }
+        name = "WallSwitch_cluster"
+        ep_attribute = "WallSwitch_cluster"
+
+        def __init__(self, *args, **kwargs):
+            """Init."""
+            self.last_tsn = -1
+            super().__init__(*args, **kwargs)
```
Adminiuga, line 90: You don't need this. Not unless it sends two consecutive commands with the same TSN, which was only tuya specific thing

zhaquirks/xiaomi/aqara/ctrl_neutral.py
```diff
@@ -53,6 +62,73 @@
...
+
+        manufacturer_server_commands = {
+            0xFD: ("press_type", (t.uint8_t,), False),
+        }
+
+        def handle_cluster_request(self, tsn, command_id, args):
+            """Handle press_types command."""
+            if tsn == self.last_tsn:
+                _LOGGER.debug("WallSwitch: ignoring duplicate frame")
+                return
```
Adminiuga, line 100: same as above. you don't need it.

zhaquirks/xiaomi/aqara/ctrl_neutral.py
```diff
@@ -53,6 +62,73 @@
...
+        def _update_attribute(self, attrid, value):
+            _LOGGER.info("%s: %s", attrid, value)
+            if attrid == 0:
+                self._current_state = PRESS_TYPES.get(value)
+                event_args = {
+                    PRESS_TYPE: self._current_state,
+                    VALUE: value,
+                }
+                self.listener_event(ZHA_SEND_EVENT, COMMAND_CLICK, event_args)
+            super()._update_attribute(attrid, self._current_state)
```
Adminiuga, line 130: so, for any attribute update, you'll be sending the last received "press type", regardless of the what attribute is being updated/reported? This doesn't look correct to me, unless i'm missing something

Adminiuga:
> And for double_click:
```
2021-01-10 20:28:15 DEBUG (MainThread) [zigpy.zcl] [0x46bb:4:0x0006] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=False disable_default_response=True> manufacturer=None tsn=149 command_id=Command.Report_Attributes>
2021-01-10 20:28:15 ERROR (MainThread) [zigpy.device] Failed to parse message (b'18950a00001002') on cluster 6, because 2 is not a valid Bool
```

So this is a bit of a problem. And atm I can't think a better approach. The problem is with the bool type, which zigpy enforces to be either 0 or 1, but Konke and in this case sends custom values. The work around for konke is to re-write attribute reports implemented in https://github.com/zigpy/zha-device-handlers/blob/3775e13d4ab5d67d558cc64204d1723d8240e9cd/zhaquirks/konke/__init__.py#L63

@dmulcahey any better ideas?

javicalle:
Third round.

Now the implementation is clean and generates the same result. Thanks for your comments and advices.
```python
    class WallSwitchOnOffCluster(OnOff, EventableCluster):
        """WallSwitchOnOffCluster: fire events corresponding to press type."""
```

And now the implementation generates an event (same event, diferent value):
```json
{
    "event_type": "zha_event",
    "data": {
        "device_ieee": "00:15:8d:00:02:c7:97:df",
        "unique_id": "00:15:8d:00:02:c7:97:df:4:0x0006",
        "device_id": "89d67ef84d0f78ea976d4afb287268ef",
        "endpoint_id": 4,
        "cluster_id": 6,
        "command": "attribute_updated",
        "args": {
            "attribute_id": 0,
            "attribute_name": "on_off",
            "value": 1
        }
    },
    "origin": "LOCAL",
    "time_fired": "2021-01-11T00:36:26.587235+00:00",
    "context": {
        "id": "030d1c25122f368dd3dc7519dab3f272",
        "parent_id": null,
        "user_id": null
    }
}
```

I guess it is enough to have a working implementation.

Adminiuga:
> @dmulcahey any better ideas?

I think just should make this a basic.enum8    in https://github.com/zigpy/zigpy/blob/5ed167206059018e25ac9bfd7c8ff0116e420270/zigpy/types/named.py#L40 Need to try this

javicalle:
When I the full picture about the events that can be implemented, I will review the implementation for `device_automation_triggers`.


> The problem is with the bool type, which zigpy enforces to be either 0 or 1

It would also be desirable to be able to differentiate `short_click` from the` long_press` and `release` events (more functionality), but at that point I don't know how complicated it can be to implement.

And if you can tell me how to avoid entities in HA ... 
Now they are not useful at all since once activated it never changes state again:
image

javicalle:
> I think just should make this a basic.enum8

But this will only prevent from the 'error', won't it?
This change will not allow receiving events of type `double_click` since the cluster `on_off` only supports 2 states (boolean).

Adminiuga:
yes, it would fix the "error", and then in the quirk you could generate double click events depending on the value of the attribute.

javicalle:
> yes, it would fix the "error", and then in the quirk you could generate double click events depending on the value of the attribute.

You may not get it right, but if the attribute is of type Boolean, it only has 2 possible values:
```
2021-01-11 01:57:46 DEBUG (MainThread) [zigpy.zcl] [0x1561:4:0x0006] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=False disable_default_response=True> manufacturer=None tsn=45 command_id=Command.Report_Attributes>
2021-01-11 01:57:46 DEBUG (MainThread) [zigpy.zcl] [0x1561:4:0x0006] ZCL request 0x000a: [[Attribute(attrid=0, value=<TypeValue type=Bool, value=Bool.false>), Attribute(attrid=0, value=<TypeValue type=Bool, value=Bool.true>)]]
2021-01-11 01:57:46 DEBUG (MainThread) [zigpy.zcl] [0x1561:4:0x0006] Attribute report received: on_off=0, on_off=1
```

Unless I can somehow change the value type I don't see how to have more events.

javicalle:
> I think just should make this a basic.enum8 in https://github.com/zigpy/zigpy/blob/5ed167206059018e25ac9bfd7c8ff0116e420270/zigpy/types/named.py#L40 Need to try this

I have tested it in my environment and reported in PR #719

javicalle:
A little update. Let me show you far has taken me my trip.
I have followed the thread of konke implementation to get this:
```python
    class WallSwitchOnOffCluster(CustomCluster, OnOff):
        """WallSwitchOnOffCluster: fire events corresponding to press type."""

        ENDPOINT_MAP = {4: "left", 5: "right", 6: "both"}
        PRESS_TYPES = { 17592186048513: COMMAND_SINGLE, 4098: COMMAND_DOUBLE, 4096: COMMAND_HOLD, 4097:COMMAND_RELEASE}

        def handle_cluster_general_request(self, header, args):
            self.debug(
                "Aqara wall switch request - handle_cluster_general_request: header: %s - args: [%s]",
                header,
                args,
            )

            if header.command_id != zigpy.zcl.foundation.Command.Report_Attributes:
                return

            attr = args[0][0]
            if attr.attrid != 0x0000:
                return

            value = attr.value.value
            event_args = {
                PRESS_TYPE: self.PRESS_TYPES.get(value, value),
                BUTTON: "button_{}".format(self.endpoint.endpoint_id),
                COMMAND_ID: value,
            }
            self.listener_event(ZHA_SEND_EVENT, event_args[PRESS_TYPE], event_args)

        def deserialize(self, data):
            """Deserialize fix for Aqara butchered Bool ZCL type."""
            hdr, temp_data = zigpy.zcl.foundation.ZCLHeader.deserialize(data)
            if (
                hdr.frame_control.is_cluster
                or hdr.command_id != zigpy.zcl.foundation.Command.Report_Attributes
            ):
                return super().deserialize(data)
            attr_id, temp_data = t.uint16_t.deserialize(temp_data)
            attr = zigpy.zcl.foundation.Attribute(
                attr_id, zigpy.zcl.foundation.TypeValue(t.uint8_t, int.from_bytes(temp_data, byteorder="big"))
            )
            return hdr, [[attr]]
```


The `EventableCluster` approach is functional and clean, but it does not allow to have (or I have not known how to get) different `hold` & `release` events.

With this implementation, it is possible to have 4 differentiated events (`hold`,`release`, `click` & `double_click`)
But on the other hand I am aware that it is a tricky implementation that does not follow the specifications, and I am not sure if it is the way to go.
Before going into the detail of the implementation, I would like to know a second opinion about it. It is my first approach to this module and I don't want to get off on the wrong foot.

Thanks in advanced.

Adminiuga:
```
BUTTON: "button_{}".format(self.endpoint.endpoint_id),
```

you don't really need this. There's already `endpoint_id` present in the `zha_event` or `unique_id` identifying what button was pressed.

```
            attr = zigpy.zcl.foundation.Attribute(
                attr_id, zigpy.zcl.foundation.TypeValue(t.uint8_t, int.from_bytes(temp_data, byteorder="big"))
            )
```

make it
```
            attr = zigpy.zcl.foundation.Attribute(
                attr_id, zigpy.zcl.foundation.TypeValue(t.uint8_t, temp_data[1]))
            )
```
This would change the commands id, but ` { 17592186048513: COMMAND_SINGLE,` is sign that something  is wrong anyway

javicalle:
> int.from_bytes(temp_data, byteorder="big")

This is the reason for the hack and one of the reasons for not sticking with the `EventableCluster` implementation.

See if I can explain it correctly.

What I have seen in my tests is that I can differentiate 4 types of events:
```
b'18560a00001000' -> hold
b'18570a00001001' -> release
b'18510a0000100000001001' -> click
b'18540a00001002' -> doubleclick
```

I have not gone into the detail of the specification, but what I have clear is that the click message corresponds (in some way) to the hold + release (and is what the `EventableCluster` implementation gives me, the 2 events):
```
b'18560a 00001000' -> hold
b'18570a 00001001' -> release
b'18510a 00001000 00001001' -> click
b'18540a 00001002' -> doubleclick
```

The current implementation (`TypeValue(t.uint8_t, temp_data[1]))`) gets bits 13-14 (from the right):
```
00 -> hold
01 -> release
00 -> click
02 -> doubleclick
```

What this implementation does is keep all the digits starting at position 11. That way I can differentiate the 4 events:
```
1000 -> hold
1001 -> release
100000001001 -> click
1002 -> doubleclick
```

Yes, it's a dirty trick. 
And I am sure it does not follow the specifications, so I prefer to consult before continuing along that line.


If the issue of the value 0x02 (# 719) can be solved, it is possible to have a no-hack implementation with 3 events, which is not bad either.

javicalle:
I have reviewed the implementation and have come to a fairly satisfactory version:
```
        PRESS_TYPES = { 4097: COMMAND_SINGLE, 2: COMMAND_DOUBLE, 0: COMMAND_HOLD, 1:COMMAND_RELEASE}

        def handle_cluster_general_request(self, header, args):
            self.error(
                "Aqara wall switch request - handle_cluster_general_request: header: %s - args: [%s]",
                header,
                args,
            )

            if header.command_id != zigpy.zcl.foundation.Command.Report_Attributes:
                return

            attr = args[0][0]
            if attr.attrid != 0x0000:
                return

            # value = attr.value.value
            value = int.from_bytes(attr.value.value, byteorder="big")
            button = self.ENDPOINT_MAP.get(self.endpoint.endpoint_id)
            press = self.PRESS_TYPES.get(value, UNKNOWN)
            action = "{}_{}".format(button, press)
            event_args = {
                PRESS_TYPE: press,
                BUTTON: button,
                VALUE: value,
            }
            self.listener_event(ZHA_SEND_EVENT, action, event_args)

        def deserialize(self, data):
            """Deserialize fix for Aqara butchered Bool ZCL type."""
            hdr, temp_data = zigpy.zcl.foundation.ZCLHeader.deserialize(data)
            if (
                hdr.frame_control.is_cluster
                or hdr.command_id != zigpy.zcl.foundation.Command.Report_Attributes
            ):
                return super().deserialize(data)
            attr_id, temp_data = t.uint16_t.deserialize(temp_data)
            attr = zigpy.zcl.foundation.Attribute(
                # attr_id, zigpy.zcl.foundation.TypeValue(t.uint8_t, int.from_bytes(temp_data, byteorder="big"))
                attr_id, zigpy.zcl.foundation.TypeValue(t.uint8_t, temp_data[1:])
            )
            return hdr, [[attr]]

```

The idea is the same as in my previous exposition, but keeping the bytes from position 13:
```
00 -> hold
01 -> release
0000001001 -> click
02 -> doubleclick
```



> ```
> BUTTON: "button_{}".format(self.endpoint.endpoint_id),
> ```
> 
> 
> you don't really need this. There's already `endpoint_id` present in the `zha_event` or `unique_id` identifying what button was pressed.
You're right.
I think that now I understand better all the part of the events and I would know how to handle them in another way, but for an average user it is more 'understandable' to manage a right or left button than to handle with `device_ieee` or` device_id`
In any case, I think that implementing the `device_automation_triggers` will make it easier for the average user, and the 'advanced' user will be able to handle events without these arguments. I will check it later.
```
    device_automation_triggers = {
        (COMMAND_SINGLE, BUTTON): {COMMAND: LEFT_SINGLE, ENDPOINT_ID: 4},
        (COMMAND_DOUBLE, BUTTON): {COMMAND: LEFT_DOUBLE, ENDPOINT_ID: 4},
        (COMMAND_HOLD, BUTTON): {COMMAND: LEFT_HOLD, ENDPOINT_ID: 4},
        (COMMAND_RELEASE, BUTTON): {COMMAND: LEFT_RELEASE, ENDPOINT_ID: 4},
    }
```

javicalle:
Another question, any advice or guidance about how make that HA don't generate entities for the new endpoints.
I have checked the MijaButton class that have the desired behavior, but I'm not able to see how it works:
https://github.com/zigpy/zha-device-handlers/blob/8379e1d18bd5eaeb699a3da5cbf48a75eaf916d7/zhaquirks/xiaomi/mija/sensor_switch.py#L55

javicalle:
> so for the "click" it just sends two reports for the same attribute. with 0x00 -- press, followed by 0x01 -- release

It is exactly what I have been able to see in some of the tests that I have been doing. 
Thank you very much for the detailed explanation, it helps me a lot to understand the specification better.


> IMO the best approach is to modify zigpy to allow not throw when bool is 0x02, then you can get rid off the `deserialize()` work around and just proccess those in the general command.

I agree that the best approach is to review the 0x02 issue.

For my part, I can release the version of the PR (with the `device_automation_triggers` part), or adapt the last version with the `deserialize ()` (only to deal with the 0x02 case) just in case the issue does not go ahead or is delayed, do you think that it would be necessary?

Thank you very much for your help. I appreciate it.

javicalle:
I think the version is ready to merge.
Any change or improvement that is considered I will review it.

Thanks a lot.

VVlasy:
@javicalle Is it possible to implement this for switches with neutral wire? Or do those not support decoupled mode?

Adminiuga (approved):
Looks good. 
I'll merge this once I push the new zigpy version into ZHA, then we'd need to bump up zha-device-handlers dependency too

dmulcahey:
Lmk when ready and I’ll get a release ready tonight for zha-device-handlers

Adminiuga:
> Lmk when ready and I’ll get a release ready tonight for zha-device-handlers

Will do. 
Unrelated: thoughts about leaving just the `model` in the signature for the Tuya remotes, so we don't have to add every rebranded manufacturer?

dmulcahey:
I have no opinions... if it works that’s fine by me but it was my understanding that the diff mfg may have different command ids etc?

Adminiuga:
that is a concerns, but AFAIK for Tuya command have been consistent for the same models. Although apparently there are thermostats with the same model, but different cluster signatures and different commands :shrug: 
Maybe should just test for one remote for now?

javicalle:
> @javicalle Is it possible to implement this for switches with neutral wire? Or do those not support decoupled mode?
> 
> I have some and I am willing to test it.

According to blakadder it should be possible:
* https://zigbee.blakadder.com/Xiaomi_QBKG12LM.html

CtrlLn class must be adapted, but probably the same implementation can be used.
* https://github.com/zigpy/zha-device-handlers/blob/b12ebc37790af34253fce4a0c549d05a8ffb995c/zhaquirks/xiaomi/aqara/ctrl_ln.py#L38

In this case, the endoints to implement would be 5, 6 and 7

Adminiuga:
pushed new zigpy version

PR #720, opened 2021-01-12
Refactor Xiaomi battery remaining calculation

Adminiuga:
Turns out #678 was quite short sighted. With max/min range of 3.0v and 2.8v and resolution of 0.1v there are only three possible battery % remaining values: 100%, 50% and 0%

This PR partially reverts #678 and does the calculation based on the reported battery voltage in mV.
This PR also changes the threshold and now 
- 100% == 3.100V
- 0% == 2.820V

I wanted to report lower % remaining when we approached the 2.8V cut off value. With this threshold, Xiaomis reporting 2.830V translates into 3.5%. With previous thresholds, 2.83V was 15% which IMO a tad high. In my experience, with Xiaomis Weather sensors approaching 2.8V the communication became wonky with the tendency to drop off more often, so I think it is better to replace batteries earlier.


newlund:
Working fine now in HA release 2021.2.0.
Thank you! :)

PR #725, opened 2021-01-14
Add quick init support for round Mijia `lumi.sensor_switch`

puddly:
The node descriptor is the same as all others but it is a massive pain to get the device to send it.


puddly:
I believe only the Mijia weather sensor and a few Aqara sensors/remotes are left.

Issue #738, opened 2021-01-22
[Device Support Request] _TZE200_zivfvd7h TS0601 Qiumi/Siterwell QM-GS361A-H04 TRV


welshcoder:
**Is your feature request related to a problem? Please describe.**
I bought a Zigbee TRV off Amazon (https://www.amazon.co.uk/gp/product/B08357D6LJ/ref=ppx_yo_dt_b_asin_title_o01_s00?ie=UTF8&psc=1), marketed as a "Qiumi Smart TRV", but is labelled "GS361A-H04" by Siterwell Electronics on the device itself. It joined the network, but it did not pick up a quirk to make it work correctly in ZHA.

**Describe the solution you'd like**
I added ```("_TZE200_zivfvd7h", "TS0601"),``` to the MoesHY368 signature at line 180 in tuya/valve.py to make the TRV appear in HA. It seems to work fine - I can set the required temperature on the device, and it reports back - so you may want to consider adding this modification to that file.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
This is the log after I got after the change to valve.py.
```
[0xcb0b] Extending timeout for 0x06 request
[0xcb0b] Node Descriptor: NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)
[0xcb0b] Discovering endpoints
Tries remaining: 3
[0xcb0b] Extending timeout for 0x07 request
[0xcb0b] Discovered endpoints: [1]
[0xcb0b:1] Discovering endpoint information
Tries remaining: 3
[0xcb0b] Extending timeout for 0x08 request
[0xcb0b:1] Discovered endpoint information: SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=81, device_version=1, input_clusters=[0, 4, 5, 61184], output_clusters=[25, 10])
Unknown cluster 61184
[0xcb0b] Extending timeout for 0x09 request
[...]

PR #735, opened 2021-01-22
Use a "wildcard" Manufacturer for Tuya TS0042 and TS0043 quirks

Adminiuga:
This PR requires new `zigpy>=0.31.0` version and uses a wildcard quirk, matching only on the model and endpoint signature.

PR #740, opened 2021-01-23
Adding Qiumi/Siterwell QM-GS361A-H04 TRV

MattWestb:
Please close https://github.com/zigpy/zha-device-handlers/issues/738 then marge.

Mvh  Mattias

### DIFF
--- a/zhaquirks/tuya/valve.py
+++ b/zhaquirks/tuya/valve.py
@@ -179,6 +179,7 @@ class MoesHY368(TuyaThermostat):
             ("_TZE200_ckud7u2l", "TS0601"),
             ("_TZE200_kfvq6avy", "TS0601"),
             ("_TZE200_c88teujp", "TS0601"),
+            ("_TZE200_zivfvd7h", "TS0601"),
         ],
         ENDPOINTS: {
             1: {


TheJulianJES:
Since the model seems to always be ``TS0601``, would it make sense to also address this using https://github.com/zigpy/zha-device-handlers/pull/735?

MattWestb:
Its sound good but i dont knowing how to do that :-((

tube0013:
> Since the model seems to always be `TS0601`, would it make sense to also address this using #735?

I just got a pair of these - <https://zigbee.blakadder.com/Moes_AM43-0_45_40-ES-EB.html> 

I have yet to join them to ZHA where I'll need to figure out a quirk but reading the z2m code they also appear to use the TS0601 model...... so a very different device with same model id.

Adminiuga:
Fixes https://github.com/zigpy/zha-device-handlers/issues/738

Adminiuga:
> Since the model seems to always be TS0601, would it make sense to also address this using #735?

#735 is more of a "proof of concept" I have those two devices in prod, so was able to test it.
The same approach could be use for TS0601, just need to verify there are no other quirks with the exact same model and exact same "endpoint" signature (same input/output clusters, device type, endpoints etc)

If someone with the said valves could test it, then we could create another PR

MattWestb:
Tuya is using the same zigbee module (and the same loaded firmware) for switches and all other devices that is talking "tuyanish" and is using it as one communicating channel for tunneling the tuya MCU that sending more or less MQTT commands thru the tunnel.
So only looking on the zigbee module hardware config (module name) and not manufacture data its  very likela not working.

Adminiuga:
> I just got a pair of these - https://zigbee.blakadder.com/Moes_AM43-0_45_40-ES-EB.html

I wish blakadder was posting the device signature too. Can you join and post the device signature. That's exactly my hesitation, Tuya using the same model for a different device. Need to compare signatures.

tube0013:
> > I just got a pair of these - https://zigbee.blakadder.com/Moes_AM43-0_45_40-ES-EB.html
> 
> I wish blakadder was posting the device signature too. Can you join and post the device signature. That's exactly my hesitation, Tuya using the same model for a different device. Need to compare signatures.

added via an issue.... #744

blakadder:
> > I just got a pair of these - https://zigbee.blakadder.com/Moes_AM43-0_45_40-ES-EB.html
> 
> I wish blakadder was posting the device signature too. Can you join and post the device signature. That's exactly my hesitation, Tuya using the same model for a different device. Need to compare signatures.

I am now :D

PR #726, opened 2021-01-15
Better Moes HY368 (tuya) implementation

xonestonex:
I've implemented almost all of the device functionality based on the device point descriptions of Zigbee2Mqtt.
The manufacturer specific stuff is also mapped to the individual clusters as manufacturer specific attributes, this should make it easier to provide more functionality in home assistant in a later step.

PR #576, opened 2020-11-09
Add Saswell 88teujp thermostat

zhulik:
Supported features:
- Get and set system mode (heat/off)
- Get and set occupied heating setpoint (desired temperature)
- Get local temperature
- Battery: the device only reports battery alert state, so the battery percentage is either 100 or 0.

Inspired by the zigbee2mqtt device handler implementation.

Thanks for review!

PS: I'll add tests after fixing all possible comments to the code.

PR #751, opened 2021-01-26
Add support for saswell sea801-z01

ha-zig:
This device "_TZE200_c88teujp", "TS0601" was classified under MoesHY368 quirk wrongly, as the attributes are different. Added new classes for saswell, Note device does not report whether valve is open or closed, similar to other Tuya battery reporting is good or bad no in between. Also not nasty hack of SASWELL_STATE_ATTR and SASWELL_MODE_ATTR to same value to get buttons working on HA thermostat interface. Basic functionallity added here, turn on device to manual, off to off you can also set temperature setpoint, device also reports room temperature. More to add here later, possible to set to auto mode and remotely programme the schedule.

### DIFF
--- a/zhaquirks/tuya/valve.py
+++ b/zhaquirks/tuya/valve.py
@@ -32,6 +32,22 @@ SITERWELL_TEMPERATURE_ATTR = 0x0203  # [0,0,0,200] current room temp (decidegree
 SITERWELL_BATTERY_ATTR = 0x0215  # [0,0,0,98] battery charge
 SITERWELL_MODE_ATTR = 0x0404  # [0] off [1] scheduled [2] manual
 
+#saswell variant
+#see protocol_TRV-Z01-80__20200901.pdf
+#see https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/lib/tuya.js
+
+SASWELL_WINDOW_DETECT_ATTR = 0x0108  # [0] inactive [1] active
+#SASWELL_FROST_DETECT_ATTR = 10 # [0] inactive [1] active #doesnt appear work
+SASWELL_TEMPERATURE_CALIBRATION_ATTR = 0x000 #0x0102 possiby
+SASWELL_CHILD_LOCK_ATTR = 0x0128 # possibly 0x0128  # [0] inactive [1] active
+SASWELL_STATE_ATTR = 0x0165 #[0] off [1] should be 0x016c but hack to make on off work instead ofenabling auto
+SASWELL_LOCAL_TEMP_ATTR = 0x0266 # [0,0,0,200] current room temp (decidegree)
+SASWELL_HEATING_SETPOINT_ATTR = 0x0267 # [0,0,0,210] target room temp (decidegree)
+SASWELL_VALVE_POSITION_ATTR = 0x0268 #0x0268 possibly
+SASWELL_BATTERY_LOW_ATTR = 0x0569 # [0] OK [1] LOW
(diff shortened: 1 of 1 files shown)


Adminiuga:
@ha-zig can you post your device signature for posterity here?

@fastcolors seems like you've added the same model in #710 Do we have two different devices with the exactly same model/manufacturer or does this new PR works good with your device?

ha-zig:
@fastcolors that patch does not work for me on, debug shows different attribs, device sig:

{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0004",
        "0x0005",
        "0x0201",
        "0x0204",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_c88teujp",
  "model": "TS0601",
  "class": "zhaquirks.tuya.valve.SaswellSEA801"
}

example of valve

https://www.aliexpress.com/item/1005001774162057.html

changes work for me for basic functionality, device is slow to send updates after mode change (it will change straight away but wont send the mode change back often for upto a minute)

MattWestb:
The timing problem you can very likely fixing with this: https://github.com/zigpy/zha-device-handlers/pull/726#issuecomment-767589485 and the comment a above it.

MattWestb:
@Adminiuga I have trying doing one matrix in the Wiki https://github.com/zigpy/zigpy/discussions/653#discussioncomment-314395 and its looks being OK then following the commits backwards. I think  its easier than having pictures for the user saying its right or not (some is working OK with wrong quirk and can throwing errors but working).

Adminiuga:
Unless there are different models with the same model/manufacturer and two users report conflicting data.

MattWestb:
@Adminiuga     **This is being one big MESS !!**

The MoesHY368 was working but luck of futures and being updated PR #726  is waiting and ("_TZE200_ckud7u2l", "TS0601"), is one true MoesHY368 and being moved to the "new MoesHY368" in the PR #726 .

("_TZE200_kfvq6avy", "TS0601"), Revolt NX-4911 II and ("_TZE200_zivfvd7h", "TS0601"), Siterwell GS361A II was in the MoesHY368 because new zigbee module and was working but cant being moved to the "new MoesHY368" then its dont have all the extras the new inplanting so its being renamed in the same PR to Siterwell  II in the PR and the "old MoesHY368" away PR #726 .
This devises looks working but can complaining of time setting is not working.

("_TZE200_c88teujp", "TS0601"), Saswell SEA801 1 was also in the MoesHY368 but was working 
bad PR #710 . 
One new PR #751 is made and shall moving the fingerprint from the "old MoesHY368".

I have verified all IDs with deCONZ and Z2M and also with photos of devices that is verified by users having the devices.

Sorry for all the problems but then the "descrambling" is made all shall being in the right place.
That is missing is updating the original Siterwell that only having basic functions (not urgent but need being done).

And its also one or 2 uncommon models that i shall putting in the matrix with picture and information of providing issues for adding them **with device signature** that was the large problem then users have adding them without no information at all in the PRs (#710).

Mvh  Mattias

dmulcahey:
Needs a rebase

ha-zig:
#576`removes need for this

PR #763, opened 2021-02-01
Expand usable pressure range of Aquara weather sensor

### LINKED ISSUE #762: [BUG] ZHA aquara pressure sensor is not updating
**Describe the bug**
I'm seeing update messages in the debug logs, but in the pressure is not updating in the ui

**Expected behavior**
measurements in the UI should match values in the logs.

**Screenshots**

image
.

**Additional context**
```
{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4151, maximum_buffer_size=127, maximum_incoming_transfer_size=100, server_mask=0, maximum_outgoing_transfer_size=100, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0302",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0402",
        "0x0403",
        "0x0405",
        "0xffff"
      ],
      "out_clusters": [
        "0x0000",
        "0x0004",
        "0xffff"
      ]
    }
  },
  "manufacturer": "LUMI",
  "model": "lumi.weather",
  "class": "zhaquirks.xiaomi.aqara.weather.Weather2"
}
```

```
2021-01-31 21:27:36 DEBUG (MainThread) [zigpy.zcl] [0x0006:1:0x0403] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=False disable_default_response=True> manufacturer=None tsn=143 command_id=Command.Report_Attributes>
2021-01-31 21:27:36 DEBUG (MainThread) [zigpy.zcl] [0x0006:1:0x0403] ZCL request 0x000a: [[Attribute(attrid=0, value=<TypeValue type=int16s, value=136>), Attribute(attrid=20, value=<TypeValue type=int8s, value=-1>), Attribute(attrid=16, value=<TypeValue type=int16s, value=1360>)]]
2021-01-31 21:27:36 DEBUG (MainThread) [zigpy.zcl] [0x0006:1:0x0403] Attribute report received: measured_value=136, 20=-1, 16=1360
```

devin122:
While the device specs list the pressure sensor going down to 30kPa,
the device still generates usable data below that.

fixes #762

### DIFF
--- a/zhaquirks/xiaomi/__init__.py
+++ b/zhaquirks/xiaomi/__init__.py
@@ -393,9 +393,9 @@ class PressureMeasurementCluster(CustomCluster, PressureMeasurement):
         self.endpoint.device.pressure_bus.add_listener(self)
 
     def _update_attribute(self, attrid, value):
-        # drop values above and below documented range for this sensor
+        # drop unreasonable values
         # value is in hectopascals
-        if attrid == self.ATTR_ID and (300 <= value <= 1100):
+        if attrid == self.ATTR_ID and (0 <= value <= 1100):
             super()._update_attribute(attrid, value)
 
     def pressure_reported(self, value):


dmulcahey:
Have you done anything to verify that the lower values are accurate? I’m not sure going beyond the specified supported ranges of the device makes sense... @Adminiuga what do you think?

Adminiuga:
I bet Xiaomi did not expect their users to use that sensor inside a vacuum packaging :) Just need to wait till someone tries to monitor the temperature inside a pressure cooker. 
But on a serious note, I'm fine with filtering the obviously wrong values, like `-1`. Probably we could filter out the `0` too, unless we expect someone to use those in the full vacuum. I'm on the side of: lets pass the values (like we do for other vendors) even if those are outside of the vendor official specification and let the user filter out the values they don't like.

devin122:
I don't have any way of measuring the actual accuracy, but the numbers I'm seeing seem plausible.
Pumping down the  vacuum bag + gradual leak:
image

The small hiccup at the beginning was me opening the bag slightly to debug the issue of why the pressure seemed to have stalled at ~300hPa. 
The jump from ~600 to ~100hPa was when I applied this PR.

Adjusting date-range to crop out the initial pump down show a mostly linear trend upward, with the deviations correlating with fluctuations in temperature.
image

I'm not sure of the actual physics involved here, but intuitively it seems reasonable.

PR #700, opened 2021-01-05
Fix button n°2 for Xiamo Aqara Opple remote 6

jokyboy25:
Fix detection of button n°2 for Xiamo Aqara Opple Remote 6 V4.
This can also apply to other version but I don't have the device to check if it's working.

PR #766, opened 2021-02-05
Fix detection of button n°2 for Xiamo Aqara Opple Remote 6 V3

ton6i:
Fix detection of button n°2 for Xiamo Aqara Opple Remote 6 V3.
Same solution as https://github.com/zigpy/zha-device-handlers/pull/700.
Please change (line 977) INPUT_CLUSTERS: [Identify.cluster_id], to INPUT_CLUSTERS: [MultistateInputCluster.cluster_id, Identify.cluster_id], 
Tested this and now button 2 works.
Many thanks.


Adminiuga:
This does not look like a PR.
Either submit a proper PR (the one with the actual changes) or open an issue. But in both cases submit the device "zigbee information" as reported by ZHA, link to where the device was purchased and if possible the values for the `software_build_id`, `date_code`  attributes etc

Adminiuga:
@dmulcahey we should start asking for the devices signature and some relevant meta information for each new quirk/issue as if it is the same device, but new firmware changes the signature then we're going to have some unhappy users one way or another. 
And of course, Xiaomi is different from everyone else, cause it doesn't have a static endpoint configuration.

ton6i:
I will.
Sorry for my mistake.

PR #796, opened 2021-02-28
Deleting  "_TZE200_c88teujp", "TS0601"

MattWestb:
Deleting  "_TZE200_c88teujp", "TS0601" from Siterwell	TRV then its one Saswell	SEA802 that have being implanted in zhaquirks/tuya/thermostat_88teujp.py #576 
It was added wrongly in #710 and have making trouble.


MattWestb:
Thanks !!

PR #799, opened 2021-03-01
Add new MODELS_INFO for BlitzWolf Tuya plugs

javicalle:
It seems like there are several MODELS_INFO for BlitzWolf Tuya plugs.
Here added the one from my 16A EU Plug


dmulcahey (approved):

javicalle:
Wow!
Thanks.

PR #800, opened 2021-03-03
Implementation of AU-A1ZBR2GW

bens0n:
Initial implementation of 2-gang dimmer switch AU-A1ZBR2GW, as per issue https://github.com/zigpy/zha-device-handlers/issues/795. Still to do: 
- choose suitable translation strings for the automation key on color temperature automation
- prevent useless entities being created (presently one switch per endpoint)


bens0n:
Is there a route forward for this one? I can't see a simple solution for the "color_up" and "color_down" strings without adding extra strings for translation. They could be derived from "dim_up" and down , plus some of the color commands, but only if some assumptions about word order are made. 

In terms of preventing the creation of junk entities, it appears that this would require changes to both this project and zha itself (autodiscovery appears to be in https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/core/discovery.py). 

Any advice on proposed routes forward gratefully received!

dmulcahey:
@Adminiuga do we need to add another flag like skip_configuration for this? the problem is that the device leverages reporting on input clusters instead of leveraging output clusters correctly. So, if we remove the input clusters on the replacement (to prevent useless entities from getting created) the output clusters don't go through binding and report configuration so the device will never work. What should we do here? We could add a flag to the output clusters to make them go through bind and configuration or we could use a flag to prevent entity discovery for the bogus input clusters... thoughts?

Adminiuga:
We could change the ep_attribute name for the server clusters? The same way we do it for Tuya? Alternative is to mask the vendor or model in zha discovery but I'm not really fond of this approach

dmulcahey:
Is anything else needed here or are we good to go?

bens0n:
The quirk works, I've been using it for the past month no problem. Beyond the issues mention in the OP (creation of junk entities, translation strings) the implementation itself works well.

Adminiuga:
in `class WallSwitchOnOffCluster(EventableCluster, OnOff)` class override `ep_attribute = "not_on_off"`

bens0n:
Implemented the suggested fix. I'm talking to aurora RE mixed up clusters, but no movement so far - just being bounced around. Will update if anything material comes of it (eg updated firmware etc)

bens0n:
For interest, Aurora finally got back to me: "Our development team is on part time furlough so any work is extremely slow at the moment.

We are looking at a way to support HomeAssistant but currently have no plans or ETA to do so.
Once we are ready to reveal any plans we will send out an email to all of our customers."

zhaquirks/aurora/aurora_dimmer.py
```diff
@@ -0,0 +1,234 @@
...
+    Basic,
+    GreenPowerProxy,
+    Identify,
+    LevelControl,
+    OnOff,
+    Ota,
+)
+from zigpy.zcl.clusters.lighting import Color
+
+from .. import EventableCluster, PowerConfigurationCluster
```
dmulcahey, line 14: can you expand the imports so they aren't relative please?

bens0n (commented):
Updated

dmulcahey (approved):

PR #802, opened 2021-03-03
Implement custom quirk loading

puddly:
This change implements a folder for custom quirks that survive zha-device-handlers updates. For example:

```yaml
zha:
  zigpy_config:
    # maybe this would go a level higher?
```

Creating `/config/custom_zha_quirks/__init__.py` and quirk submodules will allow them to take priority over the built-in ones, allowing quirks to be subclassed, overridden, and maintained independently without having to patch the `zhaquirks` Python module directly.

One notable difference is that these custom quirks are not a part of the `zhaquirks` package. They will not be able to use relative imports, so while this will not work:

```Python
from . import BOSCH
from ..const import foo
```

This will:

```Python
from zhaquirks.bosch import BOSCH
from zhaquirks.const import foo
```

This won't change their functionality when merged into `zhaquirks` but does differ from the current coding style of using relative imports.

~~Right now the newly-defined `setup` function is called immediately after it is defined, maintaining backwards compatibility.~~ ZHA will have to be modified to call `zhaquirks.setup` with the relevant section of its configuration dictionary once this "shim" is removed. This will be a breaking change for any application relying on `import zhaquirks` mutating global state.

---

~~At the moment there are no unit tests included because I can't find any way to "properly" write any, beyond ensuring that `pkgutil.walk_packages` will be called. This is due to the metaclass-based quirk registration method currently used by zigpy and there being no way to un-register  and then re-register quirks without a bit of hackery.~~

### DIFF
--- a/tests/test_ctrl_neutral1.py
+++ b/tests/test_ctrl_neutral1.py
@@ -1,7 +1,10 @@
 """Tests for xiaomi."""
 from unittest import mock
 
+import zhaquirks
 from zhaquirks.xiaomi.aqara.ctrl_neutral import CtrlNeutral
+
+zhaquirks.setup()
 
 # zigbee-herdsman:controller:endpoint Command 0x00158d00024be541/2 genOnOff.on({},
 #   {"timeout":6000,"manufacturerCode":null,"disableDefaultResponse":false})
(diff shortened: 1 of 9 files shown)


dmulcahey:
I like it... this looks good to me.

Adminiuga (approved):
Lgtm

puddly:
Since `zhaquirks` isn't really a part of zigpy, I've dropped the config key down a level:

```YAML
zha:
```

dmulcahey (approved):

PR #809, opened 2021-03-08
Update README.md

MattWestb:
Adding "deleting /config/deps/" after testing components for getting away lib miss matches.


MattWestb:
Sorry i was editing the wrong branch. Its only the Redme.md that is meant for merger.
I trying deleting the other if i can.

MattWestb:
Was not able deleting the by mistake added changes. so reposing the first changes in one new PR.

PR #811, opened 2021-03-08
Update README.md

MattWestb:
Adding "deleting /config/deps/" after testing components for getting away lib miss matches.

Second try. I deleting the first one

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,8 +450,8 @@ If you are instead using some custom python installation of Home Assistant then 
   ```bash
   pip install zha-quirks==0.0.38
   ```
-After installing components with both methods above you need deleting the installed component by hand then  you is updating your HA installation or its override the new HA files and you can getting lib miss matching.  
-Execute in your HA docker container CLI:
+After installing components with the methods listed above you will need to delete the modified files by hand. Other libraries may also be affected so it is best to remove them all.  
+To do this execute this command in the HA docker container CLI
 ```bash
 rm -r  /config/deps/
 ```
--- a/README.md
+++ b/README.md
@@ -450,6 +450,11 @@ If you are instead using some custom python installation of Home Assistant then 
   ```bash
   pip install zha-quirks==0.0.38
   ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -450,6 +450,11 @@ If you are instead using some custom python installation of Home Assistant then
   ```bash
   pip install zha-quirks==0.0.38
   ```
+After installing components with both methods above you need deleting the installed component by hand then  you is updating your HA installation or its override the new HA files and you can getting lib miss matching.
```
dmulcahey, line 453: After installing components with the methods listed above you will need to delete the installed component by hand. To do this execute this command in the HA docker container CLI

MattWestb: I  hope i have doing it right !!

Thanks for correcting my bad eng....

README.md
```diff
@@ -450,6 +450,11 @@ If you are instead using some custom python installation of Home Assistant then
   ```bash
   pip install zha-quirks==0.0.38
   ```
+After installing components with the methods listed above you will need to delete the modified files by hand. Other libraries may also be affected so it is best to remove them all.  
+To do this execute this command in the HA docker container CLI
```
Adminiuga, line 454: no need to run it in the docker container, as it it your `/config` directory which is mounted to docker

MattWestb: Very true but most user dont knowing where to finding it and the directory is hidden for the file editor that most user have installed ad-don so they cant deleting it from it. And some user is using the HASS.IO that is very locked.

Adminiuga: it's not a hidden directory and regular users most likely would have access to `/config` directory, than dev ssh access to docker.

MattWestb: Sorry i was meaning the `/deps/` and the last time i was looking its hidden in the file editor in HA (not in the file system), and most "normal" users dont knowing that its `/usr/share/hassio/homeassistant/deps` in most OS and always /config/ in docker that was intended as the installed component.
ZHA_DEPS
Changing wot you like it to be, i cant saying its right or wrong, and i was only trying getting its easier for the user getting back to one "normal" system after have installed components with pip, and you is hopefully getting lesser issues that can being solved with informative Wiki pages (that i cant writing because of my bad gramma).

PR #812, opened 2021-03-09
IKEA remotes power drain in ZHA

echauvet:
This allows to keep battery life in ZHA. More details in the topic : https://github.com/home-assistant/core/issues/43969

### DIFF
--- a/zhaquirks/ikea/dimmer.py
+++ b/zhaquirks/ikea/dimmer.py
@@ -73,7 +73,6 @@ class IkeaDimmer(CustomDevice):
                     Basic.cluster_id,
                     DoublingPowerConfigurationCluster,
                     Identify.cluster_id,
-                    PollControl.cluster_id,
                     LightLink.cluster_id,
                 ],
                 OUTPUT_CLUSTERS: [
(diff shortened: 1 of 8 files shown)


echauvet:
This PR is not needed anymore. The problem is solved using last version of HASS/ZHA and upgraded SonOff Bridge

patrickfong:
echauvet, I am using home assistant 2021.3.4 with zbbridge ncp-uart-nsw_6.7.9_115200, 9.3.1.2(zbbridge), but still seeing significant drain.  I have six buttons and some of them last a few days, but there are ones that last 24 hours.  I didn't see zigpy changes related to battery drain and there were no changes to ikea quirks.  Can you elaborate how th issue was fixed?

Thanks..!

echauvet:
@patrickfong Yes you are right, same for me after a few days, I reopened the PR.

dmulcahey:
Please split these changes into separate PR's. Ikea stuff in one and the other device in the other please.

dmulcahey:
@Adminiuga should we even do the IKEA part of this? I don't think this is the right solution for that problem...

Adminiuga:
No, we shouldn't do it an zha-device-handlers. Should do it in zha in poll channel and blacklist based on Ikea manufacturer Id -- skip changing the poll interval.

MattWestb:
I think its not needed setting "new" long pull interval after every end of check in its also possible setting short pull with time out (and then its going back to long pull after time is expired). But is also can being dangerous if the long pull is not correct stetted up in the device its can stay in short pull (the deCONZ fix of smartthings batteries draining mented in the main issue).

dmulcahey:
@Adminiuga you did the HA part of this right?

Adminiuga:
This was implemented in zha for all ikea's, instead of changing every quirk. No need for that quirk changes

PR #816, opened 2021-03-14
Add new Tuya based RGB+CCT Bulb (TS0505A)

iglu-sebastian:
Most work based off https://www.zigbee2mqtt.io/devices/TS0505A.html

Needs inter-cluster communication since it switches between Tuya's RGB modes. Works for me here.


iglu-sebastian:
Tests seem to fail caused by `import homeassistant.util.color as color_util` - is there a better way to convert HA's MOVE_TO_COLOR XY to HS directly by forcing a capability mask?

dmulcahey:
You can not import HA classes into this lib.

PR #827, opened 2021-03-24
Switch all relative imports to absolute imports

puddly:
To streamline quirk development and make it easier for new users, I've replaced all relative imports in all quirk submodules with absolute imports. I've also added a rudimentary unit test to enforce this as I couldn't find a way to have pylint or flake8 do this automatically.

The new process (in theory) will just be:

1. Configure HA to load custom quirks:
   ```YAML
   zha:
   ```
2. Copy/paste or download an existing quirk module and save it as `/config/custom_zha_quirks/new_quirk.py`. Relative imports are now gone so this will work even for deeply nested modules like `zhaquirks.xiaomi.aqara.*`.
3. Edit the quirk (e.g. by mounting the `/config/` folder over SMB) and restart HA. Existing quirks can be tweaked because custom quirks always take priority. New quirks can be tested without messing with Docker.

### DIFF
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -17,7 +17,7 @@ from zigpy.zcl.clusters.measurement import OccupancySensing
 from zigpy.zcl.clusters.security import IasZone
 from zigpy.zdo import types as zdotypes
 
-from .const import (
+from zhaquirks.const import (
     ATTRIBUTE_ID,
     ATTRIBUTE_NAME,
     CLUSTER_COMMAND,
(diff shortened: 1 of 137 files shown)


puddly:
This change will break just about every existing pull request (since they all use the current relative import convention) so I'm not too sure if it's better to have individual maintainers fix their PRs or just remove the unit test for now and add it back in once the PR backlog has been cleared.

dmulcahey:
> This change will break just about every existing pull request (since they all use the current relative import convention) so I'm not too sure if it's better to have individual maintainers fix their PRs or just remove the unit test for now and add it back in once the PR backlog has been cleared.

Folks can rebase... it isn't that big of a change for ppl to take into their PR's IMO.

MattWestb:
I agree with @dmulcahey but as one no so experience user of github it can being good if some is helping explaining how to rebase one repro that is changed so not all users is deleting the commitments and not adding it back and we is missing some code changes that is god for our systems.

PR #840, opened 2021-04-10
Fix MODELS_INFO typo for TH1300ZB and TH1400ZB

brianneate:
I tested the TH1300ZB locally and before this change it was not getting the quirks applied while after the change it was. I was not able to test the TH1400ZB as I do not own one however assume the same issue applies.

The device signature for the TH1300ZB is:
2021-04-09 20:33:27 INFO (MainThread) [zigpy.endpoint] [0x96e2:1] Discovered endpoint information: SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=769, device_version=1, input_clusters=[0, 3, 4, 5, 513, 516, 1026, 1794, 2820, 2821, 65281], output_clusters=[10, 65281, 25])


Adminiuga:
Can you post signature for your device in the 1st message?

claudegel:
This PR just clear a typo I've made when I've added TH1300ZB and TH1400ZB devices to thermostat.py
devices were added as TH11300ZB instead of TH1300ZB, TH11400ZB instead of TH1400ZB.
With that fix both devices load properly

Adminiuga (approved):
Lgtm

PR #841, opened 2021-04-11
Fix Remote AT encryption

Shulyaka:
Explicitly disable remote AT command encryption. As per the specs, it requires a secure session to be established with the destination, which we don't currently do.


Adminiuga (approved):
Lgtm

PR #842, opened 2021-04-12
Update README.md

Hedda:
Update README.md to change project description from "ZHA Device Handlers For Home Assistant" to "ZHA Device Handlers For zigpy" now that other applications such as for example Jeedom also depends on this library.

https://github.com/zigpy/zigpy/discussions/725


Hedda:
Closed without comment. What was wrong with this change?

Is the zha-device-handlers lib really only for Home Assiatant?

Issue #831, opened 2021-03-30
[BUG] Tuya/MEOS Thermostat - the time gets unsynchronized making built-in schedules unusable.


sstefanowski:
The problem:
The Tuya/MEOS Thermostat support built-in heating schedules (and change temperature up to 6 times a day). 
Unfortunately, the clock inside the device is not very precise and the time gets unsynchronized after a few days. 
ZHA should take care of the device time and re-synchronize it with some default frequency (e.g. once a day).
The device should always receive a local date/time from ZHA. This also should be taken into consideration when summer/winter time changes

Similar Zigbee solutions take care of the device clock e.g. Tuya platform and Zigbee2MQT.\
The solutions for the same issue in Zigbee2MQTT:
https://github.com/Koenkk/zigbee2mqtt/issues/5916
https://github.com/Koenkk/zigbee2mqtt/issues/5620

**To Reproduce**
Steps to reproduce the behaviour:
1. Set a correct time on the Tuya/MEOS Thermostat device manually
2. Leave it for a few days
3. Check the time on the device after a few days
4. The time on the device is no longer correct

**Expected behaviour**
When checking the time on the device it should be always correct. ZHA should take care of date/time settings and re-synchronize them periodically as the other Zigee platforms do.
The device should always receive a local date/time from ZHA. This also should be taken into consideration when summer/winter time changes

**Screenshots**
N/A

**Additional context**
N/A

Adminiuga:
Tuya should have just sticked to the standards.  Nominally nthe device is a zigbee device, but really is using a proprietary protocol.

jmoransalama:
@sstefanowski What thermostat model do you have? I have a _TZE200_aoclfnxz thermostat and I'm having the same problem.

I'm not sure, but I believe that the command to set the time from ZHA to the device is [already in the code](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L113), but it's only applied to some thermostatic valves and not to all tuya thermostats.

jmoransalama:
I was looking at the code and I think I've fixed it for my device. I've just added this line of code:
`set_time_offset = 1970`
in [this line](https://github.com/zigpy/zha-device-handlers/blob/2f9b0c33e08dfbddb703e44ce017e0a0a9315562/zhaquirks/tuya/electric_heating.py#L87) 

So now the code looks like this:
```
class MoesBHTThermostat(TuyaThermostatCluster):
[...]

PR #847, opened 2021-04-14
Fix for MEOS TRV Time setting

MattWestb:
From https://github.com/zigpy/zha-device-handlers/issues/831#issuecomment-811776478
User have reported that its working changing the time request and the TRV is getting the right time then its pulling the coordinator for it.
I dont know if its doing more bad things but for the moment its only the MOES TRV that is using the time setting (i think its made different for other models but is for the moment not used / implanted in  other then the MOES TRV).

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -132,7 +132,7 @@ class TuyaManufCluster(CustomCluster):
     manufacturer_client_commands = {
         0x0001: ("get_data", (Command,), True),
         0x0002: ("set_data_response", (Command,), True),
-        0x0024: ("set_time_request", (t.NoData,), True),
+        0x0024: ("set_time_request", (t.data16,), True),
     }
 
     def handle_cluster_request(
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -132,7 +132,7 @@ class TuyaManufCluster(CustomCluster):
     manufacturer_client_commands = {
         0x0001: ("get_data", (Command,), True),
         0x0002: ("set_data_response", (Command,), True),
-        0x0024: ("set_time_request", (TuyaTimePayload,), True),
+        0x0024: ("set_time_request", (t.NoData,), True),
     }
 
     def handle_cluster_request(


MattWestb:
I think only @xonestonex can answer if its good or bad to changing it but i think its OK then its working as expected for MOES TRV users.

MattWestb:
Second try now with `("set_time_request", (t.data16,), True),`

sstefanowski:
Hi @MattWestb,

Just to explain/justify the change...

I think the previous entry using `TuyaTimePayload `for `set_time_request` was a real bug. 
The `set_time ` (going from cordinator to the device) should just `TuyaTimePayload` but `set_time_request `(from the Device to Coordinator) should not.

This is the Tuya Dev manual for time synchronization
https://developer.tuya.com/en/docs/iot/device-development/embedded-software-development/mcu-development-access/zigbee-general-solution/tuya-zigbee-module-uart-communication-protocol?id=K9ear5khsqoty#title-18-Time%20synchronization
 
From the tables describing both requests in the documentation attached it looks like the payload in `set_time_request` has no meaning.

MattWestb:
@sstefanowski can you using the last commit and testing if it working and you is intending and is setting the time in the MOES TRV ?

sstefanowski:
OK. I've upgraded to the latest HA 2021.4.4 and replaced a single file from this commit in the installation.
Before the change, I got a well-known issue - incorrectly parsed correctly `set_time_request`

`Apr 14 17:43:44 raspberrypi hass[1390]: 2021-04-14 17:43:44 ERROR (MainThread) [zigpy.device] Failed to parse message (b'194a240422') on cluster 61184, because Data is too short to contain 1 bytes`

and after replacing the file from this commit the ERROR is gone and new entries appearing:
`Apr 14 18:00:52 raspberrypi hass[3447]: 2021-04-14 18:00:52 DEBUG (MainThread) [zhaquirks.tuya] [0x0803:1:0xef00] Got set time request (command 0x0024)`

It looks it works fine.

MattWestb:
Great !!

Hope its enough for the devs merging the PR then its looks OK :-)))

sylvaing:
works for me also

MattWestb:
Great 2 brave user is getting the (TRV) time going in the right direction :-)))

PR #849, opened 2021-04-14
Fix triple press in Aqara button

wes1993:
Fixed the triple press in Aqara sensor that won't work.


wes1993:
I have changed the file in my HA instance, in this path:
/usr/local/lib/python3.8/site-packages/zhaquirks/xiaomi/aqara/switch_aq2.py

And now works very well :-D

Issue #503, opened 2020-09-25
[Device Support Request] icasa Pulse 8S


PaulAntonDeen:
**Is your feature request related to a problem? Please describe.**
icasa Pulse 8s is not recognized by zha

**Describe the solution you'd like**
Being able to switch lights with the device

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
```{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4644, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0001",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0b05"
      ],
      "out_clusters": [
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0019",
        "0x0300",
        "0x1000"
      ]
    }
  },
  "manufacturer": "icasa",
  "model": "ICZB-KPD18S",
  "class": "zigpy.device.Device"
}
```

**Additional context**
I got working on the quirk by adapting the IKEA 5 button remote but am now stuck:

```
class IcasaPulse8S(CustomDevice):
    """Custom device representing icasa pulse 8s remote control."""

    signature = {
        # input_clusters=[0, 1, 3, 2821]
        # output_clusters=[3, 4, 5, 6, 8, 25, 768, 4096]>
        MODELS_INFO: [(ices, "ICZB-KPD18S")],
        ENDPOINTS: {
            1: {
                PROFILE_ID: zll.PROFILE_ID,
                DEVICE_TYPE: zll.DeviceType.SCENE_CONTROLLER,
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    PowerConfiguration.cluster_id,
                    Identify.cluster_id,
                    DIAGNOSTICS_CLUSTER_ID,
                ],
                OUTPUT_CLUSTERS: [
                    Identify.cluster_id,
                    Groups.cluster_id,
                    Scenes.cluster_id,
                    OnOff.cluster_id,
                    LevelControl.cluster_id,
                    Ota.cluster_id,
				#colour control,
[...]

PR #855, opened 2021-04-16
Added support for icasa ICZB-KPD18S

Emrvb:
I have made an initial implementation for support for the icasa ICZB-KPD18S wall mounted zigbee remote. This remote is mentioned here https://github.com/zigpy/zha-device-handlers/issues/503

This remote has ON and OFF buttons (hold for dimming up/down) and 6 SCENE buttons (hold for storing, push to recall).

I myself am not very familiar with zigpy in general and the device handlers in particular, so I'm left with a few questions I would like answered in order to complete this device handler.

1. To me it seems all clusters are working according to spec. Do I need to specify a replacement?
2. Should I specify endpoint_id and cluster_id in device_automation_triggers?
3. I'm a bit uncomfortable with the BUTTON_X, LONG_PRESS and SHORT_PRESS abuse. In my opinion the commands "store" and "recall" are pretty distinct and the buttons should actually be called "Scene X". How should I proceed here?
4. Hold ON and OFF activates dimming. When released we don't know which button was released. Do we somehow keep track of which button was pressed initially? I copied this part from another device implementation, hence the question.

Cheers!


Hedda:
@Emrvb FYI, if it helps there is a relative PR with button mappings that deCONZ use for iCasa Pulse 2, Pulse 4S and Pulse 8S here:

https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4160

Emrvb:
@Hedda,

Thank you for the reference, I will add those devices.

What do you think I should do to get my questions answered, raise issues?

Hedda:
> What do you think I should do to get my questions answered, raise issues?

Yes best is probably to just raise issue specific to this/these device(s) -> https://github.com/zigpy/zha-device-handlers/issues

PR #859, opened 2021-04-19
Add support for Zemismart Curtain Motor TZE200_3i3exuay

holdestmade:
Add support for Zemismart Curtain Motor TZE200_3i3exuay

The Descriptor is slightly different than others in TS0601.py so made this one and tested it works fully


dmulcahey (approved):

Arean71:
I have an issue with using zha/zhaquirks TS0601_TZE200_3i3exuay
The commands open and close works in reverse mode.
I means that command open - maded curtaines close

PR #874, opened 2021-05-01
Add Tuya based smoke detector (TS0601 _TZE200_aycxwiau)

dlq84:
Add support for Tuya based smoke detector. This one in particular can be found as Woox R7074 Smoke Alarm Kit ([https://wooxhome.com/woox-r7074-smoke-alarm-kit-p43](https://wooxhome.com/woox-r7074-smoke-alarm-kit-p43))

As a start all it can do is tell if smoke is detected or not. It is possible to detect state when a smoke detector test is triggered by the button on the device itself, but I did not implement that since it doesn't seem to have any impact in HA... It's also possible to trigger the test mode via Zigbee commands (doesn't seem to be supported by Zigbee IAS Zone).

Feedback is much appreciated. And if someone knows if it's possible to get battery level on these kinds of tuya devices I'd like to know that so I can add it.

### DIFF
--- a/zhaquirks/const.py
+++ b/zhaquirks/const.py
@@ -99,3 +99,4 @@ VALUE = "value"
 ZHA_SEND_EVENT = "zha_send_event"
 ZONE_STATE = 0
 ZONE_TYPE = 0x0001
+ZONE_STATUS = 0x0002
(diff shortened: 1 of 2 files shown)


dlq84:
Rather unintuitively Tuya/Woox in their endless wisdom decided that 1 means _clear_ and 0 means _detected_. I have verified this in the real world now and fixed this in the latest update to this PR.

Adminiuga (approved):

PR #887, opened 2021-05-17
Yet another variation of the ZM25TQ

blauret:
This one has the time cluster in the output where the other one have it in the input.


blauret:
@dmulcahey,any reason why this one is not merged?

PR #888, opened 2021-05-17
Additional Generic Curtain Motor Using Tuya Board  ( _TZE200_nogaemzt, TS0601 )

Walmtech:
Additional Motor looks same as others different manufacturer like https://www.zigbee2mqtt.io/devices/TS0601_cover.html renamed to be more generic and added additional input cluster missing from original.    

Fully tested with my device and working fine.


Walmtech:
I have another new device not in the list so will add that and resubmit.

PR #894, opened 2021-05-23
Add new Konke magnet sensor model.

mguaylam:
I’ve purchased one of those Konke magnet and they seem to have updated the model reference from 3AFE270104020015 to 3AFE280104020015.
Needed for the battery cluster.


Adminiuga:
you may want to install the black and do the formatting with black

mguaylam:
Thank you for the formatting, will do it next time. 🙂

PR #902, opened 2021-05-28
Add Tuya Covering TS130F manufacturer

axellebot:
Following https://github.com/zigpy/zha-device-handlers/issues/833#issuecomment-850381806

Adding to the existing class : 
- _TZ3000_vd43bbfq
- _TZ3000_egq7y6pr

Create new class for Zemismart device (different signature) :
- _TZ3000_ltiqubue

### DIFF
--- a/zhaquirks/tuya/ts130f.py
+++ b/zhaquirks/tuya/ts130f.py
@@ -68,7 +68,11 @@ class TuyaTS130F(CustomDevice):
 
     signature = {
         # SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=0x0202, device_version=1, input_clusters=[0, 4, 5, 6, 10, 0x0102], output_clusters=[25]))
-        MODELS_INFO: [("_TZ3000_8kzqqzu4", "TS130F")],
+        MODELS_INFO: [
+            ("_TZ3000_8kzqqzu4", "TS130F"),
+            ("_TZ3000_vd43bbfq", "TS130F"),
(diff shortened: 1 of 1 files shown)


MattWestb:
If you like adding more device that is have the same config you can looking in deCONZ or Z2M for finding the ID for the right configs.
I have doing it for tuya TRVs then all configs is the same it shall working for the ID but writing that is not being tested it very good if users is getting problems later.

Great work done and more happy users in the end !!

MattWestb:
I was looking in Z2M and its looks that the "TS130F" is only used for this type of devices that tuya is normally not doing = good.

The problem is that ZHA is needing the signature for triggering the quirk and its little work to finding then if they is posted online. deCONZ have normal screenshots that can sowing them.

If you is finding then then adding them and writing one link where you was finding it so it can being checked if getting problem in the future (it was on large mess with TRVs for some mouth ago with false IDs).

Likely is most the "original" cluster profile but not 100% then its tuya :-))

Adminiuga (approved):

moritan:
Hello, i think there is a problem with this commit.
_TZ3000_vd43bbfq is used 2 times in the file. 
When i submit my PPR it was to fixed issued with _TZ3000_vd43bbfq by adding a dedicated class.
I don't know how it will worked now with two signatures.

MattWestb:
The first is having only OTA as out cluster and the second type is having OTA and Time as out cluster so one device can only matching one of the quirks = no problems only support for 2 different (configured) Zigbee modules.

Issue #900, opened 2021-05-28
[Device Support Request] [WIP] Tuya Smart Air Box TS0601 _TZE200_8ygsuhe1


Lirrec:
This seems to be a relatively new device, containing temperature, relative humidity, CO2, VOC and formaldehyde sensors.

WIP pull request in #903, shows temperature, humidity and CO2.

**Additional context**
This seems to be already actively worked on in some other zigbee integrations, so cluster and attribute ids + conversions are already figured out:
zigbee-herdsman-converter: https://github.com/Koenkk/zigbee-herdsman-converters/pull/2612
zigbee2mqtt: https://github.com/Koenkk/zigbee2mqtt/discussions/7299

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**

```
[0x0000:zdo] ZDO request ZDOCmd.Mgmt_Permit_Joining_req: [60, <Bool.false: 0>]
[0x0000](ConBee II): started initialization
[0x0000:ZDO](ConBee II): 'async_initialize' stage succeeded
[0x0000](ConBee II): power source: Mains
[0x0000](ConBee II): completed initialization
Device 0x43f7 (60:a4:23:ff:fe:d2:0f:e8) joined the network
[0x43f7:zdo] ZDO request ZDOCmd.Device_annce: [0x43F7, 60:a4:23:ff:fe:d2:0f:e8, 142]
[0x43f7] Requesting 'Node Descriptor'
Tries remaining: 2
[0x43f7] Extending timeout for 0x35 request
[0x43f7] Node Descriptor: NodeDescriptor(byte1=1, byte2=64, mac_capability_flags=142, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)
[0x43f7] Discovering endpoints
Tries remaining: 3
[0x43f7] Discovered endpoints: [1]
[0x43f7:1] Discovering endpoint information
Tries remaining: 3
Received frame on uninitialized device 60:a4:23:ff:fe:d2:0f:e8 endpoint 1 for cluster: 61184
...
[0x9CCF:1:0x0b04]: async_update
Received frame on uninitialized device 60:a4:23:ff:fe:d2:0f:e8 endpoint 1 for cluster: 61184
...
Tries remaining: 2
[...]

PR #903, opened 2021-05-29
[WIP] Tuya Smart Air Box TS0601 _TZE200_8ygsuhe1

Lirrec:
PR for issue #900 

This is WIP and I don't expect it to be ready to merge (e.g. no formatting or codestyle review), but I wanted to add the current status of the quirks as a PR to supplement the device support request issue.

Will report temperature, humidity and CO2. Sensor values are implemented through *Concentration clusters, which are partially supported from HA 2021.5.x upwards (only CO2 so far).
VOC uses EthyleneConcentration, which is wrong and should be replaced with a more proper cluster (analogvalue possibly?)

### DIFF
--- a/zhaquirks/tuya/ts0601_smart_air_box.py
+++ b/zhaquirks/tuya/ts0601_smart_air_box.py
@@ -1,0 +1,170 @@
+"""Tuya based smart air box (temperature, relative humidity, VOC, CO2, formaldehyde."""
+from zigpy import types
+from zigpy.profiles import zha
+from zigpy.quirks import CustomDevice
+from zigpy.zcl.clusters.general import Basic, Groups, Ota, Scenes, Time
+from zigpy.zcl.clusters.measurement import CarbonDioxideConcentration, FormaldehydeConcentration, EthyleneConcentration
+
+from zhaquirks import LocalDataCluster, Bus
+from zhaquirks.const import (
+    DEVICE_TYPE,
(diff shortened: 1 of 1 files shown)


pipiche38:
For your information CO2, CH2O and VOC are in ppm (at least this is what the application shows).
The only bad thing of this device, is that it is sending all those information more-less every second and flood the entire network especially because in addition there ACK enable and Default Response

guillpocovi:
Hi, 

> For your information CO2, CH2O and VOC are in ppm (at least this is what the application shows).
> The only bad thing of this device, is that it is sending all those information more-less every second and flood the entire network especially because in addition there ACK enable and Default Response
> LAst I wonder if there is not a calibration command or something like that, as I found the measures quiet suspicious

I successfully paired my TS0601_air_quality_sensor device and I am also getting new reports extremely frequent, around 1 or 2 per second. Will it be possible to control/increase the reporting periodicity in future Z2M SW upgrades? or is this fully up to the device to decide?

pipiche38:
Very bad quality from what I investigated and really flooding the all network.
I paired it on tuya hub and didn’t see calibrate option.

The manufacturer of zigate had open one and confirmed the low quality of the hardware. 

Envoyé de mon iPhone

Adminiuga:
Closing in favor of https://github.com/zigpy/zha-device-handlers/pull/942

PR #931, opened 2021-06-21
ADD: Tuya valve random model id

koying:
Ref: https://community.home-assistant.io/t/zha-fails-after-custom-quirks-added/317056

### DIFF
--- a/zhaquirks/tuya/valve.py
+++ b/zhaquirks/tuya/valve.py
@@ -844,6 +844,7 @@ class MoesHY368_Type1(TuyaThermostat):
             ("_TZE200_ckud7u2l", "TS0601"),
             ("_TZE200_ywdxldoj", "TS0601"),
             ("_TZE200_cwnjrr72", "TS0601"),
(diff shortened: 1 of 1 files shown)


MattWestb:
Moes TRVs have extended functions that also need being added in ZHA for getting all advanced function to working.

I dont knowing if its 100% compatible with the old Moes TRVs but i think most shall working but no garante.

Then adding the TS061 version (MoesHY368_Type1) i think its best adding the second type 2 than its normally shoeing up very fast and its shall having the string `("_TYST11_b6wax7g0", "6wax7g0"),` under `MoesHY368_Type2`.

koying:
I'm just following up on the HA forum post. I don't own the device.
But the signature match perfectly Type1 and not Type2

MattWestb:
I have adding the new Moes in ZHA so it shall getting the advance function in HA GUI but i dont knowing if its 100% compatible with the old Moes TRVs.
If not some code worrier need taking one lock to fixing it but its better then not having the functionality.

PR #932, opened 2021-06-21
Adding new tuya / Moes TRV

MattWestb:
Adding the second type of the _TYST11 tuy /Moes TRV.
PR is made for adding them in ZHA for getting the extended functions Moes TRV is having.

Not  tested by my or some user but #931 is some gave testing the first version but im not 110% that all is working but i think its better then no quirk for the device.


MattWestb:
https://github.com/zigpy/zigpy/discussions/653#discussioncomment-314395 updated but picture is coming later then have finding one OK.

MattWestb:
Thanks !!

PR #955, opened 2021-07-10
Add support for Bitron/SMaBiT AV2010/32.

MidnessX:
Hello,
I've written a custom PowerConfiguration cluster for the Bitron/SMaBiT AV2010/32 that calculates the remaining battery percentage and updates the respective attribute whenever a battery voltage update is received, since the thermostat doesn't do it itself.

### DIFF
--- a/zhaquirks/bitron/__init__.py
+++ b/zhaquirks/bitron/__init__.py
@@ -1,0 +1,6 @@
+"""Bitron/SMaBiT module"""
+
+import logging
+
+BITRON = "Bitron Home"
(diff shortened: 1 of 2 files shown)


dmulcahey:
Can you use this and override the voltages? https://github.com/zigpy/zha-device-handlers/blob/f3471c606a0e5f0eac7a988b7efef1becbf31260/zhaquirks/__init__.py#L188

Also, flake8 / iSort

dmulcahey:
Thanks!

PR #971, opened 2021-07-22
Add support for Yooksmart blinds with inverted reporting percentage

bballwiz5:
For some reason, they accept a standard percentage for commanding the blinds, but invert the reported percentage.

### DIFF
--- a/zhaquirks/yooksmart/D10110blinds.py
+++ b/zhaquirks/yooksmart/D10110blinds.py
@@ -1,0 +1,87 @@
+"""Device handler for Yooksmart D10110 roller blinds."""
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster, CustomDevice
+from zigpy.zcl.clusters.closures import WindowCovering
+from zigpy.zcl.clusters.general import (
+    Basic,
+    Groups,
+    Identify,
+    Ota,
+    PollControl,
+    PowerConfiguration,
+    Scenes,
+)
+
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+
+
+class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
+    """WindowCovering cluster implementation.
+
+    This implementation inverts the reported covering percent for non standard
+    devices that don't follow the reporting spec.
+    """
+
+    cluster_id = WindowCovering.cluster_id
+    CURRENT_POSITION_LIFT_PERCENTAGE = 0x0008
+
+    def _update_attribute(self, attrid, value):
+        if attrid == self.CURRENT_POSITION_LIFT_PERCENTAGE:
+            value = 100 - value
+        super()._update_attribute(attrid, value)
+
+
+class D10110Blinds(CustomDevice):
+    """Custom device representing Yooksmart D10110 roller blinds."""
+
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=514
+        # device_version=1
+        # input_clusters=[0, 1, 3, 4, 5, 32, 258]
+        # output_clusters=[3, 25]>
+        MODELS_INFO: [
+            ("yooksmart", "D10110"),
+        ],
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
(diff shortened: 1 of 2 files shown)


bballwiz5:
These are the motors for Yoolax and Graywinds blinds on Amazon. These are the ZF version.

rohankapoorcom:
This change appears to have completely broken my Yoolax shades (which were correctly reporting the state prior to this change). With this change, they believe they are closed when they are open (and vice-versa) leading to them not accepting commands to move from Home Assistant.

I wonder if this behavior is dependent on the firmware version. My shades are `Firmware: 0x11006780`. @bballwiz5 what version are yours?

bballwiz5:
Hmm... my firmware is `Firmware: 0x11006770` so maybe they did fix it in more recent firmware versions? As far as I know, there is no way to update the firmware on these. Both the [Hubitat ](https://github.com/mwerline/hubitat/blob/main/YoolaxWindowBlinds.groovy#L250)and [SmartThings ](https://github.com/thx1200/Yoolax/blob/master/devicetypes/yoolax-shades#L231)community drivers invert these percentages too. 
I don't believe the quirks are setup to include Firmware version in the signature yet. Maybe we can add that as a PR?

rohankapoorcom:
Yeah - it certainly seems firmware dependent. Mine were just manufactured (arrived ~1 week ago). I think if we could add firmware version as part of the signature that would solve this problem for sure. @dmulcahey @Adminiuga is that something that could be added?

MattWestb:
Post the signature string from the device (or better log from ZHA reading the basic cluster then paring the device as being added in the quirk signature and not stripped as in this PR) so the devs can looking at it if its gets difference that ZHA can using for selecting the quirk.

rohankapoorcom:
Here's the signature for mine.
```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4169, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0202",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0020",
        "0x0102"
      ],
      "out_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "yooksmart",
  "model": "D10110",
}
```

bballwiz5:
Mine, unfortunately, has the same signature. Seems they updated the software version without changing anything else.

It seems zigpy is getting the firmware version from an OTA update check from the device, then stores it in the HA device registry. Zigpy doesn't store it anywhere itself and does not expose it to quirks. Not sure yet how to connect the two. Probably should open a ticket for this.

HEAT16:
> You can using local quirks that is overriding the quirk that is in HA [#693 (comment)](https://github.com/zigpy/zha-device-handlers/discussions/693#discussioncomment-857274).
> 
> Download the quirk and do the change in the local quirk folder. The local quirks is not being updated with HA and you need "manager" your self.

The whole yooksmart directory and custom quirk in /zhaquirks didn't exist prior to the August 25th commit, how would I handle removing that file entirely via the custom_zha_quirks directory?

MattWestb:
Copy the quirk file (its ending with `.py` from the commit and doing the changes in it.
Then you is installing it in the local quirk folder is ZHA using it from the local folder and you dont need doing any changes in the HA container. 

You cant deleting quirk in HA with one local quirk only overriding the installed ones so do one local quirk without the changes that is braking your device.

HEAT16:
Interestingly enough, I made the suggested change and the the directions are back to working normally BUT I still only ever get 0 or 100 out of the position. Now its at 0 when closed, but immediately jumps to 100 as soon as you start to open it. Additionally, if you send a command for it to open/close to a certain percentage, it just closes all the way and goes to 0. 

There is definitely still something weird going on.

rohankapoorcom:
@HEAT16 - try running a ZHA reconfigure on the device. That's what I had to do to get the position sorted out after swapping out the quirk.

PR #993, opened 2021-08-12
Remove all Philips Hue lighting related quirks

TheJulianJES:
REQUIRES https://github.com/zigpy/zigpy/pull/788

All Philips Hue related lighting quirks are redundant, if https://github.com/zigpy/zigpy/pull/788 is merged. 

Current (custom) attributes for Hue lighting quirks:
https://github.com/zigpy/zha-device-handlers/blob/8c61c6da4e44906183a4d96e022e67f8c2ba005d/zhaquirks/philips/__init__.py#L93-L111

Background of why it's a good idea to get all Philips Hue lighting related quirks removed:

Newer Hue firmware (https://github.com/Koenkk/zigbee-OTA/pull/47) changes the signatures of Hue lights.
For the older (ZLL profile) lights, it looks like only the manufacturer name changed.
For the newer (ZHA profile/Bluetooth) lights, it looks like cluster 64514/0xfc02 changed to 64515/0xfc03. (This may only be visible after re-pairing the light).

As the Hue quirks are not needed anymore since https://github.com/zigpy/zigpy/pull/774 and https://github.com/zigpy/zigpy/pull/788, as the attributes are directly implemented in zigpy, this PR would remove all lighting related quirks.

I haven't tested all changes yet.


FerretMonger:
Do i need to re-pair my lights for this to take effect?

TheJulianJES:
> Do i need to re-pair my lights for this to take effect?

No. You just need to be on Home Assistant 2021.9.0 or later.

There are some minor differences (different manufacturer clusters, Signify instead of Philips as the manufacturer name, ...) when re-pairing lights (after having applied firmware updates) but the core functionality of the bulb always works.

FerretMonger:
Thanks!

PR #996, opened 2021-08-16
Add the calibrate option to the attributes.

SLG:
See here for more information: https://github.com/home-assistant/core/issues/46146

I've tested it locally and it works like a charm.
I'm not sure if there should be done anything else, or that there will be any potential issues with this fix?


MattWestb:
If the device is handling the attribute 0xF003 OK the user can calibrating the device from the cluster panel that shall being OK, 
Its possible doing more things in the HA side but i think this is the most important for the moment getting in place and some can doing the HA part later if thy like for the devices.

PR #1021, opened 2021-09-03
Implemented additional time sync case for TuYa devices.

Sirse:
Haozee HY08WE thermostat (_TZE200_znzs7yaw) needs two different timestamps in one payload (timestamp in seconds from _**1/1/2000**_ for UTC time and timestamp in seconds from _**1/1/1970**_ for local time).
Thats seems like a bug, but device works with TuYa hub without problems.

[Z2M reference]( https://github.com/Koenkk/zigbee-herdsman-converters/blob/a83dd5338a20d35a3b59d54164cc371bcd8d2622/converters/fromZigbee.js#L2515-L2533):
```
    hy_set_time_request: {
        cluster: 'manuSpecificTuya',
        type: ['commandSetTimeRequest'],
        convert: async (model, msg, publish, options, meta) => {
            const OneJanuary2000 = new Date('January 01, 2000 00:00:00 UTC+00:00').getTime();
            const currentTime = new Date().getTime();
            const utcTime = Math.round((currentTime - OneJanuary2000) / 1000);
            const localTime = Math.round(currentTime / 1000) - (new Date()).getTimezoneOffset() * 60;
            const endpoint = msg.device.getEndpoint(1);
            const payload = {
                payloadSize: 8,
                payload: [
                    ...tuya.convertDecimalValueTo4ByteHexArray(utcTime),
                    ...tuya.convertDecimalValueTo4ByteHexArray(localTime),
                ],
            };
            await endpoint.command('manuSpecificTuya', 'setTime', payload, {});
        },
    },
```

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -234,6 +234,7 @@ class TuyaManufCluster(CustomCluster):
     cluster_id = TUYA_CLUSTER_ID
     ep_attribute = "tuya_manufacturer"
     set_time_offset = 0
+    set_time_local_offset = None
 
     class Command(t.Struct):
         """Tuya manufacturer cluster command."""
@@ -255,6 +256,7 @@ class TuyaManufCluster(CustomCluster):
             Zigbee payload is very similar to the UART payload which is described here: https://developer.tuya.com/en/docs/iot/device-development/access-mode-mcu/zigbee-general-solution/tuya-zigbee-module-uart-communication-protocol/tuya-zigbee-module-uart-communication-protocol?id=K9ear5khsqoty#title-10-Time%20synchronization
 
             Some devices need the timestamp in seconds from 1/1/1970 and others in seconds from 1/1/2000.
+            Also, there is devices which uses both timestamps variants (probably bug). Use set_time_local_offset var in this cases.
 
             NOTE: You need to wait for time request before setting it. You can't set time without request."""
 
@@ -305,7 +307,10 @@ class TuyaManufCluster(CustomCluster):
         )
         local_timestamp = int(
             (
-                datetime.datetime.now() - datetime.datetime(self.set_time_offset, 1, 1)
+                datetime.datetime.now()
+                - datetime.datetime(
+                    self.set_time_local_offset or self.set_time_offset, 1, 1
+                )
             ).total_seconds()
         )
         payload.extend(utc_timestamp.to_bytes(4, "big", signed=False))


Sirse:
> I don't understand how this fix works. `self.set_time_local_offset` is `None` and is not set anywhere, is it? so essentially it would always use `self.set_time_offset` same as the original code. What am I missing?

set_time_offset also initialized with 0 and skips time sync (most of devices didn't use that) - https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L284

U should define set_time_offset (already used in [valve.py](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/valve.py#L176)) or set_time_offset and set_time_local_offset both in quirk (see [my quirk draft for HY08WE](https://gist.github.com/Sirse/3cf8172935d81c5c00655650854e3160#file-ts0601_haozee-py-L83))

Adminiuga:
Thank you. Thank makes sense now.

However I'm curious, does your model really needs local time from 1970 or is it just ignoring local time? in other words, if the local time is from Jan2K, does it show incorrect time?

Adminiuga (approved):
LGTM,  but i'm on a fence if it would be more readable to have two offsets, one for UTC and another for local time, then each could be independent of each other 🤷

Sirse:
> Thank you. Thank makes sense now.
> 
> However I'm curious, does your model really needs local time from 1970 or is it just ignoring local time? in other words, if the local time is from Jan2K, does it show incorrect time? (z2m sends both local and utc times in Jan2K seconds, doesn't it?)

Z2M have a special time handler for this model (check a PR description).
Yep, if i'm send UTC and local time via standard method with offset from 1970 or from 2K device show incorrect time and stops to report data on some data points.

> LGTM, but i'm on a fence if it would be more readable to have two offsets, one for UTC and another for local time, then each could be independent of each other 🤷

Yes, I thought about it for a long time too, but in most cases (99% for now) quirks should use only set_time_offset with 2k or 1970 offset. Also, renaming seems like breaking change for probably existing custom quirks.
My implementation is flexible and allows to use set_time_offset without set_time_local_offset for most cases (device uses 1970 or 2k for both offsets). But, via setting set_time_local_offset any offset can be assigned to any timestamp.

_English is not my native language, so please be kind to my mistakes :)_

MattWestb:
The 
>Z2M have a special time handler for this model (check a PR description).
Yep, if i'm send UTC and local time via standard method with offset from 1970 or from 2K device show incorrect time and stops to report data on some data points.

looks like the problems with tuya / LIDL weathering valve that is not reporting anything then not being connected to one no tuya ZBGW :-)

PR #1024, opened 2021-09-06
Return correct response to emulated attribute writes

Adminiuga:
Return correctly formatted  response to emulated `write_attributes` commands. This wasn't a problem earlier, when ZHA was using plain patches, but if using quirks in tests it actually exposes the underlying problem and this starts failing Tuya specific tests if testing with the quirks.

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -310,7 +310,7 @@ async def test_siren_send_attribute(zigpy_device_from_quirk, quirk):
         tuya_cluster.endpoint, "request", side_effect=async_success
     ) as m1:
 
-        (status,) = await switch_cluster.command(0x0000)
+        _, status = await switch_cluster.command(0x0000)
         m1.assert_called_with(
             61184,
             1,
@@ -318,11 +318,9 @@ async def test_siren_send_attribute(zigpy_device_from_quirk, quirk):
             expect_reply=False,
             command_id=0,
         )
-        assert status == [
-            foundation.WriteAttributesStatusRecord(foundation.Status.SUCCESS)
-        ]
+        assert status == foundation.Status.SUCCESS
 
-        (status,) = await switch_cluster.command(0x0001)
+        _, status = await switch_cluster.command(0x0001)
         m1.assert_called_with(
             61184,
             2,
@@ -330,11 +328,9 @@ async def test_siren_send_attribute(zigpy_device_from_quirk, quirk):
             expect_reply=False,
             command_id=0,
         )
-        assert status == [
-            foundation.WriteAttributesStatusRecord(foundation.Status.SUCCESS)
-        ]
+        assert status == foundation.Status.SUCCESS
 
-        status = switch_cluster.command(0x0003)
+        _, status = await switch_cluster.command(0x0003)
         assert status == foundation.Status.UNSUP_CLUSTER_COMMAND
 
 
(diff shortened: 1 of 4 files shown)


Adminiuga:
i need to fix the tests

Adminiuga:
I've tested with 
```
{
  },
  "manufacturer": "_TZE200_ckud7u2l",
  "model": "TS0601",
  "class": "zhaquirks.tuya.valve.MoesHY368_Type1"
}
```
And it seem to work fine. I'm going to test again with the original version. 
I'm on the fence whether to rely on zigpy functionality in ZHA tests or just plug it again in zha tests?

Adminiuga:
with the current `zhaquirks` version I'm getting the same error as in my failing ZHA tests
```
2021-09-06 23:10:46 DEBUG (MainThread) [homeassistant.components.zha.core.channels.base][0x2BF5:1:0x0201]: wrote {'operation_preset': 2} attrs, Status: (<Status.SUCCESS: 0>,)
2021-09-06 23:10:46 DEBUG (MainThread) [homeassistant.components.zha.entity] climate.tze200_ckud7u2l_ts0601_ed9396fe_thermostat: Couldn't turn off 'away' preset
```

Some quirk cluster commands are translated into attribute writes and need to correct quirk to return default response for those and attribute write response for attribute writes.

MattWestb:
The last comment i think some user have seen on some TRVs but its looks like different Zigbee module (TZE200 or TYST11) can handling it little different ways but is more or less normal for tuya devices to having little "defects" but working. 
Great that you is digging little in the tuya implantation and fixing some old not so good things in the code :-))

Adminiuga:
@MattWestb Can you test with branch with your TRVs?

MattWestb:
Normally no problems but i is going to the car now and heading for southern Spain for Holiday and the test network is let at home so cant do it until the end of September :-((

Sorry for not can helping you !!

MattWestb:
Was stopping by IKEA Montpelier (southern France and **red** covid19 territory) and was baying 2 if the stemless steel Remotes Styrbar (E2001 i think it) that i cant getting in Austria. Italy or Spain and they was only 10€ etch :-)

Great work and i testing it on the https://github.com/zigpy/zha-device-handlers/blob/2ca8ffe2c6057e8aab8f81b44435f36aae61125a/zhaquirks/tuya/valve.py#L883
after my birthday in little more then 2 weeks then im at home if all is going well.

MattWestb:
Was stopping by IKEA Montpelier (southern France and **red** covid19 territory) and was baying 2 if the stemless steel Remotes Styrbar (E2001 i think it) that i cant getting in Austria. Italy or Spain and they was only 10€ etch :-)

Great work and i testing it on the https://github.com/zigpy/zha-device-handlers/blob/2ca8ffe2c6057e8aab8f81b44435f36aae61125a/zhaquirks/tuya/valve.py#L883
after my birthday in little more then 2 weeks then im at home if all is going well.

MattWestb:
Do i need testing it on my production network for the SiterwellGS361 or is it working OK in the dev Chanel ?

I have not bringing my test network up then they is running in docker containers on my laptop and I have not unpacking all the things from the holiday trip.

If you think / like i can taking the time and patching the production network but if you thing its no bringing any thing its better letting it running without braking things.

PR #1028, opened 2021-09-09
Add support for the Aqara TVOC monitor

dmulcahey:
This PR adds support for the Aqara TVOC monitor.


Adminiuga (approved):
Lgtm

csilv:
Some minor thoughts/suggestions:

- It might be nice to add constants for the configure_reporting values just to aid in readability.
- Would it make sense to use TemperatureMeasurementCluster and RelativeHumidityCluster from xiaomi/__init__.py? These are just lightweight wrappers to filter out invalid values, most other Aqara quirks seem to use these.
- For me, the temperature seems to mostly the device display, but the display seems to consistently show humdity a little bit higher. For example, right now I see: display humidity: 67%, zigbee humidity: 64.6%.
- Random question... the Aqara app can change a couple settings on the device... namely, units for temperature (ºC/ºF) and TVOC (mg/m3 or ppb). Think it would be possible to figure out how to accomplish that via ZHA? I assume this would need to be reverse engineered. It's not a big deal for me, I set my preferred units via the Aqara app/hub, but not everyone has a Hub.

csilv:
Also an observation... HA shows my air quality as: 53,000,000 µg/m³. In the logs this value is reported as 53.

But I'm fairly confident that the assumed units are not correct, I used the Aqara app to change my device to ppb and based on what i'm seeing on the device and the logs, pretty sure that "53" is in ppb. Which makes sense, it also reports temperature as F and I had changed that setting as well.

Probably not a big deal, i'm sure for my purposes I can come up with my own presentation for this entity, but just thought i'd mention it.

dmulcahey:
> Some minor thoughts/suggestions:
> 
> * It might be nice to add constants for the configure_reporting values just to aid in readability.
> * Would it make sense to use TemperatureMeasurementCluster and RelativeHumidityCluster from xiaomi/**init**.py? These are just lightweight wrappers to filter out invalid values, most other Aqara quirks seem to use these.
> * For me, the temperature seems to mostly match the display, but the display seems to consistently show humidity a little bit higher. For example, right now I see: display humidity: 67%, zigbee humidity: 64.6%.
> * Random question... the Aqara app can change a couple settings on the device... namely, units for temperature (ºC/ºF) and TVOC (mg/m3 or ppb). Think it would be possible to figure out how to accomplish that via ZHA? I assume this would need to be reverse engineered. It's not a big deal for me, I set my preferred units via the Aqara app/hub, but not everyone has a Hub.

1) I will add constants... I was more focused on making it work

2) honestly I am not sure. This is a Zigbee 3.0 device and for the most part they behave better than the original Xiaomi devices. If we see invalid values we can I just didn't want to assume they have the same issue. 

3) Nothing I can do about that. I noticed this with the TVOC readings too and part of me wonders if the updates happen quickly and the e-ink screen doesn't update as often as the sensor does (potentially for battery life)

4) Yes and we need to. Any chance you are willing to do some traffic sniffing with Wireshark? I don't have an aquaria hub or I would do it myself. I can also look at decent to see if they already did this. I looked at the Z2M converter and I didn't see anything for this. I can look at the docs too... there is a chance they are using standard commands for this as well.

dmulcahey:
> Also an observation... HA shows my air quality as: 53,000,000 µg/m³. In the logs this value is reported as 53.
> 
> But I'm fairly confident that the assumed units are not correct, I used the Aqara app to change my device to ppb and based on what i'm seeing on the device and the logs, pretty sure that "53" is in ppb. Which makes sense, it also reports temperature as F and I had changed that setting as well.
> 
> Probably not a big deal, i'm sure for my purposes I can come up with my own presentation for this entity, but just thought i'd mention it.

This would be fixed by item # 4 above by having the units changed. If we want to make the units in HA change dynamically we would need to change the sensor class there too.

csilv:
> 4\. Yes and we need to. Any chance you are willing to do some traffic sniffing with Wireshark? I don't have an aquaria hub or I would do it myself. I can also look at decent to see if they already did this. I looked at the Z2M converter and I didn't see anything for this. I can look at the docs too... there is a chance they are using standard commands for this as well.

I'm happy to try. No promises though, my Wireshark skills are lacking... fortunately, I do have a HUSBZB-1 stick and that looks fairly easy to setup to sniff traffic. I'll use the instructions at https://www.zigbee2mqtt.io/how_tos/how_to_sniff_zigbee_traffic.html , but let me know if you have any other tips or recommendations. Won't start on this until Sunday.... and if I fail miserably, I may just send you my Aqara Hub with a return shipping label. ;-)

dmulcahey:
@csilv I am going to merge this now. We can work on item 4 in a second PR if necessary

PR #1029, opened 2021-09-09
Fix button name for Aqara Wireless Switch

dodekaphilist:


dmulcahey:
What does this actually change? Can you share screenshots?

dodekaphilist:
Previously, button naming and button action used to be the same, now the button is called 'Turn on' and is matched with the according action (1x, 2x, .. pressed)

Before:
image

After:
image

dmulcahey:
so this is a device that is only one button... calling it turn on is wrong IMO. it's a single button that can be used for many things. Some remotes have dedicated turn on / turn off buttons and this makes sense for those (again solely in my opinion). I am open to other POV's here. @Adminiuga @puddly either of you have any thoughts on this?

Adminiuga:
Or, i do have opinions on this :) since you asked:

- yes, I agree with you, turn_on is too specific and should not be used
- we should not change existing automation nat all -- it's a breaking change, unless you can make both , the old and new one to work simultaneously
- for new quirks, we probably should use button_press for the type and single/short/double_press for subtypes

dmulcahey:
Thanks for the PR but for the reasons mentioned above we are not going to merge this.

PR #1036, opened 2021-09-15
Elko thermostat support

lodegaard:
Basic support for Elko Super TR Thermostat.
The values and meaning for the custom attributes are copied from [Deconz Rest Plugin](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/general.xml#L2034) and [Homey](https://github.com/prj84/com.elko/blob/Master/drivers/ESHSUPERTR/ST_Code/Elko%20Thermostat%20-%20Vendor%20Specific%20attributes.txt)

Tested with two thermostats and seem to be working fine.

PR #1044, opened 2021-09-22
Power reading support for Elko Super RF thermostat

tanordheim:
Builds on the Elko Super RF quirk implemented by @lodegaard in #1036, and exposes the current power consumption/effect (in watts) of the thermostat. As far as I'm able to tell this is updated once every 10 minutes, with the first update coming 10 minutes after the heating initially starts. The value produced seems to be the average consumption over the preceding 10 minutes.

WIth this, I'm able to use the [integration sensor](https://www.home-assistant.io/integrations/integration/) to produce an effect value in kWh in Home Assistant so I can use the data from this sensor in my energy dashboard.

Note that I'm not 100% sure about this implementation yet - I'm still testing, but as far as I know there is no data coming in that sets the effect back to 0 when heating stops so I might have to add something for that. I'll run this in my local HA overnight to see what the data looks like tomorrow.

### DIFF
--- a/zhaquirks/elko/__init__.py
+++ b/zhaquirks/elko/__init__.py
@@ -2,6 +2,7 @@
 
 from zigpy.quirks import CustomCluster, CustomDevice
 from zigpy.zcl.clusters.hvac import Thermostat, UserInterface
+from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
 
 from zhaquirks import Bus, LocalDataCluster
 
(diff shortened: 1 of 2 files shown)


tanordheim:
This has been running over night for me now and seems to be working as intended. From my side its ready for review.

lodegaard:
@tanordheim The code looks fine, but I would suggest changing the name for `POWER_ID` to `ACTIVE_POWER` to match the Zigbee specification and the name in `ElectricalMeasurement`.

I wasn't able to get any stable reportings of that value on any of my thermostats, only some values now and then. But nevere anything with a fixed ten minute interval.

As it is, this is probably better than leving it out, but it would be even better to configure custom attribute reporting. And I think that have to be done in homeassistant.

tanordheim:
> I wasn't able to get any stable reportings of that value on any of my thermostats, only some values now and then. But nevere anything with a fixed ten minute interval.

Strange, I seem to be getting quite consistent data from my 5 thermostats. Every time heating starts, I get effect data every 10 minutes after that until it stops heating. The last tick of data seems to come at most 10 minutes after heating stops with an effect value of 0 watt.

As for the `POWER_ID` name I can change that, that was just a name picked out of some example code in this repository that did roughly the same.

PR #1048, opened 2021-09-28
Add GreenPower (242) endpoint

lebretr:
Add the greenPower (242) endpoint for the TS130F module

See more here about the issue : https://github.com/home-assistant/core/issues/46146#issuecomment-929584833

### DIFF
--- a/zhaquirks/tuya/ts130f.py
+++ b/zhaquirks/tuya/ts130f.py
@@ -3,7 +3,15 @@ from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
 import zigpy.types as t
 from zigpy.zcl.clusters.closures import WindowCovering
-from zigpy.zcl.clusters.general import Basic, GreenPowerProxy, Groups, OnOff, Ota, Scenes, Time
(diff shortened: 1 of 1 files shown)


Adminiuga (approved):
LGTM if checks pass

MattWestb:
And the back light added 2 :-)))

What i dont like is that the signature is not reflecting the device real clusters and EP 2 is missing with its GP cluster.

It can being good later if need adding more device or if tuya (as normal) is making one device with time cluster as out instead of in.
```
 {
   "1": {
      "profile_id": 260,
      "device_type": "0x0202",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0102"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
```
Some quirk have only copy from one already added device and not fixing the clusters and its making it bad and is better not putting in information then false information then it making it only misleading.

Well done !!!

Issue #529, opened 2020-10-17
[Device Support Request] Salus SP600


hmonteiro:
I am using a [Salus SP600 smart plug](https://zigbee.blakadder.com/Salus_SP600.html) and, although it correctly shows as a smart plug with energy monitoring capablities in HA, there are some problems:
- The temperature reported is always 2x the actual value
- The instantaneous demand reported is always 1/10 of the actual value
- The total consumption (attribute `current_summ_delivered` attribute from the metering cluster) is not reported

It would be great if these values were correctly reported by the system.

The device signature is below. If you need any more informations let me know.

**Device signature**
```
{
  "node_descriptor": "NodeDescriptor(byte1=1, byte2=64, mac_capability_flags=142, manufacturer_code=4216, maximum_buffer_size=82, maximum_incoming_transfer_size=255, server_mask=0, maximum_outgoing_transfer_size=255, descriptor_capability_field=0)",
  "endpoints": {
    "9": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0402",
        "0x0702",
        "0xfc01"
      ],
      "out_clusters": [
        "0x0019"
      ]
    }
  },
  "manufacturer": "Computime",
  "model": "SP600",
  "class": "zigpy.device.Device"
}
```

davet2001:
I have the SP600, and it works ok.
After making the switch from zigbee2mqtt to zha, it still works, but now reads power 10x too high.
Could it be the same as this issue:
https://github.com/Koenkk/zigbee-herdsman-converters/issues/897

I see that in zigbee2mqtt, there is a specific handler for a certain datecode:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/d751f07919cc1122da3f0ac75053a7ba0cfe137b/converters/fromZigbee.js#L1650-L1669

I wonder whether this needs handling in a similar way in ZHA?

There are different versions of firmware for this device, as described here:
https://github.com/Koenkk/zigbee-herdsman-converters/issues/915

They behave differently.  Can similar behaviour be accommodated in ZHA?

davet2001:
I still have problems with my sp600 
-over reading power by 10x
-not reporting energy consumption (just live power and temperature)

Attached: dishwasher is not really using 25kW!
[...]

PR #1051, opened 2021-09-29
Salus SP600 10x fix

davet2001:
As per issue #529, both salus SP600 devices I have read 10x too high.

This PR adds a custom /10 scaling function to both the instantaneous_demand (W) and the new current_summ_delivered (kWh) attributes.

I tried changing the `_CONSTANT_ATTRIBUTES = {MULTIPLIER: 1, DIVISOR: 1000}` values, but this seemed to have no effect.

### DIFF
--- a/zhaquirks/salus/sp600.py
+++ b/zhaquirks/salus/sp600.py
@@ -50,7 +50,6 @@ class TemperatureMeasurementCluster(CustomCluster, TemperatureMeasurement):
     ATTR_ID = 0
 
     def _update_attribute(self, attrid, value):
-        print("QUIORK 2")
         # divide values by 2
         if attrid == self.ATTR_ID:
             value = value / 2
--- a/zhaquirks/salus/sp600.py
+++ b/zhaquirks/salus/sp600.py
@@ -34,6 +34,14 @@ class MeteringCluster(CustomCluster, Metering):
     DIVISOR = 0x0302
     _CONSTANT_ATTRIBUTES = {MULTIPLIER: 1, DIVISOR: 1000}
 
+    def _update_attribute(self, attrid, value):
+        # divide values by 10 (issue #529)
+        if attrid == 0x0000:  # current_summ_delivered
+            value = value / 10
+        elif attrid == 0x0400:  # instantaneous_demand
+            value = value / 10
+        super()._update_attribute(attrid, value)
+
 
 class TemperatureMeasurementCluster(CustomCluster, TemperatureMeasurement):
     """Temperature cluster that divides value by 2."""
@@ -42,6 +50,7 @@ class TemperatureMeasurementCluster(CustomCluster, TemperatureMeasurement):
     ATTR_ID = 0
 
     def _update_attribute(self, attrid, value):
+        print("QUIORK 2")
         # divide values by 2
         if attrid == self.ATTR_ID:
             value = value / 2


Adminiuga:
The existing thermostat may have already cached values for those, so _CONSTANT_ATTRIBUES has no effect. If you do both, then the new installation. I probably would leave only constant attributes and ask users to reconfigure and or rejoin thermostat

davet2001:
@Adminiuga thanks for looking at this.  I tried rejoining the device multiple times and _CONSTANT_ATTRIBUTES seemed to make no difference to anything.  Are you saying I should concentrate on trying to make _CONSTANT_ATTRIBUTES work and remove the `_update_attribute` function?

Adminiuga:
Can you try with the latest dev?
Also, did you just rejoin the device or removed it from HA/zigpy? It is supposed to refresh those attributes when device rejoins (for mains powered,)

Adminiuga:
Also, what's the date code for your device?
What do you read for divisor/multiplier without the quirk?

davet2001:
The date code for my device is 20170220.

If I remove the custom divisor/multiplier, then remove the zigbee device from HA, then power it off, then power back on again, and re-add it I find that the values are scaled correctly :)
In this state, HA 'manage clusters' reports divisor 10000 and multiplier 1.

So it seems the way to 'fix' this device is the remove the `_CONSTANT_ATTRIBUTES = {MULTIPLIER: 1, DIVISOR: 1000} ` line that was added.  I think it is a specific (older) date_code that needs this correction.  [Zigbee2MQTT only applies this to 20160120](https://github.com/Koenkk/zigbee-herdsman-converters/blob/2e2ba8fa2ec3bc65e84bf93890a379bd894f929d/converters/fromZigbee.js#L2048-L2072).

But I don't know how to reliably get the datecode at initialisation in zha-device-handlers.  I can read it using code within MeteringCluster like this:
```
basic_cluster = self.endpoint.in_clusters[0]
print(f"date_code={basic_cluster.get(0x0006)}")
```
...except I'm finding that  the date_code attribute is cached, and it's only a real value after I manually 'manage cluster' and 'get zigbee attribute' in HA first.

How can I make the initialisation of MeteringCluster automatically conditional on the value of the `Basic cluster`'s `date_code`?

Adminiuga:
Ok, so this has to be applied only for specific firmwares versions. 
Currently it is not possible to apply quirks based on date code. 
Does Salus release firmware updates? Maybe it is easier just to upgrade firmware? Otherwise it would need modification to ha to request basic channel to cache the date code and modifying quirk to affect divisor only for specific date codes

davet2001:
Ok, thank you :)

Yes, [salus provide firmware updates](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/lib/ota/salus.js).

In my case I have the newer firmware which doesn't need the quirk.

So I think my PR could be simplified, and my device would work perfectly but it would be a breaking change for people on older firmware. (They would need to do an OTA update which is not supported for salus in ZHA yet).

I am willing to support adding Salus OTA capability to ZHA but I'm new to the ZHA code!

dmulcahey:
@Adminiuga are we going to consider exposing the date code to quirk matching?

Adminiuga:
I was thinking about it, but datecode for quirk matching is too cumbersome. 
Better for quirk to apply dffery divisor based on the datecode.
But as a general approach, should we just allow application to override attributes? Sort of _CONSTATNT_ATTRIBUTES set by the application and not the quirk? Then users always could override and not wait for a quirk?

davet2001:
@juhanjuku I see you have a salus plug that you screenshotted in another issue, but your numbers look good.  Is it an SP600? If so, please could you tell me the firmware datecode? I'm assuming it is earlier than mine which is `Firmware: 0x02015120`

juhanjuku:
> @juhanjuku I see you have a salus plug that you screenshotted in another issue, but your numbers look good. Is it an SP600? If so, please could you tell me the firmware datecode? I'm assuming it is earlier than mine which is `Firmware: 0x02015120`

I have SPE600, firmware: 0x0109512
Inkedscreen_LI
0
Numbers are not good - smartenergy_metering_summation_delivered is not updating in Lovelace. It's read only once after restart.

davet2001:
Thanks, that's really helpful.  Looks like you have yet another firmware date code between the two that I've seen before.  Please can you check if the quirk is even being matched (if it is you can see it at the bottom of the config screen)?
image
The quirk should correct your temperature reading but it doesn't seem to be working.  For completeness, please can you also paste the contents of zigbee device signature for this device?

juhanjuku:
I don't see quirk. How I must apply quirk?
image

image

davet2001:
Ok, now it all makes sense :)  Your device is "SPE600" which is different to "SP600", hence the quirk doesn't match automatically.

I don't know a way of choosing a quirk manually.  So I think we need to add SPE600 as another entry in https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/salus/sp600.py

I have created a separate pull request for this #1068.

juhanjuku:
> Interesting... I'd never really noticed. They don't light up yellow for me either.
> 
> image
> 
> But neither do any of the others (only lights).
> 
> image
> 
> The above items are a mixture of salus SP600, IKEA and manually coded 433MHz energenie sockets.

My IKEA shows yellow
screen

juhanjuku:
> @juhanjuku Sorry I messed up my question. Please could you tell me the date_code from 'manage clusters'? I have FW 0x02015120 and date_code 20170220 image When you do get numbers, are they believable e.g. Watts value is realistic for this appliance? I see that your temperature reading is 39C which seems high. Possibly [this quirk](https://github.com/zigpy/zha-device-handlers/blob/9a5bcd45bb99d0df8b014703e0bad9efe067609c/zhaquirks/salus/sp600.py#L38-L48) is not being applied properly. (Or you are in a hot place!)

Salus temperature is fixed now
image

PR #1056, opened 2021-10-03
tuya Power State 

MattWestb:
Implanting tuya Power State in INIT then more device is needing it in the near future.

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -731,6 +731,21 @@ class TuyaSmartRemoteOnOffCluster(OnOff, EventableCluster):
             )
 
 
+class TZBPowerOnState(t.enum8):
+    """Tuya power on state enum."""
+
+    Off = 0x00
+    On = 0x01
+    LastState = 0x02
+
+
+class TuyaZBOnOffRestorePowerCluster(CustomCluster, OnOff):
+    """Tuya on off Zigbee cluster with restore state."""
+
+    attributes = OnOff.attributes.copy()
+    attributes.update({0x8002: ("power_on_state", TZBPowerOnState)})
+
+
 # Tuya Window Cover Implementation
 class TuyaManufacturerWindowCover(TuyaManufCluster):
     """Manufacturer Specific Cluster for cover device."""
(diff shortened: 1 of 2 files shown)


MattWestb:
One PR for TS011F is in the pipe from one user for being made and i like doing one for TS0101 then its better having the power on state in the INIT.

zhaquirks/tuya/__init__.py
```diff
@@ -731,6 +731,21 @@ def handle_cluster_request(
...
+    Off = 0x00
+    On = 0x01
+    LastState = 0x02
+
+
+class TuyaZBOnOffRestorePowerCluster(CustomCluster, OnOff):
+    """Tuya on off Zigbee cluster with restore state."""
+
+    attributes = OnOff.attributes.copy()
+    attributes.update({0x8002: ("power_on_state", TZBPowerOnState)})
```
Adminiuga, line 746:
```suggestion
    manufacturer_attributes = {0x8002: ("power_on_state", TZBPowerOnState)}
```

Adminiuga (changes requested):
Use manufacturer_attributes instead

zhaquirks/tuya/__init__.py
MattWestb: I dont have the device for testing but i have one user that have doing the same for the TS011F that can testing it on his device that have the same implantation or can i do the change without testing that is working ?

MattWestb:
OPS !!
I was making on PR for renaming  the old plug.p and David have merger it so i have braking the part moving the power state from it.
I never have luck re basing PRs so i deleting the part and making on new with plug.py using it from tuya INIT

MattWestb:
After this is OK and merged the #1058 can being merged that deleting the old implantation and using the new cluster for tuya power state.

MattWestb:
Looks like the change is not working as intended ;-((
https://github.com/zigpy/zha-device-handlers/pull/1057#discussion_r721217634

Putting the original back or some other code implanting ?

#1057 can testing the function then hi have one device (I have 2 TS0101 but i have not making any quirk for it but its coming then the power state is in).

Adminiuga:
Tuya just couldn't use the standard attribute, could they? 
That's fine, leave the original change in place

MattWestb:
I have putting the original code back and then its merged the ts0121_plug.py and the new TS011F can using it and more device later then adding it in there quirks.

Issue #605, opened 2020-11-27
[Device Support Request] BlitzWolf BW-SHP13 smart plug (aka Tuya TS0121)


jmoransalama:
**Is your feature request related to a problem? Please describe.**
BlitzWolf BW-SHP13 smart plug 16A EU version, same device as TuYa TS0121 smart plug.
![imagen](https://user-images.githubusercontent.com/16177130/100478771-819e9180-30ec-11eb-9ed3-7e5c56f88605.png)

Device joins zigbee network and it's recognized by home-assistant. 

Entities are created for:
- On/Off switch: It works
- Electrical power measurement (W): Values updated correctly.
- Energy metering (kWh): Entity is created but values are never updated.

This entities are missing:
- Current (A).
- Voltage (V).

**Describe the solution you'd like**
Please consider adding support for the missing values and fixing the electrical power measurement value.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
```
{
  "node_descriptor": "NodeDescriptor(byte1=1, byte2=64, mac_capability_flags=142, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0702",
        "0x0b04"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZ3000_3ooaz3ng",
  "model": "TS0121",
  "class": "zigpy.device.Device"
}
```

**Additional context**
This device doesn't support reporting of electrical measurement values, so it needs to be polled. It's [supported on zigbee2mqtt](https://github.com/Koenkk/zigbee2mqtt/issues/3812), they fixed this by polling the values every 10 seconds.

Voltage value is correctly reported if I manually get the attribute value:
![imagen](https://user-images.githubusercontent.com/16177130/100479099-7dbf3f00-30ed-11eb-943c-b23342c046e2.png)

CharlieBailly:
Hi, yes it would be great if it was fully supported as you describ it.
I also have the same device, but with a slightly different signature : (the manufacturer part differ)
[...]

PR #1058, opened 2021-10-04
Update ts0121_plug.py

MattWestb:
Updating ts0121_plug.py using tuya power state from tuya INIT that is being implanted in #1056 .

PR #1057, opened 2021-10-03
Create Quirk for TS011F

miit86:
Quirk for the TS011F 
See Issue https://github.com/zigpy/zha-device-handlers/issues/1045

PR #1062, opened 2021-10-07
Fix multiplier and divisor for Tuya TS0121

flabbamann:
The device reports current_summ_delivered values in "10W" so we must multiply it by 0.01 or divide it by 100 to get kWh in HA.

See Issue #605

### DIFF
--- a/zhaquirks/tuya/ts0121_plug.py
+++ b/zhaquirks/tuya/ts0121_plug.py
@@ -1,6 +1,6 @@
 """Tuya TS0121 plug."""
 from zigpy.profiles import zha
-from zigpy.quirks import CustomDevice
+from zigpy.quirks import CustomCluster, CustomDevice
 from zigpy.zcl.clusters.general import Basic, Groups, OnOff, Ota, Scenes, Time
 from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
 from zigpy.zcl.clusters.smartenergy import Metering
@@ -14,6 +14,14 @@ from zhaquirks.const import (
     PROFILE_ID,
 )
 from zhaquirks.tuya import TuyaZBOnOffRestorePowerCluster
+
+
+class TuyaMeteringCluster(CustomCluster, Metering):
+    """TuyaMeteringCluster divides the kWh for tuya."""
+    
+    MULTIPLIER = 0x0301
+    DIVISOR = 0x0302
+    _CONSTANT_ATTRIBUTES = {MULTIPLIER: 0.01, DIVISOR: 100}
 
 
 class Plug(CustomDevice):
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts0121_plug.py
```diff
@@ -31,6 +31,14 @@ class OnOffRestorePowerCluster(CustomCluster, OnOff):
     attributes.update({0x8002: ("power_on_state", PowerOnState)})
 
 
+class TuyaMeteringCluster(CustomCluster, Metering):
+    """TuyaMeteringCluster divides the kWh for tuya."""
+    
+    MULTIPLIER = 0x0301
+    DIVISOR = 0x0302
+    _CONSTANT_ATTRIBUTES = {MULTIPLIER: 0.01, DIVISOR: 100}
```
stattin42, line 39: Multiplier and divisor should both be of integer type and since both multiplier and divisor apply you probably want to set multiplier to 1, not to 0.01, for the desired result.

flabbamann: Hm, you're right. In my first try it did not work with multiplier 1, but now it's working. Maybe still in cache or whatever... Thanks 😃. Its now the same as in #1057. Should it be moved to init.py?

flabbamann:
> 
> 
> Does Tuya provide value for summa_formatting?

summa_formatting (0x0303) has value `bitmap8.0`

stattin42:
> 
> 
> > Does Tuya provide value for summa_formatting?
> 
> summa_formatting (0x0303) has value `bitmap8.0` multiplier and divisor are not provided

Too bad, then the summa_formatting attribute is of no help here :-(

Hedda:
FYI @Celer21 now gave feedback in https://github.com/zigpy/zha-device-handlers/issues/605 that his BlitzWolf BW-SHP13 smart plug (which supposedly is another OEM version of the Tuya TS0121) is reporting "**_kWH unit, but it number is actually in Wh_**",

MattWestb:
@pmfernandes The Outage Memory (on/off/restore) is already implanted in the TS0212 quirk "in tuya way" but its one PR waiting for moving the code to the quirk INI then more devices is needing it.
#1058

dmulcahey:
Needs a rebase

Celer21:
I update HA to 2021.11. The BW-SPH13 has now another sensors - great!
But i think, that sensors electrical_measurement_rms_current and total comsumption has to be divided by 1000.

It show, tahat my Washingmachine consume 2.400 W (correct ) and 10.000 A (that is is not correct).
Correct is 10 A.


image

image

sajovicd:
Hello, just to say I have exact same problem.

flabbamann:
current is addressed in #1131

PR #1068, opened 2021-10-13
Add Salus SPE600 support.

davet2001:
As discussed here https://github.com/zigpy/zha-device-handlers/pull/1051,

There is an additional device, Salus SPE600 which seems to work the same as SP600.

The quirk is needed (at least) for scaling the temperature correctly.


MattWestb:
Both the old and the new signature is missing all device / EP descriptions https://github.com/zigpy/zha-device-handlers#descriptors.

Then you is having one "original" device and is adding one new one can you pleas adding the missing moderate information ?

Thanks in advance !

PR #1070, opened 2021-10-14
Trying renaming tuya valve.py to ts0601_trv.py and more.

MattWestb:
Trying renaming tuya valve.py to ts0601_trv.py without braking all tests.


MattWestb:
Shall making zhaquirks.tuya.electric_heating, zhaquirks.tuya.motion and zhaquirks.tuya.siren but tomorrow.

MattWestb:
Great !!!

Only that black dont  like your change. You need doing like this:
```
        MODELS_INFO: [("_TZE200_aoclfnxz", "TS0601"), ("_TZE200_dzuqwsyg", "TS0601")],
```
or the test is failing.

You can using [Black playground](https://black.vercel.app/) for testing and fixing python code.

MattWestb:
The log you was posting is good to see that  the thermostat is working OK but for getting all possible DP (Tuya commands) you need taking the power of the device (disconnecting it from the power so its dead) and waiting one minute and connecting the power and its sending all DP is using to the host  system.

PR #1077, opened 2021-10-17
Renaming the 2 last not good named tuya quirks.

MattWestb:


MattWestb:
Last renaming of tuya quirks and  now the folder looks better (but not very good).

One thing is moving all ts0601 (tuya DP command devices) to one own folder and the rest "normal zigbee" devices to one other but that the maintainer must deciding how they like to have it.

Issue #981, opened 2021-07-29
Tuya TRV TZE200: wrong setpoint temperature


martin3000:
I have a TS0602 ,  _TZE200_b6wax7g0
If I set manually set the TRV to 25°C at the valve, home assistant shows 2.5°C. The temperature in HA is always to small by a factor of 10.
If I look at the cluster MoesThermostat -> occupied_heating_setpoint, I see 2.5, but the TRV shows 25.

MattWestb:
Its one known issue with the software version `_TZE200_b6wax7g0` and for fixing it it need  one new quirk only for it.
[HA issue](https://community.home-assistant.io/t/zha-fails-after-custom-quirks-added/317056/18)

The good thing the problem is not in ZHA its in the quirk so its easier to paying  with.
You can trying copy the `/tuya/valve.py` to your local quirk directory and editing it if you dont have more moes TRVs in  your system.

I think the converting is taking place in this section https://github.com/zigpy/zha-device-handlers/blob/21af0860992021c39b26260b6e2ae4f9c1af75d1/zhaquirks/tuya/valve.py#L202 or in this https://github.com/zigpy/zha-device-handlers/blob/21af0860992021c39b26260b6e2ae4f9c1af75d1/zhaquirks/tuya/valve.py#L345

If you finding the right place and can changing it the its needed making one new version of  MoesHY368_Type1  and 2 quirk parts (perhaps named "ZTRV-BY-100 Type 1 / 2" and making new moes cluster that is using the new changed and adding the signature from your TRV in the signature but with both cluster version (tuya is using 2 diferent Zigbee modules for there TRVs TZE200 and TYST11 that need having different signatures for being loaded and also shall having the same changes).

I hope you is better on coding then i  am (around zero or minus on my side) and shall being possible implanting but its little tricky for not braking the other 2 types that is using the mors quirk (plus 5 Siterwell TRVs).

martin3000:
Yes it is in direct_mapping_attrs, but I wonder why just my TRV version does not need the divisor whereas other models need it. This is unusual.

MattWestb:
The ZTRV-BY-100 is one very new TRV and its one updated moes type.

Z2M have adding the ZTRV-BY-100 and have doing some changes from the "normal" moes TRVs and no have updating the ZHA quirks for it and some user have saying it working OK with ZTRV-BY-100 = better then not at all. 

[...]

PR #1082, opened 2021-10-18
Fixed factor/divider for _TZE200_b6wax7g0 temperature settings

martin3000:
For moeshouse ZTRV-BY-100-WH-DE thermostat radiator valve, the setpoint temperature was wrong by a factor of 10.
To correct this, I introduced 3 new classes for this device.
The main change are these 2 lines:
MOES_TARGET_TEMP_ATTR: ("occupied_heating_setpoint", lambda value: value * 100)
DIRECT_MAPPING_ATTRS = { "occupied_heating_setpoint": (MOES_TARGET_TEMP_ATTR, lambda value: round(value / 100),   ),
In the original code, the factor/divider was 10 which I changed to 100.
As a side effect, the temperature can only be set in steps of 1 degree and not in steps of 0.5 degrees. The reason for this is not clear. But it seems that in the cluster - which is an int value - the temperature is stored in degrees and not in centidegrees.

### DIFF
--- a/zhaquirks/tuya/ts0601_trv.py
+++ b/zhaquirks/tuya/ts0601_trv.py
@@ -261,6 +261,34 @@ class MoesManufCluster(TuyaManufClusterAttributes):
                 "battery_change", 5 if value else 100
             )
 
+# this is the class for the new moes TRV "_TZE200_b6wax7g0"
+class MoesManufClusterNew(MoesManufCluster):
+    DIRECT_MAPPED_ATTRS = {
+        MOES_TEMPERATURE_ATTR: ("local_temp", lambda value: value * 10),
+        MOES_TARGET_TEMP_ATTR: ("occupied_heating_setpoint", lambda value: value * 100), #jms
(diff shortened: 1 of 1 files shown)


MattWestb (changes requested):
old 992 / new 1055 line:
`("_TZE200_b6wax7g0", "TS0601"),`
delete so the quirk dont have the same device 2 times that can being triggered.

martin3000:
fix #981

martin3000:
I inserted blank lines after docstring to fix flake8 error.
I do not understand the "black" error in https://github.com/zigpy/zha-device-handlers/runs/3962841238?check_suite_focus=true

MattWestb:
The line is being too long:
```
        MOES_TARGET_TEMP_ATTR: ("occupied_heating_setpoint", lambda value: value * 100),      
```
and need being braked down like this:
```
        MOES_TARGET_TEMP_ATTR: (
            "occupied_heating_setpoint",
            lambda value: value * 100,
        ),
```
Easiest to fixing it is pasting all cone in [Black play ground](https://black.vercel.app/) on the left window and copy the corrected from the right (working 95% for My). Its also fixing extra spaces and missed empty line and removing tabs.

PR #1089, opened 2021-10-21
xbee: experimental support for any coordinator

Shulyaka:
This PR removes the dependency on coordinator-specific `remote_at_command()` by emulating it for other coordinators.

While this should work in theory with other coordinators, I could only test the code with xbee coordinator for the lack of hardware, I simulated it by slightly modifying zigpy_xbee code for the test.


Shulyaka:
Great! If you have it already as a router, then you probably have switches for digital i/o. Just try to flip those switches and see if you get any exception. Flipping a switch generates an AT command that configures the pin. You can also double check the pin state with XCTU,

PR #1097, opened 2021-10-24
Fix TuyaManufacturerWindowCover to update cover position

smmoroz:
#744, Fix position update, for details see here https://github.com/zigpy/zha-device-handlers/issues/744#issuecomment-945191501


rob1981forest:
sorry for the likely stupid question but this affects me and my ZHA deployment in home assistant installation (2021.10.7). How do i make use of this code in my installation please ? or do i just wait and it will be incorporated into new HA releases ?

PR #1104, opened 2021-10-26
Making Siterwell TRVs using presets modes

MattWestb:
### Future changes for the TRV:
1. Making the quirk using presets _Away_, _Schedule_ and _Manual_ instead of _system mode_ (the TRV is **not** supporting on/off its always working)(need adding Siterwell presets in ZHA) and ZHA is **not** supporting Schedule (Schedule_programming_mode) as system mode only as occupancy presets.
2.  Deleting in device not implanted attribute:
 `SITERWELL_VALVE_STATE_ATTR = 0x026D  # [0,0,0,55] opening percentage`
3. Adding in device supported attribute:
 `SITERWELL_WINDOW_ALARM_ATTR = 0x0411  # [0] clear [1] alarm`
`SITERWELL_VALVE_JAMMED_ATTR = 0x0413  # [0] clear [1] alarm`
but not in the ZHA GUI (upcoming).
4. Adding switch for SiterwellChildLock in ZHA GUI but its only working setting.
5. Adding max and min heating so ZHA is showing the right scale.
6. Deleting "electric" then is one copy and paste from the original electric heating thermostat that is having one own quirk.

### Missing upcoming futures for the TRV:
-  1. Switches with status update from the local state shall being implanted for window and valve detect.
-  2. Status sensor for "window is open" and "valve is jammed" shall being implanted.
-  3. Battery attribute with type and quantity shall being implanted (AA and 2 or 3 batteies for new TRV types).

Away mode is showing idle in ZHA without setpoint temperature but it can being set on the TRV or by cluster attribute.

Look at https://github.com/zigpy/zha-device-handlers/discussions/1084 for more information.

### DIFF
--- a/zhaquirks/tuya/ts0601_trv.py
+++ b/zhaquirks/tuya/ts0601_trv.py
@@ -30,11 +30,12 @@ from zhaquirks.tuya import (
 SITERWELL_CHILD_LOCK_ATTR = 0x0107  # [0] unlocked [1] child-locked
 SITERWELL_WINDOW_DETECT_ATTR = 0x0112  # [0] inactive [1] active
 SITERWELL_VALVE_DETECT_ATTR = 0x0114  # [0] do not report [1] report
-SITERWELL_VALVE_STATE_ATTR = 0x026D  # [0,0,0,55] opening percentage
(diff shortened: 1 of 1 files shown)


MattWestb:
Its look like this is one dead end for all Siterwell TRVs is staying with 75% broken functionality ;-(

PR #1105, opened 2021-10-27
Tuya On/Off and Back light attribute changes  in INIT

MattWestb:
Adding Back light attribute for switches and switch mode as attribute on the OnOff cluster in tuya INIT.

Changing name of the cluster for getting more general.
The ts130f.py can also using the back light from INIT then its already having it in the quirk but  i dont do it now.

Switch mode is being used in TS004F https://github.com/zigpy/zigpy/issues/831 and very likely on more tuya devices like the power on state.
Consider using it in all "switch devices" then its only adding futures and not braking things if the device is not supporting it.


MattWestb:
Im very sorry but **THIS IS NOT FUNNY !!!**

If not like implanting more tuya attribute **then saying it** and i deleting this and more PR that is using it and i not adding more PR for that then its 2 more types that needing it.

But i cant making hand re basing more times in the week and getting all broken if this in not being merged !!

I dont standing 15 more days waiting and other is braking this simple change that is not braking any device and i adding more functions to many devices user is having :-(((

MattWestb:
Back light mode is confirmed working by one user https://github.com/zigpy/zha-device-handlers/issues/1106 and i have doing the same on LIDL power strip. If not supported the device is repaying with Non and if its supported you is getting the mode and can sting it 2.

Mode switch is is working on TS004F but need other things in place for getting it working the all the way. Also one user have testing it in one issued but i cant finding it for the moment.
If not supported by the device you is getting Non and if supported you is getting the mode and can also setting it in the switch.

MattWestb:
Thanks David ! ! !

PR #1116, opened 2021-10-31
Adding double press for E1812

MattWestb:
Adding double press device automation for IKEA E1812 Shortcut button (New in 2.3.080 FW).
Also changing name on the functions so its writing the same button for all device automatons and not on button and dim up button. (renaming reverted)

Tested with updated E1812.

### DIFF
--- a/zhaquirks/ikea/shortcutbtn.py
+++ b/zhaquirks/ikea/shortcutbtn.py
@@ -20,10 +20,11 @@ from zhaquirks.const import (
     CLUSTER_ID,
     COMMAND,
     COMMAND_MOVE_ON_OFF,
+    COMMAND_OFF,
     COMMAND_ON,
     COMMAND_STOP,
     DEVICE_TYPE,
-    DIM_UP,
+    DOUBLE_PRESS,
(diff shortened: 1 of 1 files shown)


Adminiuga:
Will this break existing automation with non updated buttons?

MattWestb:
The old FW is only sending `on`, `move` and `stop` so no problems then its adding only `off` command (double click).

The renaming i think is only what ZHA is showing on the device card then clicking add automation then the command is not changed but i dont know ZHA is doing the real automation from it.
E1812A
Before it was writing "DIM_UP button" for move and stop and my original was writing "Off button double clicked" so 3 different button named and only one in the "hand".
.

MattWestb:
The E1744 is using the same naming and is logic:
E1744A

dmulcahey:
Changing DIM_UP to TURN_ON is a breaking change

MattWestb:
Shall i reversing the 2 renaming DIM_UP for not braking users automatons.

MattWestb:
Dont  looks nice but working :-/
E1812B

PR #1125, opened 2021-11-03
Add `send_default_rsp` to TuyaManufacturerClusterOnOff

javicalle:
It seems that this is needed all around.
Fix `TuyaManufacturerClusterOnOff` while refactoring does not arrive.


Adminiuga (approved):
Lgtm

Issue #1038, opened 2021-09-18
[Device Support Request]MOES 2Gang inwall switch


Barmatuhin:
MOES 2gang in-wall switch have NO entities after pairing.
![image](https://user-images.githubusercontent.com/1902329/133897875-13c8195b-9c05-4130-9772-beb672d2571a.png)

**Describe the solution you'd like**
Add support for TS0601 _TZE200_g1ib5ldv for first and second gang.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_g1ib5ldv",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}

**Additional context**
![image](https://user-images.githubusercontent.com/1902329/133897944-593976da-dba3-4986-9921-b099e148b140.png)

MattWestb:
Its main powered but have not the flag sett for it.
Do you getting `zha_event`  from the development tools then pressing the gangs ?

MattWestb:
Sorry i was not seen that is one T0601 device and needing one tuya quirk for working at all :-((

Barmatuhin:
> 
> 
[...]

PR #1126, opened 2021-11-04
Tuya dp switches

MattWestb:
Renaming `ts0601_singleswitch.py` to `ts0601_switch.py` and moving the switches from `ts0601.py` and renaming it to `ts0601_cover.py` so all device types have there own quirk and not one mess.
Upcoming support of multi gang switches #1038 can then being implanted in one quirk and not spitted in many quirks.


MattWestb:
Looks like the
```
class TuyaZemismartSmartCover0601(TuyaWindowCover):
class TuyaZemismartSmartCover0601_2(TuyaWindowCover):
```
and
```
class TuyaMoesCover0601(TuyaWindowCover):
class TuyaZemismartSmartCover0601_3(TuyaWindowCover):
```
is identical on cluster and endpoint configuration (but with one typo in signature description) but i dont merger them then i dont like braking if its not 110% correct.

If some like looking if they is identical i  think its better making one `Time In`, `Time Out` and one `Without Time` and perhaps in the future the same whit ZGPP and getting 2 lesser classes that is not needed in the system.

MattWestb:
Thanks David !!

PR #976, opened 2021-07-25
Adding some more tuya TS0601 touch switches

MattWestb:
Adding:
```
         MODELS_INFO: [
            ("_TZE200_oisqyl4o", "TS0601"),
            ("_TZE200_wunufsil", "TS0601"),
            ("_TZE200_vhy3iakz", "TS0601"),
            ("_TZ3000_uim07oem", "TS0601"),
            ("_TZE200_tz32mtza", "TS0601"),
```
The quirk is made  for only one gang switches and some of the new is multiple one.
The first gang is working OK but the 2 and higher is not working :-((

For the moment is better then not working at all but its needs little patching for working OK with all gangs.
Some users is trying getting the code fixed but i think its OK for first shut for getting the users getting little use of there devices.

Feel free reacting this PR if not liking the approach or helping fixing the code for supporting multiple gang devices.

Issue #938, opened 2021-06-24
[Device Support Request] TS0601 _TZE200_fhn3negr (Termostat Knob)


JoostvdB94:
**Is your feature request related to a problem? Please describe.**
Bought a thermostat knob in China. It does pair, but does not expose any controls or status.

**Describe the solution you'd like**
I would like to be able to control the target temperature, and read current temperature and battery level

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
```
{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0, *allocate_address=True, *complex_descriptor_available=False, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False, *is_valid=True, *logical_type=<LogicalType.EndDevice: 2>, *user_descriptor_available=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0x000a",
        "0xef00"
      ],
      "out_clusters": [
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_fhn3negr",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

**Additional context**
Product: https://a.aliexpress.com/_BPiRx7
More information is also provided in another Github issue on another project.
https://github.com/Koenkk/zigbee-herdsman-converters/issues/1803

Note: I'm using ZHA via a Tasmotized Sonoff ZBBridge

MattWestb:
Its looks like [Z2M](https://github.com/Koenkk/zigbee-herdsman-converters/issues/1803) under the nale "Alt Moes TRV" have doing all the work for getting the all the functions in place but have not getting one PR for adding it in the Z2M. de(F)CONZ have not trying getting it in the system but is one version of Moes TRV but i think its its not 100% compatible with the "normal" ones.

You can trying adding the model ID in the valve.pv and testing if its working good enough.
[...]

PR #1129, opened 2021-11-07
Adding MaxSmart TRV support

MattWestb:
Is fixing #677 #976 #1061 and half #938

<html>
<body>

OEM / Brand | Model | Manufacturer Name | Model ID | ZHA Quirk | Request
-- | -- | -- | -- | -- | --
MAXSMART | MS-SH4 ZG | _TZE200_fhn3negr | TS0601 | ☑️ MaxS | 938
MAXSMART | MS-SH3 ZG | _TZE200_qc4fpmcn | TS0601 | ☑️ MaxS | 677
Moes | SH4-ZB | _TZE200_zion52ef | TS0601 | ⭕MaxS ? | 938
AWOW | MS-C-4 | _TZE200_thbr5z34 | TS0601 | ☑️ MaxS | 978
Essentials | 120144 | _TZE200_i48qyn9s | TS0601 | ☑️ MaxS | 677
Silvercrest (LIDL) | 100335076 | _TZE200_chyvmhay | TS0601 | ☑️ MaxS | 1061

</body>
</html>

Also need adding presets in ZHA that must being merged for the quirk working OK. 
https://github.com/home-assistant/core/pull/59289

All code and brain is @jacekk015 i only posing his work done for supporting this TRVs.

All "normal" test is fixed but the test  coverage is out of my knowledge :-((

### DIFF
--- a/zhaquirks/tuya/ts0601_trv_maxsmart.py
+++ b/zhaquirks/tuya/ts0601_trv_maxsmart.py
@@ -224,6 +224,7 @@ class MaxsmartManufCluster(TuyaManufClusterAttributes):
             )
 
     def away_cluster_get(self, field, attributes):
+        """Def away mode cluster."""
         year = self.endpoint.device.MaxsmartAwayYear_bus.listener_event("get_value")[0]
         month = self.endpoint.device.MaxsmartAwayMonth_bus.listener_event("get_value")[
             0
@@ -565,12 +566,15 @@ class MaxsmartAwayYear(LocalDataCluster, AnalogOutput):
         self._update_attribute(self.attridx["engineering_units"], 67)
 
     def set_value(self, value):
+        """Def set away mode year."""
         self._update_attribute(self.attridx["present_value"], value + 2000)
 
     def get_value(self):
+        """Def get away mode year."""
         return self._attr_cache.get(self.attridx["present_value"])
 
     async def write_attributes(self, attributes, manufacturer=None):
+        """Def write away mode year."""
         for attrid, value in attributes.items():
             if isinstance(attrid, str):
                 attrid = self.attridx[attrid]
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts0601_trv_maxsmart.py
```diff
@@ -0,0 +1,1294 @@
...
+            lambda value: value / 2,
+        ),
+        MAXSMART_WINDOW_DETECT_TIME_ATTR: ("window_detection_time", None),
+    }
+
+    def _update_attribute(self, attrid, value):
+        super()._update_attribute(attrid, value)
+
+        if attrid in self.DIRECT_MAPPED_ATTRS:
+            self.endpoint.device.thermostat_bus.listener_event(
```
dmulcahey, line 131: We can stop using bus and we can do something like this: https://github.com/zigpy/zha-device-handlers/blob/10dbcc4d64fdf748989e6d72ea169bc7163eb6ad/zhaquirks/xiaomi/aqara/tvoc.py#L31 I would like to eventually remove Bus from the codebase altogether

jacekk015: So this example is calling using ep_attribute name directly??

RobPlu:
do i understand it correctly that this can't be merged, because more tests need to be written? Shouldn't they also be similar to existing ones?

RobPlu:
do i understand it correctly that this can't be merged, because more tests need to be written? Shouldn't they also be similar to existing ones?

MattWestb:
Closing by demand from @jacekk015 then hi is doing the update instead and not need ding it by second hand editing.

zhaquirks/tuya/ts0601_trv_maxsmart.py
rforro: > We can stop using bus and we can do something like this:
> 
> https://github.com/zigpy/zha-device-handlers/blob/10dbcc4d64fdf748989e6d72ea169bc7163eb6ad/zhaquirks/xiaomi/aqara/tvoc.py#L31
> I would like to eventually remove Bus from the codebase altogether

The `self.endpoint.device.thermostat_bus` comes from inherited class `TuyaThermostatCluster`.

https://github.com/MattWestb/zha-device-handlers/blob/01f7ed0787319c2da85c05fa242c829f334b0cb4/zhaquirks/tuya/__init__.py#L513

At first we have to remove the bus from this class and then rewrite all tuya TRV quirks. Which cannot be tested, because nobody owns all TRVs. Or we could create new class `TuyaThermostatCluster2` without bus and slowly migrate all the TRVs.

PR #1131, opened 2021-11-07
Tuya ts0121_plug: Fix multiplier/divider for current

jmoransalama:
Fixes incorrect value for Current on tuya ts0121 plug.


flabbamann (approved):
Great, I was just about to push the same 😉. I can confirm this works with my Blitzwolf BW-SHP13.
Thank you 👍

PR #1139, opened 2021-11-11
Adding  battery AA * 2 and 3

MattWestb:
Adding battery attribute data for 2 * AA and 3 * AA for current TRV and upcoming ones.
Then its merged it can being implanted in the TRV quirks but i dont do it in this quirk then its so much problem re basing all the time.


MattWestb:
https://github.com/zigpy/zha-device-handlers/pull/1140 is depending on this and failing its tests until its merged.

Issue #1005, opened 2021-08-24
[BUG] Danfoss Ally valve: Slow actuator response [solved already in other solutions]


WolfRevo:
**Describe the bug**
The problem is well described here: https://community.home-assistant.io/t/danfoss-ally-valve-slow-actuator-response-solved/254939

So if you issue a new temperature target setpoint for the thermostat using HA+ZHA the thermostat uses the internal PID mechanism to control the valve which results in slow (and sometimes unexpected) valve control movements.

#
**To Reproduce**
1. Go to your Danfos Ally thermostat connected using ZHA
2. Change the desired temperature
3. Watch the pi_heating_demand changing over time (far to slow)

**Expected behavior**
The pi_heating_demand should change much faster

**Screenshots**
![grafik](https://user-images.githubusercontent.com/38487435/130641345-7c994dc1-d245-487d-b1e5-5507f60d2043.png)

**Additional context**
The issue is already fixed in deconz (https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4408) as well as in the Koenkk/zigbee-herdsman-converters (https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592)

So mainly the solution is well described in the official document (https://assets.danfoss.com/documents/176987/AM375549618098en-000101.pdf):

![grafik](https://user-images.githubusercontent.com/38487435/130642115-8f79652a-94fe-47d4-b838-199b1e709a25.png)

The zigbee coordinator needs to send a Setpoint Command including the HeatingSetpoint itself aswell as a bitmask (enum8 set to 1) to circumvent the thermostats PID mechism.

Based on that I started to read carefully all instructions of the zha quirks because I wanted to sort out if I can create a custom zhaquirk (based on https://github.com/zigpy/zha-device-handlers/tree/dev/zhaquirks/danfoss) fixing this issue.

While I already identified that I need to add a custom output cluster in the replacement (and I guess for that I also need to add the thermostat output cluster to the signature itself?) I didn't find any good example (first I thought the tuya custom quirks were good to understand but I honestly gave up) on how to understand:
1) how do I implement a custom command in the quirk?
2) (in this specific case) how can I access the occupied_heating_setpoint sent by HA+ZHA to the thermostat (which is based on the solution above an int16)
[...]

PR #1148, opened 2021-11-13
add D5X84YU to Danfoss quirk

hertleinj:
Bought two of the Danfoss thermostats under the oem named Popp . 
Without the updated quirks no updates could be triggered from HA.
After the changes updateing from HA works flawless (at least the last day).
The already working way thermostat -> HA still works.

Popp Danfoss was also already mentioned in #1005 

First try more than open for input. Ty.

### DIFF
--- a/zhaquirks/danfoss/__init__.py
+++ b/zhaquirks/danfoss/__init__.py
@@ -1,2 +1,3 @@
 """Module for Sinope quirks implementations."""
 DANFOSS = "Danfoss"
+D5X84YU = "D5X84YU"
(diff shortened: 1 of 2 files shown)


MattWestb:
I was trying adding it and the Hive TRV that also shall being the same hard and software but was not  getting positive feedback from the user so i was not posing the PR.
If you like you can adding it in the same PR or we doing it then user is asking for it.

https://github.com/zigpy/zigpy/issues/840#issuecomment-958823982

Well done !!

Issue #1088, opened 2021-10-21
[Device Support Request] Aqara E1 door/window sensor MCCGQ14LM


haralake:
**Is your feature request related to a problem? Please describe.**
I'm using Home assistant version 2021.10.6 with conbee 2 and  i'm managed to pair the sensor with success. The problem is that the sensor is not working properly.

**Describe the solution you'd like**
Device is not supported

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4447, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0402",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0500",
        "0xfcc0"
      ],
      "out_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "LUMI",
  "model": "lumi.magnet.acn001",
  "class": "zigpy.device.Device"
}
```

**Additional context**
![Καταγραφή στο Web_21-10-2021_14316_192 168 1 20](https://user-images.githubusercontent.com/18325407/138265094-b8548540-cffc-460e-989c-5c5e7894f63f.jpeg)

MattWestb:
Its one new device that is using isa zone and the old was using the normal on/off cluster.
```
    cluster_id = 0x0500
    name = "IAS Zone"
    ep_attribute = "ias_zone"
```
[...]

PR #1151, opened 2021-11-13
Update Aqara quirks for various Zigbee 3.0 devices

dmulcahey:
This PR adds battery support to the Aqara E1 contact sensor. The device appears to follow the ZCL spec for all clusters and the Xiaomi custom report appears to be sent on cluster `0xFCC0` via attribute `0x00F7`. The basic cluster implementation was refactored to accommodate this. 

These devices (Aqara E1 devices) require that we respond w/ a certain manufacturer id when they join the network. Z2M has produced new firmwares to make this work w/ TI coordinators. I have been able to get these to work by making a small change to bellows to set the manufacturer code to what these devices expect at network startup. We are still evaluating the best path forward to support these. 

This PR also adds support for the Aqara T1 motion sensor so that it continues to function after the changes to enable the E1 contact sensor. 

This PR also updates the TVOC quirk so that it gets the Xiaomi heartbeat data and updates the appropriate clusters.

### DIFF
--- a/zhaquirks/xiaomi/aqara/motion_agl02.py
+++ b/zhaquirks/xiaomi/aqara/motion_agl02.py
@@ -1,9 +1,9 @@
-"""Xiaomi aqara E1 contact sensor device."""
-
+"""Xiaomi aqara T1 motion sensor device."""
 from zigpy.profiles import zha
 from zigpy.zcl.clusters.general import Identify, Ota
-from zigpy.zcl.clusters.security import IasZone
+from zigpy.zcl.clusters.measurement import OccupancySensing
 
+from zhaquirks import Bus
 from zhaquirks.const import (
     DEVICE_TYPE,
     ENDPOINTS,
@@ -15,6 +15,9 @@ from zhaquirks.const import (
 from zhaquirks.xiaomi import (
     LUMI,
     BasicCluster,
+    IlluminanceMeasurementCluster,
+    MotionCluster,
+    OccupancyCluster,
     XiaomiAqaraE1Cluster,
     XiaomiCustomDevice,
     XiaomiPowerConfiguration,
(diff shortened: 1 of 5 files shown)


nsumar:
image

I've added this quirk and am still not able to use the device - I am happy to help through testing in any way I can.

zhaquirks/xiaomi/aqara/motion_agl02.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class XiaomiManufacturerCluster(XiaomiAqaraE1Cluster):
+    """Xiaomi manufacturer cluster."""
+
+    def _update_attribute(self, attrid, value):
+        super()._update_attribute(attrid, value)
+        if attrid == 274:
+            value = value - 65536
+            self.endpoint.illuminance.illuminance_reported(value)
```
Adminiuga, line 36: So illuminance update is also a trigger for  occupancy?

dmulcahey: unfortunately.... I'll link you the z2m thread

dmulcahey: Start here: https://github.com/Koenkk/zigbee2mqtt/issues/5706#issuecomment-971810515

Adminiuga: What [censored]...

robtpaton:
> image
> 
> I've added this quirk and am still not able to use the device - I am happy to help through testing in any way I can.

Hey @nsumar did you get this working in the end? Not sure if we are on the same versions of various things (I have listed my versions here https://github.com/zigpy/zha-device-handlers/issues/1088#issuecomment-999149260) but I seem to be having the same problem as you, battery stays as unknown, contact stays as off.

TIA

nsumar:
@robtpaton Yeah I did, not sure if you're using conbee ii, but I needed to upgrade the firmware using this  https://community.home-assistant.io/t/zigbee-problems/213121/5

As soon as I upgraded the firmware, I could pair the sensors, works perfectly.

nsumar:
> @robtpaton Yeah I did, not sure if you're using conbee ii, but I needed to upgrade the firmware using this  https://community.home-assistant.io/t/zigbee-problems/213121/5
> 
> As soon as I upgraded the firmware, I could pair the sensors, works perfectly.

@robtpaton read your comment on your linked post. Use the steps I provided, the deconz add-on doesn't work to update firmware. I read about a bug somewhere. Install deconz on your localmachine and then flash the latest firmware using GCFlasher

robtpaton:
Thanks @nsumar, flashed manually as you suggested and now works. For anyone else looking back at this - I'd got sidetracked using the deconz add-on, desperately hoping that would work as I thought GCFlasher wasn't usable on OSX which is all I have available to me right now. However GCFlasher is buildable on OSX so was able to use that. Had to use a USB extension lead to flash, but other than that all good

Lyr3x:
I updated to the latest Conbee II firmware `deCONZ_ConBeeII_0x26720700.bin.GCF` and added two new sensors, but both do not report battery nor does the sensor status change.

robtpaton:
@Lyr3x have you tried removing the devices from HA and re-pairing from scratch? Also did you update the firmware on the Conbee using GCFlasher?

Lyr3x:
Yes I already updated the fw of the conbee multiple times with the GCCFlasher. I was just three versions behind. 

If you mean "remove this device" via the ZHA integration and pair it again? Yes done that multiple times. Actually one started to report the state correctly but the other one does still not.

dmulcahey:
That means it wasn’t reset and joined fresh… remove it with the remove device button then fully reset the device. Then pair it to ZHA again.

Lyr3x:
How do I fully reset the device besides pressing the button a few sec ?

dmulcahey:
Hold it for approx 10 seconds

Lyr3x:
Just to understand is the fast blinking led what I am looking for? Because that's what I did. I'll give it a shot anyway tomorrow 🙂

PR #1153, opened 2021-11-13
Add command_bus to TuyaManufCluster

javicalle:
This PR adds a new `command_bus` to TuyaManufCluster.
This new bus would listen to MCU commands from other endpoints and talk to MCU.

There are some points to review:
* [x] Tests are failing. Trying to fix it
* [ ] Not sure about naming. MCU or DP commands?
* [x] `TuyaOnOff.command()` changed to `async`
This is one of those changes that you make to make it work, but that you just don't understand. I am concerned about the impact of this change (I am not fluent in Python)

The next thing will be to upload a quirk that makes use of this approach to show its use.

I have thinked about duplicate the `TuyaManufCluster` implementation and create a new subclass, but I understand that the purpose of the `TuyaManufCluster` class is precisely to send the MCU commands so I think it is the right place to implement the functionality.

IMHO the listener approach should replace the use of `self.endpoint.tuya_manufacturer.command ()` but before modifying all the implementations I prefer to validate it with a few uses (ClusterOnOff) and if it does not generate incidents, migrate to the rest of uses (WindowCover , LevelControl, Motion, ...). 
Or maybe it can be applied in those cases that require it (multi switch type). 

(Fixes: #1038 )

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -133,6 +133,80 @@ async def test_singleswitch_state_report(zigpy_device_from_quirk, quirk):
     assert switch_listener.attribute_updates[1][1] == OFF
 
 
+@pytest.mark.parametrize("quirk", (zhaquirks.tuya.ts0601_switch.TuyaDoubleSwitchTO,))
+async def test_doubleswitch_state_report(zigpy_device_from_quirk, quirk):
+    """Test tuya single switch."""
+
+    ZCL_TUYA_SWITCH_COMMAND_03 = b"\tQ\x03\x006\x01\x01\x00\x01\x01"
+    ZCL_TUYA_SWITCH_EP2_ON = b"\tQ\x02\x006\x02\x01\x00\x01\x01"
+    ZCL_TUYA_SWITCH_EP2_OFF = b"\tQ\x02\x006\x02\x01\x00\x01\x00"
+
+    switch_dev = zigpy_device_from_quirk(quirk)
+
+    switch1_cluster = switch_dev.endpoints[1].on_off
+    switch1_listener = ClusterListener(switch1_cluster)
+
+    switch2_cluster = switch_dev.endpoints[2].on_off
+    switch2_listener = ClusterListener(switch2_cluster)
+
+    tuya_cluster = switch_dev.endpoints[1].tuya_manufacturer
+
+    assert len(switch1_listener.cluster_commands) == 0
+    assert len(switch1_listener.attribute_updates) == 0
+    assert len(switch2_listener.cluster_commands) == 0
+    assert len(switch2_listener.attribute_updates) == 0
+
+    # events from channel 1 updates only EP 1
+    hdr, args = tuya_cluster.deserialize(ZCL_TUYA_SWITCH_ON)
+    tuya_cluster.handle_message(hdr, args)
+    assert len(switch1_listener.attribute_updates) == 1
+    assert len(switch2_listener.attribute_updates) == 0
+    assert switch1_listener.attribute_updates[0][0] == 0x0000
+    assert switch1_listener.attribute_updates[0][1] == ON
+
+    # events from channel 2 updates only EP 2
+    hdr, args = tuya_cluster.deserialize(ZCL_TUYA_SWITCH_EP2_ON)
+    tuya_cluster.handle_message(hdr, args)
+    assert len(switch1_listener.attribute_updates) == 1
+    assert len(switch2_listener.attribute_updates) == 1
+    assert switch2_listener.attribute_updates[0][0] == 0x0000
+    assert switch2_listener.attribute_updates[0][1] == ON
+
+    hdr, args = tuya_cluster.deserialize(ZCL_TUYA_SWITCH_OFF)
+    tuya_cluster.handle_message(hdr, args)
+    assert len(switch1_listener.attribute_updates) == 2
+    assert len(switch2_listener.attribute_updates) == 1
+    assert switch1_listener.attribute_updates[1][0] == 0x0000
+    assert switch1_listener.attribute_updates[1][1] == OFF
+
+    hdr, args = tuya_cluster.deserialize(ZCL_TUYA_SWITCH_EP2_OFF)
+    tuya_cluster.handle_message(hdr, args)
+    assert len(switch1_listener.attribute_updates) == 2
+    assert len(switch2_listener.attribute_updates) == 2
+    assert switch2_listener.attribute_updates[1][0] == 0x0000
+    assert switch2_listener.attribute_updates[1][1] == OFF
+
+    assert len(switch1_listener.cluster_commands) == 0
+    assert len(switch2_listener.cluster_commands) == 0
+
+    # command_id = 0x0003
+    hdr, args = tuya_cluster.deserialize(ZCL_TUYA_SWITCH_COMMAND_03)
+    tuya_cluster.handle_message(hdr, args)
+    assert len(switch1_listener.cluster_commands) == 0
+    assert len(switch2_listener.cluster_commands) == 0
+    # no switch attribute updated (Unsupported command)
+    assert len(switch1_listener.attribute_updates) == 2
+    assert len(switch2_listener.attribute_updates) == 2
+
+    hdr, args = tuya_cluster.deserialize(ZCL_TUYA_SET_TIME_REQUEST)
+    tuya_cluster.handle_message(hdr, args)
+    assert len(switch1_listener.cluster_commands) == 0
+    assert len(switch2_listener.cluster_commands) == 0
+    # no switch attribute updated (TUYA_SET_TIME command)
+    assert len(switch1_listener.attribute_updates) == 2
+    assert len(switch2_listener.attribute_updates) == 2
+
+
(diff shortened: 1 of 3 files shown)


MattWestb:
Comment on naming;
tuya INIT is little clouded then its one mix of DP and Zigbee functions and its not so easy see witch category they belongs.
I have trying adding new Zigbee cluster functions with ZB in the beginning for easier see its one Zigbee Cluster the OnOff function is belonging,

For the moment its looks better with quirk naming after little struggle then all DP devices (ts0601) have getting names that is grouping them and have splitting / moving devices so its not one big mess.

One good thing but its out of scope of you work is moving all DP quirks to one sub folder and one for Zigbee cluster devices so not mixing all up and can have there own INIT for DP and Zigbee quirks and dont messing all functions in one file.

But in the end im only copy / paste user and no code worrier.

And great that you is looking in the code and trying getting it better and adding our missed multi endpoints commands for our tuya DP devices !!!

javicalle:
I have checked that `async def command()` has been implemented in other cases:
* https://github.com/zigpy/zha-device-handlers/blob/9f6246bb303147b4ae7e07bfdcb6f3510f4e7ee0/zhaquirks/tuya/__init__.py#L595
* https://github.com/zigpy/zha-device-handlers/blob/9f6246bb303147b4ae7e07bfdcb6f3510f4e7ee0/zhaquirks/tuya/ts0601_siren.py#L87
* https://github.com/zigpy/zha-device-handlers/blob/9f6246bb303147b4ae7e07bfdcb6f3510f4e7ee0/zhaquirks/tuya/ts0601_trv.py#L789

So I'm going to assume the change is correct. 
I have also fixed the tests that were failing

javicalle:
`TuyaDoubleSwitchTO` added to this PR.

How is it working?
There are 2 pieces working here:
* `TuyaManufCluster` publishes a` command_bus` to the device. The events caught by the listener are handled in `tuya_mcu_command` method
* the `TuyaOnOff` cluster sends its endpoint commands to the bus to be processed

This model can be extended to devices with any amount of `TuyaOnOff` cluster/endpoint. Just add a new endpoint with a `TuyaOnOff` cluster. 
**IMPORTANT**: the endpoint number must match the MCU channel

For now it does not work with other types of Tuya clusters. IMHO once we verify that it does not generate incidents, it could be extended to other clusters that need it.
I think that modifying the `TuyaManufCluster` should not have any effect on the quirks that can use it.
Modifying the `TuyaOnOff` cluster could only affect the` ts0601_dimmer` and `ts0601_switch` devices but it should not affect

MattWestb:
Is it also working with decoupled mode switches (the light part and the remote is braked and is treated as one light and one remote) or is it needed having  both in and out cluster in replacement of the new" EPs ?

Implanting the DP commands for switch config and decoupled mode is somthing for later but its upcoming coming devices that need it (device request is made of users).

Is it possible see witch EP the MCU is sending the its command if have quirk debug activated (for helping user adding devices) ?

Great work !!

javicalle:
> Is it also working with decoupled mode switches (the light part and the remote is braked and is treated as one light and one remote) or is it needed having both in and out cluster in replacement of the new" EPs ?

It depends how the remote are working in Tuya devices. 
I assume that each switch will have its endpoint with an onoff cluster and a remote cluster.
The switch part should work with the current implementation. The remote `TuyaSmartRemoteOnOffCluster` has not been adapted so it would not work for switches other than 1.


> Is it possible see witch EP the MCU is sending the its command if have quirk debug activated (for helping user adding devices) ?

The device broadcast the switch status after any physical interaction, and in my case, debug traces for `tuya` package show these traces when you press the physical device:
```
2021-11-14 14:19:16 DEBUG (MainThread) [custom_zha_quirks.tuya] 5c:02:72:ff:fe:1a:11:75 - Received switch event message, channel: 1, state: 1
2021-11-14 14:19:16 DEBUG (MainThread) [custom_zha_quirks.tuya] 5c:02:72:ff:fe:1a:11:75 - Received switch event message, channel: 2, state: 0
2021-11-14 14:19:16 DEBUG (MainThread) [custom_zha_quirks.tuya] 5c:02:72:ff:fe:1a:11:75 - Received switch event message, channel: 782, state: 2
2021-11-14 14:19:16 DEBUG (MainThread) [custom_zha_quirks.tuya] 5c:02:72:ff:fe:1a:11:75 - Received switch event message, channel: 783, state: 2
```

These are the states for switch 1 (on) and 2 (off). So I know that the second switch is channel 2 --> endpoint 2
Before you asking, channels 783 and 782 are the device `poweron` and `backlight` modes. They are also possible to manipulate, but I will cover that in another PR later.

dmulcahey:
Can we do this without `Bus`? Something like this: 

https://github.com/zigpy/zha-device-handlers/pull/1155#discussion_r749379594 

Would this work here?

javicalle:
> Can we do this without `Bus`? Something like this:
> 
> [#1155 (comment)](https://github.com/zigpy/zha-device-handlers/pull/1155#discussion_r749379594)
> 
> Would this work here?

Not sure. The question here is that the OnOff cluster from _any_ endpoint must talk with the endpoint with the TutaManufCluster. Usually this will be endpoint 1
So, maybe, I could try something like:
```
self.endpoint.device.endpoints[1].tuya_manuf.command(whatever)
```
It seems a little complicated and against the architecture (I had read something about communication between endpoints).

I don't know if there is a way to access directly to one endpoint (similar to `ep_attribute` in cluster), I didn't find it.

javicalle:
Surely it is not clear in the detail of the PR.
This solution approach for those Tuya devices that send different commands (for different switches) through a single endpoint is to generate new 'fake' endpoints that receive the state changes and communicate to the TuyaManufCluster the commands associated with the switch they simulate. So, any 2 gang device would define an endpoint 2 to do all the work.

The existing problem until now was the communication between the different endpoints, and that is the point that the use of the Bus seeks to solve: the communication between different endpoints.

If it were possible a solution like `self.endpoint.device.the_tuya_endpoint.tuya_manufacturer.command ()` IMHO would be more appropriate than assuming that the TuyaManufCluster will be on the endpoint [1].

javicalle:
I have rechecked the `TuyaNewManufCluster` implementation and I think it would somehow serve what we are trying to get around.
I'm going to try to review it and do some tests, but I have to say that I have a hard time following the implementation to understand how I can extend it.

dmulcahey:
If you’re good with this for now I don’t mind merging it until you can assess the other cluster. Just LMK

MattWestb:
Alexei was using on other method but i dont knowing if its working sending and receiving  commands on different  end points but hi is using it  for attributes reported from the device.
It using dict and no bus:
https://github.com/zigpy/zha-device-handlers/tree/dev/zhaquirks/tuya/air

javicalle:
I have done some testing with the `TuyaNewManufCluster` implementation and although I have managed to implement multiple switches to update their state in HA, I have not been able to find a way to implement the opposite way (send the command to the device).

> Alexei was using one other method

Yes, I have look that implementation, but only works in one way (update the HA status). I haven't found any way to implement the command send.

Even so, the 'problem' of sending commands between endpoints is still there and it has to be solved in some way (either with the Bus or by recovering the endpoint).

@dmulcahey, I am not able to find another approach to the problem, so for my part I would push the merge forward if you see no problem.

MattWestb:
Perhaps Alexei have some ideas how to do it or working around the problem or making  one patch in Zigpy / ZHA for making it possible both receiving and sending commands to and from one or more "new" endpoints (that i was thinking the bus was made for doing).

jacekk015:
@javicalle Just create a reference first, and then use it instead of self
https://github.com/zigpy/zha-device-handlers/blob/9cfe00565a6167856dc258349c9bd39a7260daa2/zhaquirks/tuya/ts0601_trv_maxsmart.py#L76

```
global MaxsmartManufClusterSelf
MaxsmartManufClusterSelf = self
```
and then
```
return await MaxsmartManufClusterSelf.endpoint.tuya_manufacturer.write_attributes(
                {MAXSMART_CHILD_LOCK_ATTR: value}, manufacturer=manufacturer
            )
```
It's just still a Python

javicalle:
> Just create a reference first, and then use it instead of self
> 
> https://github.com/zigpy/zha-device-handlers/blob/9cfe00565a6167856dc258349c9bd39a7260daa2/zhaquirks/tuya/ts0601_trv_maxsmart.py#L76
> 
> ```
> global MaxsmartManufClusterSelf
> MaxsmartManufClusterSelf = self
> ```
> 

Thank you very much for the contribution. I have to say that I am not fluent in Python.

I think it is something similar to my [first attempt](https://github.com/zigpy/zha-device-handlers/issues/1038#issuecomment-958121633), but it was [not operational](https://github.com/zigpy/zha-device-handlers/issues/1038#issuecomment-962639887) if more than one device of the same type is configured.

Do you think your implementation would not have the same problem? I can't test it because I only have one test device.

jacekk015:
That's because all device types share same name. There's no other way of doing it. You need reference.
That's a replacement for bus, which is not bad in my opinion.
You need to create bus or reference - you need something.

dmulcahey:
Keep the bus for now. We can revisit this in the future.

Issue #966, opened 2021-07-17
[Device Support Request] TZE200_tz32mtza TS0601 (3 gang Moes no neutral wired switch)


Netatout:
**Is your feature request related to a problem? Please describe.**
Device pairs successfully, but no entities are created. 

**Describe the solution you'd like**
I would like a quirk that handles this device and exposes control over both gangs via ZHA. Each switch has a led that can be deactivated. If possible, also have access to this setting.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
Device 0x4286 (60:a4:23:ff:fe:aa:ec:42) joined the network
Device 60:a4:23:ff:fe:aa:ec:42 changed id (0x4286 => 0x4286)
[0x4286] Scheduling initialization
Received frame on uninitialized device <Device model=None manuf=None nwk=0x4286 ieee=60:a4:23:ff:fe:aa:ec:42 is_initialized=False> from ep 0 to ep 0, cluster ZDOCmd.Device_annce: b'\xFF\x86\x42\x42\xEC\xAA\xFE\xFF\x23\xA4\x60\x80'
[0x4286:zdo] ZDO request ZDOCmd.Device_annce: [0x4286, 60:a4:23:ff:fe:aa:ec:42, 128]
Tries remaining: 3
[0x4286] Requesting 'Node Descriptor'
Tries remaining: 2
[0x4286] Extending timeout for 0x47 request
Received frame on uninitialized device <Device model=None manuf=None nwk=0x4286 ieee=60:a4:23:ff:fe:aa:ec:42 is_initialized=False> from ep 0 to ep 0, cluster ZDOCmd.Node_Desc_rsp: b'\x47\x00\x86\x42\x02\x40\x80\x02\x10\x52\x52\x00\x00\x2C\x52\x00\x00'
[...]

Issue #913, opened 2021-06-08
[Device Support Request] _TZE200_g1ib5ldv TS0601 (2 gang Moes no neutral wired switch) ZTS-EUB


khrazik:
**Is your feature request related to a problem? Please describe.**
Device pairs successfully, but no entities are created. I tried to make a custom quirk based off similar tuya switches, but I failed. Interestingly, device seems to report only one endpoint but there are two gang switches (and physically, both work).

**Describe the solution you'd like**
I would like a quirk that handles this device and exposes control over both gangs via ZHA.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**

`2021-06-08 23:04:37 INFO (MainThread) [zigpy.application] Device 0x63cc (60:a4:23:ff:fe:d5:d4:4a) joined the network
2021-06-08 23:04:37 DEBUG (MainThread) [zigpy.application] Skip initialization for existing device 60:a4:23:ff:fe:d5:d4:4a
2021-06-08 23:04:37 DEBUG (MainThread) [zigpy.zdo] [0x63cc:zdo] ZDO request ZDOCmd.Device_annce: [0x63CC, 60:a4:23:ff:fe:d5:d4:4a, 128]
2021-06-08 23:04:39 DEBUG (MainThread) [zigpy.util] Tries remaining: 1
2021-06-08 23:04:39 DEBUG (MainThread) [zigpy.device] [0x63cc] Extending timeout for 0x47 request
2021-06-08 23:04:40 DEBUG (MainThread) [zigpy.application] Received frame on uninitialized device 60:a4:23:ff:fe:d5:d4:4a (Status.NEW) for endpoint: 1
2021-06-08 23:04:40 DEBUG (MainThread) [zigpy.device] [0x63cc] Received ZCL while uninitialized on endpoint id 1, cluster 0x000a id, hdr: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=False disable_default_response=False> manufacturer=None tsn=0 command_id=Command.Read_Attributes>, payload: b'\x07\x00'
2021-06-08 23:04:40 DEBUG (MainThread) [zigpy.device] [0x63cc] Uninitialized device command 'Command.Read_Attributes' args: [[7]]
[...]

Issue #824, opened 2021-03-20
[Device Support Request] EU Tuya ZigBee Smart Light Switch


Kalininator:
I am able to pair the device, but dont get any entities
[Product Link](https://www.aliexpress.com/item/1005002180713176.html?spm=a2g0s.9042311.0.0.7ee34c4dDeZ57Y)

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices 
screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**

```
{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_tz32mtza",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

```
[0x0000:zdo] ZDO request ZDOCmd.Mgmt_Permit_Joining_req: [60, <Bool.false: 0>]
Device 0xe670 (60:a4:23:ff:fe:94:be:74) joined the network
[0xe670] Requesting 'Node Descriptor'
Tries remaining: 2
[0xe670] Extending timeout for 0xec request
Device 0xe670 (60:a4:23:ff:fe:94:be:74) joined the network
Skip initialization for existing device 60:a4:23:ff:fe:94:be:74
Device 0xe670 (60:a4:23:ff:fe:94:be:74) joined the network
Skip initialization for existing device 60:a4:23:ff:fe:94:be:74
[0xe670:zdo] ZDO request ZDOCmd.Device_annce: [0xE670, 60:a4:23:ff:fe:94:be:74, 128]
[0xe670:zdo] ZDO request ZDOCmd.Node_Desc_req: [0x0000]
[0xe670:zdo] Unsupported ZDO request:ZDOCmd.Node_Desc_req
[0xba32] Delivery error for seq # 0xe8, on endpoint id 0 cluster 0x0021: message send failure
[...]

Issue #813, opened 2021-03-12
[Device Support Request] Tuya/MoesHouse wall switch TS0601 _TZE200_amp6tsvy


LouisHerber:
When adding a device as the MoesHouse Zigbee 3.0 wall switch to the Zigbee network using ZHA integration (e.g. : https://aliexpress.com/item/1005002061498115.html), the device will be added, but no endpoints are added.

I've already tried myself to edit the quirk TS0601.py in Tuya folder offline, then placing it in the right folder in my HA and reboot. Not sure if this would be sufficient though, but anyway I can't get it to work - logging shows that zigpy.quirks still doesn't match the re-added device to any quirk. It's probably my programming skill lacking a bit here :-) Therefore I'd like to ask some help to get it added and recognized.

2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.endpoint] [0x4697:1] Manufacturer: _TZE200_amp6tsvy
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.endpoint] [0x4697:1] Model: TS0601
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.quirks.registry] Checking quirks for _TZE200_amp6tsvy TS0601 (_mac_address_)
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.quirks.registry] Considering <class 'zhaquirks.xbee.xbee_io.XBeeSensor'>
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.quirks.registry] Fail because endpoint list mismatch: {232, 230} {1}
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.quirks.registry] Considering <class 'zhaquirks.xbee.xbee3_io.XBee3Sensor'>
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.quirks.registry] Fail because endpoint list mismatch: {232, 230} {1}
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.quirks.registry] Considering <class 'zhaquirks.smartthings.tag_v4.SmartThingsTagV4'>
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.quirks.registry] Fail because device_type mismatch on at least one endpoint
2021-03-12 12:00:57 DEBUG (MainThread) [zigpy.quirks.registry] Considering <class 'zhaquirks.smartthings.multi.SmartthingsMultiPurposeSensor'>
[...]

PR #1166, opened 2021-11-21
Add quirks for Tuya multigang switches

javicalle:
This PR add quirks for Tuya multigang switches.

It should correct the management of the On/Off cluster for the multiswitch devices that make use of the Tuya cluster (`0xEF00`) for command/status reporting.

The `MODELS_INFO` of those devices that have been reported or an issue has been reported have been added.
For those devices whose number of gangs is not clear, they have been defined as 1-gang devices until someone report correct info.

This could fix the following issues (only on/off switches):
fixes: #813
fixes: #824
fixes: #913
fixes: #972
fixes: #966

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -135,7 +135,7 @@ async def test_singleswitch_state_report(zigpy_device_from_quirk, quirk):
 
 @pytest.mark.parametrize("quirk", (zhaquirks.tuya.ts0601_switch.TuyaDoubleSwitchTO,))
 async def test_doubleswitch_state_report(zigpy_device_from_quirk, quirk):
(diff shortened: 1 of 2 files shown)


javicalle:
We have doubts about the characteristics of the following devices:
* `_TZE200_vhy3iakz`
* `_TZ3000_uim07oem`

MattWestb:
I have found both one and multiple reference but only 1 gang for them that is verified.
Z2M is having all of then in the same and no clue how many gangs.
I shall looking more in deCONZ but they dont have do many implanted (then more request that is being slated).

I think its better letting them as 1 gang until user doing request for them.

PR #1167, opened 2021-11-21
TuyaManufClusterAttributes -> write_attributes: add DEBUG logging

jacekk015:
DEBUG logger helpful for implementing new Tuya TRV devices, sometimes attributes are directly updated omitting TuyaThermostatCluster.
It can help to debug code a little better.

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -443,7 +443,7 @@ class TuyaManufClusterAttributes(TuyaManufCluster):
                 self.cluster_id,
                 cmd_payload
             )
-            
+
             await super().command(
                 TUYA_SET_DATA,
                 cmd_payload,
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/__init__.py
```diff
@@ -436,6 +436,14 @@ async def write_attributes(self, attributes, manufacturer=None):
             cmd_payload.function = 0
             cmd_payload.data = Data.from_value(record.value.value)
 
+            _LOGGER.debug(
```
Adminiuga, line 439: You can use `self.debug` on a cluster or device and it would include all identifying information, so you can omit nwk, endpoint and cluster

jacekk015: I've copied that originally from TuyaThermostatCluster
Wanna to keep same approach.
Tested locally on my setup.

PR #1202, opened 2021-12-05
Legrand lightswitch

Dreaming-Codes:
As the title says, support for Legrand`s light switch.


Dreaming-Codes:
I'm really sorry for the 3 commits of just "code beautification" but I'm without a computer and working from an ipad with vscode web is not easy

dmulcahey:
> I'm really sorry for the 3 commits of just "code beautification" but I'm without a computer and working from an ipad with vscode web is not easy

no need to apologize!

PR #1226, opened 2021-12-20
Add support for Polled clusters and use it for ts011f plug Metering

3v1n0:
ts011f plug is [known](https://www.zigbee2mqtt.io/devices/TS011F_plug_3.html) not to properly report Metering information and they need to be polled periodically.

I've implemented this initially just for this plug implementation, but given it can be useful for other devices too, I decided to generalize it a bit (allowing different polling frequency per each attribute, that makes things to complicate a bit as I wanted to prioritize batched queries anyways...).

Let me know if this is not wanted, in any case commits are atomic so I could squash them or just use the initial ones.

### DIFF
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -415,6 +415,142 @@ class DeviceMonitor(Bus):
         self.listener_event("device_removed")
 
 
+class PolledCluster(CustomCluster):
+    """PolledCluster Class to periodically poll cluster attributes."""
+
+    DEFAULT_FREQUENCY = 60
+    POLL_ATTRIBUTES = tuple()
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the cluster."""
+        super().__init__(*args, **kwargs)
+        self._device_monitor = DeviceMonitor(self.endpoint.device)
+        self._device_monitor.add_listener(self)
+        self._timer_handles = {}
+        self._poll_tasks = {}
+        self._poll_attribs = tuple(self.attridx[a] for a in self.POLL_ATTRIBUTES)
+        self._poll_freq = {a: self.DEFAULT_FREQUENCY for a in self._poll_attribs}
+        self._loop = asyncio.get_running_loop()
+
+    def device_initialized(self):
+        """Device monitor callback handler for device initialization."""
+        self._stop_polling_timers()
+        self._start_polling_timers()
+
+    def device_removed(self):
+        """Device monitor callback handler for device removed."""
+        self._stop_polling_timers()
+
+    def _attr_frequency(self, attrid):
+        return self._poll_freq.get(attrid, self.DEFAULT_FREQUENCY)
+
+    def _get_update_frequencies(self):
+        return set(self._poll_freq.values())
+
+    def _attribs_by_frequency(self, frequency):
+        return [a for a, freq in self._poll_freq.items() if freq == frequency]
+
+    def _has_frequency_timer_running(self, frequency):
+        same_frequency_attrs = self._attribs_by_frequency(frequency)
+        return any(a in self._timer_handles for a in same_frequency_attrs)
+
+    def _start_polling_timers(self):
+        assert not self._timer_handles
+
+        for freq in self._get_update_frequencies():
+            self._start_polling_timers_by_frequency(freq)
+
+    def _start_polling_timers_by_frequency(self, freq):
+        if self._has_frequency_timer_running(freq):
+            return
+
+        attribs = self._attribs_by_frequency(freq)
+        timer = self._loop.call_later(freq, self._queue_poll_attribs, attribs, freq)
+        self._timer_handles.update({a: timer for a in attribs})
+
+    def _stop_polling_timers(self):
+        for timer in self._timer_handles.values():
+            timer.cancel()
+        self._timer_handles = {}
+
+        for task in self._poll_tasks.values():
+            task.cancel()
+        self._poll_tasks = {}
+
+    async def _poll_attribs_async(self, attribs, freq):
+        cancelled = False
+        try:
+            await self.read_attributes(attribs)
+        except asyncio.CancelledError:
+            cancelled = True
+        except asyncio.TimeoutError:
+            _LOGGER.warn(
+                "Timed out while polling device %s attributes %s",
+                self.endpoint.device.ieee,
+                attribs,
+            )
+        except Exception:
+            _LOGGER.exception(
+                "Failed to poll device %s attributes", self.endpoint.device.ieee
+            )
+        finally:
+            self._poll_tasks.update({a: None for a in attribs})
+            if not cancelled:
+                self._start_polling_timers_by_frequency(freq)
+
+    def _queue_poll_attribs(self, attribs, freq):
+        attribs_to_poll = []
+        for attrid in attribs:
+            if self._timer_handles.pop(attrid, None) and not self._poll_tasks.get(
+                attrid
+            ):
+                attribs_to_poll.append(attrid)
+
+        if attribs_to_poll:
+            task = self._loop.create_task(
+                self._poll_attribs_async(attribs_to_poll, freq)
+            )
+            self._poll_tasks.update({a: task for a in attribs_to_poll})
+
+    def _configure_reporting(self, args, manufacturer=None):
+        """Configure reporting."""
+        frequencies = set()
+        for arg in [a for a in args if a.attrid in self._poll_attribs]:
+            self._skip_update(arg.attrid, restart=False)
+            frequency = min(
+                max(arg.min_interval, self.DEFAULT_FREQUENCY), arg.max_interval
+            )
+            for freq in self._get_update_frequencies():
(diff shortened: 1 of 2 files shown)


dmulcahey:
I think this is a great contribution. Any chance you can add tests for the device monitor and polling cluster implementations

dmulcahey:
@Adminiuga / @puddly any thoughts on this one?

dmulcahey:
We may want to start documenting reusable parts like this also… probably time to start creating development docs aside from the getting started stuff to make it easier for folks to contribute

Adminiuga:
IMO we're taking the wrong approach with Tuya -- we are over complicating the quirks, just to make Tuya to appear as a regular ZCL compliant device and they are not. 

I think the quirks just should define the data point types in the manufacturer specific channel and essentially in ZHA we  define  the entities based on the data points, manufacturer and model (like we somewhat do already for Tuya valves)

In other words instead of the quirk reading the attribute regularly and exposing that data over the metering cluster, let's have a polled ZHA sensor entity which reads the  consumption data point directly and exposes it as a meeting sensor?

3v1n0:
@dmulcahey yeah, I was thinking indeed to add tests... However I also wanted a prior feedback on the idea first :)

@Adminuga, sure I see that approach as something fair indeed. I wanted to get something quickly working for my setup so I started at the point where I had to touch less aspects.

However, in the ZHA approach: I would think that zha-device-handlers should still provide an hint to ZHA about whether they want a polled sensor for a entity or not, no?

As I assume other devices like this may be around and it would be annoying to handle this specifically for them all, while just changing a quick bit could be easier and quicker.

Adminiuga:
> However, in the ZHA approach: I would think that zha-device-handlers should still provide an hint to ZHA about whether they want a polled sensor for a entity or not, no

How does this quirk know that it has to poll the "consumption" cluster/data point? Zha could use the same criteria. 
And depending how it's done, may take same approach bas with HUE lights which are also polled.

3v1n0:
Well in this case we're just hardcoding this information...
But I had the feeling that this was also the point of this repository: add all the device specific information so that upper levels can just process the data accordingly.

Thus, while I'm fine to have ZHA to handle the polling itself, IMHO would be better if the device handlers would expose to it (and other zygpy customers, if any) such detail. In this way, there's not device-specific code hardcoded in ZHA, but only class-specific (i.e. `device.pid == 'foo' and device.gid == 'bar'` vs `device.cluster_foo.needs_polling()`.

I was wondering if in some cases there should be also a "fallback" mode for ZHA to just try to poll (at least the powered) devices when some endpoint doesn't get updated for a long time, but not sure if that would be correct to cover all the cases.

dmulcahey:
> > However, in the ZHA approach: I would think that zha-device-handlers should still provide an hint to ZHA about whether they want a polled sensor for a entity or not, no
> 
> How does this quirk know that it has to poll the "consumption" cluster/data point? Zha could use the same criteria. And depending how it's done, may take same approach bas with HUE lights which are also polled.

while I see your point I'll add that doing it here would allow this functionality to work for the other home automation platforms that are adopting Zigpy and it would allow us to keep mfg specifics out of HA. If we can make this clean enough it may make sense to migrate what is already in HA to quirks... either way will work it's more a matter of how do we want this to work?

Adminiuga:
> it may make sense to migrate what is already in HA to quirks

As you mentioned, there have to be device/manufacturer specifics and whether those should be inside HA/ZHA is another topic which I don't have strong feeling about.

Quirks are fine for fixing a small deviation/non standard behavior here or there, but Tuya proved, bringing their devices to look like a normal ZCL compliant implementation could be done, but quickly gets out of hand and increases complexity.

In other words, zigpy/quirks should be left to deal with ZigBee and ZCL layer interfacing (like defining new attributes, commands or fixing/overriding some attribute values), and manufacturer/device specific behavior/implementation should be handled at the layer above it. Maybe split ZHA channels out of HA and morph them into proto-entites/channels?

smiba:
My Blitzwolf SHP-15 power plugs are crippled because as this pull request describes, they do not automatically report the total power consumed and require polling.

In the meanwhile, is there a way I can get ZHA to poll this manually? (By adding a timed automation or even an external script etc.). Restarting the ZHA integration retrieves this value once, but I can't really just go around and restart ZHA every 4 hours or so...

dumpfheimer:
I wrote a python script that uses the home assistant websocket to detect the plugs and poll the energy consumption: https://github.com/dumpfheimer/hass_zha_attr_poll_workaround

Maybe this is enough for you or at least gives you an idea on how to make your own workaround.

SuperBigB:
I have a lot of these plugs. How can I implement this?

dumpfheimer:
Hi @3v1n0

May I ask, how many of these plugs do you have in your network?

The reason why I am asking is: I made a workaround over websockets and polled the energy for all plugs in a regular interval. I noticed that they jam the network when too many are polled at the same time (I have > 10 of them).

I then added a minimum time between polling the devices (10 or 20s not sure anymore) which improved network reliability again. The difference was subtle but definately noticable (mostly remote controll button events were dropped).

Has anyone experienced the same issue?
If this is repreducable it might be considered in the polling cluster.

SuperBigB:
I have about 15 of these plugs. Made a workaround using ZHA Toolkit, and a automation for attribute read that runs every 5 minutes. Have not experienced any issues, but I don't  have much other traffic on my zigbee network, yet.

dumpfheimer:
Do all the plugs poll at the same time / pretty much at the same time?
Are you also on a ZNP coordinator?

I was polling the attributes every minute and I think it was the worsed when 1 or 2 of the plugs were not available / unplugged and it was trying and retrying to reach them.

SuperBigB:
They're all in the same automation, so pretty much polled at the same time. I use the Sonoff zigbee WiFi-bridge with Tasmota FW.
I have not activated any re-tries, as it's not critical.

dumpfheimer:
Okay, thanks.
I will try to replicate the issues as soon as this PR is merged. Maybe it was something else.

mattjes04:
With this patch the TS011F-Plugs are working fine. Problem is with every update the fix is gone and need to be replaced. Is there an approach to make this fix persistent?

SuperBigB:
Try the ZHA toolkit and the following automation:

alias: Update energy
description: ''
trigger:
  - platform: time_pattern
    minutes: /5
condition: []
action:
  - service: zha_toolkit.attr_read
    data:
      ieee: sensor.varmepumpe_smartenergy_metering_summation_delivered
      cluster: 1794
      attribute: 0
mode: single

mattjes04:
> Try the ZHA toolkit and the following automation:
> 
> `alias: Update energy description: '' trigger:
> 
> * platform: time_pattern
>   minutes: /5
>   condition: []
>   action:
> * service: zha_toolkit.attr_read
>   data:
>   ieee: sensor.varmepumpe_smartenergy_metering_summation_delivered
>   cluster: 1794
>   attribute: 0
>   mode: single`

Thank you for your solution! With the downside though, that I have to do this for every plug  ;)

dumpfheimer:
If you are willing to run a separate script that uses a websocket connection (you will need an API key) to hass you can check out https://github.com/dumpfheimer/hass_zha_attr_poll_workaround
It should automatically detect the devices.

SuperBigB:
> Thank you for your solution! With the downside though, that I have to do this for every plug ;)

I have them all listed in the same automation.

veitw:
@3v1n0 Thank you very much for your solution and I hope it will be integrated in zigpy soon.
This works perfectly with my TS011F OEM plug "Blitzwolf BW-SHP15" from Banggood.

I noticed that my plug also does not send updates about its On/Off status on changes, e.g. initial state when replugging or when using the plug's button, but after reading the on_off attribute from the TuyaZBOnOffAttributeCluster cluster, the status is updated correctly. So also pulling the On/Off status could be worth for full compatibility.

tbaumann:
> Try the ZHA toolkit and the following automation:
> 
> [...]

Happy to use this workaround for the meanwhile but looking forward seeing this quirks merged into release soon. Thanks

dmulcahey:
we're going to go with this approach: https://github.com/home-assistant/core/pull/71527

Thanks for the contribution though

dmulcahey:
we're going to go with this approach: https://github.com/home-assistant/core/pull/71527

Thanks for the contribution though

lukasmedia:
> ZHA

How do i add this automation? So far, i only use the gui for automations: /config/automation/edit/new

veitw:
With the polling feature that has been merged into ZHA, an automation should no longer be necessary, as long as the polling quirk is enabled for the device. So if it does not work out of the box for your device, you should open an issue to get your device added.

To use an automation in the meanwhile, you also can add this via GUI, either by setting up the time_pattern trigger and zha_toolkit service call action by GUI, or by switching the GUI editor to yaml mode via three dots menu, the yaml above, "Edit as yaml", and pasting the yaml from above. After pasting the yaml source, you may also switch back to GUI mode via three dots menu and continue editing the automation via GUI.

PR #1227, opened 2021-12-24
Dm2550zb support

claudegel:
Added new dimmer device from Sinopé DM2550ZB with electrical measurement option


zhaquirks/sinope/light.py
```diff
@@ -148,3 +149,67 @@ class SinopeDM2500ZB(SinopeTechnologieslight):
...
+                ],
+            }
+        },
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha_p.PROFILE_ID,
+                DEVICE_TYPE: zha_p.DeviceType.DIMMABLE_LIGHT,
```
dmulcahey, line 193: I assume this quirk is only necessary to make this device match the light platform in ZHA. @Adminiuga the original device type is: `zha_p.DeviceType.DIMMER_SWITCH` Should we add a device type entry to discovery for this to make it go to light by default?

claudegel:
The DM2550ZB is a little different then the DM2500ZB that is already in the quirks. The DM2550ZB have a new input cluster for electrical measurement that the old one DM2500ZB don't have. That's why a added a new class SinopeDM2550ZB

dmulcahey (approved):
Thanks!

PR #1229, opened 2021-12-24
Custom thermostat cluster

claudegel:
Added backlight AutoDim attribute for thermostats in manufacturer_attributes to allow for change backlight state from allways on to auto


dmulcahey (approved):
Thanks!

PR #1239, opened 2021-12-27
extend Xiaomi Aqara temperature sensor range. actually it works below documented values

djp1-hub:
it  works in outdoor and currently shows values below -20. I suppose we must extend sensor range.

### DIFF
--- a/zhaquirks/xiaomi/__init__.py
+++ b/zhaquirks/xiaomi/__init__.py
@@ -428,7 +428,7 @@ class TemperatureMeasurementCluster(CustomCluster, TemperatureMeasurement):
     def _update_attribute(self, attrid, value):
         # drop values above and below documented range for this sensor
         # value is in centi degrees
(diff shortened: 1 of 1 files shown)


Adminiuga (commented):
Maybe cap at -5000?

dmulcahey:
@Adminiuga @puddly I don't have strong feelings either way about this... do either of you?

Adminiuga:
I don't. I was running it as an outdoor sensor, but only a few times it got below -20C and it was working fine, but battery not holding much in the winter time

puddly:
I think this is fine, the only "bad" value I recall seeing is `-10000` (when the battery gets low).

djp1-hub:
Confirm value -10000 when battery is low. Maxell CR2032 batteries operating temperature range:  -40 to+125.

dmulcahey (approved):
Thanks!

PR #1241, opened 2021-12-27
add plug LIDL Silvercrest HG06337 / SAPZ 1 A1 (Tuya TS011F)

jlayec:
ability to change the behavior of the led on the plug and the power on state on the plug LIDL Silvercrest HG06337 / SAPZ 1 A1 (based on Tuya TS011F)

for information jeedom format parameters associated
```
"config" : [
  {"endpoint":11,"cluster":6,"attribute":32769,"name":"Light mode","type":"select","values":[
    {"value":0,"name":"Led disabled"},
    {"value":1,"name":"Led enabled"},
    {"value":2,"name":"Led enabled but inverted"}
  ]},
  {"endpoint":11,"cluster":6,"attribute":32770,"name":"Power on state","type":"select","values":[
    {"value":0,"name":"Off"},
    {"value":1,"name":"On"},
    {"value":2,"name":"Last state"}
  ]}
],
```

### DIFF
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -92,6 +92,64 @@ class Plug(CustomDevice):
     }
 
 
+class Plug_1AC(CustomDevice):
+    """Tuya plug without metering with restore power state support."""
+
+    signature = {
+        MODEL: "TS011F",
+        ENDPOINTS: {
+            # <SimpleDescriptor endpoint=11 profile=260 device_type=266
+            # device_version=1
+            # input_clusters=[0, 3, 4, 5, 6]
+            # output_clusters=[10, 25]>
+            11: {
+                PROFILE_ID: zha.PROFILE_ID,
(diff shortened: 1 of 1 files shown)


MattWestb:
### Looks great !!!!

MattWestb:
I have implanting the "backlight mode" from the clover implanting but i have not finding good names on the 3 modes and i think tuya is using it in different ways in there devices like one dimmered mode for the night (on mains powered switches) and off but i dont knowing if it good putting different names on them in the tuya INIT then the function is not the same on other devices.
If finding more devices with the same function as your perhaps we shall changing name on then.

jlayec:
on my device, the behavior of the attribute backlight_mode = 0x8001 is the following
    Mode_0 = 0x00 --> Led disabled, always off
    Mode_1 = 0x01 --> Led enabled, on when plug is power on
    Mode_2 = 0x02 --> Led enabled but inverted, on when plug is power off

dmulcahey (approved):
Thanks!

Issue #834, opened 2021-04-08
Re-Adding temperature sensor to Xiaomi aqara contact sensor device


Styx85:
I know it is a quite old issue #92 but isn't the temperature sensor in this case quite useful, even if it is unreliable in the absolute reported values.
I think the actual value is quite uninteresting, but the relative change could be used to detect a sun exposure of a window and therefore steer a window cover (cover down if temperature rises a certain absolute or relative value).

Any thoughts on readding the temperature feature?

Or has anyone solved that kind of issue without adding another sensor?

puddly:
I think this can be exposed as a virtual `DeviceTemperature` cluster (0x0002).

PR #1248, opened 2022-01-04
Restore xiaomi dev temp the correct way

rforro:
I think that rather than disabling the temperature sensor inside of various xiaomi sensors, we should leave this to user. So I decided to restore it, but not as environment temperature sensor. I've chosen for this device temperature cluster (what it actually is), so it should be more or less zigbee standard compliant.

Currently I implemented this for only following 2 devices, which I won, Aqara Water Leak and Aqara Window Sensor. I could extend it to all devices from #92 but I'd like to here your feedback first.

P.S. If this will be merged I'll do the necessary changes to `zha/sensor.py` and `zha/core/registries.py`

#834


zhaquirks/xiaomi/__init__.py
```diff
@@ -414,6 +419,16 @@ class MotionCluster(LocalDataCluster, MotionOnEvent):
     reset_s: int = 70
 
 
+class DeviceTemperatureCluster(LocalDataCluster, DeviceTemperature):
+    """Device Temperature Cluster."""
+
+    cluster_id = DeviceTemperature.cluster_id
+
+    def update_attribute(self, attrid, value):
+        """attribute value reported"""
```
MattWestb, line 428: """Attribute value reported."""

rforro: That's not the main problem why is pipeline failing

MattWestb: The py test is failing in other files that you have **not** making changes in and that test its broken or some other have doing bad commitments.

"Unchanged files with check annotations **Beta**"

MattWestb (approved):

rforro:
The pipeline is OK, but don't merge it yet. The thing is, if the device IS publishing it's device temperature, but is missing this DeviceTemperature cluster, by calling `update_attribute` it will produce an error. To fix it I need to know which Aqara devices are publishing its internal temperature or somehow check if DeviceTemperature  was implemented.

MattWestb:
The test have doing some "magic" :-))

For old devices you can looking in the old commits witch devices that have being deleting the temperature measurement cluster but for new i think its easiest looking if Z2M have it implanted in the devices = then is normally is working.

rforro:
I've added this cluster to all devices from #92

zhaquirks/xiaomi/__init__.py
```diff
@@ -414,6 +419,16 @@ class MotionCluster(LocalDataCluster, MotionOnEvent):
     reset_s: int = 70
 
 
+class DeviceTemperatureCluster(LocalDataCluster, DeviceTemperature):
+    """Device Temperature Cluster."""
+
+    cluster_id = DeviceTemperature.cluster_id
+
+    def update_attribute(self, attrid, value):
```
Adminiuga, line 427: This should already be in the latest zigpy, so no need to redefine it here.

rforro: Without `def update_attribute(self, attrid, value):` it wasn't working on my "testbench" HA 2012.12.07

rforro: You were right, sorry

Adminiuga (approved):
LGTM

Adminiuga:
This still requires changes to ZHA to expose this as sensors.

rforro:
> This still requires changes to ZHA to expose this as sensors.

Yes, I've mentioned it, but I wasn't sure if this PR will be accepted.
> P.S. If this will be merged I'll do the necessary changes to `zha/sensor.py` and `zha/core/registries.py`

PR #1252, opened 2022-01-05
Update plug_mmeu01.py

zoic21:
Add maeu01 new model for xiaomi plug


puddly:
@zoic21 a few people are having issues with `lumi.plug.maeu01` after this change. According to their reports, removing the quirk for this device "fixes" the problem to some extent, which I think has to do with `SKIP_CONFIGURATION` and ZHA possibly not setting up attribute reporting for this device.

If you own this specific model, can you double check the functionality of your quirk?

zoic21:
Hello,

I need to check but without this change device does not report power and consumption... Maybe there is two firmware and each firmware work differently

PR #1256, opened 2022-01-05
Add TS130 device

zoic21:


zhaquirks/tuya/ts130f.py
```diff
@@ -293,3 +293,42 @@ class TuyaTS130GP(CustomDevice):
             },
         },
     }
+
+
+class TuyaTS130FTI2(CustomDevice):
+    """Tuya smart curtain roller shutter Time In."""
+
+    signature = {
+        # SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=0x0203, device_version=1, input_clusters=[0, 4, 5, 6, 10, 0x0102], output_clusters=[25]))
```
dmulcahey, line 302: Does this quirk work? The comment and the signature don't appear to match. The comment states 6 clusters and the signature only has 5.

dmulcahey (changes requested):

zhaquirks/tuya/ts130f.py
dmulcahey: @zoic21 I want to cut a release before the beta on Wednesday. Any chance you can address this?

Issue #1249, opened 2022-01-04
[BUG] Livarno Lux E27 G95 Filament Bulb doesn't support color change, but that is available in ZHA


RasmusLauritsen:
**Is your feature request related to a problem? Please describe.**
The Livarno Lux E27 G95 Filament bulb only supports dimming and on/off. In ZHA it shows up with support for color change as well.
It is this bulb: https://zigbee.blakadder.com/Lidl_HG06463B.html

**Describe the solution you'd like**
I would like to remove the color change option, as the bulb doesn't support it.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
```
New device 0x2958 (84:2e:14:ff:fe:9b:2b:1f) joined the network
[0x2958] Scheduling initialization
Received frame on uninitialized device <Device model=None manuf=None nwk=0x2958 ieee=84:2e:14:ff:fe:9b:2b:1f is_initialized=False> from ep 0 to ep 0, cluster 19: b'\x83X)\x1f+\x9b\xfe\xff\x14.\x84\x8e'
[0x2958:zdo] ZDO request ZDOCmd.Device_annce: [0x2958, 84:2e:14:ff:fe:9b:2b:1f, 142]
Tries remaining: 3
[0x2958] Requesting 'Node Descriptor'
Tries remaining: 2
[0x2958] Extending timeout for 0xa8 request
Received frame on uninitialized device <Device model=None manuf=None nwk=0x2958 ieee=84:2e:14:ff:fe:9b:2b:1f is_initialized=False> from ep 0 to ep 0, cluster 32770: b'\xa8\x00X)\x01@\x8e\x02\x10RR\x00\x00,R\x00\x00'
[...]

PR #1264, opened 2022-01-07
adding devices

gromgsxr:
just correcting the lidl bulb reporting colour when its only dimmable see issue #1249 and adding a tuya single channel led controller as dimmable and not colour capable


MattWestb:
Use https://black.vercel.app/ for fixing the formation and most test errors is going away.
(put the code in the left window and you is getting the fixed in the right one and use copy and pace)

PR #1266, opened 2022-01-08
Add new quirk for sinope SP2600ZB/SP2610ZB devices 

EtienneMD:
These plugs would report energy usage in Wh instead of kWh in ZHA.

### DIFF
--- a/zhaquirks/sinope/switch.py
+++ b/zhaquirks/sinope/switch.py
@@ -18,6 +18,7 @@ from zhaquirks.sinope import SINOPE
 
 SINOPE_MANUFACTURER_CLUSTER_ID = 0xFF01
 
+
 class CustomMeteringCluster(CustomCluster, Metering):
     """Custom Metering Cluster."""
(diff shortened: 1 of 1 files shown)


EtienneMD:
This quirks is to add support for Sinope SP2600ZB and SP2610ZB devices where energy usage is wrongly reported. Device send Wh instead of kWh for energy measurement reporting.

Adminiuga:
It would affect all attribute values. Maybe better just correct attribute values for divisor attributes? Check Tuya ts0121 osmart plug quirk for example.

dmulcahey:
@EtienneMD see here for an example: https://github.com/zigpy/zha-device-handlers/blob/6d50032e3cfb16a55a5a0d748fc84bad0c7e2029/zhaquirks/tuya/__init__.py#L866

EtienneMD:
Thank you both for steering me in the right direction. Appears to be working fine so far on my devices.

PR #1269, opened 2022-01-09
Manufacturer cluster specific attributes for light and dimmer

claudegel:
Added many manufacturer attributes for light and dimmer in manufacturer cluster to allow for device parameters configuration.


claudegel:
This have been merged  but the manufacturer_attributes that was added do not appear for light or dimmer in ZHA manage cluster of home assistant. I've also added many  manufacturer_attributes for thermostats and they all show up for thermostats. Is there a bug somehere that I miss ?

MattWestb:
Perhaps adding the new cluster to the devices in the replacement section ?
https://github.com/zigpy/zha-device-handlers/blob/23a166902e50d5c32019e0d4d32a968715041bb6/zhaquirks/sinope/light.py#L37

claudegel:
I did the same as for the sinope thermostats and the custom attributes for manufacturer cluster are available in ZHA. Is there a debug I can load to find out why when the sinope.light is loaded the custom attributes are not added ?
Could it be because those attributes are defined in the same class SinopeTechnologiesManufacturerCluster and when the sinope.thermostat is loaded in ZHA after the sinope.light it overwrite that class ?

claudegel:
Stupid me. yes adding in the replacement section did it. Thank you

PR #1271, opened 2022-01-09
Add more support for aqara magic cube

tribela:
- Add "from face" attribute when flip (eg: flip from face 2 to face 3)
- Add wake up event (when pick up the device while it was standing still)

### DIFF
--- a/zhaquirks/xiaomi/aqara/cube.py
+++ b/zhaquirks/xiaomi/aqara/cube.py
@@ -37,10 +37,14 @@ from zhaquirks.xiaomi import (
 )
 
 ACTIVATED_FACE = "activated_face"
+FROM_FACE = "from_face"
 DESCRIPTION = "description"
 DROP = "drop"
 DROP_VALUE = 3
 DROPPED = "device_dropped"
+WAKE = "wake"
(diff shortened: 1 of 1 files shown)


zhaquirks/xiaomi/aqara/cube.py
```diff
@@ -37,10 +37,14 @@
...
 
 ACTIVATED_FACE = "activated_face"
+FROM_FACE = "from_face"
 DESCRIPTION = "description"
 DROP = "drop"
 DROP_VALUE = 3
 DROPPED = "device_dropped"
+WAKE = "wake"
+WAKE_VALUE = 2
+WAKED = "device_waked"
```
dmulcahey, line 47: Can we make this `device_awake`

zhaquirks/xiaomi/aqara/cube.py
```diff
@@ -37,10 +37,14 @@
 )
 
 ACTIVATED_FACE = "activated_face"
+FROM_FACE = "from_face"
```
dmulcahey, line 40: can we make this `deactivated_face`

dmulcahey (changes requested):

PR #1276, opened 2022-01-10
Adding Group in cluster on TS004F

MattWestb:
If binding the remote to one group i not working.
Was testing adding the group cluster and ZHA is showing it in group and is possible being added in groups.
Then added its sending light commands (OnOff Dimm up and down) to the group without problems !!
OnOff cluster is deleted for not getting one light in the device card and looks being OK and shall not being added.

One PS for user if pressing right upper and lower bottom long its changing mode to scene and one more time to commands as the mode attribute on OnOff cluster but in scene mode its only using EP1 do not so useful until we have getting the INIT working of the 3 new endpoints in ZHA.


dmulcahey:
Is this good to go?

MattWestb:
Its working on my TS004F in light mode i can "binding" / adding it to one light group and its controlling the lights. 👍

PR #1293, opened 2022-01-17
Add Quirk for Linkind Leak Sensor

tube0013:
Allows setting siren/led behavior

For reference:

```
Endpoint                  1
Cluster                   Basic (0x0000)
Attribute                 0x400a
ACK request               TRUE
Manuf. specific           TRUE
Disable default response  TRUE
Data type                 8-bit unsigned int
Data:    
  Siren disabled, LED disabled     0 (0x00)
  Siren disabled, LED enabled      1 (0x01)
  Siren enabled, LED disabled      2 (0x02)
  Siren enabled, LED enabled       3 (0x03)
```

### DIFF
--- a/zhaquirks/linkind/__init__.py
+++ b/zhaquirks/linkind/__init__.py
@@ -1,0 +1,10 @@
+"""Linkind sensors."""
+from zigpy.quirks import CustomCluster
+import zigpy.types as t
+from zigpy.zcl.clusters.general import Basic
+
+
+class LinkindBasicCluster(CustomCluster, Basic):
+    """Linkind Basic cluster."""
+
+    manufacturer_attributes = {0x0400A: ("linkind", t.uint8_t)}
--- a/zhaquirks/linkind/__init__.py
+++ b/zhaquirks/linkind/__init__.py
@@ -4,7 +4,17 @@ import zigpy.types as t
 from zigpy.zcl.clusters.general import Basic
 
 
+class LinkindLeakMode(t.enum8):
+    """Linkind Leak Sensor mode enum."""
+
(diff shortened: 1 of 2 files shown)


MattWestb:
You can making the attributes with names so you is getting the function name then reading it like we have doing with tuy devices.
https://github.com/zigpy/zha-device-handlers/blob/dcee551996a0a9e388b6afb1e054dada46897ee5/zhaquirks/tuya/__init__.py#L827-L857
The problem is i dont knowing how to "tagging" it with manufacture code but we need learning doing that later for getting ZHA handing all special attributes we have implanting.

dmulcahey:
> 

as long as it is defined as a manufacturer attribute the code is sent automatically

javicalle:
> > 
> as long as it is defined as a manufacturer attribute the code is sent automatically

The 'manufacturer attribute' condition it is set by the definition of attribute or by the attribute ID?

I mean, manufacturer attribute can be set as:
```python
class OccupancyCluster(CustomCluster, OccupancySensing):
    """Philips occupancy cluster."""

    manufacturer_attributes = {
        0x0030: ("sensitivity", t.uint8_t),
        0x0031: ("sensitivity_max", t.uint8_t),
    }
```

but also have read something, somewhere about attributes with ID >= 0xFF00 or 0xFE00 (not sure about this point).

warre01:
Hi Guys can this be used Smarthings IDE also issues here with initial setup joining in App.....App later showed "Thing" 2 separate lines 2 separate/different zigbee ID & Network ID Connected...IDE Shows Thing "Connected" Edit shows Type- 2015 Samsung Smart TV- Published....I do have that TV- Thanks!!

Issue #1299, opened 2022-01-21
[Device Support Request] TS0121 _TZ3000_8nkb7mof still gives incorrect readings


ali1234:
This device pairs and all functions work, but it gives incorrect current and power summation readings. The current is wrong by a factor of 1000 and the power is wrong by a factor of 100. Recent patches [1](https://github.com/zigpy/zha-device-handlers/pull/1131), [2](https://github.com/zigpy/zha-device-handlers/pull/1062) claim to resolve this issue for TS0121, but are not applied to this device for some reason - presumably because its signature is slightly different from what the quirk expects (see below).

With zhaquirks-0.0.65, after deleting and re-pairing the device:

![image](https://user-images.githubusercontent.com/651970/150607995-b8021f48-2040-4a67-a6d6-76edbf1a3768.png)

Signature:

```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x010a",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0702",
        "0x0b04",
        "0xe000",
        "0xe001"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZ3000_8nkb7mof",
  "model": "TS0121",
  "class": "zigpy.device.Device"
}
```
[...]

PR #1300, opened 2022-01-22
Add new model of TS0121 Tuya plug

Adminiuga:
Add a new model of TS0121 Tuya plug. 
Fixes https://github.com/zigpy/zha-device-handlers/issues/1299 
I don't have such a plug so this needs testing to confirm it is working correctly.

### DIFF
--- a/zhaquirks/tuya/ts0121_plug.py
+++ b/zhaquirks/tuya/ts0121_plug.py
@@ -1,7 +1,7 @@
 """Tuya TS0121 plug."""
 from zigpy.profiles import zha
 from zigpy.quirks import CustomDevice
-from zigpy.zcl.clusters.general import Basic, Groups, OnOff, Ota, Scenes, Time
+from zigpy.zcl.clusters.general import Basic, Groups, Identify, OnOff, Ota, Scenes, Time
 from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
 from zigpy.zcl.clusters.smartenergy import Metering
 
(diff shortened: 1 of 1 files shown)


ali1234:
This certainly causes the quirk to be applied to my devices. The custom clusters are now shown in the cluster manager. The cluster manager works now and I can get the value of current_divisor, and it is 1000 as expected when the quirk is activated.

Tomorrow I will verify that the readings really are correct using my 2kW fan heater.

zhaquirks/tuya/ts0121_plug.py
```diff
@@ -62,3 +64,65 @@ class Plug(CustomDevice):
...
+
+class TS0121B(CustomDevice):
+    """Tuya TS0121 plug with restore tuya power state support and ZGP endpoint."""
+
+    signature = {
+        MODEL: "TS0121",
+        ENDPOINTS: {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=81
+            # device_version=1
+            # input_clusters=[0, 4, 5, 6, 9, 1794, 2820]
```
javicalle, line 77: The device signature is wrong.
Maybe copy&paste?

javicalle (changes requested):

zhaquirks/tuya/ts0121_plug.py
ali1234: `input_clusters` comment is wrong for both devices too. Neither defines 0x0009 "Alarms" and the new one is obviously completely different...

ali1234:
After running heater for 6 minutes I have this:

image

These readings seem to be correct.

zhaquirks/tuya/ts0121_plug.py
```diff
@@ -62,3 +64,65 @@ class Plug(CustomDevice):
...
+            # output_clusters=[10, 25]>
```
Adminiuga, line 78:
```suggestion
```

zhaquirks/tuya/ts0121_plug.py
Adminiuga: fixed

javicalle:
LGTM

PR #1305, opened 2022-01-24
Fix Xiaomi devices

dmulcahey:


puddly:
Related PR: https://github.com/zigpy/zha-device-handlers/pull/1248

Issue #1236, opened 2021-12-27
[Device Support Request] Integrating a motion, temperature & humidity sensor in ZHA model NAS-PD07 manufactured by NEO


sassiahmed:
Hi all,

I want to integrate a motion sensor manufactured by NEO in ZHA.
Here is the description : https://zigbee.blakadder.com/Neo_NAS-PD07.html

When I try to integrate it with ZHA, there is no entity found.

I tried the quirks (Tuya/ts0601_motion.py : https://github.com/zigpy/zha-device-handlers/blob/5228c54d70d5eee0be09ff6ca2eeb81a538eefa1/zhaquirks/tuya/ts0601_motion.py ==> but nothing.

I modified the ts0601_motion.py file adding a new class NeoMotion with the right signature and import the clusters to use (inspired by ts0601_trv.py) found in the same repo.
I have the motion sensor entity but didn't work (always clear)

Here is the zigbee signature :
```

{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_7hfcudw5",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}

```

and here is my ts0601_motion.py modified :

```

"""BlitzWolf IS-3/Tuya motion rechargeable occupancy sensor."""

from typing import Optional, Tuple, Union

from zigpy.profiles import zha
[...]

PR #1315, opened 2022-01-30
Add some ts0601_motion devices (with Tuya MCU Cluster approach)

javicalle:
This PR will add support for some `ts0601_motion` devices.

The new devices will follow the new 'Tuya MCU Cluster' approach to update device attributes.

Fixes: #1236
Reported motion sensor 'NAS-PD07' manufactured by NEO.
Most of the code came from 'herdsman-converters'
Those DPs that have been possible are reported as a new cluster.
Those that I have not found a match for are reported as attributes of the MCU cluster.

Not sure about possible `tamper` values. Probably there are diferent values >0 with other meanings.

Fixes: #1312
Unknown sensor, from unknown manufacturer.
The functionality of most DPs is unknown. They are probably used to configure device settings.
The known DPs (`occupancy`, `illuminance`, `target_distance`) have been implemented according to the information provided by the manufacturer.

### DIFF
--- a/zhaquirks/tuya/ts0601_motion.py
+++ b/zhaquirks/tuya/ts0601_motion.py
@@ -89,6 +89,7 @@ class NeoMotionManufCluster(TuyaNewManufCluster):
     manufacturer_attributes = {
         0xEF01: ("battery_level", NeoBatteryLevel),  # ramdom attribute ID
         0x0004: ("tamper", t.uint8_t),  # same ID as IasZone.ZoneStatus.Tamper
+        0xEF0D: ("dp_113", t.enum8),  # ramdom attribute ID
     }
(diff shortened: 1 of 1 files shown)


MattWestb:
The tamper attribute looks very OK.
The battery is little tricky then its also having USB powered "mode".
If it was not having it some TRVs is converting battery low warning to 20% battery and no warming 100% (*2 for ZCL compatibility) so i think you have doing the best what can being done.

One way is having the "tuya battery attribute" and also  doing the % converting (then USB = 100% *2) to the normal battery attribute for getting the normal ZHA battery sensor showing "somthing" and the user can getting the real sate from tuya battery attribute if they like that.

PR #1326, opened 2022-02-03
Add support for new Stelpro SORB fan heater.

EtienneMD:
This heater comes with an integrated Stelpro Maestro thermostat (SMT402AD) but since it lacks a humidity sensor, it continuously reports 0% relative humidity in ZHA.

This quirk removes the RelativeHumidity cluster from ZHA.

### DIFF
--- a/zhaquirks/stelpro/__init__.py
+++ b/zhaquirks/stelpro/__init__.py
@@ -1,0 +1,2 @@
+"""Module for Stelpro quirks."""
+STELPRO = "Stelpro"
--- a/zhaquirks/stelpro/thermostat.py
+++ b/zhaquirks/stelpro/thermostat.py
@@ -1,0 +1,67 @@
+"""Module to handle Stelpro thermostats and heaters."""
+
+import zigpy.profiles.zha as zha_p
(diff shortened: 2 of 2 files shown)


dmulcahey:
do we know if this is reported in a non standard way instead of on the correct cluster? Have you checked z2m docs to see what they support? Just curious.

EtienneMD:
I could not find anything on this particular fan in Z2M. Documentation for the device makes no mention of any humidity reading capability.
 All I found online was this [Smarthings handler](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/cadf20d8b4e94a36ad85895f0db2110baf6454fe/devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy#L456) that disables the humidity reading from both the Sonoma and Orleans(SORB) fans.
I have not looked at DECONZ support.

EtienneMD:
If you have any ideas as to what I should look for, I'd be happy to give it a shot. 

As this is meant for bathrooms, humidity reading could be quite useful for automations. I will try scouring the Zigbee.db for something that could correlate to humidity.

PR #1192, opened 2021-11-29
Maxsmart, Beca, Siterwell and Moes TRV support/update

jacekk015:
According to matrix
https://github.com/zigpy/zigpy/discussions/653
new TRVs has support for ZHA and some got update of functionality.

Siterwell tested locally by @MattWestb
https://github.com/zigpy/zha-device-handlers/discussions/1084#discussioncomment-1701973

Beca tested locally by @iedex and @Youpimatin
https://github.com/zigpy/zha-device-handlers/issues/1123#issuecomment-980314632

Maxsmart written and tested locally by me - I own 12pcs of them
Maxsmart -> Silvercrest Tuya Lidl clone tested by @JoostvdB94 @lukasf98 @baldisos
https://github.com/zigpy/zha-device-handlers/issues/1061#issuecomment-981156618

Moes TRV extended using same approach as above Tuyas, split to separate file
Zonnsmart not touched code, spitted to separate file

Pull request made by @MattWestb should be Closed, newer code version is here
https://github.com/zigpy/zha-device-handlers/pull/1129

PR #1339, opened 2022-02-09
Add Quirk for Linkind DoorSensor D0003

arnemauer:
Adds tampering sensor as a separate on/off sensor (because HA acts on ALARM_1 and ALARM_2 but ignores other bits like battery and tamper in the IAS_zone byte).
Schermafbeelding 2022-02-09 164702


arnemauer:
Can somebody tell me why the isort check is failing?

MattWestb:
Its 4 lines with minus and 4 with plus. The without is for reference so you is knowing where the code is located (no line numbers).
```
 from typing import Any, List, Optional, Union
 
 import zigpy.profiles.zha
-from zigpy.zcl import foundation
-
-from zigpy.quirks import CustomDevice, CustomCluster
-from zhaquirks import Bus
-from zigpy.zcl.clusters.homeautomation import Diagnostic
+from zigpy.quirks import CustomCluster, CustomDevice
 import zigpy.types as t
-
 from zigpy.zcl.clusters.general import (
     Basic,
     Identify,
@@ -20,8 +16,10 @@ from zigpy.zcl.clusters.general import (
     PollControl,
     PowerConfiguration,
 )
+from zigpy.zcl.clusters.homeautomation import Diagnostic
 from zigpy.zcl.clusters.security import IasZone
 
+from zhaquirks import Bus
 from zhaquirks.const import (
     DEVICE_TYPE,
     ENDPOINTS,
```
Plus two empty line that is wrong.

Delete line 8 - 12:
```

from zigpy.quirks import CustomDevice, CustomCluster
from zhaquirks import Bus
from zigpy.zcl.clusters.homeautomation import Diagnostic
```
plus empty line 14
And putting them in there its pus before them.

MattWestb:
## All in !!!

MattWestb:
The largest problem is that HA is using one costume setting so online cant fixing it but in ground its sorting all imports but its not liking 110% normal alphabet sorting.

My record is over 10 treys for getting all OK ;-(

arnemauer:
@MattWestb 
1. I also want to add the other properties of the IASZone byte like lowbattery, trouble, batteryDefect. Is it wise to make multiple endpoints with an onoff cluster? Or is there an other way to get multiple onoff cluster within one endpoint?
2. Is it possible to give a cluster a sensor name, so homeassistant automaticly gives a logic name to a cluster?

MattWestb:
Have you looking on PowerConfiguration cluster that many battery powered devices is having.
Its have different status reports and also alarms and is easy implanting.
https://github.com/zigpy/zigpy/blob/e91ac25d968403fd8e26c828e2177cfa2797973a/zigpy/zcl/clusters/general.py#L232 but i think you is not getting any GUI elements from then (= i dont knowing but i think its not implanted in ZHA).

Also the Zigbee standard is having the attribute but i think ZHA have not implanting then in the GUI.
https://github.com/zigpy/zigpy/blob/e91ac25d968403fd8e26c828e2177cfa2797973a/zigpy/zcl/clusters/security.py#L41-L53

tuya TRVs have adding multiple EPs for making it possible have more on/off cluster but its not so easy to do but its working.

Parsing names to ZHA we have trying adding "discretion attributes" but was getting no go from the maintainers so its not possible to do that for the moment.
Pending PR that cant being merged then we cant writing the tests :-(((
https://github.com/zigpy/zha-device-handlers/pull/1192

arnemauer:
@[MattWestb](https://github.com/MattWestb) Can we merge this pull request?

MattWestb:
For my its OK if its working with your device but i think the maintainers dont like that then the `CI / Run tests Python 3.x` test is not being OK and i cant helping with that then its too deep for my knowledge ;-(

MattWestb:
@javicalle can you taking one look then you have time if the quirk is looking OK ?

Thanks in advance !!

javicalle:
I have reviewed the code and I have doubts that it can actually work.

```python
        """Handle a cluster command received on this cluster."""
        if hdr.command_id == 0:
            state = bool(
                args[0] & 4
            )  # check if fourth bit (tamper alarm) is set, if so  set bit of Alarm 1 high. (HA only checks Alarm_1 and Alarm_2 and ignores tamper/battery...
            self.endpoint.device.ias_bus.listener_event("update_state", state)
```
This part handles the cluster commands, but only takes into account the fourth bit. If I'm not mistaken, the rest of the sensor events are not being taken into account.

There are some problems with the adaptation of the code to the latest versions. That part can be fixed, but I think the approach needs an overhaul.

Regards.

javicalle:
BTW, related to https://github.com/home-assistant/core/issues/72828

PR #1340, opened 2022-02-09
Support for Tuya TS0601-family gas leakage sensor

sergey-suloev:


zhaquirks/tuya/air/ts0601_air_quality.py
```diff
@@ -103,7 +102,7 @@ class TuyaCO2SensorGPP(CustomDevice):
                 PROFILE_ID: 41440,
                 DEVICE_TYPE: 97,
                 INPUT_CLUSTERS: [],
-                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
+                OUTPUT_CLUSTERS: [33],
```
javicalle, line 105: Why?
I think is better keep the `GreenPowerProxy` reference

javicalle (changes requested):

zhaquirks/tuya/air/ts0601_air_quality.py
```diff
@@ -47,7 +46,7 @@ class TuyaCO2Sensor(CustomDevice):
                 ],
                 OUTPUT_CLUSTERS: [Time.cluster_id, Ota.cluster_id],
             }
-        },
+        }
```
javicalle, line 49: Don't change.
Black complains about it:
* https://github.com/zigpy/zha-device-handlers/runs/5167595067?check_suite_focus=true

zhaquirks/tuya/air/ts0601_air_quality.py
```diff
@@ -32,8 +32,7 @@ class TuyaCO2Sensor(CustomDevice):
         # input_clusters=[0, 4, 5, 61184],
         # output_clusters=[25, 10])
         MODELS_INFO: [
-            ("_TZE200_8ygsuhe1", "TS0601"),
-            ("_TZE200_yvx5lh6k", "TS0601"),
+            ("_TZE200_8ygsuhe1", "TS0601")
         ],
```
javicalle, line 36: Black also complains here, replace with:
```yaml
        MODELS_INFO: [("_TZE200_8ygsuhe1", "TS0601")],
```

javicalle (changes requested):

zhaquirks/tuya/air/ts0601_air_quality.py
dmulcahey: +1

zhaquirks/tuya/air/ts0601_air_quality.py
javicalle: This change has nothing to do with "gas leakage sensor".

Please manage it in a separate PR so we can review it properly.

javicalle (changes requested):

PR #1345, opened 2022-02-12
Add support for Tuya Gas leakage sensor

sergey-suloev:
The Gas leakage sensor works well but there is still an issue while adding a new automation in HomeAssistant: the New Automation dialog show 'smoke' events instead of gas. This needs to be fixed additionally.


zhaquirks/tuya/air/ts0601_air_quality.py
```diff
@@ -33,7 +33,6 @@ class TuyaCO2Sensor(CustomDevice):
         # output_clusters=[25, 10])
         MODELS_INFO: [
             ("_TZE200_8ygsuhe1", "TS0601"),
-            ("_TZE200_yvx5lh6k", "TS0601"),
```
javicalle, line 36: Looks like change has sneaked in around here again...

javicalle (changes requested):

Issue #1343, opened 2022-02-12
[BUG] Tuya Air Quality detector exposes only VOC level sensor


sergey-suloev:
Hi, the Tuya air quality detector located here  zhaquirks/tuya/air/ts0601_air_quality.py  doesn't expose all supported sensors.
See attached screenshot.

![Screenshot from 2022-02-12 11-41-43](https://user-images.githubusercontent.com/6279627/153704215-5fb6d17a-0b57-4e8a-bc7b-4d32d7157b6c.png)

sergey-suloev:
```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0x0402",
        "0x0405",
        "0x040d",
        "0x042b",
        "0x042e",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZE200_yvx5lh6k",
  "model": "TS0601",
  "class": "zhaquirks.tuya.air.ts0601_air_quality.TuyaCO2SensorGPP"
}
```

sergey-suloev:
It seems like the problem was discussed here https://github.com/zigpy/zha-device-handlers/issues/1004#issuecomment-1002206182

javicalle:
[...]

PR #1346, opened 2022-02-12
Tuya Air quality: ambiguous models info

sergey-suloev:
Resolves an ambiguity within Tuya Air Quality quirks when more > 1 class support the device TS0601/_TZE200_yvx5lh6k

### DIFF
--- a/zhaquirks/tuya/air/ts0601_air_quality.py
+++ b/zhaquirks/tuya/air/ts0601_air_quality.py
@@ -33,7 +33,6 @@ class TuyaCO2Sensor(CustomDevice):
         # output_clusters=[25, 10])
         MODELS_INFO: [
             ("_TZE200_8ygsuhe1", "TS0601"),
-            ("_TZE200_yvx5lh6k", "TS0601"),
         ],
         ENDPOINTS: {
             1: {


javicalle:
For future references:

From the code I believe that only devices with the same clusters must match:
https://github.com/zigpy/zigpy/blob/e91ac25d968403fd8e26c828e2177cfa2797973a/zigpy/quirks/registry.py#L87-L97

but MattWestb's tests indicate that this is not the case:
* https://github.com/zigpy/zha-device-handlers/issues/1343#issuecomment-1037243099

javicalle (approved):
LGTM

MattWestb:
Its looks like the same MCU is being used with 2 different versions of the Zigbee modules (or more correct the same Zigbee module but with different EP / cluster settings in the firmware) but that shall not being any problem then the system is only loading the version with the right EP/cluster settings is its doing it right.

MattWestb:
For reference we have testing the TS004F with 2 device IDs and matching MODELS_INFO and EP/cluster its loading the right quirk.

If having one device class with the same EP / cluster settings and MODEL later (under) in the code its being used if the MODELS_INFO is **not** matching = OK.
That is needed for the TS004F then adding more functions to devices and not braking "old" implemented function.

Working quirk with "multi class type select" that working OK for My and one brave user is here https://github.com/zigpy/zha-device-handlers/issues/1322 but is still very WIP.

sergey-suloev:
> 

We can't use the same ID in two implementation classes. It was removed from the old class in favor of new devices supported by TuyaCO2SensorGPP class.

MattWestb:
Om 95% sure its being wrong.

The 2 classes is having different EP / cluster config so its only being loaded for the class that is having the right EP/cluster setting.

If the device is having GPP its loading the GPP class and if the device is not having GPP its loading the class without GPP.

If your device with GPP have more functions then implanting it in the GPP class but leave the class without GPP for not braking thos devices.


Can you pleas showing that the GPP class is not being loaded if its one class with the same device ID and different EP/cluster setting and class name is before it in the quirk.

Before your changes we was having users with devices with the same MCU ("_TZE200_yvx5lh6k", "TS0601") working OK with Zigbee modules with GPP and without GPP endpoint.

sergey-suloev:
@MattWestb 
I have many devices with GPP.  They are not working without this change.

MattWestb:
If ZHA is not liking having 2 classes with different EP/cluster and the same signature / device ID then i think you need doing one new quirk for your device so the system is trying that also and not the "first one" that  is not working for your devices.

Example: 
`quirk_1` with one device class with out GPP EP and device id 1234.
`quirk_2` with one device class with GPP EP and device ID 1234.

MattWestb:
One more strange thing is that 2 user have reporting that the https://github.com/zigpy/zha-device-handlers/pull/1280 is working for there devices with GPP in January and have not complaining its not working on HA 2022-02 then it was added in HA.
Perhaps they is still using local quirk and not the one in HA.

Issue #1303, opened 2022-01-23
[BUG] LIDL TS0505A/_TZ3000_dbou1ap4 - "erratic"


mdeweerd:
ℹ️ The fix (custom quirk) is [further below](https://github.com/zigpy/zha-device-handlers/issues/1303#issuecomment-1020371189) and [instructions a bit further](https://github.com/zigpy/zha-device-handlers/issues/1303#issuecomment-1039714323)

**Description of the behavior**
~"Unstable"~ Erratic ~UI~ behavior of LIDL  TS0505A/_TZ3000_dbou1ap4 .

Edit: given the current state of analysis the On/Off attribute of the On/Off cluster is incorrectly representing the bulb's state.

**To Reproduce**
Pictures say more than 1000 words:

The UI switches the light to off when the intensity is changed (there are some other operations in this gif animation, but watch closely when the intensity changes);
![LigthButtonResetWithoutAdaptive2](https://user-images.githubusercontent.com/1504752/150676306-decb1357-3fdf-4977-bd41-430595c79f7c.gif)

The light can be switched on and off normally.
![LigthButtonResetWithoutAdaptive](https://user-images.githubusercontent.com/1504752/150676302-75b25eac-cc63-4656-ac2c-9c2cffd80d86.gif)

I tried another light (TI Sample Light) and that's behaving normally.

**Expected behavior**
The intensity change should not result in the device switching off

**Additional context**

I probably upgraded HAOS to 2021.12.3 a few days before reporting this in the Adaptive Lighting component.

I filed an [issue for the Adaptive Lighting component](https://github.com/basnijholt/adaptive-lighting/issues/245).  The behavior is worse when that is enabled, but I think the core issue is not really there..

This might be related to the appearance of https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/lidl/rgbcct.py with #1177 .

These are the attributes available with their values (when they can be read) - [TS0505A__TZ3000_dbou1ap4_ff23a460_scan_results.txt](https://github.com/zigpy/zha-device-handlers/files/7921232/TS0505A__TZ3000_dbou1ap4_ff23a460_scan_results.txt)

The question now is:

- What kind of adjustment could be made to the quirk to cope with this;
- Are there FW updates for the bulb?

mdeweerd:
I added and changed the quirk locally, removing the reference to the LicldRGBCCTColorCluster:
```python
[...]

PR #1360, opened 2022-02-15
#1303 Ignore OnOff attribute change to Off for LIDL/rgbcct

mdeweerd:
At least three users (including me) experience issue with the on-off attribute and ignoring incoming changes fixes that.
See #1303 .

### DIFF
--- a/zhaquirks/lidl/rgbcct.py
+++ b/zhaquirks/lidl/rgbcct.py
@@ -1,6 +1,9 @@
 """Quirk for LIDL RGB+CCT bulb."""
+import logging
+
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
+from zigpy.types.named import Bool
 from zigpy.zcl.clusters.general import (
     Basic,
     GreenPowerProxy,
@@ -24,12 +27,15 @@ from zhaquirks.const import (
     PROFILE_ID,
 )
 
+_LOGGER = logging.getLogger(__name__)
+
 
 class LidlOnOffCluster(CustomCluster, OnOff):
     """Lidl OnOff custom cluster."""
 
     def _update_attribute(self, attrid, value):
-        if attrid != 0:
+        _LOGGER.debug(f"LIDL ATTR {attrid} {value!r} {value!=Bool.false}")
+        if attrid != 0 or value != Bool.false:
             super()._update_attribute(attrid, value)
         # else ignore update
 
--- a/zhaquirks/lidl/rgbcct.py
+++ b/zhaquirks/lidl/rgbcct.py
@@ -25,6 +25,15 @@ from zhaquirks.const import (
 )
 
 
+class LidlOnOffCluster(CustomCluster, OnOff):
+    """Lidl OnOff custom cluster."""
+
+    def _update_attribute(self, attrid, value):
+        if attrid != 0:
+            super()._update_attribute(attrid, value)
+        # else ignore update
+
+
 class LidlRGBCCTColorCluster(CustomCluster, Color):
     """Lidl RGB+CCT Lighting custom cluster."""
 
@@ -81,7 +90,7 @@ class RGBCCTLight(CustomDevice):
                     Identify.cluster_id,
                     Groups.cluster_id,
                     Scenes.cluster_id,
-                    OnOff.cluster_id,
+                    LidlOnOffCluster,
                     LevelControl.cluster_id,
                     LidlRGBCCTColorCluster,
                     LightLink.cluster_id,


mdeweerd:
@dmulcahey It probably prevents the state being updated - with inconsistent data .

In https://github.com/zigpy/zha-device-handlers/issues/1303#issuecomment-1020554042 I share a series of commands and states sent back by the light.

The only "improvement" that could be made is to set the light On when the On state is received - On  seems to be On in the trace.   I am not sure anymore, but I think that was not true all the time either. 

One example in the trace:
```
    Command: On (0x01)
        On/off Control: Off (0x00)
        On/off Control: On (0x01)
    Command: Move to Level with OnOff (0x04)
        On/off Control: Off (0x00)
```

The ligit is "on' all the time but it sends Off two times.

The light is controlled by commands, which is the same thing that a remote would send.

I agree, the UI would not know about what the remote is doing.  But using the on/of state provided by the light is not making it any wiser.

I could not determine a rule to determine wheter the light was on or of.  Using the state returned by the light was making the HA UI impractical.

In this case, the only way to have the UI know about it is to define the automation in HA, and not use direct control.

MattWestb:
If getting off state reported from the device after sending commands to it is it also reporting off **if** reading the on/off attribute (more or less pulling that is being done with Philips HUE lights) ?

mdeweerd:
@dmulcahey I did check Read Attributes

Here is a ZIP file with a filtered sniffer log (relevant clusters only, tested light only, no acks):
[LIDL.zip](https://github.com/zigpy/zha-device-handlers/files/8081668/LIDL.zip).

mdeweerd:
I made an update so that only the off state is ignored.

I added a debug statement so that this is easily trackable.

when turning the light on, I get:
```
2022-02-17 18:41:43 DEBUG (MainThread) [rgbcct] LIDL ATTR 0 <Bool.true: 1> True
2022-02-17 18:41:43 DEBUG (MainThread) [rgbcct] LIDL ATTR 0 <Bool.false: 0> False
```

So the last notification is "Off" when turning the ligth on.

mdeweerd:
It looks line unbinding the coordinator keeps the OnOff reporting working when changing OnOff, not when turing the light on with an intensity change.
https://github.com/zigpy/zha-device-handlers/issues/1303#issuecomment-1046216657

MattWestb:
I was dehacking my LIDL ZBGW and updating it and paired one LIDL RGBW E14 light and using my TS004F as dimmer remote bounded to the light group and its works OK.
But LIDL APPis not updating the status then using the remote but then sending commands (both OnOff and dimmer) and from the APP its updating OK. 
One LIDL indoor plug is reporting OK in the APP then toggling it on the button so i think LIDL / tuya is knowing the problem.
The light is on firmware 1.1.0 if looking in the APP and is not being updated.

mdeweerd:
@MattWestb Thanks for checking.

So the "Tuya"* app isn't updating either.  A question is: is it required for zigpy?  There is a way to do it: remove all bindings with the coordinator (and disable requests for binding with the coordinator) and then the on/off reports are ok.

So ideally, the quirk would do that:
- On (initial) setup: remove all bindings towards the coordinator;
- Catch the binding requests and ignore any request to bind to the coordinator (so that HA does not break it).

And then, no hack is needed with regards to the on/off attribute.

MattWestb:
Xiaomi / Aqara is using one parameter for not doing all config in setup but i dont knowing how its working:
https://github.com/zigpy/zha-device-handlers/blob/e4e62504738ced0c54c983ee5094a3a6af0a741d/zhaquirks/xiaomi/aqara/magnet_aq2.py#L65-L66
Perhaps its working on tuya devices 2 and i think Z2M is doing minimal config of  tuya devices.

mdeweerd:
Skipping the configuration does not unconfigure it for devices that are already configured, but it's a start.

To know what happens, follow the code...

https://github.com/zigpy/zha-device-handlers/blob/e4e62504738ced0c54c983ee5094a3a6af0a741d/zhaquirks/const.py#L92

https://github.com/zigpy/zha-device-handlers/blob/e4e62504738ced0c54c983ee5094a3a6af0a741d/zhaquirks/const.py#L2-L12

https://github.com/zigpy/zigpy/blob/dev/zigpy/quirks/__init__.py#L6-L16

https://github.com/zigpy/zigpy/blob/dev/zigpy/const.py#L12

And finally:
https://github.com/zigpy/zigpy/search?q=SIG_SKIP_CONFIG
to select
https://github.com/zigpy/zigpy/blob/2f8f841a2444bac97e0aa38fa6c4859072096c44/zigpy/quirks/__init__.py#L83

Meaning that it's not handled in zigpy...

So, in ha-core:
https://github.com/home-assistant/core/search?q=skip_configuration

we can see that it's wrapped/aliased in:
https://github.com/home-assistant/core/blob/5e90e178aaa5ba63f5291b8804043f1b4a7d97b2/homeassistant/components/zha/core/device.py#L268-L270

and used in several locations, such a:

https://github.com/home-assistant/core/blob/fbad453c89d4a9a1332cf410addc4ec0c05f3f7a/homeassistant/components/zha/core/channels/base.py#L248-L261

And skip_configuration is defined in the channel pool as:
https://github.com/home-assistant/core/blob/fa7e787415e3a29e0687700bebf011d74873f58b/homeassistant/components/zha/core/channels/__init__.py#L243-L246

Which is not so ideal coding as it does not use devices.skip_configuration .



So skip_configuration skips the bindings which may be a solution for previously unconfigured devices and has to be validated.

And what about the others: they have to manually unbind their devices?  Or remove and add their devices again after adding the quirk (either manually or after a future update) as kind of explained in this [very old joke](https://www.reddit.com/r/Jokes/comments/a26tzt/four_engineers_get_into_a_car_the_car_wont_start/)

MattWestb:
I think one reset to "factors hard way" is needed for getting groups and binding going away in the device (Im not 110% sure) if not sending commands from the quirk deleting them (if its possible from the quirk).

I have reading the info but i dont remember in what Zigbee paper ;-((

By the way what was the Electronic engineer doing ? 
OTA update of the MCU (Motor Control Unit) with BT ? ;-))

MattWestb:
I was reversing my hacked LIDL ZBGW and added one LIDL E14 RGBWW and one remote (TS004F Knob) and changing all parameters in the APP its looks OK but all changes made of the remote is not reflected in the APP = the light is doing no or very bad reporting to the coordinator than being paired with one tuya ZBGW.
So its in the end tuya devices.

mdeweerd:
> By the way what was the Electronic engineer doing ?
> OTA update of the MCU (Motor Control Unit) with BT ? ;-))

I am not sure what you refer to, but it wasn't me (I am an EE ;-) ).

Factory reset: I tried it at one point, but it did no help (I did not retry to check).
zha_toolkit has a function to unbind everything which works and helps.

I think it's an issue with the Tuya devices, but when the binding is removed, I am under the strong impression that the on/off attribute is reported correctly when switching on and off (and I think not when switching on or off through level control).

So the question really is : what is ok for integrating this as a standard quirk?

MattWestb:
I is having problem and have making one bug report that all my LIDL RGBWW devices have stopping changing color only working with CT and its looks they is ignoring the color command if sending one on command before that ZHA is doing all the time.

And also discovered that both my devices is sending off to ZHA if dimming them down to min (1%) is the light still on and ZHA is saying them is off so i think my devices is having the same ugly bug.
If turning on and off from ZHA all looks OK.

I was thinking normal dim up and down shall not truing the light on / off only if using step with on/off.

Issue #1373, opened 2022-02-18
[Device Support Request] LoraTap SS6400ZB-VS TS0044


frank8the9tank:
**Is your feature request related to a problem? Please describe.**
I have bought a zigbee 4 button remote from aliexpress, the device is succesfully connection to HA in the ZHA integration.
I see 4 entities for the 4 buttons, but no state change when i press a button.
I have asked for help on the Home assistant forum and for the ZHA integration on github.
Those guys point me here.
[HA forum](https://community.home-assistant.io/t/zigbee-4-button-remote-added-succesfull-but-not-working/394116/13)
[ZHA github](https://github.com/home-assistant/core/issues/66818)

**Describe the solution you'd like**
I like to have the 4 buttons to work in home assistant.
Trough an entity or an event, as long i can trigger an automation. 
With the 3 modes that that is supports, single click, double click, and long press.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
[0xa33b] Delivery error for seq # 0xbd, on endpoint id 0 cluster 0x0034: message send failure
Sending 'zdo_leave_req' failed: [0xa33b:0:0x0034]: Message send failure
[0x41d3:1:0x0b04] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=True disable_default_response=True> manufacturer=None tsn=150 command_id=Command.Report_Attributes>
[0x41d3:1:0x0b04] ZCL request 0x000a: [[Attribute(attrid=1291, value=<TypeValue type=int16s, value=0>)]]
[0x41d3:1:0x0b04] Attribute report received: active_power=0
[0x41d3:1:0x0b04] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=True disable_default_response=True> manufacturer=None tsn=151 command_id=Command.Report_Attributes>
[0x41d3:1:0x0b04] ZCL request 0x000a: [[Attribute(attrid=1288, value=<TypeValue type=uint16_t, value=0>)]]
[0x41d3:1:0x0b04] Attribute report received: rms_current=0
[0x41D3:1:0x0b04]: async_update
[...]

PR #1374, opened 2022-02-19
Adding one more TS0044 version

MattWestb:
One more TS0044 cluster version. Have moving the OnOff cluster to out in replace and deleting the power-config on EP 2-3 then it was not reporting battery OK with all 4 active.
So no no working switches on the device card and working battery reporting.

Fixes: #1373


MattWestb:
Shall i doing the same fixing on the other device classes ?
Deleting power config on all cluster and only having it on EP 1 so the battery reporting it  working better.
Moving OnOff from in to out cluster so not getting not working switches in ZHA GUI.
Also the same on the ts0041, ts0042 and ts0043 ?

MattWestb:
Was missing that the unknown cluster was defined in INIT and putting it in for future use.

Issue #1375, opened 2022-02-19
[Device Support Request] TS0042 by _TZ3000_wjr2jjrm


bluechris:
Describe the solution you'd like
I have a Wireless Switch 2 GANG from Tuya that advertise with Tuya Hub that Have suport for Short Tap, Double Tap and Long Press. I bought them because i saw in this post here in github that this devices are supported 

https://github.com/zigpy/zha-device-handlers/issues/729

but unfortunately they can be added to ZHA but i dont get buttons but 4 switches that doesnt do anything. I suppose something is need to be added to ZHA for them to work correctly.
I bought them from [Aliexpress](https://www.aliexpress.com/item/1005002847761106.html?spm=a2g0o.9042311.0.0.1b8f4c4dCEWGQ2)

manufacturer": "_TZ3000_wjr2jjrm",
model": "TS0042",

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be

{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0006",
        "0xe000"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    },
    "3": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    },
[...]

PR #1376, opened 2022-02-20
Adding new ts0042 model

MattWestb:
Adding one new model with **4 endpoints** as the ts0044 but is only using the 2 first (hard and software recycling on high level from tuya then using the same for all switches and only changing the device ID !!).
Putting back the device type and moving OnOff to out and deleting power config from all EP but keeping it on EP1 so battery reporting is working better.

Fixes: https://github.com/zigpy/zha-device-handlers/issues/1375

The old is untouched but need being updated in the same way.

### DIFF
--- a/zhaquirks/tuya/ts0042.py
+++ b/zhaquirks/tuya/ts0042.py
@@ -167,7 +167,7 @@ class TuyaSmartRemote0042TOPlusA(CustomDevice):
                     Basic.cluster_id,
                     PowerConfiguration.cluster_id,
                     OnOff.cluster_id,
-                    TuyaZBE000Cluster,
+                    TuyaZBE000Cluster.cluster_id,
                 ],
                 OUTPUT_CLUSTERS: [Time.cluster_id, Ota.cluster_id],
             },
--- a/zhaquirks/tuya/ts0042.py
+++ b/zhaquirks/tuya/ts0042.py
@@ -19,7 +19,7 @@ from zhaquirks.const import (
     PROFILE_ID,
     SHORT_PRESS,
 )
-from zhaquirks.tuya import TuyaSmartRemoteOnOffCluster
+from zhaquirks.tuya import TuyaSmartRemoteOnOffCluster, TuyaZBE000Cluster,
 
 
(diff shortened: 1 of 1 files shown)


javicalle:
> The old is untouched but need **being updated in the same way**.

@MattWestb , does that mean that the `TuyaSmartRemoteOnOffCluster` cluster would not be needed to maintain the functionality of these devices?

It sounds like it could be completely removed (which would be an improvement).

MattWestb:
`TuyaSmartRemoteOnOffCluster` is being used of all ts004x devices (5 quirks) and its needed then the device is sending no ZCL commands 0xFD on cluster 0x0006 with press types: https://github.com/zigpy/zha-device-handlers/blob/f9e25f410da29eb56c9996ca0fcc43bab36c3ba6/zhaquirks/tuya/__init__.py#L777-L783
Also adding sending default response so the switch is not repeating the commands and doing delays / blocking the device then the command is out of ZCL..

I only doing the trick deleting the cluster 0x0006 from in and adding it as one out cluster on the same endpoint (that it shall have being) and we dont getting any no working switch in ZHA GUI but its still working and is tested on 3 devices we have trying (ts0044, ts0042 and ts004f).

As you see its real tyua devices = you is getting what you is paying for.

javicalle:
Sorry, you're right, I got confused when reviewing the changes you had made.
I don't know why I thought you were using the `OnOff` cluster and not the `TuyaSmartRemoteOnOffCluster` 🤦🏻‍♂️

MattWestb:
Its tuya devices so many things is little more then strange.
The ts004f rotating knob is sending command 0xFC with 0x00: ROTATE_RIGHT, 0x01: ROTATE_LEFT and 0x02: ROTATE_STOP on cluster 0x0006 but its not ready then all commands is not encoded and implanted. And i have  ordering one from France so can testing and sniffing if its needed but its still on the way.

I think its one part of housekeeping fixing bad things like this (deleting not working functions that making problem for users).

MattWestb:
Using the unknown cluster from INIT that makes if we is finding some functions of it then the device is getting the function we is implanting in the future.

PR #1378, opened 2022-02-20
Add support for a new TS0044 variety (Tuya 4-gang switch)

aecay:
This reports its model as _TZ3000_ee8nrt2l.  It is similar to the
existing `TuyaSmartRemote0044TOPlus`, but where that variety places an
input cluster with id 0xe000 on endpoint 1, this kind places it instead
on endpoints 2-4.

### DIFF
--- a/zhaquirks/tuya/ts0044.py
+++ b/zhaquirks/tuya/ts0044.py
@@ -321,3 +321,98 @@ class TuyaSmartRemote0044TOPlus(CustomDevice, Tuya4ButtonTriggers):
             },
         },
     }
+
+class TuyaSmartRemote0044TOPlusVarietyTwo(CustomDevice, Tuya4ButtonTriggers):
+    """Tuya 4-button remote device with time on out cluster."""
+
+    signature = {
+        # SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=0, device_version=1, input_clusters=[0, 1, 6], output_clusters=[10, 25]))
+        # SizePrefixedSimpleDescriptor(endpoint=2, profile=260, device_type=0, device_version=1, input_clusters=[1, 6, 57344], output_clusters=[])
+        # SizePrefixedSimpleDescriptor(endpoint=3, profile=260, device_type=0, device_version=1, input_clusters=[1, 6, 57344], output_clusters=[])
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts0044.py
```diff
@@ -321,3 +321,98 @@ class TuyaSmartRemote0044TOPlus(CustomDevice, Tuya4ButtonTriggers):
             },
         },
     }
+
+class TuyaSmartRemote0044TOPlusVarietyTwo(CustomDevice, Tuya4ButtonTriggers):
```
aecay, line 325: I was a bit unsure of the name to use here...I'm sure it would be easy to improve on my choice; suggestions welcome

MattWestb:
The clusters 57344 is wrong in hex it shall being 0xE000 so `TuyaZBE000Cluster` is not working.
I have adding the same cluster setting / endpoints in https://github.com/zigpy/zha-device-handlers/pull/1374. 
Can you testing if its working on your device 2 ?
I have not looking on all lines in your quirk but i think the rest is OK but its having some problem i have fixed in the linked PR.

aecay:
> The clusters 57344 is wrong in hex it shall being 0xE000

I don't understand this part; 57344 and 0xE000 are equal.

> I have adding the same cluster setting / endpoints in https://github.com/zigpy/zha-device-handlers/pull/1374.  Can you testing if its working on your device 2 ?

I will test whether it works and report back.

It looks like the difference between your PR and mine is that you have put `TuyaSmartRemoteOnOffCluster` in the `OUT_CLUSTERS`.  I don't know enough to comment on whether this is correct, although the docs seem to suggest that your way is better.  But I'm not sure what I should be looking for when testing, to see if that setting is correctly applied.

MattWestb:
Sorry i was wrong then was thinking it was one other unknown cluster.
You have doing the right and importing it from INIT and i was not doing it then i was thinking wrong.
The putting the OnOff cluster as out is making the ZHA is not doing 4 switches in the GUI that is not working.

If it loading OK then you shall getting events if listening to  `zha_event` and also you shall having device automatons on the device card  (clicking on the + on the upper right automatons) and they shall working is using them.

aecay:
@MattWestb - happy to report that the configuration from #1347 works with my device :+1:

PR #1381, opened 2022-02-20
Extended ZONNSMART TRV functionality & added _TZE200_hue3yfsn

rforro:
After quite some time I've finally added _TZE200_hue3yfsn into ZONNSMART class, which is basically the same device in new design. Thanks to it I could extend the current ZONNSMART class with new functionalities like:
- modes
- temperature offset
- child lock
- boost mode
- online mode (for 1 minute device publishes all values including current temp)
- opened window sensor
- opened window temperature

Because I don't own the old TRV TV01-ZG _TZE200_e9ba97vf I couldn't tested the new features, but by reading manual on internet it should works.

Here is a picture from ha:

Clipboard01

This PR should close following issue #1027


Kettenknecht:
Great. Thank you very much. Comes this automatically with new HA Release next Wednesday?

rforro:
> Great. Thank you very much. Comes this automatically with new HA Release next Wednesday?

Quirk probably yes, because it was packed into 0.0.67 version which is included in HA 2022.3.0b branch, but not the thermostat presets. For those you have to wait.

Issue #1407, opened 2022-03-03
Tuya Smart Air box - TS0601 _TZE200_yvx5lh6k 


kan84:
Hi,

I have been using Tuya Smart Air box - TS0601 _TZE200_yvx5lh6k for couple of months with Homeassistant without any issue.
Yesterday when i update homeassistant to Home Assistant 2022.3.0 all of the four entities became unavailable.

After device was reset, it was visible in Homeassistant to be added as a new device. As soon as device was added and interview was complete device entities shows as unavailable. 
I have attached the device signature below for the device let me know is there are logs which i can pull  

Device Signature

{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_yvx5lh6k",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}

MattWestb:
`"class": "zigpy.device.Device"` = the system have not loading the quirk for the device.
Then the device signature have not being changed it cant being the device but one user have deleting your device in the quirk that is making it not working longer https://github.com/zigpy/zha-device-handlers/pull/1346.

[...]

Issue #1405, opened 2022-03-02
[BUG] airbox breaks in HA 2022.3


pbvdven:
When update ha to 2022.3.0 my smart airbox loses first stats all entitys as unavailable then i tried to pair it agian but  it loses all its entitys. i have 2 of these one still works after updating. when i move back to ha 2022.2.9 the both work again.

this is the failing one
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0x0402",
        "0x0405",
        "0x040d",
        "0x042b",
        "0x042e",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_yvx5lh6k",
  "model": "TS0601",
  "class": "zhaquirks.tuya.air.ts0601_air_quality.TuyaCO2Sensor"
}

this one works

{
[...]

Issue #953, opened 2021-07-10
[Device Support Request] TS0601 _TZE200_yvx5lh6k (Smart Air Box)


depl0y:
**Is your feature request related to a problem? Please describe.**
Bought a Smart Air Box which should measure CO2, Formaldehyde, temperature and humidity. It does pair, but it does not expose any status.

**Describe the solution you'd like**
Read the available values.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.0: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_yvx5lh6k",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

**Additional context**
Product: https://www.aliexpress.com/item/1005002511043244.html
Blakadder link: https://zigbee.blakadder.com/Tuya_RSH-AirBox01.html

depl0y:
Also, it seems like another variant of this: https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/air/ts0601_air_quality.py

depl0y:
Updating the file mentioned above to include the following line: 

```
[...]

PR #1410, opened 2022-03-05
Reversing braking #1346

MattWestb:
Reversing #1346 that is braking working devices for many user that is confirmed working OK.
If like implanting other devices that is working in one other way pleas do it  without braking devices that have working for long  time.

### DIFF
--- a/zhaquirks/tuya/air/ts0601_air_quality.py
+++ b/zhaquirks/tuya/air/ts0601_air_quality.py
@@ -33,7 +33,8 @@ class TuyaCO2Sensor(CustomDevice):
         # output_clusters=[25, 10])
         MODELS_INFO: [
             ("_TZE200_8ygsuhe1", "TS0601"),
-            ("_TZE200_yvx5lh6k", "TS0601"),        ],
+            ("_TZE200_yvx5lh6k", "TS0601"),
+        ],
         ENDPOINTS: {
             1: {
                 PROFILE_ID: zha.PROFILE_ID,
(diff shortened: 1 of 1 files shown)


MattWestb:
Reported braking: #1407 #1405 = 2 users
Reported working before #1346: #953 = 4 users
So for the moment  1 / 6 plus unknown number of users.

kan84:
#1346 has broken Tuya Smart Air box - TS0601 _TZE200_yvx5lh6k, which has been working flawlessly without any issue. 
The patch work was to manually add custom zha pyquirk. Hopefully this change is reversed and released with next home assistant version.

Nixellion:
I can confirm, this happened to me as well. Been pulling my hair out trying to figure out what went wrong, I already had one half-faulty Tuya Smart Air box - TS0601, so my first assumption was that it's something wrong with the device.

Thank god I found this. I don't want to mess with a manual workaround, I hope it will be fixed in the next version.

Thank you.

MattWestb:
The user with the new problem device must fining one way fixing it and doing one PR that is not braking many working devices for user.
Very likely ist need one own quirk file that is not getting the interference between the device is having more config of the Zigbee module.
Good its working for you with the fix.

You can downloading and configuring local quirks and putting it in there and restarting HA and it shall working or waiting some weeks for getting  it back in the next HA relese.

MPJK77:
Hi,

Are you still working on a solution?
I have a Tuya Zigbee Air detection box (5 in 1) TS0601_TZE200_yvx5lh6k that is only showing VOC. level.
(https://nl.aliexpress.com/item/1005004958270735.html?spm=a2g0o.order_detail.order_detail_item.3.a7f76d76whYxOg&gatewayAdapt=glo2nld)
How do I manage to show the other indicators? (T, CO2, R.V. etc.)

Thanks in advance!

PR #1412, opened 2022-03-05
Adding tuya rotating to OnOff cluster.

MattWestb:
Adding tuya rotating to the OnOff cluster that is needed for updating the TS004f quirk with the new rotating knob https://github.com/zigpy/zha-device-handlers/issues/1322.
Is also moving `class TuyaZBOnOffAttributeCluster` before the `class TuyaSmartRemoteOnOffCluster` so can using its sub classes of defined attributes.

Two question to David: 
STOP is defined because tuya have it in the protocol but its not used in this devices for the moment and in defined in INIT but its also one COMMAND_STOP in constant for quirk but i leaving it like this.

I dont like the SHORT_PRESS, DOUBLE_PRESS and LONG_PRESS that can using the COMMAND_BUTTON_SINGLE, COMMAND_BUTTON_DOUBLE and COMMAND_BUTTON_HOLD alternative also COMMAND_CLICK, COMMAND_DOUBLE and COMMAND_HOLD.

I think we dont need 3 versions of the same thing but i dont chaining it if you not saying doing it.

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -16,11 +16,11 @@ from zigpy.zcl.clusters.smartenergy import Metering
 from zhaquirks import Bus, EventableCluster, LocalDataCluster
 from zhaquirks.const import (
     DOUBLE_PRESS,
+    LEFT,
     LONG_PRESS,
+    RIGHT,
     SHORT_PRESS,
     ZHA_SEND_EVENT,
-    RIGHT,
-    LEFT,
 )
 
 # ---------------------------------------------------------
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -14,7 +14,14 @@ from zigpy.zcl.clusters.hvac import Thermostat, UserInterface
 from zigpy.zcl.clusters.smartenergy import Metering
 
 from zhaquirks import Bus, EventableCluster, LocalDataCluster
-from zhaquirks.const import DOUBLE_PRESS, LONG_PRESS, SHORT_PRESS, ZHA_SEND_EVENT
+from zhaquirks.const import (
+    DOUBLE_PRESS,
+    LONG_PRESS,
+    SHORT_PRESS,
+    ZHA_SEND_EVENT,
+    RIGHT,
+    LEFT,
+)
 
 # ---------------------------------------------------------
 # Tuya Custom Cluster ID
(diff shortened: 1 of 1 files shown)


javicalle:
Hi @MattWestb,
 I haven't read all the linked issue, but doesn't make more sense implement the rotate command at the `LevelControl` cluster?
Or is this another "Tuya doing things the Tuya way" and sending rotate commands in the `OnOff` cluster?

MattWestb:
If you like doing that but its not making any sense then its making "tuya rotate commands" in events mode and not dim up and down that is being made in the command mode and the commands is being sent on cluster 0x0006.
```
{
    "event_type": "zha_event",
    "data": {
        "device_ieee": "00:3c:84:ff:fe:b1:5b:b3",
        "unique_id": "00:3c:84:ff:fe:b1:5b:b3:1:0x0006",
        "device_id": "cf4d64a8db61e2dd336a30f6ae769553",
        "endpoint_id": 1,
        "cluster_id": 6,
        "command": "rotate_type",
        "args": [
            0
        ]
    },
```
Also the "tiya scene commands" in event mode is sent on the  0x00006 cluster and is not Zigbee scene commands only data 0x00, 0x01 and 0x02 that user is calling scenes.

If redoing the rotating then the "scene" shall also being  redone but as i was saying its not bringing any good in the system and not for the users.

In the end we only using the events from this commands and is not making anything in ZHA GUI then its only sending events.

Perhaps the Knob can needing little more work for getting the Light commands working OK with group binding (that is very likely not working) and perhaps need the same patch as the Dimmer switch / remote.

I "extracting" 31 events both command and event mode for making device automatons for this 2 devices but i like getting little feedback from user before posting on PR for it.
https://github.com/zigpy/zha-device-handlers/issues/1322#issuecomment-1059771390

Its real "tuya Zigbee" devices so most thing is not standard and i need getting reading and writing attribute for getting the new endpoints 100% for the user but i cant getting it working with my minimal knowledge of coding so user must paring the dimmer switch with tuya ZBGW for getting them working ;-((

MattWestb:
Thanks David !

I making one PR for updating the TS004F and adding the rotating knob that using all function now in INIT and constants.

Can i moving the rotate from IKEA INIT (its i constants now) in the same PR or do you like have it on one separate PR ?
I can testing it then i have the device before posting it.

dmulcahey:
Separately in another PR after the IKEA one is merged please. Thanks for all of the work you are doing!!

PR #1414, opened 2022-03-06
Test device signature fixture and documentation

challs:
I have been working on support for some additional devices. As part of that, I wanted to be able to easily test if my device signatures would be matched without having to manually test them each time.

So I have created a test fixture that can be used to check if a given device signature will be matched to the given quirk. I wanted to use the real matching algorithm to do this to avoid differences in behaviour between the tests and real usage. In order to do this, I had mock out various parts related to devices.

I have added a test for a device which I have and is already implemented (TS0121 plug) to show how the tests work. I have also created a new section in the README about the tests. Hopefully this will encourage some of the developers implementing new devices to have a go at writing tests with their submissions.

I have some further tests for devices which I am working on but I will submit those together with their code: I wanted to separate the fixture out so that the discussion here can focus around the proposed fixture and documentation, without needing to discuss device implementations at the same time.

For the sample test, I wondered about using the `@pytest.mark.parametrize("quirk", [...])` style that a lot of tests in `test_tuya.py` seem to use, but decided against it so that people don't need to learn that concept before they start writing a test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,6 +422,119 @@ If you look at another example for the same device:
 
 You can see a pattern that illustrates how to match a more complex event. In this case the step command is used for the dim up and dim down buttons so we need to match more of the event data to uniquely match the event.
 
+## Testing using unit tests
+
+The tests use the [pytest](https://docs.pytest.org/en/latest/) framework.
+
+### Getting started
+
+To get set up, you need install the test dependencies:
+
+```bash
+pip install -r requirements_test_all.txt
+```
+
+### Running the tests
+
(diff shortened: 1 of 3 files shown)


challs:
Many thanks for the initial review.

> I'd also like to make a quirk generator that uses the content of the dialog as well

Ah, yes that would make sense and make it easier to get started on a new quirk :+1: 

> I say clean this PR up

* Rebased against today's dev branch
* Added a commit to clean up sorting and linting issues.

PR #1426, opened 2022-03-10
Expose device events for the puck stopping

elupus:
Adds the STOP event from a symfonisk device as a device action.

The symfonisk puck will send a left/right event when starting to spin, then a stop event when it is back at standstill.

* This change is at the moment untested, and perhaps it needs fixes for translations in home assistant.
* I have tested with raw zha events and and a blueprint against KNX relative dimming where it works well.


MattWestb:
I is trying getting the tuya TS004F rotating knob working with one quirk and its have one tuya mode and one dimmer mode.
I was missing the stop command in DA for this device but i have not time looking on it.
Is its possible getting fast roaring left and right ? 
I think its sending one other command for that and can being used (thinking = not knowing / remember).
Also the stop can being selected if its stop from right and left if implanting the direction it was having being rotated as is doe in the Open / Croce remote is have getting.

Its only suggestions if you have time and like getting more out of the nice device.

Great work done !!!

elupus:
I've not seen any fast rotate on this device. It is still tricky to hook up in home assistant. 

I'm not sure I follow what you said about stop direction. The device doesn't seem to indicate which direction is stopping.

MattWestb:
The fast i can have misunderstanding then i have not testing it with the SYMFONISK but the old round gyro is sending one command for slow rotating and one other for fast rotating.

True its not direction of the stop command.
But you always have one `move` command before with direction 0 or 1 = Right / Left.
If you like you can using this logic: 
https://github.com/zigpy/zha-device-handlers/blob/e3fa034284437595d1997326a8597086d623f413/zhaquirks/ikea/opencloseremote.py#L47from the open / close and saving  the direction and using it for the stop command and adding the last "saved" direction.

No must but only if you like.

MattWestb (approved):
Nice done !!

dmulcahey:
LGTM, @elupus if you are set w/ the change we can merge it.

elupus:
I dont think there is much benefit to knowing which direction did stop. A stop is a stop. There is a lack of support for these type of continuous commands in home assistant anyway. So a blueprint will be needed to make any good use of it.

MattWestb:
Its not needed but the IKEA Open/Close is having it as device automation and i thinking doing the same TS004F Knob.

Im not but many user like have so much function they can for triggering automatons in there system but in the end you need on long list on the device for all and its not practical.

Better merging this and later i moving the rotating from IKEA INIT to constants so it can being user for more rotating devices :-)

dmulcahey:
@elupus ill merge this as soon as I get the green light from you. (Just in case you’re still soliciting feedback)

elupus:
I'm good with this solution.

dmulcahey:
Thanks

elupus:
@dmulcahey @MattWestb any of you have the other IKEA dimmer? It seems it uses same args for left/right, so seem very likely it also exposes stop.

MattWestb:
I have 2 of the "retro" dimmers and i can looking later how it working but if i remember its sending nearly the same commands but i was wrong then its sending one other for fast rotating and the symfonisk not then its sending toggle on pressing.
It can also being that the args is wrong for the device automation but i fixing that if is so.
I must first getting the tuya TS004F rotating Knob working without  tuya ZBGW and it can taking little time.

elupus:
This really looks to be standard LevelControl events. I don't think these device triggers should be added by quirks at all. It should be done by the devices just having LevelControl cluster. The slow fast is just the rate parameter, it should be a property of the device trigger rather than a separate trigger.

dmulcahey:
@elupus just using the presence of clusters would cause more triggers to show that aren’t functional. So many devices don’t follow specs. We’re also working on some changes that may affect how this all works. Also, can we move the rest of this to an issue or a discussion so we aren’t chatting on a merged PR? Thanks!

MattWestb:
I think it wold being good is ZHA was having device automatons tigers for standard light controllers functions so not need doing quirk for all devices if they is not needing it for the normal Zigbee functionality.

Also new usier is not knowing how to using one light controller for automation (i dont do it then better using one HA switch for that) and is no good instruction what i have seen.

David can you open one discussion with appropriate topic so user can feeding there input and getting more good ideas ?
Im not so good on that then im more one spammer (as you knowing).

PR #1427, opened 2022-03-11
Lidl ts011f - WIP

mdeweerd:


dmulcahey:
Should we do this in an extension of the basic cluster? We have other devices where we handle things like this in the cluster implementation itself. Then if it needs to be reapplied the user can just issue the reconfigure call (override the bind method, call super first the. Issue these commands)

dmulcahey:
Should we do this in an extension of the basic cluster? We have other devices where we handle things like this in the cluster implementation itself. Then if it needs to be reapplied the user can just issue the reconfigure call (override the bind method, call super first the. Issue these commands)

MattWestb:
With router and no sleepers its shall working OK.
But the problem is sleepers in combination with "Camelots " (transforming its hardware config) that must being made then the device is in INIT mode.
If also sending one leave with rejoining and casting the spell then its doing device accouterments its very likely working but must being tested.

mdeweerd:
Maybe that works as well - FYI discussion are ongoing here: #1428 .

This is actually changing the device signature and idealy triggers a device reinitialisation (which is more than "reconfiguring") but the current zigpy code is skipping too many initializations.

MattWestb:
I suggesting moving the discussion to the issue liked for not bombing the PR for no use.

dmulcahey:
> Maybe that works as well - FYI discussion are ongoing here: #1428 .
> 
> This is actually changing the device signature and idealy triggers a device reinitialisation (which is more than "reconfiguring") but the current zigpy code is skipping too many initializations.

ok I didn’t realize that was changing the signature. I thought puddly had a solution for that that worked in quirks with forcing the device to leave and rejoin. (I thought that’s how you guys found the command bug with that)

mdeweerd:
@dmulcahey The leave with rejoin command bug was discovered related to this because @MattWestb reported this did not work in zha-toolkit (possibly still zha-custom) while he was trying to get the device to rejoin to force a "rediscovery".
But it never was in a quirk.

@MattWestb  I moved the code to a class independent method and for the fun of it I use a wand instead of a task and added a call to rejoin - with retries (I did not wrap the read_attributes, I suppose the current zigpy version supports tries).

No guarantee that the code does not need some tweaks.  (Suggestion: test it by tweaking tuya.py in zha-toolkit ;-) which avoid restarts ).

zhaquirks/lidl/ts011f_plug.py
```diff
@@ -0,0 +1,127 @@
...
+    app = dev.application
+    # Delete the device from the database
+    app.listener_event("device_removed", dev)
+
+    # Delete the device from zigpy
+    app.devices.pop(dev.ieee, None)
+
+
+def tuya_magic_spell(dev, tries=100) -> None:
+    """Set up the magic spell asynchrously."""
```
dmulcahey, line 63: spelling: asynchronously

zhaquirks/lidl/ts011f_plug.py
```diff
@@ -0,0 +1,127 @@
...
+
+    app = dev.application
+    # Delete the device from the database
+    app.listener_event("device_removed", dev)
+
+    # Delete the device from zigpy
+    app.devices.pop(dev.ieee, None)
+
+
+def tuya_magic_spell(dev, tries=100) -> None:
```
dmulcahey, line 62: Is 100 excessive?

zhaquirks/lidl/ts011f_plug.py
dmulcahey, line 63: nit: spelling. Codespell may need tweaks too. No need to look at that for this PR. Just making a note for myself.

zhaquirks/lidl/ts011f_plug.py
```diff
@@ -0,0 +1,127 @@
...
+    """Initialize device so that all endpoints become available."""
+    import inspect
+
+    basic_cluster = dev.endpoints[1].in_clusters[0]
+
+    # The magic spell is needed only once.
+    # TODO: Improve by doing this only once (successfully).
+
+    # Magic spell - part 1
+    attr_to_read = [4, 0, 1, 5, 7, 0xFFFE]
```
dmulcahey, line 38: Are there names for the attributes? I know it’s verbose but I prefer words over magic #’s

MattWestb: 0-7 is ZCL = yes, and the last is not implanted then we dont knowing what is doing but we can putting it in tuya INIT as "tuya-magic" if you like have all node defined.
More info: https://github.com/zigpy/zigpy/discussions/823#discussioncomment-1539469

zhaquirks/lidl/ts011f_plug.py
mdeweerd: I use 100 when configuring sleepy devices - 20 wasn't enough - as far as I remember the "difficult" one responded after about 50 tries.  The cool thing about it was that I wasn't even on location when that happened.

I set 100 since @MattWestb wants to try it on sleepy devices, it is set to 3 for the plug.

The loop stops as soon as the packet is confirmed.

dmulcahey: For sleepy devices folks should be waking them up to pair them and keeping them awake (this is the guidance that is provided).

> The cool thing about it was that I wasn't even on location when that happened.

I don't think we want to encourage users to attempt to configure sleepy devices remotely. This falls into the just because you can doesn't mean you should camp IMHO

Side note: 
I think we are debating creating some sort of command queue for devices. The thought is we will queue up commands for devices that aren't reachable / are sleeping to be processed when they check in. This is probably a ways off for now though.

zhaquirks/lidl/ts011f_plug.py
mdeweerd: @dmulcahey I agree about naming them, for me the first goal is to confirm this works (the "spell" is ok, but rejoin & rediscover not yet).  After that things can cleaned up, moved to tuya/__init__.py, define TuyaBasicCluster that recognizes when the spell is needed, and maybe also stop the fun and rename magic spell and wand to something else but I don't have an idea yet and at the same time I like the current name.

dmulcahey: I don't have an issue w/ whimsical names for things like this. It's silly that we even have to do it at all so silly / fun names make sense :)

zhaquirks/lidl/ts011f_plug.py
MattWestb: Sleepers shall always being waked up and better implanting hits in the GUI for user doing it and trying sending somthing to Xiaomi sleepers is very welcome trying it.

Only thing i think can being good is implanting queue is for ZB3 sleepers that is having working pull control and making check ins (Z2M have doing it but not all the way) but can being good for some cases but real sleepers is not made for receiving commands only then being joined and then only sleeping ans sending little now and then.
"Half sleepers" (end device with radio on the idle) like TRVs and some switches can benefiting for queue if its good implanted but i dont  have so much experience of them.

zhaquirks/lidl/ts011f_plug.py
MattWestb: Like Alexeis tuya `PITA00` and `PITA01` cluster from the beginning ;-)

I think in the end working names that can being understand then working with them is the best but its very WIP.

zhaquirks/lidl/ts011f_plug.py
mdeweerd: I have no issue with "nit(picking)", some "collegues" of mine do.  I appreciate the level of exigence.

As far as I know, "magic spell" is the adjective "magic" and the noun "spell" so I do not understand why codespell needs an update for spelling (unless you want to have 'codespell' renamed to 'codespelling' ;-) ) .

Or, as functions should start with a verb it would have to be "spell_tuya_magic" ?.

dmulcahey: Lol I think asynchronously is missing an n

zhaquirks/lidl/ts011f_plug.py
mdeweerd: A command queue is definitively usefull, but I am not sure it will be sufficient for sleepy devices where the parent is a router.

I haven't digged in that part of ZB yet, but I am under the impression that the pending message has to be waiting at the router so that it can reply to the device when it checks in.  
I am not sure that it's relayed up to the coordinator before the device goes to sleep again.

One may need to act on a report that needs an ack so that the ack and the pending message is sent "at the same time".  Or, know the periodicity of the checkins and send the command just prior to that so that it is still waiting in the router.  + there is Green Power.

MattWestb: If the coordinator is ZB3 and the SED and its parent and having pull control implanted chickins shall working OK but only if all is supporting pull control so no old CC-2431 coordinators or routers with HA zigbee stack.

zhaquirks/lidl/ts011f_plug.py
mdeweerd: Right, overlooked that one.

dmulcahey:
Where do we stand with this?

dmulcahey:
> * How to move this to a custom basic cluster (what we tried did not work because the basic cluster apparently does not imply binding)?

Yeah I forgot we removed that from binding because it was causing binding tables on some devices to become full. My bad on that. Don't worry about that for now. We can figure something else out later.

MattWestb:
I think most devices (but not all) that need tuya magic is using OnOff cluster so one or both used on/off cluster (normal Zigbee not PD / MCU) in tuya INIT ?
Also power configuration cluster can being patched with it then its n´many battery powered device (but not only) that is needing it.

And adding it in the quirk for other (like the parkside whater valve).

dmulcahey:
I am going to merge this as is. We can deal with any cleanup later.

zhaquirks/lidl/ts011f_plug.py
```diff
@@ -0,0 +1,166 @@
...
+    async def bind(self):
+        """Bind cluster."""
+
+        _LOGGER.debug(
+            f"Requesting Tuya Magic Spell for {self.ieee!r} in basic bind method"
+        )
+        tries = 3
+        await asyncio.create_task(cast_tuya_magic_spell_task(self, tries=tries))
+
+        return await super().bind()
```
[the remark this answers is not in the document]
mdeweerd: I was seeking some agreement on adding this to Tuya's init, the suggestion was to use a Custom BasicCluster, but that did not work because there is no binding for that cluster, so I was waiting for feedback.   It was released to make it available sooner to the users.

I wrote the code so that it can be moved elsewhere, but it was local to not change too many files during development.

MattWestb:
This quirk is not working in reality if not patching HA docker container.
The reason is that the device is having one original tuya quirk that is being loaded after the LIDL one and therefore is being the quirk ZHA is using.
My production system running HA 2011.04.X is having: `Quirk: zhaquirks.tuya.ts011f_plug.Plug_3AC_4USB` and is only having one local quirk for my tuya TRVs.

I still think its crazy implanting quirks for the same device in different places lik Z2M is doing then MOES having one own device converter and LIDL is having one and must looking around to finding if its defined in one or more places and in the end all is tuya device.

I recommending the maintainer moving all LIDL device to tuya quirks for making more logical for users and devs.

mdeweerd:
@MattWestb Isn't this because once it is enchanted, it corresponds to the Tuya quirk - and as long as it only has only one endpoint, the LIDL quirk is used?

For me that behavior is fine.

It's ok to implement everything in Tuya, but I think there should be a readme in the lidl directory to point to that so that it's easy to determine that the implementation is elsewhere.

MattWestb:
You is very right then i have not updating the firmware on my power strips so its having 2 more endpoints and the ZGPP EP so its using the tuya quirk and not  the LIDL one.

Its being the same problem then starting adding quirks for tuya DP / MCU device that LIDL is using and all the MCU things is in tuya directory or must doing identical implementations for it.

By the way great work with coding and getting tricky things working !!

PR #1431, opened 2022-03-13
IKEA ROTATED from const.py

MattWestb:
Use ROTATED from const.py and deleting it in IKEA INIT and change the import for symfonisk.py so its using it from const.py.

Edit: I was forgetting that the retro dimer was using it 2 :-((


MattWestb:
Ops !!

PR #1432, opened 2022-03-13
Add Neo siren without sensors (_TZE200_t1blo2bj)

javicalle:
This PR adds the Neo siren `_TZE200_t1blo2bj`.

'New' MCU approach has been used here.
This device requires no manufacturer code in messages, so methods has been overwritten.

Fixes: #1379


javicalle:
Enums for `volume` and `melody` has been added.

PR #1435, opened 2022-03-16
Inovelli New Zigbee Dimmer

InovelliUSA:
Prepping files for the release of our new zigbee dimmer.


zhaquirks/inovelli/VZM31-SN.py
```diff
@@ -0,0 +1,108 @@
...
+            },
+        },
+    }
+
+    replacement = {
+        MODELS_INFO: [("Inovelli", "VZM31-SN")],
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: DeviceType.DIMMABLE_LIGHT,
```
dmulcahey, line 74: @Adminiuga should we map `DeviceType.DIMMER_SWITCH` to the light platform so that quirks don't have to change the device type like this?

dmulcahey:
@InovelliUSA I believe you need to remove the - from the file name as well

InovelliUSA:
Thanks, we will check it out.

zhaquirks/inovelli/VZM31-SN.py
kreene1987: PLEASE do. In development of this device I had to override in the very beginning to even get any dim levels. All dimmers need 0-100% brightness_pct attribute, etc, as provided by the light domain.

Adminiuga: isn't `DeviceType.DIMMER_SWITCH` a "remote" device? Device really should report itself as "Dimmable_Light". 
Check Section 7.5.5 of Zigbee HA Profile, document 05-3520-29. DIMMER_SWITCH is mandated to have client on/off, level control and identity clusters. 
image

Adminiuga: @kreene1987 what kind of dimmer is it? two gang or single gang?

dmulcahey: > isn't `DeviceType.DIMMER_SWITCH` a "remote" device? Device really should report itself as "Dimmable_Light".
> Check Section 7.5.5 of Zigbee HA Profile, document [05-3520-29](tel:05-3520-29). DIMMER_SWITCH is mandated to have client on/off, level control and identity clusters.
> image

Great point. I didn’t check this before hand.

kreene1987: It is a single gang device. I might need a consult on this from @InovelliUSA but I believe we selected the class we did because this device has scene capabilities (multitap up to 5x) and the functionality of the dimmer_switch did not present those, although per the screenshot above it does seem that these are optional. @codyhackw perhaps we can try this offline and report back. We would need both scenes and groups to match the functions of the unit.

Adminiuga: If it is a single gang device, then the endpoint id 1 should be of a DIMMING_LIGHT device type, and the 2nd endpoint should be of a DIMMER_SWITCH device type. Similar to how GE46857 implemented it, none of the others got it right.

kreene1987: > Similar to how GE46857 implemented it, none of the others got it right.

Can you link me to where this quirk is stored? I couldn't find GE or Jasco as a manuf in the quirks folder.

We'll get right on it for testing.

Adminiuga: Yes, I meant that the correct fix is to update the device firmware. There's no quirk for GE 45857, but you can find signature in https://github.com/home-assistant/core/blob/bc194cd209d4261bf7e15aa89eac09be5685c3db/tests/components/zha/zha_devices_list.py#L1674

InovelliUSA: > Why not change the firmware of the device so that it reports itself as the correct device types? You can do it with a quirk but if you can change the firmware to be more accurate it may help other platforms as well.

Thanks for the feedback all, we will discuss with the engineer.

zhaquirks/inovelli/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+            0x0100: ("Local_Protection", t.Bool),
+            0x0101: ("Remote_Protection", t.Bool),
+            0x0102: ("Output_Mode", t.Bool),
+            0x0103: ("On_Off_LED_Mode", t.Bool),
+            0x0104: ("Firmware_Progress_LED", t.Bool),
+        }
+    )
+
+    server_commands = {
+        0x00: ("button_event", (t.uint8_t, t.uint8_t), False),
```
puddly, line 141: To ZCL command definitions, we've added parameter names:

```Python

...

0x00: foundation.ZCLCommandDef("button_event", {"param_name1": t.uint8_t, "param_name2": t.uint8_t}, is_reply=False, is_manufacturer_specific=True),
                                                ^^^^^^^^^^^^^                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

zhaquirks/inovelli/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+    """Inovelli VZM31-SN custom cluster."""
+
+    cluster_id = 0xFC31
+    name = "InovelliVZM31SNCluster"
+    ep_attribute = "inovelli_vzm31sn_cluster"
+
+    attributes = ManufacturerSpecificCluster.attributes.copy()
+    attributes.update(
+        {
+            0x0001: ("Dimming_Speed_Up_Remote", t.uint8_t),
```
puddly, line 75: To ZCL attribute definitions, make them snake case and add a `, True` at the end of the definition tuple to indicate that the attribute is manufacturer-specific.

```Python
0x0001: ("dimming_speed_up_remote", t.uint8_t, True),
          ^       ^     ^  ^                 ^^^^^^
```

If you have more information about the access controls on each specific attribute, you can use the full attribute definition object instead of the shorthand tuple:

```Python

...
0x0001: foundation.ZCLAttributeDef("dimming_speed_up_remote", t.uint8_t, access="rw", is_manufacturer_specific=True)
```

zhaquirks/inovelli/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+        ] = None,
+    ):
+        """Handle a cluster request."""
+        _LOGGER.debug(
+            "%s: handle_cluster_request - Command: %s Data: %s",
+            self.name,
+            hdr.command_id,
+            args,
+        )
+        if hdr.command_id == 0x00:
```
puddly, line 170: You can use `self.commands_by_name["button_event"].command_id` instead of `0x00`

zhaquirks/inovelli/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+
+            button = BUTTONS.get(args[0])
```
puddly, line 172: And similarly, `args.name_of_command_parameter` instead of `args[0]`.

zhaquirks/inovelli/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+            press_type = PRESS_TYPES.get(args[1])
```
puddly, line 173: If you don't expect the provided `PRESS_TYPE` to be anything other than the integers 1-6, maybe use `PRESS_TYPES[args.press_type]`.

puddly (changes requested):
Thanks for the PR!

We've recently restructured our ZCL command and attribute definitions to hold more information. Could you make some changes to ensure this PR can be merged?

zhaquirks/inovelli/__init__.py
codyhackw: I should be able to tackle these along with the rest here soon fortunately. I'll have to test the manufacturer-specific piece as I don't think that worked previously for me though.

Out of curiosity, is there any alternative method that lets us have a more user-friendly name in the UI? Not sure if we'd need to go the full blueprint route for something more user-friendly, but these attributes will be used to configure the switch and edit parameters, etc. I'm new to messing with ZHA quirks so not sure if I just haven't figured it out yet, but it seems like I'm a bit limited in terms of information provided in Manage Clusters, is that the point of blueprints?

dmulcahey: The manage clusters UI is really like a developer mode. We are working on adding configuration entity support to ZHA.

codyhackw: Got it, that makes sense and looking forward to it, thanks!

kreene1987: > The manage clusters UI is really like a developer mode. We are working on adding configuration entity support to ZHA.

Oh heck yes. This will be huge since this device has like 60+ parameters. Z-wave JS obviously is a fantastic template to go off of...

codyhackw:
> Thanks for the PR!
> 
> We've recently restructured our ZCL command and attribute definitions to hold more information. Could you make some changes to ensure this PR can be merged?

@puddly I believe the most recent update should have the requested changes covered if you're able to confirm?

PR #1436, opened 2022-03-18
Adding TS0004

MattWestb:
Adding quad switch module with power on state and external switch type.
My module is having one Telink Zigbee module and not one nice MG21 but looks working OK after shipping form "Chinese Spain" in 4 weeks ;-(

### DIFF
--- a/zhaquirks/tuya/ts000x.py
+++ b/zhaquirks/tuya/ts000x.py
@@ -284,3 +284,143 @@ class Switch_3G_GPP(CustomDevice):
             },
         },
     }
+
+
+class Switch_4G_GPP(CustomDevice):
+    """Tuya 4 gang switch module with restore power state support."""
+
+    signature = {
+        MODEL: "TS0004",
+        ENDPOINTS: {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=256
+            # device_version=1
+            # input_clusters=[0, 3, 4, 5, 6, 57344, 57345]
+            # output_clusters=[10, 25]>
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    TuyaZBE000Cluster.cluster_id,
+                    TuyaZBExternalSwitchTypeCluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [Time.cluster_id, Ota.cluster_id],
+            },
+            # <SimpleDescriptor endpoint=2 profile=260 device_type=256
+            # device_version=1
+            # input_clusters=[4, 5, 6, 57345]
+            # output_clusters=[]>
+            2: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                INPUT_CLUSTERS: [
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    TuyaZBExternalSwitchTypeCluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
+            # <SimpleDescriptor endpoint=3 profile=260 device_type=256
+            # device_version=1
+            # input_clusters=[4, 5, 6, 57345]
+            # output_clusters=[]>
+            3: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                INPUT_CLUSTERS: [
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    TuyaZBExternalSwitchTypeCluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
+            # <SimpleDescriptor endpoint=3 profile=260 device_type=256
+            # device_version=1
+            # input_clusters=[4, 5, 6, 57345]
+            # output_clusters=[]>
+            4: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                INPUT_CLUSTERS: [
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    TuyaZBExternalSwitchTypeCluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
+            # <SimpleDescriptor endpoint=242 profile=41440 device_type=97
+            # device_version=0
+            # input_clusters=[]
+            # output_clusters=[33]>
+            242: {
+                PROFILE_ID: 41440,
+                DEVICE_TYPE: 97,
+                INPUT_CLUSTERS: [],
+                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
+            },
+        },
+    }
+    replacement = {
+        ENDPOINTS: {
+            1: {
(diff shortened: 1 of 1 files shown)


MattWestb:
PS: For user that expecting one Rolls-Royce is thing still one tuya device so the power on state on EP is controlling it for all 4 EPs and its possible setting it on EP 2-4 but you is getting `non` as replay then reading it and that is doing you cant have different power on state on different switches only the same for all (from EP1) ;-(((

MattWestb:
David i getting warning then restarting HA:
```
Logger: zigpy.zcl
Source: /usr/local/lib/python3.9/site-packages/zigpy/zcl/__init__.py:95
First occurred: 08:25:44 (5 occurrences)
Last logged: 08:25:44

Unknown cluster 57344
Unknown cluster 57345
```
Both cluster is defined in tuya INIT and being imported in the quirk OK but `/zigpy/zcl/__init__.py` is complaining so its looks its being forwarded as one ZCL cluster but i dont understand what is doing that then its configured as one `CustomCluster`.

I have implanting it but the user was saying it was working OK but i have not  testing it before i was getting  this devices.

For my is the device / quirk working OK but i dont like the warning on restart of HA.
Perhaps they need being  defined in one other way for working OK without getting the warning ?

javicalle:
I have seen similar WARNING messages in other quirks and have always related it to some message received before ZHA initializes the quirk or perhaps the initialization of the device from the HA registry.
I am speaking without knowing the facts and simply assuming the possible reason for the message. Although it would be preferable not to have it, I have never given it more importance.

MattWestb:
If the quirk for one tuya DP/MCU device is not having the quirk OK loaded its normal getting it then the commands is not hocked by the quirk and the ZCL part is getting strange no TCL commands.
The strange thing is that the quirk is working and so the cluster is possible changing the the switch working mode.

I think the cluster is not 100% OK defined in the quirk system but as long it working i can living with it.

javicalle:
What I mean is that (maybe) zhaquirk is loading the quirk later in time. So in firts sign, ZCL complains about the unknown cluster, but later (once the quirk is loaded) device is working as expected.

I'm not fluent with all the ZHA initialization, but this is the kind of log that make me think that these can be the behavior (full trace [here](https://github.com/zigpy/zha-device-handlers/issues/1033#issuecomment-975182332)):
```
[0x5d4a] Discovered endpoints: [1]
[0x5d4a] Initializing endpoints [<Endpoint id=1 in=[] out=[] status=<Status.NEW: 0>>]
[0x5d4a:1] Discovering endpoint information
Tries remaining: 3
Received frame on uninitialized device <Device model=None manuf=None nwk=0x5D4A ieee=84:fd:27:ff:fe:60:95:9d is_initialized=False> from ep 0 to ep 0, cluster ZDOCmd.Simple_Desc_rsp: b'\x33\x00\x4A\x5D\x14\x01\x04\x01\x51\x00\x01\x04\x00\x00\x04\x00\x05\x00\x00\xEF\x02\x19\x00\x0A\x00'
[0x5d4a:1] Discovered endpoint information: SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=81, device_version=1, input_clusters=[0, 4, 5, 61184], output_clusters=[25, 10])
Unknown cluster 61184
[0x5d4a:1:0x0000] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=True disable_default_response=True> manufacturer=None tsn=52 command_id=Command.Read_Attributes_rsp>
[0x5d4a] Read model 'TS0601' and manufacturer '_TZE200_e3oitdyu' from <Endpoint id=1 in=[basic:0x0000, groups:0x0004, scenes:0x0005, None:0xEF00] out=[ota:0x0019, time:0x000A] status=<Status.ZDO_INIT: 1>>
[0x5d4a] Discovered basic device information for <Device model='TS0601' manuf='_TZE200_e3oitdyu' nwk=0x5D4A ieee=84:fd:27:ff:fe:60:95:9d is_initialized=True>
Device is initialized <Device model='TS0601' manuf='_TZE200_e3oitdyu' nwk=0x5D4A ieee=84:fd:27:ff:fe:60:95:9d is_initialized=True>
Checking quirks for _TZE200_e3oitdyu TS0601 (84:fd:27:ff:fe:60:95:9d)
Considering <class 'ts0601_dimmer_test.TuyaDoubleSwitchDimmer'>
Found custom device replacement for 84:fd:27:ff:fe:60:95:9d: <class 'ts0601_dimmer_test.TuyaDoubleSwitchDimmer'>
device - 0x5D4A:84:fd:27:ff:fe:60:95:9d entering async_device_initialized - is_new_join: True
device - 0x5D4A:84:fd:27:ff:fe:60:95:9d has joined the ZHA zigbee network
[0x5D4A](TS0601): started configuration
[0x5D4A:ZDO](TS0601): 'async_configure' stage succeeded
[0x5D4A:1:0x0000]: finished channel configuration
```

The warning trace occurs before the quirk class is considered.

MattWestb:
From the last restart of HA:
```
2022-03-19 15:56:32 DEBUG (MainThread) [zigpy.appdb] Current database version is v7
2022-03-19 15:56:32 DEBUG (MainThread) [zigpy.appdb] Loading application state
2022-03-19 15:56:32 WARNING (MainThread) [zigpy.zcl] Unknown cluster 57344
2022-03-19 15:56:32 WARNING (MainThread) [zigpy.zcl] Unknown cluster 57345
2022-03-19 15:56:32 WARNING (MainThread) [zigpy.zcl] Unknown cluster 57345
2022-03-19 15:56:32 WARNING (MainThread) [zigpy.zcl] Unknown cluster 57345
2022-03-19 15:56:32 WARNING (MainThread) [zigpy.zcl] Unknown cluster 57345
```
All quirks is loaded and the system is starting and its reading the DB and the cluster have saved states that is triggering the warning and its before it have starting "matching" devices in the db and loading the quirks.

Perhaps using the wrong class for the cluster ?

dmulcahey:
fairly certain the warning is innocuous: https://github.com/zigpy/zigpy/blob/0c4b9a8a25c5eef5b008d66c2e56aae3f8d5e769/zigpy/zcl/__init__.py#L210

javicalle:
> They may be avoidable if you make the Tuya manufacturer specific clusters extend: https://github.com/zigpy/zigpy/blob/0c4b9a8a25c5eef5b008d66c2e56aae3f8d5e769/zigpy/zcl/clusters/manufacturer_specific.py#L7 if the cluster ids are valid for the range defined in that.

No, not extending `ManufacturerSpecificCluster`:
```python
class TuyaZBExternalSwitchTypeCluster(CustomCluster):
    """Tuya External Switch Type Cluster."""

    name = "Tuya External Switch Type Cluster"
    cluster_id = TUYA_CLUSTER_E001_ID
    ep_attribute = "tuya_external_switch_type"
    attributes = {0xD030: ("external_switch_type", ExternalSwitchType)}
```

But cluster ids aren't valid for the `ManufacturerSpecificCluster` range:
```python
TUYA_CLUSTER_E000_ID = 0xE000
TUYA_CLUSTER_E001_ID = 0xE001
```

The message seems innocuous enough. What could be wrong with not extending `ManufacturerSpecificCluster`?

MattWestb:
Its one normal tuya thing making "new" cluster that is out of ZCL and is also out or manufacture cluster range. Also the DP / MCU cluster is out of range but  we is not saving any attribute for it the the DB so the warning is not being triggered.

Perhaps degrading the warning to one info and its not looking so bad for user using this devices that is having attributes saved in this "illegal ranges".

puddly:
> I am not sure what the purpose of the warning is tbh.

Hmm. It's supposed to say that a cluster is not in any range that zigpy expects: clusters should be in one of the ranges defined in the spec for "normal" clusters, or in the manufacturer-specific range.  It looks like Tuya is using one outside of both but the warning is logged before the quirk has a chance to load.

I think @MattWestb's suggestion is good, it shouldn't be a warning.

MattWestb:
One way is adding the "tuya" clusters in Zigpy as separate range as manufacture cluster but then we is braking the standard and its not bringing any more functionality in the system so better deranging the warning.

The warning have being good to see if users have getting the quirk loaded (for all tuya devices with no standard clusters) but is only one side effect and have not with how the system is working OK or not.

PR #1464, opened 2022-04-07
Fix return values for `command` overrides

puddly:
Many quirks had invalid return values when overriding the cluster's `command` method. ZHA was recently updated to properly check this return value, which is causing runtime errors with quirks (mostly Tuya).

This change should fix the underlying problem. And also make MyPy pass by disabling all failing checks...

Related issue: https://github.com/home-assistant/core/issues/69498

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -66,6 +66,22 @@ error_summary = True
 install_types = True
 non_interactive = True
 
+disable_error_code =
+    arg-type,
+    assignment,
+    attr-defined,
+    call-arg,
+    dict-item,
+    index,
+    misc,
+    no-any-return,
+    no-untyped-call,
+    no-untyped-def,
+    override,
+    return-value,
+    union-attr,
+    var-annotated
+
 [pydocstyle]
 ignore =
     D202,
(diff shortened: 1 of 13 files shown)


javicalle:
Hi @puddly thanks for the review. I haven't had a chance to check it out yet.

Have you been able to verify that the change fixes the detected error? Or just checked at tests level?

javicalle:
Changes has been pseudo-tested here:
* https://github.com/home-assistant/core/issues/69498#issuecomment-1091592472

javicalle:
There is also in `zigpy` a `send_default_rsp` method, but requires an `ZCLHeader`:
https://github.com/zigpy/zigpy/blob/76b97fd0fbeb45d818e1dd2eb89a11868016e178/zigpy/zcl/__init__.py#L824-L837

Not usefull in fixing tuya calls.

javicalle:
@puddly Is there anything I can help with here?

puddly:
If you have any Tuya devices to test these changes, that would be very helpful.

I also have a patch for ZHA in the works, if you want to test it out as well: https://github.com/home-assistant/core/compare/dev...puddly:puddly/command-schema-compat

javicalle:
Unfortunately I only have one environment available (my production environment) and it is not updated to the latest version (I have an N-1 version policy). Yes I know, what a programmer!
The zigpy version upgrade prevents me from testing the latest changes.

But I'll try to check as much as I can to see if I see anything (and learn along the way)

dmulcahey (approved):

PR #1465, opened 2022-04-07
Use `stop_with_on_off` instead of `stop` where appropriate

puddly:
Zigpy recently had its two `LevelControl.stop` commands split into two: one with an ID of 0x03 and one with an ID of 0x07. This is an ambiguity in the spec:

image

The ZLL spec, however, calls the second command `stop (with On/Off)`. This seems correct and matches the code-only behavior of the devices.

---

The spec clarifies this later on:

> This command has two entries in Table 3-5, one for the Move to Level, Move and Set commands, and one
> for their 'with On/Off' counterparts. This is solely for symmetry, to allow easy choice of one or other set of
> commands – the Stop commands are identical, because the dependency on On/Off is determined by the original command that is being stopped.


MattWestb (approved):
4 and 5 button looks like `zha_event` after patching the extra args = OK
Great work !!!

PR #1468, opened 2022-04-07
Fix `manufacturer` type for `TuyaClusterData`

javicalle:
Fix for the `TuyaClusterData` the `manufacturer` type.


javicalle:
Required change for make work `manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID`

puddly:
Can you add a unit test for this change?

javicalle:
Ummmm, thanks for your suggestion but this change is only for the `TuyaClusterData` struct.
I can look to add tests for all the class attributes.

javicalle:
@puddly test for manufacturer attribute added.
I'm not sure other tests would be valuable.

It's fine to you to continue with the version as is?

PR #1471, opened 2022-04-08
fix IKEA 5 btn remote E1810

dumpfheimer:
Trying to fix [this bug](https://github.com/home-assistant/core/issues/69375)


dmulcahey:
We have a solution in the works for this issue. Thanks for attempting to correct our mistakes!!

PR #1472, opened 2022-04-08
fixing device automation triggers for HA 2022.04.1

dumpfheimer:
This was a regex search and replace to make ARGS compatible with optional arguments by converting them to a PARAMS dictionary (numbered).

Is supposed to fix as may issues as possible related to https://github.com/home-assistant/core/issues/69375


puddly:
@dumpfheimer have you confirmed this change works?

`{v: k for v, k in enumerate([0, 26, 5])}` is just `dict(enumerate([0, 26, 5]))`, which generates `{0: 0, 1: 26, 3: 5}`. The `"params"` dictionary has only string keys so I'm not sure how this will be able to extract the *n*-th value from the dictionary.

dumpfheimer:
Only with an Ikea 5 button remote.

I believe that in the event data it ends up with positional and named entries but I could be mistaking.

dumpfheimer:
If you know how this can be tested by an average user it would be great if we could provide them with steps how they can test it to gain confidence in these changes

puddly:
Creating a custom quirk containing just the fixed Python module for the `ikea/*.py` device will allow for local testing but considering there are only 45 quirks using device automation triggers with `ARGS`, what do you think about just fixing them to use `PARAMS` properly, with the expected keys? I'm going to try patching a unit test to generate these dictionaries with string keys, maybe it will be relatively quick.

dumpfheimer:
That would of course be better and would be the next step. I just wanted to have a solution as fast as possible (and I did not have time for a "proper" fix)

puddly:
Hmm, the lax matching of `PARAMS` may be a problem: some quirks could be relying on `ARGS: [1]` being *exact*. For example, if `[1, 2]` is sent, `ARGS: [1]` won't match, but `PARAMS: {"key": 1}` will.

dumpfheimer:
You're right that might be possible. Did you see such a case?

puddly:
I have not, it just occurred to me that this is a possibility.

dumpfheimer:
Okay. I know this is far from optimal and it should definitely not stay this way but I think a fix should be published as fast as possible even if it does not work 100%. Remote control not working is imho pretty much a worsed case. That is why I am eager to get something out. Don't think about the poor men standing in front of a broken remote. Think about the wives next to them 😬

puddly:
My approach didn't work, the automation triggers are too free-form to be programmatically migrated. The good news is that only a few commands are affected (basically anything with newly-added optional parameters):

```
COMMAND_STEP
COMMAND_STEP_COLOR_TEMP
COMMAND_STEP_SATURATION
COMMAND_STEP_HUE
COMMAND_MOVE
```

Narrowing the scope of this change to just these commands is safer and will hopefully ensure no other quirks will break while trying to fix the IKEA remotes. There are thankfully only 91 of them so I'll do it with the actual parameter names.

dumpfheimer:
Not needed

PR #1480, opened 2022-04-09
Remove info about dependency manipulation

dmulcahey:
This causes way more harm than help. We will no longer support this type of modification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,54 +535,6 @@ def test_ts0121_signature(assert_signature_matches_quirk):
     assert_signature_matches_quirk(zhaquirks.tuya.ts0121_plug.Plug, signature)
 ```
 
-# Testing new releases
-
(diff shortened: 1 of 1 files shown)


MattWestb:
Sometime its the same with local quirks i  trying remeber the user deleting it then the PR is merged.
Perhaps implanting one warning then ZHA is starting "Warning have found local quirk XXX that can being **DANGERUS** for your system !!" or some thing ?

dmulcahey:
> Sometime its the same with local quirks i trying remeber the user deleting it then the PR is merged.
> Perhaps implanting one warning then ZHA is starting "Warning have found local quirk XXX that can being **DANGERUS** for your system !!" or some thing ?

I think we will implement diagnostics in HA for ZHA. Then we will have better info for issues. We can also make it spit out warnings too. IMO these are less dangerous and they are easily removable. Messing with the deps / containers has all sorts of unintended consequences.

Issue #1476, opened 2022-04-09
[Device Support Request] dimmer switch TS0601 _TZE200_1agwnems


yftj:
Hi hope you are well, I'd like to request support for [these dimmer switches](https://www.ozsmartthings.com.au/products/zigbee-dimmer-switch) (TS0601). They have a middle switch and a brightness increase on-top and decrease in the bottom. Please let me know if there's anything else you need. I'm new to GitHub so forgive my formatting. The zigbee signature is:

'''
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_1agwnems",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}

'''

MattWestb:
Your device need being added in one quirk for working in ZHA and we have getting its working with most tuya dimmers but its little more work getting all in the right place.
Look in this issue https://github.com/zigpy/zha-device-handlers/issues/1302 for the progress and if you can helping testing the quirk for getting it working OK and being added in zha as soon is possible then its many user like having it so there devices is working in ZHA.

yftj:
[...]

Issue #1461, opened 2022-04-04
Moes 3 Gang Light Dimmer Switch ZS-USD3-WH-MS (Tuya "_TZE200_vm1gyrso", "TS0601")


freemanm:
I like to request support for the Moer Zigbee Dimmer Light Switch. see signature detail below.
THe switch I have is a 3-gang version

It is currently pick up as a router and not an End-Device. THis is a [link](https://www.moeshouse.com/collections/smart-dimmer-switch/products/new-zigbee-2-gang-smart-light-dimmer-switch-independent-control-smart-life-tuya-app-relay-status-backlight-switch-off-rf-remote-control-works-with-alexa-google-voice-assistants?variant=39795547045969) to the vendor website.

**Signature**
```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZE200_vm1gyrso",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

javicalle:
A lot of info (thanks for that) but not exactly the one I expected.

That's what we are going to do, you need to enable the `custom_zha_quirks` in your installation. Here is an explanation how to do:
[...]

Issue #1011, opened 2021-08-25
[Device Support Request] Moes Zigbee Dimmer Module MS-105Z


thechad0351:
Hello, 
When trying to add these zigbee dimmer modules via ZHA it does not add any entities, only the device. 
Are you able to assist with support for this device/entities?

{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_la2c2uo9",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}

Thanks in advance,
Chad.

thechad0351:
Thanks for the reply.
I'm only just starting out with home assistant, so not quite sure what to
do with setting up quirks.
How long until the next update for when these will be supported natively in
ZHA?

Cheers,
Craig

MattWestb:
Its merged in to the beta of next HA so you can waiting some week and its being released and shall working out of the box with your device :-)))

thechad0351:
Thank you.

thechad0351:
Hi,

I Updated to the new version of Home Assistant today and the devices are
still not being recognized.
Was the device support added as planned?
Moes MS-105Z 1 Gang Dimmer Module

Cheers,
Craig.

javicalle:
`_TZE200_la2c2uo9` must be supported in HA 2022.02

[...]

Issue #730, opened 2021-01-19
[Device Support Request] TS0601 by _TZE200_dfxkcots


achaulin:
I have a dimmer WDM-1ZBA-EU that have no objects after pairing. It's integrated in Z2M, but not in ZHA. (https://zigbee.blakadder.com/Earda_EDM-1ZAB-EU.html)

**Device info:**

TS0601
by _TZE200_dfxkcots
Zigbee info
IEEE: bc:33:ac:ff:fe:5a:8e:53
Nwk: 0x571a
Device Type: Router
LQI: 144
RSSI: -64
Last Seen: 2021-01-19T11:58:22
Power Source: Mains

**Zigbee device signature**

```
{
  "node_descriptor": "NodeDescriptor(byte1=1, byte2=64, mac_capability_flags=142, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_dfxkcots",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

**Pairing log**

Sent 'mgmt_permit_joining_req' to 58:8e:81:ff:fe:ba:41:82: [<Status.SUCCESS: 0>]
[0x3a87:1:0x0020] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=CLUSTER_COMMAND manufacturer_specific=False is_reply=True disable_default_response=False> manufacturer=None tsn=64 command_id=0>
[0x3a87:1:0x0020] Data remains after deserializing ZCL frame
[0x3a87:1:0x0020] ZCL request 0x0000: []
[0x3a87:1:0x0020] No handler for cluster command 0
[0x3A87:1:0x0020]: Received 64 tsn command 'checkin': []
[0x3a87] Extending timeout for 0x40 request
[...]

PR #1489, opened 2022-04-13
Implementations for `NO_MANUFACTURER_ID` dimmers

javicalle:
This PR adds device definitions for dimmers that require send commands with the `NO_MANUFACTURER_ID` (or what is the same `manufacturer_specific=False`).

A few comments:
* `tuya/mcu/TuyaLevelControl` now handles `move_to_level_with_on_off` command, sending an `on_off` command before the `current_level` command.
* the `TuyaLevelControlManufCluster` handles DPs up to 3 gangs.
* the new `NoManufacturerCluster` will force the `NO_MANUFACTURER_ID` in commands. This is needed for recent dimmers
* at some point I wanted to migrate the remaining classes to the new implementation and I have done now. `TuyaSingleSwitchDimmer` devices will use the same implementation as the rest
* some definitions are using the `NoManufacturerCluster` implementation and others no. It can be important (or not) when adding new devices to the existing implementations

Related issues:
* Fixes: #730
* Fixes: #1011 
* Fixes: #1103
* Fixes: #1302
* Fixes: #1461
* Fixes: #1476

This PR is good to go in next minor version.

### DIFF
--- a/zhaquirks/tuya/mcu/__init__.py
+++ b/zhaquirks/tuya/mcu/__init__.py
@@ -397,6 +397,20 @@ class TuyaLevelControl(LevelControl, TuyaLocalCluster):
             command_id,
             args,
         )
+        # (move_to_level_with_on_off --> send the on_off command first)
+        if command_id  ==  0x0004:
+            cluster_data = TuyaClusterData(
+                endpoint_id=self.endpoint.endpoint_id,
+                cluster_attr="on_off",
+                attr_value=args[1],
+                expect_reply=expect_reply,
+                manufacturer=manufacturer,
+            )
+            self.endpoint.device.command_bus.listener_event(
+                TUYA_MCU_COMMAND,
+                cluster_data,
+            )
+
         # (move_to_level, move, move_to_level_with_on_off)
         if command_id in (0x0000, 0x0001, 0x0004):
             cluster_data = TuyaClusterData(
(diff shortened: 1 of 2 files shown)


yftj:
Hey, I recently had my device support request closed via this thread. I assume a fix has been implemented? Any work I need to do to add the fix or will an update to ZHA via Homeassistant fix this. Thank you so much for your work!

javicalle:
Hi @yftj,
 the fix has been added to the library mainstream and (probably) will be included in the next library version.
And this new version will be included in future versions of HA. It could be in the next fix or in the future v2022.05.

However, there is no simple criteria to define when the new version of the library will be generated or when it will be included in HA, it depends on multiple factors.

yftj:
Just tested today with the new update to HAOS, I have full control of my dimmers now (dimmer switch TS0601 _TZE200_1agwnems)! Thank you very much! There is however a small problem, changing the brightness slider causes the light to momentarily turn off almost like a reset for a fraction of a second (time it takes to blink almost) and then from full brightness goes down to the selected brightness. Expected result would be for the dimmer to change from eg. 40% to 50% instead it goes from 40% to off to 100% and slowly dims to 50%. Sorry if this isn't the right place to report a bug.

javicalle:
@yftj Please, open a new issue with your device info and attach the logs from your problem. The logs will tell us if the behaviour comes from HA, from ZHA or if is a device issue.

PR #1497, opened 2022-04-15
create button.py

weihuan1111:


weihuan1111 (commented):
update

weihuan1111 (commented):
update

weihuan1111 (commented):
update

weihuan1111 (commented):
update

weihuan1111:
@dmulcahey please help to review the merge require,thanks

weihuan1111:
@dmulcahey  hello, can you tell me how to make this commit take effect?

dmulcahey:
Update to the latest HA

puddly:
@weihuan1111 in this PR, did you remove `LevelControl` and replace the `Ota` output cluster with `Basic` intentionally, or was this a mistake? The device supports OTA.

weihuan1111:
@puddly What is the meaning of replacement in the code? Is it neccessary to remove the repacement?

dmulcahey:
Replacement allows you to change standard clusters for modified versions. It also allows you to add or remove clusters if they aren’t functional or if they’re missing. Lots of devices don’t follow the specifications and we need a way to correct them

weihuan1111:
> Replacement allows you to change standard clusters for modified versions. It also allows you to add or remove clusters if they aren’t functional or if they’re missing. Lots of devices don’t follow the specifications and we need a way to correct them

OK， i will remove the repacement and test it

MattWestb:
For one Zigbbee Light controller is sending all commands from out cluster so OnOff and Level control must being on the out part.

One very good example of bad devices is tuya TS004F that have both on in and out and is making ZHA thinking its one light that can being controller from the GUI = false then its one remote that only sending commands.
In the quirk we have deleting them in the replacement from the in and its sending commands OK from the out cluster and ZHA is not making one light in the GUI that  can sending commands to it (that is not possible then its one sleeping device) and all is working OK (the group cluster is little special then the device is not supporting binding and need it as in cluster for adding groups like lights but its one other story).

PR #1508, opened 2022-04-17
Update issue templates to include proper formatting for device signature

TheJulianJES:
These changes should avoid any confusion on whether removing and re-joining a device is required for just getting the device signature. It should also help with the proper formatting of the device signature.

(If further debug logs of a device working are required for a specific quirk, they can still be requested individually. I don't think it makes sense to include instructions to get debug logs in the issue template, as most simple quirks don't require these and grabbing them can be somewhat complicating for beginners. Just grabbing the device signature should be enough for a start.)

I'm open for changes of course.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -23,5 +23,15 @@ A clear and concise description of what you expected to happen.
 **Screenshots**
 If applicable, add screenshots to help explain your problem.
 
+**Device signature - this can be acquired by clicking on the "Zigbee Device Signature" button in the device settings**
+```
+Paste the device signature here
+```
+
+**Additional logs**
+```
+Paste any additional debug logs here
+```
+
 **Additional context**
-Add any other context about the problem here. Please include full debug logs as well as the device signature.
+Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -23,15 +23,32 @@ A clear and concise description of what you expected to happen.
 **Screenshots**
 If applicable, add screenshots to help explain your problem.
 
-**Device signature - this can be acquired by clicking on the "Zigbee Device Signature" button in the device settings**
+<details>
+<summary>Device signature - this can be acquired by clicking on the "Zigbee Device Signature" button in the device settings</summary>
+
 ```
-Paste the device signature here
+Paste the device signature here.
 ```
 
-**Additional logs**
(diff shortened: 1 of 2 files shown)


dmulcahey:
Awesome. Can you add a section for device diagnostics too please?

TheJulianJES:
Right, I forgot we have these now.
Perhaps the device signature section could even be removed (in the future), since it should be included in the diagnostics. Personally, I'd probably still keep it around for a bit though, as some users may not be up-to-date and it's a little easier to get the signature on mobile/tablets or similar? (and the signature is also a bit more compact)

Should diagnostic information perhaps be "hidden" by default, so the issue doesn't get as long?
-> Which of the following "designs" would be better?

---

**Diagnostic information - this can be acquired by clicking on the "Download Diagnostics" button in the device settings**
```
Paste the diagnostic information here
somewhat long text here will be always visible
```

---

<details>
<summary>Diagnostic information - this can be acquired by clicking on the "Download Diagnostics" button in the device settings</summary>

```
Paste the diagnostic information here
somewhat long text here will be mostly hidden
```

</details>

dmulcahey:
I like the expandable one. Would it make sense to do that for the signature and the diagnostics data?

MattWestb:
Like this:
<details>
  <summary>Device Diagnostics</summary>
Put the Device diagnostics here  
</details>

Code:
```
<details>
  <summary>Device Diagnostics</summary>
Put the Device diagnostics here  
</details>
```

TheJulianJES:
> I like the expandable one. Would it make sense to do that for the signature and the diagnostics data?

Yeah, I think that's good. Maybe also do it for logs at this point (like below)?

The "code" isn't as nice with the expandable sections (updated PR) but I think it should be do-able for the average user. (The Home Assistant Core template doesn't look "easy" at first glance for example.)

---

<details>
<summary>Device signature - this can be acquired by clicking on the "Zigbee Device Signature" button in the device settings</summary>

```
Paste the device signature here.
```

</details>

<details>
<summary>Diagnostic information - this can be acquired by clicking on the "Download Diagnostics" button in the device settings</summary>

```
```

</details>

<details>

```
Paste any additional debug logs here.
```

</details>

TheJulianJES:
Also, apparently the Home Assistant core repo now uses the new GitHub [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms).
That can probably be implemented for this repo or all ``zigpy`` repos at some point (but I don't have the time to do it atm).

These changes should still be a bit better than the current templates.

PR #1513, opened 2022-04-18
Removing comment that is not needed + new device ID

MattWestb:
The DA is now  working after the ZCLR7.5 update then using PARMS instead of ARGS so no need for comments on the code.

### DIFF
--- a/zhaquirks/tuya/ts004f.py
+++ b/zhaquirks/tuya/ts004f.py
@@ -68,7 +68,7 @@ class TuyaSmartRemote004FROK(CustomDevice):
     """Tuya Smart (rotating) Knob device."""
 
     signature = {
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts004f.py
```diff
@@ -68,9 +68,9 @@ class TuyaSmartRemote004FROK(CustomDevice):
...
+        # "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size82, descriptor_capability_field=0, *allocate_address=True, *complex_descriptor_available=False, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False, *is_valid=True, *logical_type=<LogicalType.EndDevice: 2>, *user_descriptor_available=False)",
```
javicalle, line 71: `maximum_outgoing_transfer_size=82`

dmulcahey (approved):

PR #1515, opened 2022-04-20
Fix King Of Fans quirk returning radio status code

puddly:
Fixes https://github.com/home-assistant/core/issues/69893

I don't own the device and had to rewrite its unit test so while tests pass and it seems like it should work, I can't verify that it does.

Would appreciate it if someone can test it out so this PR can be merged.


tube0013:
Tested this and no longer get the error like in https://github.com/home-assistant/core/issues/69893 ( I don't have light hooked up to the fan, but the light entity switches on and off fine with no errors)

PR #1526, opened 2022-04-24
Verify quirks subclassing ZCL clusters extend them

### LINKED ISSUE #1520: [BUG] Upgrading Home Assistant to 2022.4.4 breaks Elko Super TR thermostat.
**Describe the bug**

Upgrading Home Assistant to 2022.4.4 breaks Elko Super TR thermostat.

**To Reproduce**

Upgrade Home Assistant from an earlier version to 2022.4.4 no longer allows control of the thermostat. Commands fail with "Failed to call service climate/set_hvac_mode. 'system_mode'" 

**Expected behavior**

Pressing the on button on the default thermostat card should enable the heater. Fails with error message.

<details>
<summary>Device signature </summary>

```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4919, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0201",
        "0x0204",
        "0x0b04"
      ],
      "out_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "ELKO",
  "model": "Super TR",
  "class": "zhaquirks.elko.smart_super_thermostat.ElkoSuperTRThermostat"
}
```

</details>

<details>

```File "/usr/src/homeassistant/homeassistant/components/zha/climate.py", line 425, in async_set_hvac_mode
if await self._thrm.async_set_operation_mode(HVAC_MODE_2_SYSTEM[hvac_mode]):
File "/usr/src/homeassistant/homeassistant/components/zha/core/channels/hvac.py", line 243, in async_set_operation_mode
if not await self.write_attributes({"system_mode": mode}):
File "/usr/src/homeassistant/homeassistant/components/zha/core/channels/hvac.py", line 292, in write_attributes
res = await self.cluster.write_attributes(data, **kwargs)
File "/usr/local/lib/python3.9/site-packages/zhaquirks/elko/smart_super_thermostat.py", line 110, in write_attributes
return await super().write_attributes(attributes, manufacturer=manufacturer)
File "/usr/local/lib/python3.9/site-packages/zigpy/zcl/__init__.py", line 535, in write_attributes
attrs = self._write_attr_records(attributes)
File "/usr/local/lib/python3.9/site-packages/zigpy/zcl/__init__.py", line 502, in _write_attr_records
attr_def = self.find_attribute(attrid)
File "/usr/local/lib/python3.9/site-packages/zigpy/zcl/__init__.py", line 186, in find_attribute
return self.attributes_by_name[name_or_id]
KeyError: 'system_mode'
```

</details>

puddly:
Fixes #1520

Currently a few unit tests are failing:

- [ ] `zhaquirks.tuya.mcu.TuyaOnOff`
- [ ] `zhaquirks.tuya.ts0601_dimmer.TuyaOnOffNM`
- [ ] `zhaquirks.xbee.XBeeCommon.DigitalIOCluster`

@javicalle for the Tuya clusters, them subclassing `zcl.OnOff` but then exposing only a single `on_off` attribute seems like it will break code expecting the other attributes to exist. Can the `OnOff` subclass be replaced with just `CustomCluster` or is this intentional?

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -62,6 +62,8 @@ for manufacturer in zq._DEVICE_REGISTRY._registry.values():
 
 del quirk, model_quirk_list, manufacturer
 
+ALL_ZIGPY_CLUSTERS = frozenset(zcl.clusters.CLUSTERS_BY_NAME.values())
+
 
 SIGNATURE_ALLOWED = {
     ENDPOINTS,
@@ -658,3 +660,36 @@ def test_quirk_device_automation_triggers_unique(quirk):
                 [f" * {event} <- {trigger}" for trigger, event in triggers_and_events]
             )
             fail_func(f"Triggers are not unique for {quirk}:\n{triggers_text}")
+
+
+@pytest.mark.parametrize("quirk", ALL_QUIRK_CLASSES)
+def test_attributes_updated_not_replaced(quirk: CustomDevice) -> None:
+    """Verify no quirks subclass a ZCL cluster but delete its attributes list."""
+
+    for ep_id, ep_data in quirk.replacement[ENDPOINTS].items():
+        for cluster in ep_data.get(INPUT_CLUSTERS, []) + ep_data.get(
+            OUTPUT_CLUSTERS, []
+        ):
+            if isinstance(cluster, int) or not issubclass(cluster, zcl.Cluster):
+                continue
+            elif cluster in ALL_ZIGPY_CLUSTERS:
+                continue
+
+            assert issubclass(cluster, zigpy.quirks.CustomCluster)
+
+            base_clusters = set(cluster.__mro__) & ALL_ZIGPY_CLUSTERS
+
+            # Completely custom cluster
+            if len(base_clusters) == 0:
+                continue
+            elif len(base_clusters) > 1:
+                pytest.fail(f"Cluster has more than one zigpy base class: {cluster}")
+
+            base_cluster = list(base_clusters)[0]
+
+            # Ensure the attribute IDs are extended
+            if not set(base_cluster.attributes) <= set(cluster.attributes):
+                pytest.fail(
+                    f"Cluster deletes parent class's attributes"
+                    f" instead of extending them: {cluster}"
+                )
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -662,7 +662,18 @@ def test_quirk_device_automation_triggers_unique(quirk):
             fail_func(f"Triggers are not unique for {quirk}:\n{triggers_text}")
 
 
-@pytest.mark.parametrize("quirk", ALL_QUIRK_CLASSES)
+@pytest.mark.parametrize(
+    "quirk",
+    [
+        quirk_cls
+        for quirk_cls in ALL_QUIRK_CLASSES
+        if quirk_cls
+        not in (
+            zhaquirks.xbee.xbee_io.XBeeSensor,
+            zhaquirks.xbee.xbee3_io.XBee3Sensor,
+        )
+    ],
+)
 def test_attributes_updated_not_replaced(quirk: CustomDevice) -> None:
     """Verify no quirks subclass a ZCL cluster but delete its attributes list."""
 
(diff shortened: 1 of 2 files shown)


javicalle:
Most of the Tuya manufacturer cluster only handle a few attributes, for example for the `OnOff` cluster only `on_off` attribute. I wasn't sure if it was better to remove attributes that the device wouldn't be able to handle. Seeing that it is common for devices to define all the attributes even if they do not implement them, now I believe that the appropriate approach is to extend the cluster 'as is' and then detect those possible use cases not covered.

I think that the code is more understandable if it extends from an `OnOff` cluster but I don't know if it must fulfill a minimum of functionality. For example, I recently discovered an issue (in this case with a dimmer) with the `move_to_level_with_on_off` method that was not implemented on Tuya devices, but was used by the HA UI.

There is a risk that these devices present more problems in integrations (for example with HA) because they do not correctly implement some of the attributes/methods of a cluster, but I believe that the integration will only be possible if standard Zigbee clusters and attributes are used. Otherwise (for example using a `CustomCluster`) I think the devices would not show up in HA which would be more confusing for users and generate more issues.

So, I can fix it in a separate PR or you can include it in this same PR, what do you prefer?

javicalle:
I have checked it, and you can just remove this code from `zhaquirks.tuya.mcu.TuyaOnOff`:
```python
    attributes = {
        ATTR_ON_OFF: ("on_off", t.Bool),
    }
```

This must fix the unit tests for both Tuya classes.

puddly:
Interesting. I guess as long as the attribute and command definitions match what the device actually supports I think your current approach is correct, I'll update the unit test to skip this cluster.

@dmulcahey thoughts? Which approach makes it easier for ZHA/ZHAWS to properly expose device functionality for devices like this? Should we emulate the bare minimum?

dmulcahey:
> @dmulcahey thoughts? Which approach makes it easier for ZHA/ZHAWS to properly expose device functionality for devices like this? Should we emulate the bare minimum?

I think so.

javicalle:
> Should we emulate the bare minimum?

But, how can anyone to know what is this 'bare minumum'?

dmulcahey:
> > Should we emulate the bare minimum?
> 
> But, how can anyone to know what is this 'bare minumum'?

I think we only leverage commands marked mandatory in the spec but I’d have to double check

javicalle:
@puddly limit the functionality of a Zigbee cluster I think it can only generate problems, especially if we take into account that some of them may be mandatory, so IMHO there should be no exceptions in the validations.

Perhaps the question would be whether it is feasible to fully implement a quirk of a cluster that the device does not define. Surely an `OnOff` cluster is not too complicated:
image

Or for example a classic such as the `PowerConfiguration` cluster full of optional attributes and no methods.

But implementing the `LevelControl` cluster is not as simple as translating the values of the `current_level`:
image

So far the `TuyaLevelControl` cluster only implements the commands 0x00, 0x01 and 0x04:
* https://github.com/zigpy/zha-device-handlers/blob/ae8377749383648bc8b806de0b3e1e258b6ef85a/zhaquirks/tuya/mcu/__init__.py#L415

And I'm not sure I _can_ correctly implement the rest of the mandatory commands. 
But I'm not sure I _want_ to do it either, considering that the manufacturer has consciously decided not to implement (for whatever reason) those Zigbee clusters and make use of the manufacturer's cluster.

In any case, it is not a 'problem' only with Tuya devices, I believe that this issue is transversal to any quirk of any manufacturer

PR #1546, opened 2022-05-04
Update ts0601_dimmer.py

intra-aud:
Signature added to dimmer config and confirmed working.


intra-aud:
Adjusting PR, please disregard.

Issue #1399, opened 2022-02-27
[Device Support Request] Zemismart Curtain ZM79E-DT - position inverted


rjbungay:
**Is your feature request related to a problem? Please describe.**
I recently purchased 2 Zemismart Curtain Tracks ZM79E-DT having checked that they were compatible with ZHA on the zigbee.blakadder website. I am running Home Assistant 2022.2.9 supervised on a Pi 4 4GB with a SonOff ZBBridge running ZHA.

If I just add the Zemismart to ZHA the ZM79E is recognised but fails due to "input cluster mismatch on at least one endpoint" and shows with no control entities, just LQI and RSSI.

I have managed to identify that this is down to the ts0601_cover.py showing a different cluster configuration for the _TZE200_3i3exuay with "time" as an input cluster rather than an output cluster. I have managed to edit a copy of ts0601_cover.py show "time" as an output cluster and load this as a custom quirk.

From 0601_cover.py with "time" shown as an input cluster:-
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    Time.cluster_id,
                    Groups.cluster_id,
                    Scenes.cluster_id,
                    TuyaManufCluster.cluster_id,
                ],
                OUTPUT_CLUSTERS: [Ota.cluster_id],

The ZM79E is now recognised and loaded as a cover device with control entities. Howver, the controls don't work as the position report appears to work inverted 0=open, 100=closed. I am able to invert this with the remote control but now the Up/Down controls on the remote and HA are inverted. This means that the curtains can't be controlled in HA as the cover thinks it is already at the end stop and "greys" out the control needed to move it in the correct direction.

I am not a programmer or familiar with ZHA/quirks so am now out of my depth. I have done a lot of googleing and think that this should be correctable by editing the __init__.py file to add "_TZE200_3i3exuay" to the inverted by default list and/or re-mapping the 0x0000, 0001, 0002 Up, Down, Stop commands as needed. I have tried editting these parameters in a copy of the __init__.py file and saving this in the custom quirks directory but this doen't seem to be doing anything to affect the position reporting or altering the direction of the Up/Down controls. As this is a supwervised install I don't believe I can access the original files in the \usr\ directory to edit them. Am I doing this correctly or is there something I'm missing/doing wrong with the init file?

[...]

PR #1549, opened 2022-05-05
Fixing issues with _TZE200_3i3exuay

datebo:
I had the same issue as zigpy/zha-device-handlers#1399 

I implemented the changes proposed there except that the only thing that needed to be inverted was the open and close command. I tested the code and it works for me. 

This is my first pull request, so please advise me in case I didn't follow any procedures correctly.


javicalle:
Please, don't modify the existing quirk because that will break another user's device.
I think that in your case it would be enough to add your device in the `TuyaMoesCover0601` quirk:
* https://github.com/zigpy/zha-device-handlers/blob/6f5c5c9143436f05d4dbc89b0568fe9fd4010ed4/zhaquirks/tuya/ts0601_cover.py#L164

And yes, the same device can be in 2 diferent quirk.

If you modify the inverted behaviour, someone will complain for sure, because its cover will then be inverted. (I have always wondered how this is managed because the direction of the shutter will depend on how it is installed and each one will do it differently)

You may be able to configure the behavior by modifying some of the attributes of the `TuyaWindowCoverControl` cluster:
* `motor_direction`
* `cover_inverted`

Regards.

datebo:
Thanks for the feedback. I now understand I was probably under the false impression that all devices with the same model number would behave the same. You live and learn:) And it is good to know a device can be in two quirks, that should then be an easy fix.

I'll see if I can fix it this weekend. Inverting the controls might be a challenge but let's see where it takes me;)

datebo:
> Please, don't modify the existing quirk because that will break another user's device. I think that in your case it would be enough to add your device in the `TuyaMoesCover0601` quirk:
> 
> * https://github.com/zigpy/zha-device-handlers/blob/6f5c5c9143436f05d4dbc89b0568fe9fd4010ed4/zhaquirks/tuya/ts0601_cover.py#L164
> 

This worked like a charm. 

I am struggling however with the inverted controls. The cover controls are now hardcoded in a global variable `TUYA_COVER_COMMAND` based on the "manufacturer" to identify the specific device. As we have established however, the behavior appears not to be unique. So this would require a more fundamental change to the code. I am currently not knowledgeable enough to decide what would be the best approach to do so. 

Currently the best ideas that I have are to either change the `TUYA_COVER_COMMAND` to be a combination of the "manufacturer" and the selected quirk. Or alternatively edit the `TuyaWindowCover` class so the commands can be configured in the quirks instead of in the __init__ file. The last option seems to me to be the cleanest since you move all your configurations to the ts0601_cover file and out of the __init__ file.

What would you advise @javicalle ?

javicalle:
I haven't look at the code in detail, but my impression is that there are 2 aspects to control:
* the device control: what happens when you push the open/close controls. Is the device reacting accordl to the commands?
* the device visualization: what is reporting the device when moves. When the device opens, it shows as open in HA?

These aspects are the ones that I expect that the `motor_direction` and `cover_inverted` attributes would control.
Have you try to modify its values?
Which aspect would you to fix?

My guess will be that the `cover_inverted` will be the one that invert the device control, but just try to know which one can be.

Be aware that this comment in code, could mean that it is not implemented for the open/close commands:
https://github.com/zigpy/zha-device-handlers/blob/b802c1fb2cf2682f9a4722bfb57a1958cad9dad7/zhaquirks/tuya/__init__.py#L1116-L1117
But I believe that the `set_position` function is implemented. In your tests play also with the position bar to understand the attributes funcrion.

datebo:
The `cover_inverted` is used for how the cover reports it's status, ie to indicate whether 100% means fully opened or fully closed. And this already seems to be implemented well:

https://github.com/zigpy/zha-device-handlers/blob/b802c1fb2cf2682f9a4722bfb57a1958cad9dad7/zhaquirks/tuya/__init__.py#L1126-L1132

The motor_direction doesn't seem to be used in the code after it is being defined. I doubt however that it would lead to a general solution. If I may believe the code it appears it is for some devices not just the inverting of the controls but also the stop command is scrambled into the mix:

https://github.com/zigpy/zha-device-handlers/blob/b802c1fb2cf2682f9a4722bfb57a1958cad9dad7/zhaquirks/tuya/__init__.py#L103-L110

javicalle:
This part of the code handles how messages are sent to the device:
https://github.com/zigpy/zha-device-handlers/blob/b802c1fb2cf2682f9a4722bfb57a1958cad9dad7/zhaquirks/tuya/__init__.py#L1085-L1092

And this one, how is handled the device reports:
https://github.com/zigpy/zha-device-handlers/blob/b802c1fb2cf2682f9a4722bfb57a1958cad9dad7/zhaquirks/tuya/__init__.py#L985-L993

> The motor_direction doesn't seem to be used in the code after it is being defined.

That's true, but my guess is that the cover internally invert the cover rotation for the open/close commands. Have you tried to update?
I understand that in your case, the position report is correct, but the controls from HA are reversed, so isn't that what you need to change?
Don't be afraid to update the attribute values from the device vies (but first get the actual value 😉 )

I may be missing something (I'm not familiar with this code), but there has to be some way to 'fix' it other than each user changing the order of the commands definitions.

javicalle:
> also the stop command is scrambled into the mix:

Ouch!
I haven't paid attention to this. That would be a big problem.

datebo:
Okay, I managed to invert the motor on my device. It took a lot of time and I didn't manage to do it using Zigbee, but I did find instructions how to do it with the remote (for which I first had to buy batteries).

Long story short, in the end I only had to add one line, but now it is good to go.

javicalle:
Just bucause it can be useful to other users, has it been necessary to adjust the cover to the remote commands? Or have you used the remote to adjust the cover according to HA?

I think that knowing how you have managed to fix it can help other users.

javicalle (approved):
LGTM

datebo:
> Just bucause it can be useful to other users, has it been necessary to adjust the cover to the remote commands? Or have you used the remote to adjust the cover according to HA?
> 
> I think that knowing how you have managed to fix it can help other users.

Good question. Just to paint the complete picture, the remote is not Zigbee based, it is a normal RF remote. With some procedure using the remote, the controls were inverted and it also caused the cover to react differently to the open/close Zigbee commands that I sent via HA. So I believe the cover inverted itself.

Since it is possible to invert the cover using the RF remote it might also be possible to do this using Zigbee, but I couldn't figure out how.

javicalle:
Thank you very much for clarifying. 🎉 

It seems that a good starting point should be to have the shutter working correctly with the controller before trying to adjust the behavior in HA. Good to know.

Thanks for your work and patience.

datebo:
Just in case someone is looking at this topic for answers. After a while I have been able to pick up my project again, but I have been too quick with my testing last time. Inverting the motor on the curtain also inverts the way it reports the status of the curtain. So either the controls are inverted or the curtain position.

The PR still makes the curtain available in HA, so it is very useful since you don't need a custom quirk anymore. But still have to find a way to get the position inverted without disrupting the old version of the same model.

javicalle:
There are 2 attributes `cover_inverted ` and `motor_direction`.
My guess is that one will change the cover movement and other could change the report value (that is, how is reported to HA).

Once again, I am neither familiar with code or Tuya covers. So pick my comments with caution.

PR #1511, opened 2022-04-18
IKEA Controllers PARMS for 4Bt 5Bt and 5Bt HA no ZCLS commands

MattWestb:
Was missing that some not ZCL light commands was using args and some extra that was not used and need being added and also fixing the last ARGS -> PARMS

PR #1334, opened 2022-02-07
Ensure all device automation triggers are unique

puddly:
Inspired by https://github.com/home-assistant/core/issues/65968

Some quirks do not have unique automation triggers and need to be fixed:

 - [ ] `AuroraDimmerBatteryPowered`
 - [ ] `IkeaTradfriRemote1`
 - [ ] `IkeaTradfriRemote2BtnZLL`
 - [ ] `IkeaTradfriRemote2Btn`
 - [ ] `SwitchAQ3V2`
 - [ ] `SwitchAQ3`
 - [ ] `SwitchAQ3B`

PR #1552, opened 2022-05-06
SYMFONISK II

MattWestb:
Adding new cluster config for IKEA SYMFONISK that looks being the same as was with Shortcut button (different parameters flashed in user data that changing the cluster settings).
Fixing https://github.com/home-assistant/core/issues/71355

### DIFF
--- a/zhaquirks/ikea/symfonisk.py
+++ b/zhaquirks/ikea/symfonisk.py
@@ -10,6 +10,7 @@ from zigpy.zcl.clusters.general import (
     Ota,
     PollControl,
     PowerConfiguration,
+    Scenes,
 )
 from zigpy.zcl.clusters.lightlink import LightLink
 
@@ -39,8 +40,10 @@ from zhaquirks.const import (
 )
 from zhaquirks.ikea import IKEA, PowerConfiguration1CRCluster
 
+IKEA_CLUSTER_ID = 0xFC7C  # decimal = 64636
 
-class IkeaSYMFONISK(CustomDevice):
+
+class IkeaSYMFONISK1(CustomDevice):
     """Custom device representing IKEA of Sweden TRADFRI remote control."""
 
     signature = {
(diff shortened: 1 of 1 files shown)


MattWestb:
Is `CI / Run tests Python 3.7 (pull_request)` broken ???

dmulcahey:
Coveralls appears to have issues

dmulcahey:
Hrm I wonder if we killed a token we shouldn’t have…

zhaquirks/ikea/symfonisk.py
```diff
@@ -132,3 +135,101 @@ class IkeaSYMFONISK(CustomDevice):
...
+            CLUSTER_ID: 8,
+            ENDPOINT_ID: 1,
+            PARAMS: {"step_mode": 0, "step_size": 1, "transition_time": 0},
+        },
+        (TRIPLE_PRESS, TURN_ON): {
+            COMMAND: COMMAND_STEP,
+            CLUSTER_ID: 8,
+            ENDPOINT_ID: 1,
+            PARAMS: {"step_mode": 1, "step_size": 1, "transition_time": 0},
+        },
```
TheJulianJES, line 234: Since the device triggers seem to be identical, maybe just use:
```python
device_automation_triggers = IkeaSYMFONISK1.device_automation_triggers.copy()
```
(``copy()`` probably isn't needed but it's also used in ``fivebtnremotezha.py``)

MattWestb:
Thanks for fixing the tests **David** !!

@TheJulianJES I was thinking using this from the 2Bt:
https://github.com/zigpy/zha-device-handlers/blob/b802c1fb2cf2682f9a4722bfb57a1958cad9dad7/zhaquirks/ikea/twobtnremote.py#L183

But the maintainers dont like using not unique device triggers per https://github.com/zigpy/zha-device-handlers/pull/1334.
I was asking if i shall fixing the 3 IKEA ones but was not getting and response  of it so i have not doing it and i dont doing things that cant being merged like https://github.com/zigpy/zha-device-handlers/pull/1511 that was not being merged in HA 2022.5 that was very urgent (very sad for our users that is using the remotes for clicking around and in the end moving to Z2M).

dmulcahey:
I think there is confusion. What @TheJulianJES suggests is correct.

dmulcahey:
The puddly PR is about not letting a single device define the same trigger more than once in its triggers list

MattWestb:
I like the "TheJulianJES Fix" then its reusing good code and not spamming all quirks with  duplicate code without adding more functions to the system.

MattWestb:
PS: the updated quirk is not tested with one real device but i think its shall working OK.

MattWestb:
David i think you have one more broken test that is blocking  one PR https://github.com/zigpy/zha-device-handlers/pull/1549.

dmulcahey:
> David i think you have one more broken test that is blocking one PR #1549.

It just needs a rebase. Same issue as this one

LCMilstein:
Hi, all. I got a Symfonisk remote and have been trying to follow all of this, but to no avail. The device is added via zha, but there are only 2 diagnostic entities and no actions. I downloaded the custom quirk from [here](https://github.com/home-assistant/core/issues/71355), and tried that, but nothing changed. Is there a code line I need to update? Shall I provide my diagnostics in this thread? How can I best help/get mine working?

dmulcahey:
There are no entities for remotes. Listen for zha_event or leverage device triggers in automations

TheJulianJES:
@LCMilstein If you don't have proper device automation triggers when selecting the remote as a device trigger in a new automation, then create a new issue on this repo please (and fill out the template).

If you (also) don't get any zha_event events, a new issue would be good too for debugging this.

Issue #1566, opened 2022-05-15
[BUG] Tuya data processing - ZCL frame multiple attributes


rforro:
**Describe the bug**
Lidl Parkside water valve (under development), which is as usual a Tuya device publishes one ZCL frame with two attributes+values. This structure cannot be deserialized correctly using any of two Tuya Manufucturing Clusters. In one case the second one is completely ignored (a warning is issued, see log) in other case both attributes will get value from the second one.

Here is the complete ZCL Frame: `b'\x09\x7B\x02\x01\x0F\x01\x01\x00\x01\x01\x05\x02\x00\x04\x00\x00\x00\x07'` first attribute is onoff `\x01\x01\x00\x01\x01`, second attribute is timer duration `\x05\x02\x00\x04\x00\x00\x00\x07`.

I could narrow the deserializing problem to `class TuyaData` https://github.com/zigpy/zha-device-handlers/blob/b802c1fb2cf2682f9a4722bfb57a1958cad9dad7/zhaquirks/tuya/__init__.py#L205-L210 (I would like to stick with `TuyaNewManufCluster`)

In theory the TuyaData class could during deserializing check if there is some more data to parse, but how to run this process again? I have to idea how to solve it, do any of you can point me to right direction?

Complete log
```
2022-05-15 07:00:23 DEBUG (MainThread) [zigpy.zcl] [0x2F34:1:0xef00] Received ZCL frame: b'\x09\x7B\x02\x01\x0F\x01\x01\x00\x01\x01\x05\x02\x00\x04\x00\x00\x00\x07'
2022-05-15 07:00:23 DEBUG (MainThread) [zigpy.zcl] [0x2F34:1:0xef00] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl(frame_type=<FrameType.CLUSTER_COMMAND: 1>, is_manufacturer_specific=0, is_reply=1, disable_default_response=0, reserved=0, *is_cluster=True, *is_general=False), tsn=123, command_id=2, *is_reply=True)
2022-05-15 07:00:23 DEBUG (MainThread) [zigpy.zcl] [0x2F34:1:0xef00] Decoded ZCL frame: TuyaWaterValveManufCluster:set_data_response(data=TuyaCommand(status=1, tsn=15, dp=1, data=TuyaData(dp_type=<TuyaDPType.BOOL: 1>, function=0, raw=b'\x01', *payload=<Bool.true: 1>)))
[...]

Issue #1452, opened 2022-03-31
[Device support request] Parkside PSBZS A1  frost alert


SzymonZy:
**Is your feature request related to a problem? Please describe.**
At the moment Parkside PSBZS A1  has no reset on frost alert in ZHA. If temp fall below 5 deg device is locked and you have to remove batteries to make it work again.

**Describe the solution you'd like**
Have additional entity on which you can call reset of frost alert. 

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen.**
Details of sniffed traffic can be found on zigbee2mqtt github issues discussion: 
https://github.com/Koenkk/zigbee2mqtt/issues/7695
current id in ZHA:
`{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_htnnfasr",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}`

**Additional context**
Current integration is not complete and not suitable for productive usage in real life.

rforro:
It has been worked on see https://github.com/zigpy/zha-device-handlers/issues/933#issuecomment-1127249857

Issue #933, opened 2021-06-22
[Device Support Request] Parkside Smart Watering Timer


Fenix999:
I would like to add this Parkside Smart Watering Timer

The actual bottom in home assistant works but I don't have any feedback if I press it on the device, also it seems to have a timer function that isn't available.

Device 0xeec8 (84:71:27:ff:fe:d3:3b:e3) joined the network
Device 84:71:27:ff:fe:d3:3b:e3 changed id (0x0b07 => 0xeec8)
[0xeec8:zdo] ZDO request ZDOCmd.Device_annce: [0xEEC8, 84:71:27:ff:fe:d3:3b:e3, 128]
[0xeec8] Extending timeout for 0x72 request
Checking quirks for _TZE200_htnnfasr TS0601 (84:71:27:ff:fe:d3:3b:e3)
Considering <class 'zhaquirks.xbee.xbee_io.XBeeSensor'>
Fail because endpoint list mismatch: {232, 230} {1}
Considering <class 'zhaquirks.xbee.xbee3_io.XBee3Sensor'>
Fail because endpoint list mismatch: {232, 230} {1}
Considering <class 'zhaquirks.smartthings.tag_v4.SmartThingsTagV4'>
Fail because device_type mismatch on at least one endpoint
Considering <class 'zhaquirks.smartthings.multi.SmartthingsMultiPurposeSensor'>
Fail because device_type mismatch on at least one endpoint
Considering <class 'zhaquirks.netvox.z308e3ed.Z308E3ED'>
Fail because device_type mismatch on at least one endpoint
Considering <class 'zhaquirks.gledopto.soposhgu10.SoposhGU10'>
Fail because endpoint list mismatch: {11, 13} {1}
Considering <class 'bellows.zigbee.application.EZSPCoordinator'>
Fail because device_type mismatch on at least one endpoint
device - 61128:84:71:27:ff:fe:d3:3b:e3 entering async_device_initialized - is_new_join: True
device - 61128:84:71:27:ff:fe:d3:3b:e3 has joined the ZHA zigbee network
[61128](TS0601): started configuration
[61128:ZDO](TS0601): 'async_configure' stage succeeded
[0xeec8] Extending timeout for 0x74 request
[61128:1:0x0000]: finished channel configuration
[61128:1:0x0019]: finished channel configuration
[61128:1:0x0006]: bound 'on_off' cluster: Status.SUCCESS
[0xeec8] Extending timeout for 0x76 request
[...]

PR #1574, opened 2022-05-21
Add Device Support: Parkside water valve _TZE200_htnnfasr

rforro:
# Adding support for LIDL PARKSIDE Water Valve

Following is supported by the quirk:

- [x] On/Off state reporting
- [x] Change open timer duration
- [x] Countdown reporting
- [x] Frost lock reporting
- [x] Frost lock reseting

Support not planned:

- [ ] Week Scheduler (use HA automation instead)

This PR closes this https://github.com/zigpy/zha-device-handlers/issues/933 and https://github.com/zigpy/zha-device-handlers/issues/1452

This can be merged before solving following issue https://github.com/zigpy/zha-device-handlers/issues/1566 because quirk will stay as is.

image


rforro:
Please look at the folder structure. I put the quirk in LIDL folder, but pytest is uses regular tuya naming`test_tuya_water_valve.py` (future valves can be put there)

MattWestb:
I have pointing out before that the LIDK folder shall being merged to the tuya for **not** getting one "Z2M" structure then having 15 different names / folders for the more or less the same device,
Like MOES is using nearly all tuya devices and making one folder for them is useless the most device is also sold by other brands and in the end all it tuya devices.

### GREAT WORK DONE  Augsburg !!

javicalle:
I'm not sure now, but I believe that there's no need for the `TuyaNewPowerConfigurationCluster`.
The `TuyaLocalCluster` class is needed for write actions on fake clusters, but being th `PowerConfiguration` a read-only cluster, I'd say it's not needed. 
Could you validate that the `DoublingPowerConfigurationCluster` cluster is enough?

There is also another approach that could be usefull for casting that Tuya spell around the devices.
My suggestion would be:
* new `BewitchedDevice` class (or the name that you consider more appropriate):
```python
class BewitchedDevice(CustomDevice):
    """Cast a magic spell for Tuya devices that needed."""

        """Initialize with task."""

        # cast_tuya_magic_spell(self, tries=3)
        self._init_plug_task = asyncio.create_task(self.spell())

    async def spell(self) -> None:
        """Initialize device so that all endpoints become available."""
        basic_cluster = self.endpoints[1].in_clusters[0]

        _LOGGER.debug("Device class will cast Tuya Magic Spell")
        attr_to_read = [4, 0, 1, 5, 7, 0xFFFE]
        await basic_cluster.read_attributes(attr_to_read)
```
* use the new class wherever it is needed:
```python
class ParksidePSBZS(BewitchedDevice, CustomDevice):
    """LIDL Parkside water without implemented scheduler."""

    signature = {
        MODELS_INFO: [("_TZE200_htnnfasr", "TS0601")],  # HG06875
        ENDPOINTS: {
            # <SimpleDescriptor endpoint=1 profile=260 device_type=0
            .../...
```

Some comments:
* IMHO @MattWestb has earned the right to name the new class
* if I am not wrong, the spell cast is only needed in MCU devices, so I would put the new class in `zhaquirks.tuya.mcu.__init__.py`


What do you think?

javicalle:
Ahhh! and I agree to put the class inside the Tuya folder(`tuya/ts0601_valve.py`). Probably I would keep the `ParksidePSBZS` for now until we can see if it is a particular device or is a generic one.

MattWestb:
@javicalle We was first needing it for the famous TS004F DSR but our hubart friends is using it on many MCU devices that looks need it being casted for getting all function working OK (most temp and humidity sensors with display) and also the updated LIDL power strips is needing it for using all 3 endpoint (its one "normal" Zigbee device).

I think it can being good implanting the tMS (tuya Magic Spell) casting for all MCU devices but its need testing so not braking devices if its changing the device functionality of the device.
Perhaps implanting it in your new MCU and converting all MCU device to it then you is getting them ready for it so having control if somthing is not working OK and can fixing it easy and not forcing all MCU devices at one time now and old implementation) and getting large problems with broken devices for our users.

One side note of tuya MCU devices is the Zigbee module from the factory not having its real manufacture name then its being flashed its using one standard factory ones but then connecting to one tuya MCU its getting one updated manufacture name its storing in the NVM (flash region "user data") and is using it in the Zigbee network (have getting one dumped user data and flash from one tuya TRV that is verifying the data settings of the Zigbee module). And i think the tMS casting is also making some things being sent from the zigbee module to the MCU for getting it initiated OK.

rforro:
> Could you validate that the DoublingPowerConfigurationCluster cluster is enough?

For some reason it wasn't enough, because I've begun with that class. I saw it in the other Irrigation valve PR, but had no success.

javicalle:
> For some reason it wasn't enough, because I've begun with that class. I saw it in the other Irrigation valve PR, but had no success.

Ok, then I would prefer a new without the 'new' part. Just `TuyaPowerConfigurationCluster` or maybe `TuyaDoublePowerConfigurationCluster`.


> I don't like the generic word **valve**, because it can be the Thermostatic Valve or Gas Valve. **ts0601_water_valve** is easier to understand.

I see your point, but IMO the question would be which devices use this implementation and not 'what are these devices used for' (water, gas, ...) Recently, another 'valve' has appeared that can be used for water or gas:
* #1571
* https://zigbee.blakadder.com/IOTPerfect_PF-PM02D-TYZ.html

IMO I would put all the valves in the same class. But it's not something that bothers me too much.

rforro:
> Ok, then I would prefer a new without the 'new' part. Just TuyaPowerConfigurationCluster or maybe TuyaDoublePowerConfigurationCluster.

The original `TuyaPowerConfigurationCluster` is doubling as well, i think its normal in tuya world.
https://github.com/zigpy/zha-device-handlers/blob/c7f2c61fb71b0a4b01e17779644270ef9fc11326/zhaquirks/tuya/__init__.py#L793-L805

MattWestb:
The tuya MCU is sending battery power left as 100% that is OK for there cloud but Zigbee standard is saying it shall being sent 200% then the battery is full.
.
Some "wild ones" Zigbee devices is using 100% (all IKEA sleepers) and need one quirk for fixing it sending 200% to ZHA and its looks nice 100% in the HA GUI and not 50% without the quirk.

rforro:
all proposed changes have been applied, but two things:

Do any of you know how to test following methods:
```
 async def bind(self):
        """
        Bind cluster.
        When adding this device tuya gateway issues factory reset,
        we just need to reset the frost lock, because its state is unknown to us.
        """
        result = await super().bind()
        await self.write_attributes({self.attributes_by_name["frost_lock_reset"].id: 0})
        return result
```
and
```
        """Initialize with task."""
        self._init_device_task = asyncio.create_task(self.spell())

    async def spell(self) -> None:
        """Initialize device so that all endpoints become available."""
        attr_to_read = [4, 0, 1, 5, 7, 0xFFFE]
        basic_cluster = self.endpoints[1].in_clusters[0]
        await basic_cluster.read_attributes(attr_to_read)
```
it can be merged without this, but coverage percentage is lower.

PR #1573, opened 2022-05-20
Remove all unnecessary parameters from `MOVE_` and `STEP_` triggers

puddly:
Using `move_mode`/`step_mode` alone should be enough to tell the events apart, without unnecessarily over-constraining them with things like `rate`.

Should fix #1569.

PR #1578, opened 2022-05-23
Fix TS004F quirk broken by #1573

puddly:
Closes #1334

This PR includes a unit test to ensure that any future automation trigger changes won't cause any new duplicates.


MattWestb:
Thanks Puddly !!
I dont have time testing it but i think its shall working as before if you have done the codding well ;-)

MattWestb:
Was doing one fast test with DA for slow and fast rotating with light controller mode and all is working as before !!!

Thanks **P** !!

PR #1541, opened 2022-04-29
Add support for Aqara P1 motion sensor

dmulcahey:
**TODO:**

- [x] motion sensitivity
- [x] detection interval
- [ ] illuminance update w/o motion

Fixes: #1498

PR #1587, opened 2022-05-26
Small tweaks to Aqara P1 motion sensor

dmulcahey:


zhaquirks/xiaomi/aqara/motion_ac02.py
```diff
@@ -71,11 +71,24 @@ class LocalIlluminanceMeasurementCluster(
 ):
     """Local lluminance measurement cluster."""
 
+    def illuminance_reported(self, value):
+        """Illuminance reported."""
+        if 0 > value or value > 0xFFFF:
+            _LOGGER.debug("Discarding invalid illuminance value: %s", value)
+            return
```
TheJulianJES, line 78: After some discussion in https://github.com/zigpy/zha-device-handlers/pull/1541 and in https://github.com/Koenkk/zigbee2mqtt/issues/12596, it seems like the really high values are only sent when the sensor is without motion in a very dark place. Perhaps it makes sense to explicitly set the light level to 0 when such a high value is received. I've yet to receive them in the bright. (Although I mostly tested with the sensor being in the dark.)

The lowest invalid value I've gotten so far is ``65531``.
```python
if 0 > value or value > 0xFFFA:
    _LOGGER.debug("Received invalid illuminance value: %s. Setting illuminance value to 0", value)
    value = 0
```
Perhaps the limit should be reduced further (0xFFAA/65450), as values that high are basically impossible.
The highest legit value I got so far was ``2484``.

Issue #1590, opened 2022-05-27
[Device Support Request]


elektrinis:
**Is your feature request related to a problem? Please describe.**
There is a new presence detector available, which should solve many automation problems. However it is not supported on zha yet.

**Describe the solution you'd like**
Support for new device

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_ztc6ggyl",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

**Additional context**
https://www.aliexpress.com/item/1005004170513399.html

Hedda:
Nice! This "Tuya Zigbee Human Presence Detector" and other microwave/millimeter wave motion sensors from Tuya for smart human body presence a.k.a. occupancy and vacancy detection (product models: MIR-HE200-ZB, MYQ-ZHFD02, MSA050-Z1, MSA042D-Z1, MSA035-Z1, and MYQ-ZHFD01) are listed for global worldwide sale (unlike competing [Aqara FP1](https://github.com/zigpy/zha-device-handlers/issues/1533) product):

https://expo.tuya.com/product/952001

https://expo.tuya.com/product/1031059

https://expo.tuya.com/product/842246

https://expo.tuya.com/product/902056

[...]

PR #1591, opened 2022-05-27
Added new Tuya millimeter wave sensor signature

TheJulianJES:
Detections seem to be working now: https://discord.com/channels/330944238910963714/551843655682490399/979719516588212264

Possibly fix https://github.com/zigpy/zha-device-handlers/issues/1590

### DIFF
--- a/zhaquirks/tuya/ts0601_motion.py
+++ b/zhaquirks/tuya/ts0601_motion.py
@@ -339,7 +339,7 @@ class NeoMotion(CustomDevice):
 
 
 class MmwRadarMotion(CustomDevice):
-    """NAS-PD07 occupancy sensor."""
+    """Millimeter wave occupancy sensor."""
 
     signature = {
         #  endpoint=1, profile=260, device_type=81, device_version=1,
@@ -348,6 +348,7 @@ class MmwRadarMotion(CustomDevice):
             ("_TZE200_ar0slwnd", "TS0601"),
             ("_TZE200_sfiy5tfs", "TS0601"),
             ("_TZE200_mrf6vtua", "TS0601"),
+            ("_TZE200_ztc6ggyl", "TS0601"),
         ],
         ENDPOINTS: {
             1: {
@@ -385,7 +386,7 @@ class MmwRadarMotion(CustomDevice):
 
 
 class MmwRadarMotionGPP(CustomDevice):
-    """BW-IS3 occupancy sensor."""
+    """Millimeter wave occupancy sensor."""
 
     signature = {
         #  endpoint=1, profile=260, device_type=81, device_version=1,


dmulcahey (approved):

sparkydave1981:
I realise this is merged however it seems that this device comes in a few variations and my one isn't working.
I have basically the same sensor but from MoesHouse which shows as _TS0601
by _TZE200_ikvncluo_ in ZHA.  I don't get the actual sensor entities, only 2 diagnostics.  

Wondering if there is something i can do to get this working as-is or if a new PR is required.

smittong:
Hello.  I have been able to get this working for model code TZE200_ikvncluo by adding this to the second to last quirk in the ts0601_motion.py file. I had to remove and add the device again and get a motion and lux sensor working though.  Only problem is I can't access the configuration options.  This device has configurable parameters of min and max distance and sensitivity.  Is it possible to access these with an updated quirk?  I want to change these values so the cat doesn't get detected.  This is the product I have here:  https://www.ebay.com/itm/224996019306

sparkydave1981:
@smittong, can you please explain how you got it working?  I didn't get anywhere following the links above

smittong:
Hello @sparkydave1981 sure, this is what I did:

I'm using a Sonoff Zigbee 3.0 USB dongle and did this:

1. Made a new directory \config\custom_zha_quirks

2. Updated the configuration.yaml file to include:

zha:

3. added the product code into the ts0601_motion.py file like this: (just added one extra row with the product details)

        MODELS_INFO: [
            ("_TZE200_ar0slwnd", "TS0601"),
            ("_TZE200_sfiy5tfs", "TS0601"),
            ("_TZE200_mrf6vtua", "TS0601"),
            ("_TZE200_ztc6ggyl", "TS0601"),
            ("_TZE200_ikvncluo", "TS0601"),

and added the modified file into the new quirks directory.

I removed and then re-added the presence sensor (but not sure if I need to do that or not.

My problem is I need to adjust the sensitivity and max detect distance.  When I connected it directly with SmartLife app I get configuration options which I really need with zha as well.

This is how it came up for me after doing the above:

image

javicalle:
> This device has configurable parameters of min and max distance and sensitivity. Is it possible to access these with an updated quirk?

Yes it would be possible to access to these config but there is no UI for them.
All 'known' attributes could be accessed in the "admin clusters" option from the device view.
Select the `MmwRadarManufCluster` and you would get and set the attributes values from here.
The problem is that attributes don't have friendly names because attributes couldn't get matched with funtionallity. So there are some `dp_NN` attributes there and one of them could be sensitivity, min and max distance.

Probably the values can be extracted from here:
* https://github.com/Koenkk/zigbee-herdsman-converters/pull/4179/files#diff-64f3be399b00ec5c9aa89849f08b5d84ee33038fb4f28c0d0d55e978a390427b

And if someone can perform some tests and get the correct ones, a friendly name could be setted in the quirk for the attributes (better than these nosense `dp_NN`).

dmulcahey:
We can also add config entities in ZHA for them

sparkydave1981:
> Hello @sparkydave1981 sure, this is what I did:

Thanks so much, worked perfectly 👍

sparkydave1981:
> And if someone can perform some tests and get the correct ones, a friendly name could be setted in the quirk for the attributes

I'm happy to do the testing if you can guide me a little.

This is what I see for that cluster:
image
image


So I take it we need to line those up with these somehow?
```
// TuYa Smart Human Presense Sensor
    tshpsPresenceState: 1,
    tshpscSensitivity: 2,
    tshpsMinimumRange: 3,
    tshpsMaximumRange: 4,
    tshpsTargetDistance: 9,
    tshpsDetectionDelay: 101,
    tshpsFadingTime: 102,
    tshpsIlluminanceLux: 104,
    tshpsCLI: 103, // not recognize
    tshpsSelfTest: 6, // not recognize
```

sparkydave1981:
Since this issue has already been merged I think my investigatino should continue under[ this issue](https://github.com/zigpy/zha-device-handlers/issues/1645).

Issue #1215, opened 2021-12-13
[Device Support Request] Ikea Starkvind Air Purifier


kcleong:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**

When adding the Starkvind in ZHA/home assistant only the fan is recognized. The device also support other sensors but those are not exposed. Would it be possible to also expose the other sensors which are supported by this device; pm25, air_quality, led_enable, child_lock, replace_filter.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**

```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4476, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0007",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0202",
        "0xfc57",
        "0xfc7d"
      ],
      "out_clusters": [
        "0x0019",
        "0x0400",
        "0x042a"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "IKEA of Sweden",
  "model": "STARKVIND Air purifier",
  "class": "zigpy.device.Device"
}
```
```
New device 0x4d79 (b4:e3:f9:ff:fe:2b:07:3f) joined the network
[0x4d79] Scheduling initialization
[...]

PR #1619, opened 2022-06-13
Device handler for IKEA of Sweden STARKVIND Air purifier

arnemauer:
Quirk for the IKEA Starkvind Air purifier! #1215

**Supports:**
- All fan modes: Off, Auto, 1, 1.5, 2, 2.5, 3, 3.5, 4, 4.5 & 5
- Air quality PM 2.5
- Device run time
- Filter run time (of the current placed filter)
- Filter status (need to replace filter) corresponds with the replace filter light on the Starkvind
- Child Lock (disables all controles)  corresponds with the child lock light on the Starkvind
- Disable Leds (disables all leds on the Starkvind)
- Filter lifetime (you can set the lifetime of a filter, default is 6 months = 259200 minuts). If filter run time exceeds this value, Filter status is true (and you need to replace the filter)

**This quirk also needs some adjustments in the ZHA code! 
See this pull request: https://github.com/home-assistant/core/pull/73450**

### DIFF
--- a/zhaquirks/ikea/starkvind.py
+++ b/zhaquirks/ikea/starkvind.py
@@ -1,5 +1,6 @@
 """Device handler for IKEA of Sweden STARKVIND Air purifier."""
 import asyncio
+
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
 import zigpy.types as t
@@ -13,10 +14,8 @@ from zigpy.zcl.clusters.general import (
     Scenes,
 )
 from zigpy.zcl.clusters.hvac import Fan
-from zigpy.zcl.clusters.measurement import (
-    PM25,
-    IlluminanceMeasurement,
-)
+from zigpy.zcl.clusters.measurement import PM25, IlluminanceMeasurement
+
 from zhaquirks import Bus
 from zhaquirks.const import (
     DEVICE_TYPE,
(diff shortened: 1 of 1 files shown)


MattWestb:
I think the default reporting rate is little over kill for the air purifier and PM25 cluster that can spamming the network if the sensors is updating oft but it must being fixed in the ZHA core and cant being made in the quirk.

### GREAT WORK DONE !!!

@dmulcahey do you knowing where the default reporting is configured in ZHA core so the device is not spamming the network for the 3 new cluster ?

dmulcahey:
Each channel (basically a handler for a cluster) defines the report configs per attribute that it cares about. For example: https://github.com/home-assistant/core/blob/c660fae8d8dfd190186fc404ca77cc1bc4c0cda3/homeassistant/components/zha/core/channels/general.py#L49

arnemauer:
@MattWestb @dmulcahey Hmm, yeah it will send every minute a attribute report. Maybe change it to every 5 minuts or more? What is an usable interval?

See this pull request for a IkeaAirPurifier Cluster:
https://github.com/home-assistant/core/blob/12735b874e4d57c8f816cb8c7991d92e066f90e1/homeassistant/components/zha/core/channels/manufacturerspecific.py#L122

MattWestb:
I think 1 in the minute is OK then its main powered.
I was thinking the runtime counter was updated every second but then its every minute.
Also the PM25 is "filtered" so not sending so oft (Vindriktning is pulling the sensor 7 times every 20 seconds so raw data is 21 / min if not being filtered) so its also OK and getting nice graph in HA GUI :-))).

Then letting it and getting it merged !!

arnemauer:
* Added a test for de ikea air purifier.
* Instead of using the fan cluster, i've created a Ikea Air Purifier channel/cluster
* Updated sensors to work with the new channel/cluster
* Also updated the quirk to work with this changes :)

This quirk needs to be merged first before merging (https://github.com/home-assistant/core/pull/73450) to pass the new test.

PR #1639, opened 2022-06-29
Refactoring Tuya Cover inverting

datebo:
This is a continuation of the problem discussed in my earlier PR: #1549. And I expect this will also resolve issue #1399

First of I would welcome some extra people testing my code. I tested it on my device and this works fine, but this could effect more devices than just mine.

What did I do? I made some changes to move the definition of which devices should be inverted or require inverted controls away from the __init__ file. This has two advantages:

1. The code is cleaner in my opinion. Because now the definition of the device is all done in the ts0601_cover file and no longer in the __init__ file.
2. I can now invert the position on my device, even though a device with the same manufacturer/model doesn't need to be inverted. But since they have different signatures they can be separately defined.

I welcome any feedback because I am relatively new to this.


datebo:
>Is it also possible moving the key definition from the INIT to the device quirk ? 

What do you mean with "key definitions"? If you mean the definition of the open/close/stop command, then I indeed moved that out of the INIT too. It defaults to the most common setup for the tuya covers but can be overridden in the ts0601_cover file.

javicalle:
I also prefer the idea of having the specific configuration of the shutter all together in the same file (in addition to the advantages it brings).
I don't know the reason that led to the current implementation, but I don't see a problem with the new approach.

My comments to the new implementation:
* put the new attributes at the beginning of the class (before the `signature` part). I think that way it is easier to detect the special cases.


And now a personal thought of mine:

_I'm having a hard time understanding the use of the `TUYA_COVER_INVERTED_BY_DEFAULT` attribute. I understand the implementation and its use, but I keep thinking that it is a property that depends on how the person implementing the quirk has his cover installed and I suspect that its use is not compatible if a remote control is used in addition to HA control.
Be careful, I am not saying that the functionality of reversing the rotation of the blind does not make sense. I think the `motor_direction` and `cover_inverted` attributes are necessary, but I think the use of the `tuya_cover_inverted_by_default` attribute is a shortcut to avoid having to assign the proper value to those variables._

@datebo I'm not arguing about your code I'm speaking about the already existing `TUYA_COVER_INVERTED_BY_DEFAULT` attribute and use.

datebo:
> * put the new attributes at the beginning of the class (before the `signature` part). I think that way it is easier to detect the special cases.

That makes sense, I'll find some time later today to adjust this. Thanks for the suggestion!

> _I'm having a hard time understanding the use of the `TUYA_COVER_INVERTED_BY_DEFAULT` attribute. I understand the implementation and its use, but I keep thinking that it is a property that depends on how the person implementing the quirk has his cover installed and I suspect that its use is not compatible if a remote control is used in addition to HA control. Be careful, I am not saying that the functionality of reversing the rotation of the blind does not make sense. I think the `motor_direction` and `cover_inverted` attributes are necessary, but I think the use of the `tuya_cover_inverted_by_default` attribute is a shortcut to avoid having to assign the proper value to those variables._

The problem here is that the `motor_direction` and `cover_inverted` attributes are not implemented in my device. And I assume other devices have the same issue. Because my device doesn't support these attributes I can't verify, but by reading the code my assumption is that `motor_direction` is used to invert the motor/controls and `cover_inverted` reports whether it is actually inverted. Since my device (and I presume others as well) don't support these attributes we have to either invert the controls or the position in the code. Because when inverting the motor/controls on the device itself it also inverts the way that it reports the position, resulting in always one or the other being inverted if we don't correct for this in the code.

javicalle (approved):
LGTM

javicalle (changes requested):
There is no neeed of `PROFILE_ID: zha.PROFILE_ID,` in the `replacement` part.

datebo:
Those lines were already in there, they are not added by me. I don't know what they do exactly and since it is also not in the quirk of my device I can't test the consequences of removing those lines.

I don't mind removing them for you though, but I can't take any responsibility for it. Just give me a minute and I'll remove them.

datebo:
> There is no neeed of `PROFILE_ID: zha.PROFILE_ID,` in the `replacement` part.

@javicalle I went through the code and found a total of 7 instead of 4, so I removed all of them. I presume that is okay.

MattWestb:
@datebo tuya is messing many devices with strange functions as the TS130F smart curtain switches some is having calibrating with time and other is having from the motor how many turns the motor have being doing for max open and close.
So very likely its the same with the TS0601 devices have different function implanted and some is working and other not.

Great that you is thinking or other device for not braking the functionality of device and i hope some user with other devices can helping testing more for getting all working well.+

### Great work done !!

javicalle:
> I don't mind removing them for you though, but I can't take any responsibility for it.

I understand your concern (I have been in your shoes many times), but don't worry, I can take responsibility and any incidence of that change.

datebo:
Thanks for all your support @javicalle !

javicalle:
There's one thing I'd like to try if you're up for it.

If I'm not mistaken, you have a remote control that allows you to change the direction of the motor. 
I know that the command is RF and we are not going to see traces in HA when you operate it.
But it is possible that when you interact with the cover it reports its new state to HA. If so, it may be that there is something valuable in that information when you change the `motor_direction` with the remote. 

Could you catch and attach the logs from the remote operation report?

javicalle:
Thanks @datebo, I think that something can be done with that info, but will be in another issue.

Thanks for your work, it's a pleasure to contribute with helpful people like you.

javicalle:
### For future references

I believe that there is Tuya custom command to activate the 'inverted' mode.

The previous logs only shows the report for the command call (the command has been called with a RF remote). The device reports to the `DP=6` with `value=0` both times, so there's no report (in DP 6) for the motor direction, just for the command trigered. 

Here is the detail:

* `command_id=1` is the Tuya 'get_data' command
```
Decoded ZCL frame header: 
  ZCLHeader(
    frame_control=FrameControl(frame_type=<FrameType.CLUSTER_COMMAND: 1>, 
    is_manufacturer_specific=0, 
    is_reply=1, 
    disable_default_response=0, 
    reserved=0, 
    *is_cluster=True, 
    *is_general=False), 
  tsn=17, 
  command_id=1, 
  *is_reply=True)
```

* the 'classic' `TuyaManufCluster` decodes the payload in a diferent way
  * that `command_id=256` corresponds to `DP=6, DataType=VALUE`
  * the `data=[1, 0]` is telling us that the reported value is `0` always no matter which way is the motor direction
```
TuyaManufacturerWindowCover:get_data(
  param=Command(
    status=0, 
    tsn=0,
    command_id=261,
    function=0,
    data=[1, 0]
  )
)
```


So I still believe that there must be an unknown Tuya command to toggle the motor direction for Tuya covers but with the current info I have not been able to find it.

Note: it seems that `herdsman-converters` is using DP 5 (`motor_direction`) to reverse the cover, but we haven't get it working in ZHA.

Issue #1655, opened 2022-07-15
[Device Support Request] Aqara MAEU01 Plug


TheJulianJES:
**Is your feature request related to a problem? Please describe.**
The Aqara MAEU01 is somewhat supported. There are multiple firmware revisions however. Newer firmware doesn't support ``active_power`` correctly.
If ``OppleMode`` is set to ``0``, ``active_power`` always reads as ``0``.
If ``OppleMode`` is set to ``1``, the attribute simply reads as ``None``.

For Aqara E1 sensors, the following patch is required for Z-Stack: https://github.com/Koenkk/Z-Stack-firmware/blob/0cea4d898afaa26ec1fe8550fd6cd6469b332ee4/coordinator/Z-Stack_3.x.0/firmware.patch#L2025-L2038
Bellows has a similar patch for EZSP coordinators.

However, without this patch, ``active_power`` actually returns the correct power consumption (if ``OppleMode`` is ``0``).
So either Z-Stack firmware could be modified to return normal manufacturer codes for the plug or Xiaomi mains-powered devices maybe? Or an alternative solution could be figured out.
I've gotten this working somewhat. The plug behaves similar to the Aqara MAUS01 plug (if Xiaomi manufacturer codes were returned during initial pairing).
However, implementing the changes seem to break MAEU01 plugs with older firmware (and the firmware update link wasn't found yet).

I don't think it's possible to match a quirk depending on ``Basic``cluster/``app_version`` attribute though?

**Describe the solution you'd like**
Get proper power consumption for the plug. I'm currently working on this but just writing up all my findings here.

<details>

<summary>Device signature</summary>

Note: This is already a somewhat modified signature and a custom quirk. The plug can give a total of three different signatures. Even if an endpoint with an AnalogCluster doesn't show in the signature, it still works...
```yaml
{
[...]

Issue #1654, opened 2022-07-15
[Device Support Request] lumi.plug.maus01 (ZNCZ12LM)


ghost:
I have seen similar devices listed as issues here but not quite this exact model number ZNCZ12LM.
It is an [Aqara Smart Plug](https://www.amazon.com/gp/aw/d/B07CJ2MM6Z?psc=1&ref=ppx_pop_mob_b_asin_title). It will join to the network and can be controlled but much of the energy reporting metrics are showing as unavailable

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4447, maximum_buffer_size=127, maximum_incoming_transfer_size=100, server_mask=0, maximum_outgoing_transfer_size=100, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0002",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x000a",
        "0x0010",
        "0x0b04"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0009",
      "in_clusters": [
        "0x000c"
      ],
      "out_clusters": [
        "0x0004",
        "0x000c"
      ]
    },
    "3": {
      "profile_id": 260,
      "device_type": "0x0053",
      "in_clusters": [
        "0x000c"
      ],
      "out_clusters": [
        "0x000c"
      ]
    },
    "100": {
      "profile_id": 260,
      "device_type": "0x0107",
      "in_clusters": [
        "0x000f"
      ],
      "out_clusters": [
        "0x0004",
[...]

PR #1211, opened 2021-12-09
Update lumi.plug to report total energy

Brazen00:
I've updated the xiaomi quirk for lumi.plug and added a metering cluster using the data from the second analog cluster. It works with my plug in my Docker install of HA.

The only issue that I'm still having is that the second analog sensor and the current_summation_delivered sensor in HA only update when I manually request the present value from the analog sensor via "Manage Clusters". This is different from the first analog sensor (power) and the power sensor from ElectricalMeasurementCluster which updates whenever the power drawn by the plug changes.

Is there a way of periodically polling the analog sensor within HA?

PR #1656, opened 2022-07-17
Fix power measurement for Aqara MAEU01 plugs with newer firmware

TheJulianJES:
## Notes
This quirk will allow for the power outage memory to be enabled with a config entity in HA with:
- https://github.com/home-assistant/core/pull/80444

This quirk will probably only work if:
- ~~the MAEU01 plug has ``app_version = 41`` (couldn't confirm with older firmware versions) *AND*~~ (should work with all firmware versions now)
- the coordinator sends out Xiaomi manufacturer codes when pairing (likely only an issue for ``maeu01`` plugs with new firmware and as the sensors of those weren't working at all before, it's not really an issue. Older plugs seem to use OppleMode instead of needing Xiaomi manufacturer codes at pairing to determine whether or not to send the special Xiaomi messages)

This quirk (as of right now) will break power measuring in other MAEU01 plugs if:
- ~~the plug is running older firmware *OR*~~
- the coordinator doesn't send out Xiaomi manufacturer codes when pairing (likely fixed for older plugs by turning on OppleMode)

If anyone needs the quirk right now, you can install this as a "custom quirk". See the instructions at the bottom.

Would fix https://github.com/zigpy/zha-device-handlers/issues/1655

As of right now, the newest firmware for the MAEU01 can be found here: https://github.com/Koenkk/zigbee-OTA/pull/140 (41)
Older firmware is available here: https://github.com/Koenkk/zigbee-OTA/pull/159 (32)
For downgrading your firmware, refer to the "Plugs randomly toggling on/off when using IKEA remotes" section of this post.

It's possible that this also fixes an issue with MAUS01 plugs (https://github.com/zigpy/zha-device-handlers/issues/1654) where sensors aren't created because there's no default value.

## Changes
- add new signatures (yes, even if the plug isn't reporting an endpoint with an AnalogInput cluster, it's there)
- Improve custom Xiaomi ElectricalMeasurementCluster:
- - add lumi.plug.maeu01
- - put default values for sensors
- - switch to report voltage on ``rms_voltage`` instead of wrongly used ``instantaneous_voltage`` attribute (sensor isn't created for instantaneous voltage (and it's wrong, as we deal with AC voltage here and the sensor always reports the "max" of the sin wave)
- - set ac_power_divisor to 10 and multiple power_reported value to workaround a rounding error in ZHA (initial attribute report gets correctly rounded: 100.9 => 101 but during "polling" of the fake DataCluster, it only reads 100 which causes a weird jump in graphs. Another reason we want this: it also adds one decimal point. See: for the rounding issue, see: https://github.com/zigpy/zigpy/issues/1041)
- use the custom Xiaomi ElectricalMeasurementCluster for MAEU01 plugs (previously used for MMEU01 and MAUS01 plugs)
- renamed some things
- now set OppleMode to 1 (enabled), as this sends the custom Xiaomi attribute reports on plugs with older firmware when enabled. Newer firmware versions don't have this attribute and always have this "enabled" (will require re-pairing to change "automatically" for older plugs)
- add power outage memory
- ...

## TODO
- ~~make this work without breaking power measurement on older plugs paired without Xiaomi manufacturer codes/older firmware (or force everyone to update firmware?) (how to determine in which state we are?)~~
- ~~provide custom quirk download links~~
- ~~also get the alternative signatures of the older MMEU01 plug~~
- search ZHA for where that rounding/cutoff issue happens (currently tracking in https://github.com/zigpy/zigpy/issues/1041))
- figure out if AnalogInput cluster on endpoint 31 is used for anything (at least US plugs use it, maybe EU too. Do the US plugs, like the EU plugs, also send total consumption in the special Xiaomi message?)
- ~~It's possible that ``current_summ_delivered`` needs to be polled, as it doesn't seem to update (Reference:  https://github.com/zigpy/zha-device-handlers/pull/1211)~~
- ~~check if we still need to write ``OppleMode`` to ``0``. ``1`` might be better? (default for ``MMEU01`` plugs with older firmware. For those, this quirk seems to work fine with both. ``OppleMode`` seems to be removed on newer firmware versions).~~ Edit: We want ``OppleMode`` set to ``1`` if it exists, so special Xiaomi attribute reports are sent
- ~~figure out if https://github.com/Koenkk/zigbee2mqtt/issues/13903 can be solved. (IKEA remotes routing through these plugs trigger toggle)~~ Edit: by downgrading firmware (refer to the "Plugs randomly toggling on/off when using IKEA remotes" section below)
- ...?

## Testing
1. Completely remove the device from HA
2. Do the following:

In the ``configuration.yaml``, add this:
```yaml
zha:
```
In the ``/config`` directory, create a folder called ``zha_custom_quirks``, put the ``tjj_custom_aqara_plugs`` folder from this ZIP file in it: [zha_custom_quirks.zip](https://github.com/zigpy/zha-device-handlers/files/9552533/zha_custom_quirks.zip)

In the end, the folder structure should look like this:
``/config/zha_custom_quirks/tjj_custom_aqara_plugs``
The last folder should contain four files:
``__init__.py, plug.py, plug_eu.py, plug_maus01.py``

4. Restart Home Assistant
5. Re-pair the device and see if the power measurement sensor works (it's likely others (like power_factor get created which can be disabled, as they'll never work))
6. Also see if "total power consumption" works now (only EU models atm).
7. Even if the power measurement sensor works now, please restart Home Assistant and let me know if any additional useless sensors are created.

## Plugs randomly toggling on/off when using IKEA remotes
There's an issue in the v41 firmware on the plug where IKEA remotes toggle the plug (because the plugs always listens to messages sent for group 0). This is somewhat fixed in current versions of ``zhaquirks`` and this custom quirk by making the plug explicitly leave group 0. However, the plug still reacts when messages from those remotes are routed through it.
See the following comment for further explanation and a link for how to downgrade your plug to a "non-broken version": https://github.com/zigpy/zha-device-handlers/pull/1656#issuecomment-1244750465


zhaquirks/xiaomi/__init__.py
```diff
@@ -563,9 +564,17 @@ def __init__(self, *args, **kwargs):
...
+            self._update_attribute(self.POWER_ID, 0)
+        if self.VOLTAGE_ID not in self._attr_cache:
+            self._update_attribute(self.VOLTAGE_ID, 0)
+        if self.CONSUMPTION_ID not in self._attr_cache:
+            self._update_attribute(self.CONSUMPTION_ID, 0)
+
     def power_reported(self, value):
         """Power reported."""
-        self._update_attribute(self.POWER_ID, value)
+        self._update_attribute(self.POWER_ID, value * 10)
```
pgrondek, line 577: This change made my another aqara outlet (mmeu01 square one) report 10* more usage than normal

TheJulianJES: That should be fixed by either re-pairing the device or reading the "ac_power_divisor" attribute on the ElectricalMeasurement cluster through ZHA -> Devices -> Plug -> Manage Clusters -> ElectricalMeasurement -> ac_power_divisor -> read attribute.

pgrondek: Reparing is workaround, I don't think this is good user experience to ask user to re-pair working device,
I think the solution is to remove this multiplication and change ac_power_divisor back to 1, I've changed it locally and both devices shows power consumption correctly for me

TheJulianJES: I agree. If possible, the attribute should be re-read automatically. There is another issue with it being left on 1 (different rounding when values like 1.9W are reported which causes a jump from 2W to 1W after some time with unchanged consumption).

If consumption varies a lot when low, it's also nice to see values in between 0W and 1W (and it not constantly jumping around).

TheJulianJES:
I've updated the [custom quirk](https://github.com/zigpy/zha-device-handlers/files/9466319/zha_custom_quirks.zip) and total power consumption *might* work after some time now.
(EU models only atm: ``maeu01`` should work. ``mmeu01`` could potentially also work(?))

fmiermont:
I've updated the quirks and let you know if the total power consumption is now working !

One quick question, the plug has the hability to resume power on power loss. This is not the default behaviour (does not resume power by default). With Z2M, we had access to this settings directly. It's also described on the device page : 
Power_outage_memory (binary)

Enable/disable the power outage memory, this recovers the on/off mode after power failure. Value can be found in the published state on the power_outage_memory property. To read (/get) the value publish a message to topic zigbee2mqtt/FRIENDLY_NAME/get with payload {"power_outage_memory": ""}. To write (/set) a value publish a message to topic zigbee2mqtt/FRIENDLY_NAME/set with payload {"power_outage_memory": NEW_VALUE}. If value equals true power_outage_memory is ON, if false OFF.
[#](https://www.zigbee2mqtt.io/devices/SP-EUC01.html#voltage-numeric)
Voltage (numeric)

With ZHA, I cannot find it, nor using "Set Cluster" options. Do you have any idea how should I do that ?

Regards,

TheJulianJES:
I updated [the quirks](https://github.com/zigpy/zha-device-handlers/files/9492225/zha_custom_quirks.zip). (No need to re-pair when coming from the previous version).

Under ``Manage Clusters`` -> ``OppleCluster``, you'll now find ``power_outage_memory``.
When writing the attribute, use ``0`` for ``default off`` or ``1`` for ``restore``.

fmiermont:
Thanks ! 
But it does not seems to do anything. Getting the value returns "Bool.false". If I set this to 1 and then try to get value, it returns "none".
If I close/ reopen Manage Cluster and get the value, returns "Bool.false" again ?

Smartenergysummation has started to returns value !

TheJulianJES:
Mhm, that's weird. It works on the plugs I tested it with. Reading the attribute before re-opening the window also shouldn't be able to get another value (None) at all. Can you turn on ZHA debug logs and see if there's anything in the logs when reading or writing the attribute?

Also, make sure you're not using the lower text field with "manufacturer code override". That should always stay empty.

So only use the upper one with writing 0 or 1. When reading the attribute, it will show in the same text field (but Bool.false/Bool.true).

fmiermont:
Ok, now I feel dumb, i've never used custer attributes to write attributes so I was using the "Manufacturer code Override" because the "value" was filed with 'Bool.false'.

Works like a charm using the correct textfield !

Jojonintendo:
Looking good here too on my 2 lumi.plug.maeu01. Only one useless sensor is created, but I think you already mentioned that (sensor.lumi_lumi_plug_maeu01_electricalmeasurementpowerfactor_2). The rest seems to work fine and update as well, many thanks!

TheJulianJES:
@fmiermont It's not really a user-friendly way. If/when this quirk gets merged officially, an option could be added to HA itself, so the clusters menu doesn't even need to be used.

oskarannas:
My two maeu01 plugs also works with "Electricalmeasurement", "Smartenergysummation" and the new "power_outage_memory" setting. Thanks! :)

fmiermont:
@TheJulianJES yes I know this is not the user friendly way to set this parameter but this is also the basics of "advanced" zwave settings and I'm usually familiar with this (but i'm not using HA for a long time). 
Anyway that works flawlessly for me now, perfect !

xam-ps:
Thanks a lot for your work! It's working for me as well 👍 (MAEU01, app_version = 41)

TheJulianJES:
FYI for anyone that is using IKEA remotes and firmware v41 on the ``MAEU01``/``SP-EUC01`` plug:
There's an issue in the v41 firmware version where the plug toggles on/off when IKEA remotes are used with their messages routed through the plug. 
Before https://github.com/zigpy/zha-device-handlers/pull/1706 (changes are also included in this modified quirk), the plug on v41 always toggled if an IKEA remote was used **anywhere** on the network. With those changes and v41 plugs, it'll now only happen if the messages are routed through that plug. (Again, an issue in Aqara's v41 firmware).

The plug includes a rudimentary downgrade check that can be bypassed. I've successfully downgraded one of the plugs I'm testing with from v41 to v32. Be aware that you could brick your plug!
Instructions for downgrading are here: https://github.com/Koenkk/zigbee2mqtt/issues/13903#issuecomment-1244720072
I'd only recommend trying the downgrade if you're running into this issue. (v32 doesn't seem to have the same issue as v41)

The [quirk is now also updated](https://github.com/zigpy/zha-device-handlers/files/9552533/zha_custom_quirks.zip) to support ``maeu01`` plugs with older v32 firmware. This is done by explicitly enabling "OppleMode" now. Newer firmware versions don't have this attribute anymore and it's basically always enabled. So, the plugs send special Xiaomi attribute reports. Since this modified quirk properly works with those now, we want to enable "OppleMode" if that attribute exists.
(Note: re-pairing and/or updating the custom quirk might be required in some cases when the sensors don't update)

mvaessen:
I tested the latest quirk with 3x v41 plugs, all work great. Thanks for the quirk!

muddenhed:
Hi

Just wanted to give my input on this, i have mixed luck with the original shipped in 2022.9.7 of Home assistant and this one, in the original one the connection to my plugs with .41 firmware is stable, but no power readings what so ever, in this this one i get power readings but the plugs constantly goes unavailable but i get power readings from them.

i have 3 with .41 firmare and 2 with .32 and the old ones work with either quirks.

I have removed one of the plugs after i added this and reinstalled it, same problem.

Just wanted you to know that it does not work for every .41 plug sadly.

If there is something specific you want me to post about my plugs i will gladly do so.

Regards / Marcus

TheJulianJES:
@muddenhed The quirk can't affect the connection stability. It only changes how the messages are parsed when coming from the plug. With these plugs, I haven't encountered any stability issues so far. What coordinator/stick are you using?

(Also FYI, you can upgrade or downgrade all plugs to whatever firmware version you want (32 or 41))

muddenhed:
Hmm, intresting, wieard that it starts acting like that then.
I'm using ConBee II on a Raspebby PI, I will try to downgrade one of them if i can.

Sorry for wasting your time on this, thought it containted some connection polling code in this, but know better now.

Thanks.
/Marcus

TheJulianJES:
> Sorry for wasting your time on this, thought it containted some connection polling code in this, but know better now.

You didn't! Thanks for all the feedback.

> I'm using ConBee II on a Raspebby PI

Mhm, I've heard of quite a few issues with ConBee sticks now. You could try to upgrade the firmware on the stick and see if that helps.
Also make sure that the ConBee is on a USB extension and not near any USB 3 devices (like a USB-SSD plugged into the Pi, as that will cause massive interference in the 2,4 GHz range).

> downgrade one of them if i can.

Process should basically be to download the modified firmware image, put the image in a zha_ota folder that's registered in teh configuration.yaml, send a cluster command to notify the device of the OTA, wait until the device updates, delete the image from the zha_ota folder and restart HA (so the device doesn't "upgrade"(downgrade) again).
Let me know if you need help with this.

muddenhed:
> > Sorry for wasting your time on this, thought it containted some connection polling code in this, but know better now.
> 
> You didn't! Thanks for all the feedback.
> 
> > I'm using ConBee II on a Raspebby PI
> 
> Mhm, I've heard of quite a few issues with ConBee sticks now. You could try to upgrade the firmware on the stick and see if that helps. Also make sure that the ConBee is on a USB extension and not near any USB 3 devices (like a USB-SSD plugged into the Pi, as that will cause massive interference in the 2,4 GHz range).
> 
> > downgrade one of them if i can.
> 
> Process should basically be to download the modified firmware image, put the image in a zha_ota folder that's registered in the configuration.yaml, send a cluster command to notify the device of the OTA, wait until the device updates, delete the image from the zha_ota folder and restart HA (so the device doesn't "upgrade"(downgrade) again). Let me know if you need help with this.

Thanks for the tips, i will check it out, i acutally manage to do a workaround by creating a timed automation that triggers the indentify button on them every 2 minutes and they have been far more stable now, no dropouts since then(6h+).

But it is better to have them working without the automation. :-)

/Marcus

Animajosser:
Just wanted to say, I have a few maeu01 (I think firmware 0.0.0_0032) connected to a Conbee II (firmware 0x26780700) and I would love to help test if necessary. Up to now the quirk seems to work fine. :+1:

zhaquirks/xiaomi/aqara/plug_eu.py
```diff
@@ -48,19 +48,44 @@
...
+        await self.write_attributes(self.attr_config, manufacturer=OPPLE_MFG_CODE)
+        return result
+
+
 class PlugMMEU01(XiaomiCustomDevice):
     """lumi.plug.mmeu01 plug."""
 
     def __init__(self, *args, **kwargs):
         """Init."""
+        asyncio.create_task(remove_from_ep(self))
```
dmulcahey, line 88: we should probably use `self.create_catching_task` here

TheJulianJES:
The quirk is now included starting with Home Assistant Core **2022.10.5** (and later).
It's recommended to **uninstall the custom quirk** before you upgrade to that version.
If you've already installed 2022.10.5, then just delete the custom quirk, reboot HA, and the "official quirk" should be used by ZHA.

If you didn't install the custom quirk and have issues with your Aqara MAEU01 or MMEU01 plugs now, please remove the device from ZHA and re-pair it. If you still have issues then, create a new issue in this repo and tag me.

FSHelgeland:
@TheJulianJES

I just got three of these plugs, first tried to install one of them this evening. It did not work so I got searching for why, and found this thread. Now I have updated to: 
Home Assistant 2022.10.5
Supervisor 2022.10.0
Operating System 9.3
Grensesnitt 20221010.0 – latest

But it still do not function, I have tried to install one of the other ones I got but the result is still the same. I got ZHA and the Raspbee II, where do I go wrong?

Best regards Freddy.

Maggi0r:
I'm facing the same problem. I have 2 switches of this type. One is recognized, the other is not recognized or already recognized but I don't get any values displayed. The other connector works with no problems. My HA is 2202.10.5, everything is up to date. Greetings, Maggi

FSHelgeland:
This is actually really frustrating. Have spend hours searching and I do have to conclude that my knowledge isnt good enough. Hope this is posible to solve then.

FSHelgeland:
@TheJulianJES any idea how I can solve this? Today I updated the firmware on my Raspbee II, but still none of my plugs reporting any measuring, only function is on off. 
Home Assistant 2022.11.1
Supervisor 2022.10.2
Operating System 9.3
Grensesnitt 20221102.1 – latest

TheJulianJES:
> If you didn't install the custom quirk and have issues with your Aqara MAEU01 or MMEU01 plugs now, please **_remove the device from ZHA and re-pair it_**. If you still have issues then, **_create a new issue in this repo and tag me_**.

@FSHelgeland So, you **re-paired the plug after updating your Raspbee firmware** and it still doesn't work?
If so, then please create another issue: https://github.com/zigpy/zha-device-handlers/issues/new/choose
Make sure to fill out the template and tag me (@TheJulianJES) there.

sipy:
Hello @TheJulianJES , I have Home Assistant 2022.11.1 as well and only the switch is working. Then I tried installing the quirk but it does not work anyway...

TheJulianJES:
> So, you **re-paired the plug after updating your Raspbee firmware** and it still doesn't work?
> If so, then please create another issue: https://github.com/zigpy/zha-device-handlers/issues/new/choose
> Make sure to fill out the template and tag me (@TheJulianJES) there.

@sipy Did you do the above?

Also, remove **any custom quirks** for the plug. They're all in Home Assistant 2022.10.5 or later (so also in 2022.11.1).
After having removed all custom quirks, remove the device from ZHA and then pair it again.

sipy:
Yep @TheJulianJES , tried them all and still not working. I am now trying to downgrade the Aqara firmware from V41 to V32. I'll let you know if it works.

TheJulianJES:
Also, like I mentioned, please create a new issue here and fill-out the template: https://github.com/zigpy/zha-device-handlers/issues/new/choose

FSHelgeland:
@TheJulianJES Shame on me... Like my students I told you I had done something that I havent. But no I have updated my Raspbee II firmware to the latest one, and now I am able to read at least the active power, and also the correct device temperature. Cant read the voltage, but I dont need it, as long as I can read the active power I am satisfied. Thank you for solving this.

TheJulianJES:
@FSHelgeland Nice, glad it works for you now!

(Voltage always takes some time to appear after initial pairing due to how Xiaomi's messages work and I'm also not sure how accurate it is lol)

sipy:
Downgrading the Aqara firmware from V41 to V32 following this guide https://github.com/Koenkk/zigbee2mqtt/issues/13903#issuecomment-1244720072 solved the issue. Thank you.

TheJulianJES:
@sipy Yeah, that's works around the issue you were having. It's the recommended firmware anyway, but, so far, it looks like your Raspbee firmware didn't correctly update if the older plug firmware works and the newer one doesn't. The newer firmware versions requires Xiaomi manufacturing codes to be sent using pairing. That's implemented in deCONZ firmwares since 1 or 2 years.

I'm glad that it finally works for you.

sipy:
Thanks @TheJulianJES , in order to be sure that I will face no issue in the future I should disable any updates of the Aqara plug firmware, right? How can I ensure that this will not happen?

Thanks

TheJulianJES:
@sipy Just remove all Xiaomi plug related OTA files from your ZHA OTA directory and don't put any new ones in there and you're set.

MrGlad8:
I just bought  a new Aqara Smartplug and it doesnt show up any values! Tried the quirk here but it didnt work. Any workaround?

zhaquirks/xiaomi/__init__.py
hakong: The power measurement on LLKZMK11LM (lumi.relay.c2acn01) is also wrong after this change. It's showing 100W for a 10W device.

TheJulianJES: @hakong No, check the AC power divisor: ``0x0605``. It's set to ``10`` instead of ``1`` now. Both of these changes were made so that there's one decimal.

PR #1664, opened 2022-07-28
Remove Yooksmart quirk

### LINKED ISSUE #1306: [BUG] New yooksmart D10110 inverted with quirk
**Describe the bug**
I purchased a new yooksmart D10110 cover and paired with home assistant. The controls
seemed inverted and I had to move the bar twice in order to get it to move. I read reports
in the past with the suggestion to unpair and pair again, tried multiple times with no luck.
So I disabled the quirk (apologies for the brute force: moved the file to a different directory
and reloaded) and it works now. For completeness:
Before:
- buttons up and down wouldn't work
- available button would be inverted (e.g.: cover was all the way down and the down button was enabled)
- in order to control the cover I'd move the progress bar all the way to 0 or to 100 then the opposite in order to work
After:
- buttons up and down work
- enabled button matches the direction of the cover: if open, it shows down button enabled

**To Reproduce**
Behavior is consistent across multiple pair/unpair cycles and full home assistant instance restarts

**Additional context**
Something that is possible, since the cover is new, is that they corrected the behavior in their firmware
and the quirk isn't needed anymore.
This device has:  Firmware: 0x10013001

I can provide any debugging necessary. I'm using homeassistant official virtual machine image and keeping
it up to date.

Editted: formatting

jonnylangefeld:
There is actually no issue with the percentage reported on the Yooksmart shades.

Fix #1306


puddly:
Are there firmware revisions still affected by this? I think instead of removing the quirk, we can just add a way to stop applying it depending on the device's current firmware version.

jonnylangefeld:
@puddly that makes sense to me. Can you link me to something that explains how to exclude firmware versions?
I have firmware versions `0x11013001` and `0x11206780` in use and they both don't need the quirk. 
Happy to change this PR to only exclude those versions.

puddly:
There's no way to do it at the moment, it will have to be added to zigpy.

rwalker777:
DO NOT approve this PR.  Please see comment here https://github.com/zigpy/zha-device-handlers/issues/1306#issuecomment-1372402947

rwalker777:
I was playing with this and reset my blind that was reporting normally... now it is reporting reversed.  So all mine are now reversed... I think there is a bug in their firmware that sometimes fails to set the direction and thus reverses the blind.  Resetting it and setting up again fixes it.  I didn't even reset it's zigbee configuration just the up/down/direction part.  I agree removing the quirk and then instructing people if they run into this to reset the blind and run through setting limits and direction again.

TheJulianJES:
Rebased. IMO, we should merge this, as the initial PR was already full of complaints:
- https://github.com/zigpy/zha-device-handlers/pull/971

And the issue also complains about the quirk inverting the blinds:
- https://github.com/zigpy/zha-device-handlers/issues/1306

rwalker777 previously mentioned that the PR shouldn't be merged, but now says the quirk should be removed, as the pairing seemingly didn't complete successfully the first time.

Z2M also seems to have removed their "inverted" option for this (regardless of firmware): https://github.com/Koenkk/zigbee-herdsman-converters/commit/a25ef0708ed9bd64609f646b1425aa211698fc7d

Ideally, the blinds can be re-paired and then work properly. Everyone who's currently modifying zha-quirks directly (to remove the quirk) and those who are using a custom quirk to override the built-in one can stop modifying their stuff.

In the unlikely case that some people need the inverted reporting (which would have been "long" noticed with Z2M IMO), the quirk can temporarily be installed back as a custom quirk: https://github.com/zigpy/zha-device-handlers/blob/6e5e5384c5a0a2788458c2e1a20e105e24398e72/zhaquirks/yooksmart/D10110blinds.py

TheJulianJES (approved):

jonnylangefeld:
Glad to see this moving forward. And it [seems to work](https://github.com/zigpy/zha-device-handlers/issues/1306#issuecomment-1378035472) for @rwalker777 now as well.

TheJulianJES:
Read up more on the issue and it seems like there were only complaints in the Z2M repo when the cover was (wrongfully) inverted.

The *fix* will likely land with HA Core 2023.4.0. So you can remove any custom quirks that override the built-in one then or stop modifying your HA installation when upgraded. I'll monitor the issues closely at that time, but don't expect any.

PR #1676, opened 2022-08-03
#940: Add Aqara H1 Wireless Remote Double Rocker Version WRS-R02, variant 3

Alcolo47:
- With no model info
- endpoint 1 declared as color dimmer (261)
- No input/output declared

Big hack line 304: `MODELS_INFO: [(LUMI, "lumi.remote.b28ac1"), (None, None)],`
But, how can I do otherwise when I receive manufacturer=None and model=None ...

In the plugin-zigbee, I change into `lumi.remote.b28ac1.json` logical ids with 
`"logicalId": "1::18::0"` and `"logicalId": "2::18::0"`


javicalle:
I have had to pair some devices several times because they did not report the manufacturer or model correctly.
Have you tried pairing the device again? Make sure to keep your device awake. You can usually do this by briefly pressing the button while the pairing is taking place.

dmulcahey:
none for these values is from a failed pairing. We will not add that to quirks

TheJulianJES:
Closing for now. The device should be picked up as one of the other variants already.
Make sure you're running the somewhat up-to-date firmware on your coordinator (for Aqara manufacturer codes) and try re-pairing the device multiple times while pressing buttons on it.

A pairing with `manufacturer=None`, `model=None`, and no clusters reported is a failed or incomplete pairing.

Issue #1702, opened 2022-08-21
[Device Support Request] NOUS E6 TS0601 _TZE200_nnrfa68v


RhavoX:
**Is your feature request related to a problem? Please describe.**
The NOUS E6 Thermometer and Humidity sensor does not provide any values, as well as the time on it is not synced.
I have modified the current sensor quirk a bit and made temperature, humidity and battery level work. Whatever I do though the time synchronization has no effect. In this issue I will provide the code I have now, maybe someone will have some ideas :)

I have also checked the tuya zigbee communication spec but I didn't find anything useful there.

**Describe the solution you'd like**
I would like for the device to be officially supported :). And maybe additionally to suport the time sync feature.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0302",
      "in_clusters": [
        "0x0001",
        "0x0402",
        "0x0405",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_nnrfa68v",
  "model": "TS0601",
  "class": "ts0601_sensor_cst.TuyaNousE6TempHumiditySensor"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2022.8.5",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.10.5",
    "docker": true,
    "arch": "x86_64",
[...]

PR #1737, opened 2022-09-08
Implement `handle_set_time_request` in `TuyaMCUCluster`

javicalle:
Fully implement the `handle_set_time_request` in `TuyaMCUCluster`.

The code has been taken from the `TuyaManufCluster` and must have the same behavior that the already supported devices (`set_time_offset` and `set_time_local_offset`):
* https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L377-L398

Related to: #1702

All credit goes to @RhavoX who perform the first code and tests.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,4 +1,5 @@
 """Quirks common helpers."""
+import datetime
 
 
 ZCL_IAS_MOTION_COMMAND = b"\t!\x00\x01\x00\x00\x00\x00\x00"
@@ -21,3 +22,19 @@ class ClusterListener:
     def cluster_command(self, tsn, commdand_id, args):
         """Command received listener."""
         self.cluster_commands.append((tsn, commdand_id, args))
+
+
+class MockDatetime(datetime.datetime):
+    """Override for datetime functions."""
+
+    @classmethod
+    def now(cls):
+        """Return testvalue."""
+
+        return cls(1970, 1, 1, 1, 0, 0)
+
+    @classmethod
+    def utcnow(cls):
+        """Return testvalue."""
+
+        return cls(1970, 1, 1, 2, 0, 0)
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,7 +1,6 @@
 """Quirks common helpers."""
 import datetime
 
-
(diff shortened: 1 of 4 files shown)


RhavoX:
@javicalle hi, isn't the `manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID` param missing here?

javicalle:
> isn't the `manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID` param missing here?

Although I am becoming more and more certain that Tuya is not using the manufacturer code in any case, the implementation according to the specifications should use it. That's why we don't override it here.

The cases where it has been detected that it is necessary to override it, we do it with the `NO_MANUFACTURER_ID`, but it is done in a specific class. Like here:
* https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/mcu/__init__.py#L311-L313

It can be confusing to review the implementations because there are 2 architectures implementing the quirks. I am taking care of the one that uses the `TuyaMCUCluster`

When I can fix the tests I will review your case and tell you how it would be implemented.

javicalle:
I don't know why test coverage is failing.
Tests has been added and in my local environment the coverage for tuya classes are better than here:
```
zhaquirks/tuya/__init__.py                        551    152    72%
zhaquirks/tuya/air/__init__.py                     18      0   100%
zhaquirks/tuya/air/ts0601_air_quality.py           11      0   100%
zhaquirks/tuya/mcu/__init__.py                    179      0   100%
zhaquirks/tuya/ts000x.py                           18      0   100%
```
🤷🏻‍♂️

javicalle:
I desist from trying to fix the coverage, although I don't know if that could be a problem in future PRs (if this PR is merged without complying with, any future PRs, it will not comply). I can't understand how can the coverage be lower than before, or why complain about code that in my local environment is covereded.

I'm going to try to review the tests of the Tuya devices in another PR because I don't see the point in continuing to add tests that are not related to the PR.

dmulcahey:
I’ll take a peek in a bit.

javicalle:
Now all the checks are OK.
For me is ready to go.

dmulcahey (approved):

PR #1740, opened 2022-09-09
Add missing model info quirk for TUYA TZE200

Palid:
This commit adds a missing model info quirk for TUYA TZE200 without which it'll only show VOClevel sensor instead of all the available sensors on the device.

### DIFF
--- a/zhaquirks/tuya/air/ts0601_air_quality.py
+++ b/zhaquirks/tuya/air/ts0601_air_quality.py
@@ -34,6 +34,7 @@ class TuyaCO2Sensor(CustomDevice):
         MODELS_INFO: [
             ("_TZE200_8ygsuhe1", "TS0601"),
             ("_TZE200_yvx5lh6k", "TS0601"),
+            ("_TZE200_dwcarsat", "TS0601"),
         ],
         ENDPOINTS: {
             1: {


javicalle (approved):
LGTM

javicalle:
Could you include a link to the device manufacturer?
This way other users will be able to identify if your device is the same.

Palid:
@javicalle it's a follow-up to https://github.com/zigpy/zha-device-handlers/pull/1367

Had to add this line otherwise most of the sensors were not found.
Please let me know if I should amend the commit with details about previous PR, or with the manufacturer details.

javicalle:
No, there is no need.

I suppose that is some kind of upgraded version with the GPP cluster.

It's this the device?:
* https://zigbee.blakadder.com/Tuya_DCR-KQG.html

Palid:
Yep, that's this one, however it seems that I might need to do some filtering over the CO2 meter, as it's giving out garbage data.

javicalle:
> it's giving out garbage data.

Sometimes the device needs to be repaired to fix the garbage data (random 0 values).

> it seems that this device requires a different `dp_to_attribute` marcher than the previous ones

It would be a pity because that would mean to duplicate also the `TuyaCO2Sensor` class.


> is that even doable in quirks?

I believe that is not supported in the present implementation:
* https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L1423-L1439

gudvinr:
You can see a bit of history in zigbee2mqtt implementation: Koenkk/zigbee2mqtt#11033

They filter garbage values and there is correct value mapping too

Palid:
Alright, it means that I need to figure out some basic denouncing and filtering for the data there. Hoped that there's something for that in ZHA codebase already, but I can write some simple helpers myself.

The zigbee2mqtt issue is of this device, so I can try to mirror the mappings too, might be that we won't need another mapping class.

dmulcahey (approved):

pamatika:
Any update on this fix?
I have also _TZE200_dwcarsat and only voc value available.

Issue #1188, opened 2021-11-28
FeiBit Light switch showing as dimmable when it's not


Smiley1244:
Hello,

Please forgive me I'm very new to ZHA but i'm in the process of moving all my Hue devices to ZHA.
I have a couple of FeiBit switches which worked fine in Hue and worked 100% in ZHA until a recent update. With the update (sorry I don't remember which one, RL has got in the way and it's been some time since the change). They went from being std on/off to reporting as dimmable which they are not.

In some circumstances the brightness gets set to 1 which really only affects the bightness of the icon in HA, everything works fine so it's just an annoyance but i thought I'd report it as it doesn't seem correct.

The devices I have are the 
FeiBit FNB56-ZSW02LX2.0 2 way switch
```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4478, maximum_buffer_size=127, maximum_incoming_transfer_size=90, server_mask=10752, maximum_outgoing_transfer_size=90, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "11": {
      "profile_id": 49246,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x1000"
      ],
      "out_clusters": [
        "0x0019"
      ]
    },
    "12": {
      "profile_id": 49246,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008"
      ],
      "out_clusters": [
        "0x0019"
      ]
    }
  },
  "manufacturer": "FeiBit",
  "model": "FNB56-ZSW02LX2.0",
  "class": "zigpy.device.Device"
}
```

[...]

PR #1745, opened 2022-09-12
Quirks for FeiBit light switches

yruss972:
* FNB56-ZSW01LX2.0
* FNB56-ZSW02LX2.0
* FNB56-ZSW03LX2.0

Switches also marketed by ZemiSmart under the KS-811 model

### DIFF
--- a/tests/test_feibit.py
+++ b/tests/test_feibit.py
@@ -1,0 +1,130 @@
+"""Tests for Feibit quirks."""
+
+
+import zhaquirks
+import zhaquirks.feibit.fnb56
+
+zhaquirks.setup()
+
+
+def test_zsw01lx20_signature(assert_signature_matches_quirk):
+    """Test FeiBit 1 gang light switch."""
+
+    signature = {
+        "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4478, maximum_buffer_size=127, maximum_incoming_transfer_size=90, server_mask=10752, maximum_outgoing_transfer_size=90, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
+        "endpoints": {
+            "11": {
+                "profile_id": 49246,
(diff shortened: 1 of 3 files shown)


dmulcahey:
We’re these not working as is? I only see profile swaps from ZLL to ZHA. This shouldn’t be needed. What is the difference in experience in ZHA with and without this change?

yruss972:
@dmulcahey Thanks for taking a look. 

I admit this was a break glass attempt to get a switch working and I don't understand myself why it helped.

These switches have been working well for me in Deconz for years.
I switched over to ZHA via Raspbee a couple months ago and had about 12 of them working OOTB, though ZHA tends to show some irrelevant options.
Last week I added two more of these switches (2x1 gang) in one room and only one would work at a time- really weird.
I have multiple working routers in the same room and more in adjacent rooms so I didn't suspect any reception issues.
Replacing the HW didn't make any difference and I successfully paired the same HW somewhere else.

I originally thought the issue was some routing limitation since this was probably my 33rd device on the network. I tried adding via a working device in the room, etc. but it doesn't seem to mean anything with a zigpy-deconz backend? Everything always shows as connected directly to the coordinator.

When nothing worked, I wrote a custom quirk, originally not even swapping from ZLL to ZHA. Only when I made the swap to ZHA, did I manage to successfully pair both switches in the room.

Since it worked I figured I'd share. Maybe there is some added or different resilience/retry logic which only takes effect via the ZHA code path?

dmulcahey:
Routers will change that limit. I am at 160 devices in prod

yruss972:
I really have no idea- I've got some 20+ routers out of ~32/33 devices.
The error messages I'm getting look the same as before but on a different device in a completely different part of the house:
```
2022-09-12 21:04:55.377 WARNING (MainThread) [homeassistant.components.zha.core.channels.base] [0xFE79:2:0x0006]: async_initialize: all attempts have failed: [DeliveryError('[0xfe79:2:0x0006]: Message send failure'), DeliveryError('[0xfe79:2:0x0006]: Message send failure'), DeliveryError('[0xfe79:2:0x0006]: Message send failure'), DeliveryError('[0xfe79:2:0x0006]: Message send failure')]
```

This time, the misbehaving device is a E220-KR4N0Z0-HA power strip which has an annoying issue of showing as lights rather than sockets/switches so maybe a quirk is a good idea anyway. I'll try one locally and see if the problems shift to some other quirkless device again.

yruss972:
Circling back, I wrote up a custom quirk for the now misbehaving devices here: #1749 and now they are all paired again without issues. I really don't know how/if the quirks are making a difference but it seems to do something.

TheJulianJES:
I see you also removed the ``LevelControl`` cluster. Currently tracking a similar issue here: https://github.com/zigpy/zha-device-handlers/issues/1188#issuecomment-1371281949

Was the problem that the devices showed a brightness slider in HA, even though they only switch on/off?

MattWestb:
Changing ZLL to ZH can making problems then its some fundamental different how the mesh is working but the device is still doing there thing so if its one ZLL router its still doing the router things "ala ZLL" but ZHA is thinking its one ZH router.

PR #1754, opened 2022-09-14
Add the `on_off` call in the `move_to_level_with_on_off` command

javicalle:
Implement the `on_off` call in the `move_to_level_with_on_off` command for the MCU Tuya devices. 
The previous implementation was removed in #1748. Originally introduced in #1489

After some discussion in https://github.com/zigpy/zha-device-handlers/pull/1748#issuecomment-1244931971 I really believe that the implementation must be done inside the handler library better than expect to add all the devices in the ZHA implementation. This way the handlers library would fully implement the `move_to_level_with_on_off` for the MCU devices.

Fixes: #1694

### DIFF
--- a/zhaquirks/tuya/mcu/__init__.py
+++ b/zhaquirks/tuya/mcu/__init__.py
@@ -420,8 +420,21 @@ class TuyaLevelControl(LevelControl, TuyaLocalCluster):
             command_id,
             args,
         )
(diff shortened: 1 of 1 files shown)


dmulcahey (approved):

javicalle:
Thanks David.

Issue #1761, opened 2022-09-16
[Device Support Request]  Zigbee Roller Shade _TZE200_7eue9vhc   


codypet:
**Is your feature request related to a problem? Please describe.**
Request cover support for Zigbee roller shades _TZE200_7eue9vhc [...]

**Describe the solution you'd like**
Add cover support for the _TZE200_7eue9vhc roller shade

<details>

<summary>Device signature</summary>

```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_7eue9vhc",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2022.9.4",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.10.5",
    "docker": true,
    "arch": "x86_64",
    "timezone": "America/New_York",
    "os_name": "Linux",
    "os_version": "5.15.60",
    "supervisor": "2022.08.6",
    "host_os": "Home Assistant OS 8.5",
    "docker_version": "20.10.14",
    "chassis": "vm",
    "run_as_root": true
  },
  "custom_components": {
    "samsungtv_smart": {
      "version": "0.7.6",
      "requirements": [
[...]

PR #1763, opened 2022-09-17
Add support for Tuya _TZE200_7eue9vhc

codypet:
Line 373 Added _TZE200_7eue9vhc in response to issue #1761
Fixes: #1761


javicalle:
Have you verified if the behavior in HA is correct?
Mainly that the behavior in the UI corresponds to the movement of the shutter both when you operate it from HA and from the physical switch.

And can you reference the issue in your first post like this:
```
Fixes: #1761
```

That way the issue will be closed once the PR is merged.

codypet:
@javicalle Yep.  Works like a charm.  I have the zha_quirks folder in my instance that has this updated ts0601_cover file the way I uploaded it and it works brilliantly.

javicalle (approved):
LGTM

dmulcahey (approved):

PR #1779, opened 2022-09-21
Handle Tuya multiple attributes reports

javicalle:
This PR enable handle reports with multiple attributes from Tuya devices.

All credits goes to @puddly:
* https://github.com/zigpy/zha-device-handlers/issues/1566#issuecomment-1133066260

My first approach was to include the `dp` attribute inside the already existing `TuyaData` class, but most of the deserialization tests break and I haven't been able to fix them.

The tests of the new functionality have been added.


javicalle:
Thanks

Issue #1786, opened 2022-09-25
Tesla Smart Thermostatic Valve TV100 [Device Support Request]


dohnalekp:
Hi, i buy two Tesla Smart Thermostatic Valve TV100. Adding to home assistant is OK, but no entities show.
A have the same working device. 

**Working device zigbee signature:**

{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x000d",
        "0x000f",
        "0x0201",
        "0x0204",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0007",
      "in_clusters": [
        "0x0006",
        "0x000d"
      ],
      "out_clusters": []
    },
    "3": {
      "profile_id": 260,
      "device_type": "0x0007",
      "in_clusters": [
        "0x0006"
      ],
      "out_clusters": []
    }
  },
  "manufacturer": "_TZE200_husqqvux",
  "model": "TS0601",
  "class": "zhaquirks.tuya.ts0601_trv.ZonnsmartTV01_ZG"
}

**New device zigbee signature:**
{
[...]

PR #1792, opened 2022-09-27
Add `_TZE200_lnbfnyxd` TRV

javicalle:
Adds the `manufacturer_code` for the ' Tesla Smart Thermostatic Valve TV100'
I've also sorted the list alphabetically.

Fixes: #1786


MattWestb:
ZONNSMART TRVs is needing being added in ZHA for getting the presets working here:
https://github.com/home-assistant/core/blob/9084beda328ee29d98357379414a6c8d39db0af0/homeassistant/components/zha/climate.py#L760-L767

Also we have missing adding some and can you also adding the missing "brands" from the quirk ?

javicalle:
Thanks for pointing it out.

Would everything be correct now?

PR #1795, opened 2022-09-29
Add support for _TZE200_yi4jtqq1 illuminance sensor.

gmsoft-tuxicoman:
This is a simple illuminance sensort with a maximum of 1000 lumen. So it's only really suited for inside, not outside.

The quirk provide both the raw lumen value as well as the low/medium/high brightness level.

### DIFF
--- a/zhaquirks/tuya/ts0601_illuminance.py
+++ b/zhaquirks/tuya/ts0601_illuminance.py
@@ -17,19 +17,16 @@ from zhaquirks.const import (
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-
 from zhaquirks.tuya import TuyaLocalCluster
-from zhaquirks.tuya.mcu import (
-    DPToAttributeMapping,
-    TuyaDPType,
-    TuyaMCUCluster,
-)
+from zhaquirks.tuya.mcu import DPToAttributeMapping, TuyaDPType, TuyaMCUCluster
 
 TUYA_BRIGHTNESS_LEVEL_DP = 0x01  # 0-2 "Low, Medium, High"
 TUYA_ILLUMINANCE_DP = 0x02  # [0, 0, 3, 232] illuminance
 
+
 class BrightnessLevel(t.enum8):
     """Brightness level enum."""
+
     LOW = 0x00
     MEDIUM = 0x01
     HIGH = 0x02
--- a/zhaquirks/tuya/ts0601_illuminance.py
+++ b/zhaquirks/tuya/ts0601_illuminance.py
@@ -17,7 +17,13 @@ from zhaquirks.const import (
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-from zhaquirks.tuya import DPToAttributeMapping, TuyaLocalCluster, TuyaNewManufCluster
+
+from zhaquirks.tuya import TuyaLocalCluster
+from zhaquirks.tuya.mcu import (
+    DPToAttributeMapping,
+    TuyaDPType,
+    TuyaMCUCluster,
+)
 
 TUYA_BRIGHTNESS_LEVEL_DP = 0x01  # 0-2 "Low, Medium, High"
 TUYA_ILLUMINANCE_DP = 0x02  # [0, 0, 3, 232] illuminance
@@ -40,19 +46,21 @@ class TuyaIlluminanceMeasurement(IlluminanceMeasurement, TuyaLocalCluster):
     )
 
 
-class TuyaIlluminanceCluster(TuyaNewManufCluster):
+class TuyaIlluminanceCluster(TuyaMCUCluster):
     """Tuya Illuminance cluster."""
 
     dp_to_attribute: Dict[int, DPToAttributeMapping] = {
         TUYA_BRIGHTNESS_LEVEL_DP: DPToAttributeMapping(
             TuyaIlluminanceMeasurement.ep_attribute,
             "manufacturer_brightness_level",
-            lambda x: BrightnessLevel(x),
+            dp_type=TuyaDPType.ENUM,
+            converter=lambda x: BrightnessLevel(x),
         ),
         TUYA_ILLUMINANCE_DP: DPToAttributeMapping(
             TuyaIlluminanceMeasurement.ep_attribute,
             "measured_value",
-            lambda x: (10000.0 * math.log10(x) + 1.0 if x != 0 else 0),
+            dp_type=TuyaDPType.VALUE,
+            converter=lambda x: (10000.0 * math.log10(x) + 1.0 if x != 0 else 0),
         ),
     }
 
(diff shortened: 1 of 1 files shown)


dmulcahey:
Please rebase this and install / run the precommit checks

dmulcahey:
@javicalle lmk what you think when you have a moment please. Thanks!

zhaquirks/tuya/ts0601_illuminance.py
```diff
@@ -0,0 +1,115 @@
...
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+from zhaquirks.tuya import DPToAttributeMapping, TuyaLocalCluster, TuyaNewManufCluster
+
+TUYA_BRIGHTNESS_LEVEL_ATTR = 0x01  # 0-2 "Low, Medium, High"
+TUYA_ILLUMINANCE_ATTR = 0x02  # [0, 0, 3, 232] illuminance
```
javicalle, line 23: No problem with setting as constants, but I would prefer not to use the `_ATTR` sufix because it can be confusing (they are not clusters attributes).
Replace all around with a DataPoint nomenclature:
```python
TUYA_BRIGHTNESS_LEVEL_DP = 0x01  # 0-2 "Low, Medium, High"
TUYA_ILLUMINANCE_DP = 0x02  # [0, 0, 3, 232] illuminance
```

zhaquirks/tuya/ts0601_illuminance.py
```diff
@@ -0,0 +1,115 @@
...
+TUYA_BRIGHTNESS_LEVEL_STR = ["Low", "Medium", "High"]
+
+
+class TuyaIlluminanceMeasurement(IlluminanceMeasurement, TuyaLocalCluster):
+    """Tuya local IlluminanceMeasurement cluster."""
+
+    attributes = IlluminanceMeasurement.attributes.copy()
+    attributes.update(
+        {
+            0xFF00: ("brightness_level", t.CharacterString, True),
```
javicalle, line 34: The illuminanceMeasurement cluster don't have any reference to `brightness_level` so not sure if that could be the correct attribute name 🤔 
We can keep the reference for now but I would replace for `manufacturer_brightness_level` or `manufacturer_measured_value`.

zhaquirks/tuya/ts0601_illuminance.py
```diff
@@ -0,0 +1,115 @@
...
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+from zhaquirks.tuya import DPToAttributeMapping, TuyaLocalCluster, TuyaNewManufCluster
+
+TUYA_BRIGHTNESS_LEVEL_ATTR = 0x01  # 0-2 "Low, Medium, High"
+TUYA_ILLUMINANCE_ATTR = 0x02  # [0, 0, 3, 232] illuminance
+
+TUYA_BRIGHTNESS_LEVEL_STR = ["Low", "Medium", "High"]
```
javicalle, line 25: What about if we use:
```python
class BrightnessLevel(t.enum8):
    """Brightness level enum."""

    LOW = 0x00
    MEDIUM = 0x01
    HIGH = 0x02
```

That way you can define the attribute as:
```python
            0xFF00: ("brightness_level", BrightnessLevel, True),
```

And convert as:
```python
        TUYA_BRIGHTNESS_LEVEL_ATTR: DPToAttributeMapping(
            TuyaIlluminanceMeasurement.ep_attribute,
            "brightness_level",
            lambda x: BrightnessLevel(x),
        ),
```

javicalle (changes requested):
These would be the minimum requirements to review.

zhaquirks/tuya/ts0601_illuminance.py
```diff
@@ -0,0 +1,115 @@
...
+            TuyaIlluminanceMeasurement.ep_attribute,
+            "measured_value",
+            lambda x: (10000.0 * math.log10(x) + 1.0 if x != 0 else 0),
+        ),
+    }
+
+    data_point_handlers = {
+        TUYA_BRIGHTNESS_LEVEL_ATTR: "_dp_2_attr_update",
+        TUYA_ILLUMINANCE_ATTR: "_dp_2_attr_update",
+    }
```
javicalle, line 58: You will need to change the imports statements to the new ones:
```python
from zhaquirks.tuya import TuyaLocalCluster
    DPToAttributeMapping,
    TuyaDPType,
    TuyaMCUCluster,
)
```

Change your `TuyaIlluminanceCluster` class declaration to something like:
```python
class TuyaIlluminanceCluster(TuyaMCUCluster):
    """Tuya Illuminance cluster."""

        TUYA_BRIGHTNESS_LEVEL_ATTR: DPToAttributeMapping(
            TuyaIlluminanceMeasurement.ep_attribute,
            "brightness_level",  # or the new name for attribute
            dp_type=TuyaDPType.ENUM,
            converter=lambda x: BrightnessLevel(x),
        ),
        TUYA_ILLUMINANCE_ATTR: DPToAttributeMapping(
            TuyaIlluminanceMeasurement.ep_attribute,
            "measured_value",
            dp_type=TuyaDPType.VALUE,
            converter=lambda x: (10000.0 * math.log10(x) + 1.0 if x != 0 else 0),
        ),
    }

    data_point_handlers = {
        TUYA_BRIGHTNESS_LEVEL_ATTR: "_dp_2_attr_update",
        TUYA_ILLUMINANCE_ATTR: "_dp_2_attr_update",
    }
```

Thats all.

javicalle (changes requested):
I intend to migrate the class hierarchy for these types of devices, so I am going to ask you to make the following changes.
These changes are not a prerequisite, but they will help us with future maintenance.

If you're going to tackle the changes, I'd suggest making the other changes first, then attacking these changes.

javicalle:
Not happy with another `ts0601_whatever.py` file, but not sure if there is a better fit. Keep as is for now.

The `TuyaIlluminanceMeasurement` must be moved to a generic class. 
I like the security code check, but not sure if it is correct (I believe that would be something like `if x > 1 else 0` or maybe a parenthesis is needed here).
Keep as is and I will move in another PR.

The device doesn't have a battery report?
The new implementation would log any not implemented report (warning level).

gmsoft-tuxicoman:
I've made all the changes.

This device doesn't have a battery sensor as it's USB powered.

zhaquirks/tuya/ts0601_illuminance.py
```diff
@@ -0,0 +1,124 @@
...
+    HIGH = 0x02
+
+
+class TuyaIlluminanceMeasurement(IlluminanceMeasurement, TuyaLocalCluster):
+    """Tuya local IlluminanceMeasurement cluster."""
+
+    attributes = IlluminanceMeasurement.attributes.copy()
+    attributes.update(
+        {
+            0xFF00: ("brightness_level", BrightnessLevel, True),
```
javicalle, line 41: Maybe `manufacturer_brightness_level` here?

Must be the same value as in `dp_to_attribute` mapping.

javicalle (changes requested):
Just one left.

javicalle (approved):
LGTM

Thanks for your work and patience.

dmulcahey (approved):

Issue #1582, opened 2022-05-24
[Device Support Request] TS130F - _TZ3000_kmsbwdol


Oulewappeur:
**Is your feature request related to a problem? Please describe.**
This device gets recognized and connected by ZHA out of the box. 
It's a switch for two shutters/curtains/covers with physical buttons (up, pause, down) for both outputs of them
However, the percentages are the wrong way around (so when it's closed it says its 0% and when it's open it says it's 100%), and I'm not able to calibrate properly.

I've been trying to fix it myself, but I have never done this before and I'm probably doing it completely wrong!

There are also two toggles which seem to do nothing
And the rssi is not shown either.

Product that I purchased:
[https://nl.aliexpress.com/item/1005003308586552.html?srcSns=sns_Copy&spreadType=socialShare&bizType=ProductDetail&social_params=60161453943&aff_fcid=b68f242928344655ae91979f35502ba9-1653409970693-02737-_u76szN&tt=MG&aff_fsk=_u76szN&aff_platform=default&sk=_u76szN&aff_trace_key=b68f242928344655ae91979f35502ba9-1653409970693-02737-_u76szN&shareId=60161453943&businessType=ProductDetail&platform=AE&terminal_id=a4e2cde960db4c23a78d82206988e934&afSmartRedirect=y](url)

**Describe the solution you'd like**
A roller shutter that works properly would be nice!

<details> <summary>Device signature </summary> 

```
 {
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0202",
      "in_clusters": [
[...]

PR #1807, opened 2022-10-04
Update TS130f covers quirk

javicalle:
A couple of changes here:
1. remove the duplicated quirk definition `TuyaMoesDiyRf433TS130F` (exactly the same as `TuyaTS130FTI2`)
2. add a new `TuyaTS130Double_GP` quirk

Fix: #1582

### DIFF
--- a/zhaquirks/tuya/ts130f.py
+++ b/zhaquirks/tuya/ts130f.py
@@ -159,51 +159,6 @@ class TuyaZemismartTS130F(CustomDevice):
     }
 
 
-class TuyaMoesDiyRf433TS130F(CustomDevice):
-    """Moes Tuya ZigBee DIY RF433 Smart Curtain Switch Module."""
-
-    signature = {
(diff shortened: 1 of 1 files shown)


javicalle:
The files compare is not helping here.

I have moved the `TuyaTS130FTI2` upwards to the `TuyaMoesDiyRf433TS130F` definition (and I have replaced it).
And github is comparing the new `TuyaTS130Double_GP` against the moved `TuyaTS130FTI2`, but the `TuyaTS130Double_GP` is fully new definition.

MattWestb:
OK i understand the view is telling its updated the device class and in the real (if expanding all the quirk) its one new class "made" of one duplicated device class.

Sorry for making trouble and god work done as normal !!

javicalle:
No problem. It is great to have other eyes that review what I do.
Thank you. 🙇🏻‍♂️

PR #1816, opened 2022-10-09
Disable schedule on Tuya/Saswell TRV (TS601/_TZE200_c88teujp)

### LINKED ISSUE #1815: [BUG] Tuya/Saswell TRV (TS601/_TZE200_c88teujp) stuck in schedule mode
**Describe the bug**
Some Tuya TRV's, Sasswell and it clones in particular, like to enable schedule mode in unexpected conditions.

I have 6 TRV's of this type and I have used them with Tuya app ("Smart Life") then with Zigbee2MQTT and now with ZHA. After the migration to ZHA my automations stopped working because my set-points were constantly overridden by values from a schedule that I have previously set in Tuya app. I've tried pairing with Tuya GW again and disabling the schedule but devices seem to enable the schedule by default upon pairing with ZHA and/or possibly after switching from "off" to "heat" mode.

This renders TRV's that have been previously configured with a schedule uncontrollable via ZHA.

I do have a working PoC of a quick-fix for this issue and I'll try to publish a PR with it.

**To Reproduce**
1. Pair TRV with Tuya GW using Tuya app
2. Program any schedule using Tuya app
3. Pair TRV with ZHA
4. Get frustrated because values programmed in the schedule keep overriding your desired set-points

**Expected behavior**
I expect TRV to keep the requested set-point,

<details>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0004",
        "0x0005",
        "0x0201",
        "0xef00"
      ],
      "out_clusters": [
        "0x0001",
        "0x000a",
        "0x0019",
        "0x0201",
        "0xef00"
      ]
    }
  },
  "manufacturer": "_TZE200_c88teujp",
  "model": "TS0601",
  "class": "zhaquirks.tuya.ts0601_trv_sas.Thermostat_TZE200_c88teujp"
}
```

</details>

**Additional context**

This issue was already noticed and worked-around by Zigbee2MQTT developers, see code highlighted below:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/converters/toZigbee.js#L5877:L5879

blazewicz:
This work-around is intended to fix #1815. Its based on solution used in Zigbee2MQTT linked below:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/converters/toZigbee.js#L5877:L5879

I've tested it locally with 6 devices and it fixed the issue.

### DIFF
--- a/zhaquirks/tuya/ts0601_trv_sas.py
+++ b/zhaquirks/tuya/ts0601_trv_sas.py
@@ -4,18 +4,9 @@ import logging
 
 from zigpy.profiles import zha
 import zigpy.types as t
-from zigpy.zcl.clusters.general import (
-    Basic,
-    Groups,
-    Identify,
-    Ota,
-    PowerConfiguration,
-    Scenes,
-    Time,
-)
+from zigpy.zcl.clusters.general import Basic, Groups, Identify, Ota, Scenes, Time
 from zigpy.zcl.clusters.hvac import Thermostat
 
-from zhaquirks import LocalDataCluster
 from zhaquirks.const import (
     DEVICE_TYPE,
     ENDPOINTS,
@@ -26,21 +17,18 @@ from zhaquirks.const import (
 )
 from zhaquirks.tuya import (
     TuyaManufClusterAttributes,
+    TuyaPowerConfigurationCluster,
     TuyaThermostat,
     TuyaThermostatCluster,
 )
 
 _LOGGER = logging.getLogger(__name__)
 
-OCCUPIED_HEATING_SETPOINT_REPORTED = "occupied_heating_setpoint_reported"
-SYSTEM_MODE_REPORTED = "system_mode_reported"
-LOCAL_TEMP_REPORTED = "local_temp_reported"
-BATTERY_REPORTED = "battery_reported"
-
-OCCUPIED_HEATING_SETPOINT_COMMAND_ID = 615
-SYSTEM_MODE_COMMAND_ID = 357
-LOCAL_TEMP_COMMAND_ID = 614
-BATTERY_STATE_COMMAND_ID = 1385
+SASSWEL_HEATING_SETPOINT_ATTR = 0x0267
+SASSWEL_STATE_ATTR = 0x0165
+SASSWEL_LOCAL_TEMP_ATTR = 0x0266
+SASSWEL_BATTERY_LOW_ATTR = 0x0569
+SASSWEL_SCHEDULE_ENABLE_ATTR = 0x016C
 
 CTRL_SEQ_OF_OPER_ATTR = 0x001B
 
@@ -49,61 +37,60 @@ MAX_HEAT_SETPOINT_ATTR = 0x0016
 
 
 class ManufacturerThermostatCluster(TuyaManufClusterAttributes):
-    """Manufacturer thermostat cluster."""
+    """Tuya manufacturer specific cluster."""
+
+    class State(t.enum8):
+        """State option."""
+
+        Off = 0x00
+        On = 0x01
+
+    class BatteryState(t.enum8):
+        """Battery state option."""
+
+        Normal = 0x00
+        Low = 0x01
+
+    class ScheduleState(t.enum8):
+        """Schedule state option."""
+
+        Disabled = 0x00
+        Enabled = 0x01
 
     attributes = TuyaManufClusterAttributes.attributes.copy()
     attributes.update(
         {
-            SYSTEM_MODE_COMMAND_ID: ("system_mode", t.uint8_t, True),
-            OCCUPIED_HEATING_SETPOINT_COMMAND_ID: (
-                "occupied_heating_setpoint",
-                t.uint32_t,
-                True,
-            ),
-            LOCAL_TEMP_COMMAND_ID: ("local_temperature", t.uint32_t, True),
-            BATTERY_STATE_COMMAND_ID: ("battery_state", t.uint8_t, True),
+            SASSWEL_STATE_ATTR: ("state", State, True),
+            SASSWEL_HEATING_SETPOINT_ATTR: ("heating_setpoint", t.uint32_t, True),
+            SASSWEL_LOCAL_TEMP_ATTR: ("local_temperature", t.uint32_t, True),
+            SASSWEL_BATTERY_LOW_ATTR: ("battery_state", BatteryState, True),
+            SASSWEL_SCHEDULE_ENABLE_ATTR: ("schedule_enabled", ScheduleState, True),
         }
     )
 
+    TEMPERATURE_ATTRS = {
+        SASSWEL_HEATING_SETPOINT_ATTR: "occupied_heating_setpoint",
+        SASSWEL_LOCAL_TEMP_ATTR: "local_temperature",
+    }
+
     def _update_attribute(self, attrid, value):
         super()._update_attribute(attrid, value)
-        if attrid == OCCUPIED_HEATING_SETPOINT_COMMAND_ID:
+        if attrid in self.TEMPERATURE_ATTRS:
             self.endpoint.device.thermostat_bus.listener_event(
-                OCCUPIED_HEATING_SETPOINT_REPORTED, value
+                "temperature_change",
+                self.TEMPERATURE_ATTRS[attrid],
+                value * 10,
             )
-        elif attrid == SYSTEM_MODE_COMMAND_ID:
+        elif attrid == SASSWEL_STATE_ATTR:
             self.endpoint.device.thermostat_bus.listener_event(
-                SYSTEM_MODE_REPORTED, value
+                "system_mode_reported", value
             )
-        elif attrid == BATTERY_STATE_COMMAND_ID:
-            self.endpoint.device.battery_bus.listener_event(BATTERY_REPORTED, value)
-        elif attrid == LOCAL_TEMP_COMMAND_ID:
-            self.endpoint.device.thermostat_bus.listener_event(
-                LOCAL_TEMP_REPORTED, value
+        elif attrid == SASSWEL_BATTERY_LOW_ATTR:
+            # this device doesn't have battery level reporting, only battery low alert
+            # when the alert is active (1) we report 0% and 100% otherwise (0)
+            self.endpoint.device.battery_bus.listener_event(
+                "battery_change", 0 if value == 1 else 100
             )
-
-
-class PowerConfigurationCluster(LocalDataCluster, PowerConfiguration):
-    """Power configuration cluster."""
-
-    cluster_id = PowerConfiguration.cluster_id
-
-    def __init__(self, *args, **kwargs):
-        """Init."""
-        super().__init__(*args, **kwargs)
-        self.endpoint.device.battery_bus.add_listener(self)
-
-    def battery_reported(self, value):
-        """Handle reported battery state."""
-        _LOGGER.debug("reported battery alert: %d", value)
-        if value == 1:  # alert
-            self._update_attribute(
-                self.attributes_by_name["battery_percentage_remaining"].id, 0
-            )  # report 0% battery
-        else:
-            self._update_attribute(
-                self.attributes_by_name["battery_percentage_remaining"].id, 200
-            )  # report 100% battery
 
 
 class ThermostatCluster(TuyaThermostatCluster):
@@ -116,25 +103,6 @@ class ThermostatCluster(TuyaThermostatCluster):
         MAX_HEAT_SETPOINT_ATTR: 3000,
         CTRL_SEQ_OF_OPER_ATTR: Thermostat.ControlSequenceOfOperation.Heating_Only,  # the device supports heating mode
     }
-
-    def __init__(self, *args, **kwargs):
-        """Init."""
-        super().__init__(*args, **kwargs)
-        self.endpoint.device.thermostat_bus.add_listener(self)
-
-    def occupied_heating_setpoint_reported(self, value):
-        """Handle reported occupied heating setpoint state."""
-        self._update_attribute(
-            self.attributes_by_name["occupied_heating_setpoint"].id, value * 10
-        )
-        _LOGGER.debug("reported occupied heating setpoint: %d", value)
-
-    def local_temp_reported(self, value):
-        """Handle reported local temperature."""
-        self._update_attribute(
-            self.attributes_by_name["local_temperature"].id, value * 10
-        )
-        _LOGGER.debug("reported local temperature: %d", value)
 
     def system_mode_reported(self, value):
         """Handle reported system mode."""
@@ -160,23 +128,21 @@ class ThermostatCluster(TuyaThermostatCluster):
 
         if attribute == "occupied_heating_setpoint":
             # centidegree to decidegree
-            return {OCCUPIED_HEATING_SETPOINT_COMMAND_ID: round(value / 10)}
+            return {SASSWEL_HEATING_SETPOINT_ATTR: round(value / 10)}
 
         if attribute == "system_mode":
+            # this quirk does not support programmig modes so we force the schedule mode to be always off
+            # more details: https://github.com/zigpy/zha-device-handlers/issues/1815
             if value == self.SystemMode.Off:
-                return {SYSTEM_MODE_COMMAND_ID: 0}
+                return {SASSWEL_STATE_ATTR: 0, SASSWEL_SCHEDULE_ENABLE_ATTR: 0}
             if value == self.SystemMode.Heat:
-                return {SYSTEM_MODE_COMMAND_ID: 1}
+                return {SASSWEL_STATE_ATTR: 1, SASSWEL_SCHEDULE_ENABLE_ATTR: 0}
 
 
 class Thermostat_TYST11_c88teujp(TuyaThermostat):
     """Saswell 88teujp thermostat valve."""
 
     signature = {
-        # <SimpleDescriptor endpoint=1 profile=260 device_type=0
-        # device_version=0
-        # input_clusters=[0, 3]
-        # output_clusters=[3, 25]>
         MODELS_INFO: [
             ("_TYST11_KGbxAXL2", "GbxAXL2"),
             ("_TYST11_c88teujp", "88teujp"),
@@ -186,6 +152,10 @@ class Thermostat_TYST11_c88teujp(TuyaThermostat):
             ("_TYST11_zuhszj9s", "uhszj9s"),
         ],
         ENDPOINTS: {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=0
+            # device_version=0
+            # input_clusters=[0, 3]
+            # output_clusters=[3, 25]>
             1: {
                 PROFILE_ID: zha.PROFILE_ID,
                 DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
@@ -208,16 +178,14 @@ class Thermostat_TYST11_c88teujp(TuyaThermostat):
                 DEVICE_TYPE: zha.DeviceType.THERMOSTAT,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
+                    TuyaPowerConfigurationCluster,
                     Identify.cluster_id,
-                    ManufacturerThermostatCluster,
-                    PowerConfigurationCluster,
                     ThermostatCluster,
+                    ManufacturerThermostatCluster,
                 ],
                 OUTPUT_CLUSTERS: [
                     Identify.cluster_id,
                     Ota.cluster_id,
-                    PowerConfigurationCluster,
-                    ThermostatCluster,
                 ],
             }
         },
@@ -228,8 +196,6 @@ class Thermostat_TZE200_c88teujp(TuyaThermostat):
     """Saswell 88teujp thermostat valve."""
 
     signature = {
-        #  endpoint=1 profile=260 device_type=81 device_version=0 input_clusters=[0, 4, 5, 61184]
-        #  output_clusters=[10, 25]>
         MODELS_INFO: [
             ("_TZE200_c88teujp", "TS0601"),
             ("_TZE200_azqp6ssj", "TS0601"),
@@ -239,6 +205,10 @@ class Thermostat_TZE200_c88teujp(TuyaThermostat):
             ("_TZE200_2ekuz3dz", "TS0601"),
         ],
         ENDPOINTS: {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=81
+            # device_version=0
+            # input_clusters=[0, 4, 5, 61184]
+            # output_clusters=[10, 25]>
             1: {
                 PROFILE_ID: zha.PROFILE_ID,
                 DEVICE_TYPE: zha.DeviceType.SMART_PLUG,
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -81,6 +85,20 @@ def _update_attribute(self, attrid, value):
             self.endpoint.device.thermostat_bus.listener_event(
                 LOCAL_TEMP_REPORTED, value
             )
+        elif attrid == SCHEDULE_ENABLED_COMMAND_ID:
+            self.endpoint.device.thermostat_bus.listener_event(
+                SCHEDULE_STATE_REPORTED, value
+            )
+
+    async def write_attributes(self, attributes, manufacturer=None):
+        if "system_mode" in attributes or self.attributes_by_name["system_mode"].id in attributes:
```
blazewicz, line 94: Could someone explain to me why `attributes` passed to that method has numerical `command_id` instead of the command name as keys?

I expected:

```python
{"system_mode": 1}
```

but it receives:

```python
{357: 1}
```

blazewicz: Ok, I've figured it out. Its overwritten in `TuyaThermostatCluster:write_attributes()` with value from `ThermostatCluster:map_attribute()`.

https://github.com/zigpy/zha-device-handlers/blob/2292c4edfd5dc0ca6aac4439923c11cefbe5006c/zhaquirks/tuya/__init__.py#L643

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -269,6 +295,7 @@ class Thermostat_TZE200_c88teujp(TuyaThermostat):
                 OUTPUT_CLUSTERS: [
                     Time.cluster_id,
                     Ota.cluster_id,
+                    ManufacturerThermostatCluster,
```
javicalle, line 298: Is that needed? Can't you write the attribute from the input clusters `ManufacturerThermostatCluster`?

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -81,6 +85,25 @@ def _update_attribute(self, attrid, value):
...
+                SCHEDULE_STATE_REPORTED, value
+            )
+
+    async def write_attributes(self, attributes, manufacturer=None):
+        """Implement writeable attributes."""
+
+        if "system_mode" in attributes:
+
+            async def _disable_schedule():
+                await asyncio.sleep(3)
```
javicalle, line 99: I think that Z2M adds a delay because there is 2 writes. Have you tried without the `sleep`?.

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -81,6 +85,25 @@ def _update_attribute(self, attrid, value):
...
+                await self.write_attributes(
+                    {"schedule_enabled": 0}, manufacturer=manufacturer
+                )
+
+            asyncio.get_running_loop().create_task(_disable_schedule())
```
javicalle, line 104: Maybe you can do the same with:
```python
        if "system_mode" in attributes:
            await self.write_attributes({"schedule_enabled": 0}, manufacturer=manufacturer)

        return await super().write_attributes(attributes, manufacturer)
```

(Despite I don't know the relationship between the `system_mode` and `schedule_enabled` attributes) 🤷🏻‍♂️

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -155,18 +178,25 @@ def system_mode_reported(self, value):
             )
             _LOGGER.debug("reported system_mode: off")
 
+    def schedule_state_reported(self, value):
+        """Handle reported schedule state."""
+        if value == 1:
+            _LOGGER.debug("reported schedule state: enabled")
+        else:
+            _LOGGER.debug("reported schedule state: disabled")
```
javicalle, line 186: The implementation is updating the `schedule_enabled` attribute when the `system_mode` is updated. Why don't just let the user write the attribute here?
```python
    def schedule_state_reported(self, value):
        """Handle reported schedule state."""

        self._update_attribute(self.attributes_by_name["schedule_enabled"].id, value)

```

dmulcahey: That will only change the value in the cache. Just making sure that’s known

zhaquirks/tuya/ts0601_trv_sas.py
blazewicz: It is not, thanks!

zhaquirks/tuya/ts0601_trv_sas.py
blazewicz: I think they added it to work-around some race condition in device firmware. I've tried to reproduce this condition but I've failed to. For now I'll remove this delay.

zhaquirks/tuya/ts0601_trv_sas.py
blazewicz: Changed to send state and schedule mode simultaneously.

zhaquirks/tuya/ts0601_trv_sas.py
blazewicz: This quirk does not implement schedule programming therefore IMO it should automatically force schedule mode to be off, without involving the user.

javicalle: It's fine to me.

Just to say that the `SASSWEL_SCHEDULE_ENABLE_ATTR: 0` is performed here:
```python
    def map_attribute(self, attribute, value):
        """Map standardized attribute value to dict of manufacturer values."""

        if attribute == "occupied_heating_setpoint":
            # centidegree to decidegree
            return {SASSWEL_HEATING_SETPOINT_ATTR: round(value / 10)}

        if attribute == "system_mode":
            # this quirk does not support programmig modes so we force the schedule mode to be always off
            # more details: https://github.com/zigpy/zha-device-handlers/issues/1815
            if value == self.SystemMode.Off:
                return {SASSWEL_STATE_ATTR: 0, SASSWEL_SCHEDULE_ENABLE_ATTR: 0}
            if value == self.SystemMode.Heat:
                return {SASSWEL_STATE_ATTR: 1, SASSWEL_SCHEDULE_ENABLE_ATTR: 0}
```

and takes place every time that the `system_mode` is written to the `Off` or `Heat` values. I don't know if there can be other values for the `system_mode`.

javicalle:
For future references, these are the relevant lines:
* https://github.com/Koenkk/zigbee-herdsman-converters/blob/96563772e92eb459bf52c10e241c3a725968bcab/converters/toZigbee.js#L5924-L5934

blazewicz:
> Just to be sure.
> 
> Are that change indented to be applied to all the devices in that quirk?
> 
> That will be all these devices:
> 
> ```
>             ("_TZE200_c88teujp", "TS0601"),
>             ("_TZE200_azqp6ssj", "TS0601"),
>             ("_TZE200_yw7cahqs", "TS0601"),
>             ("_TZE200_9gvruqf5", "TS0601"),
>             ("_TZE200_zuhszj9s", "TS0601"),
>             ("_TZE200_2ekuz3dz", "TS0601"),
>             ("_TYST11_KGbxAXL2", "GbxAXL2"),
>             ("_TYST11_c88teujp", "88teujp"),
>             ("_TYST11_azqp6ssj", "zqp6ssj"),
>             ("_TYST11_yw7cahqs", "w7cahqs"),
>             ("_TYST11_9gvruqf5", "gvruqf5"),
>             ("_TYST11_zuhszj9s", "uhszj9s"),
> ```

Fair point, I've assumed these were all clones.

All except these three share the same configuration in Z2M:

* `_TZE200_2ekuz3dz`
* `_TYST11_azqp6ssj`
* `_TYST11_9gvruqf5`

Here's the list:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/96563772e92eb459bf52c10e241c3a725968bcab/devices/saswell.js#L11:L22

The `_TYST11_azqp6ssj` seems to be yet another sibling of `_TZE200_c88teujp` according to https://zigbee.blakadder.com/RTX_ZB-RT1.html.

I couldn't find any information regarding `_TYST11_9gvruqf5`.

The `_TZE200_2ekuz3dz` seems to be a completely different device type and I think it should get a separate quirk. More on that device: https://zigbee.blakadder.com/Beok_TGR85-ZB.html

I have no way of testing them all of course.

blazewicz:
Here's the configuration for `_TZE200_2ekuz3dz` from Z2M:

https://github.com/Koenkk/zigbee-herdsman-converters/blob/177326f7af1c42cd5e2f89998a03effd042780c6/devices/tuya.js#L2803:L2840

javicalle:
But... according to Z2M, your device also have some kind of scheduler:
* https://github.com/Koenkk/zigbee-herdsman-converters/blob/96563772e92eb459bf52c10e241c3a725968bcab/devices/saswell.js#L34-L40

and:
```python
        toZigbee: [tz.saswell_thermostat_current_heating_setpoint, ... , tz.tuya_thermostat_weekly_schedule],
```

javicalle:
You have said that the scheduled was setted in the Tuya application, right?

But as the quirk can not manage the scheduled for the device you need to disable it always to prevent that undesired behavior when HA interacts with the device.

If HA could manage the scheduler (in a good manner) that quirk will not be needed.

Would that be more or less the purpose of the quirk?

Quidamdk:
How to i disable schedule???

blazewicz:
> You have said that the scheduled was setted in the Tuya application, right?
> 
> But as the quirk can not manage the scheduled for the device you need to disable it always to prevent that undesired behavior when HA interacts with the device.
> 
> If HA could manage the scheduler (in a good manner) that quirk will not be needed.
> 
> Would that be more or less the purpose of the quirk?

Yes, that is correct.

Whenever someone finds time and will to implement the full support for the schedules this quirk may be upgraded. Until then my patch makes it usable in the basic form.

Quidamdk:
Mvh.Jørgen HansenDen 19. okt. 2022 kl. 16.43 skrev Krzysztof Błażewicz ***@***.***>:﻿

You have said that the scheduled was setted in the Tuya application, right?
But as the quirk can not manage the scheduled for the device you need to disable it always to prevent that undesired behavior when HA interacts with the device.
If HA could manage the scheduler (in a good manner) that quirk will not be needed.
Would that be more or less the purpose of the quirk?

Yes, that is correct.
Whenever someone finds time and will to implement the full support for the schedules this quirk may be upgraded. Until then my patch makes it usable in the basic form.

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -160,23 +128,21 @@ def map_attribute(self, attribute, value):
 
         if attribute == "occupied_heating_setpoint":
             # centidegree to decidegree
-            return {OCCUPIED_HEATING_SETPOINT_COMMAND_ID: round(value / 10)}
+            return {SASSWEL_HEATING_SETPOINT_ATTR: round(value / 10)}
```
javicalle, line 131: I feel like there is something wrong here.

At first sign there is an asymetric behavior with the temp attributes (`local_temperature` and `occupied_heating_setpoint`).
The `_update_attribute` is updating both with `value * 10`.
And then here in `map_attribute` is only updating the `occupied_heating_setpoint `(value/10)`.
Are you able to check if this part is fine?

javicalle (changes requested):
A few more comments

jclsn:
Would be really great if this could be merged soon. We just hit 0°C and my thermostats are not turning on because of this.

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -49,63 +37,62 @@
...
-            SYSTEM_MODE_COMMAND_ID: ("system_mode", t.uint8_t, True),
-            OCCUPIED_HEATING_SETPOINT_COMMAND_ID: (
-                "occupied_heating_setpoint",
-                t.uint32_t,
-                True,
-            ),
-            LOCAL_TEMP_COMMAND_ID: ("local_temperature", t.uint32_t, True),
-            BATTERY_STATE_COMMAND_ID: ("battery_state", t.uint8_t, True),
+            SASSWEL_STATE_ATTR: ("state", State, True),
+            SASSWEL_HEATING_SETPOINT_ATTR: ("heating_setpoint", t.uint32_t, True),
```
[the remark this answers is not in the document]
blazewicz: I see how that's tricky, let me try to explain.

ZCL Thermostat class uses `occupied_heating_setpoint` this is the attribute name which zigpy sends to the quirk when user requests setpoint change. Unit used by ZCL is centidegree Celsius (22,4 °C = 2240)

Tuya thermostat uses different set of attributes and the name used by Tuya developers is simply "heating setpoint", I've used different name to avoid confusion between ZCL and Tuya attributes which use different units. Unit used by Tuya is decidegree Celsius (22,4 °C = 224).

When zigpy updates the value for ZCL Thermostat, (i.e. when user changes setpoint in HA dashboard) it calls `ThermostatCluster.map_attribute()` with `attribute = "occupied_heating_setpoint"`. This method converts the value from centidegree to decidegree and translates ZCL's `occupied_heating_setpoint` to Tuya's `SASSWEL_HEATING_SETPOINT_ATTR`.

On the other hand device reports two temperature values with attrids: `SASSWEL_HEATING_SETPOINT_ATTR` and `SASSWEL_LOCAL_TEMP_ATTR`. First is the current heating setpoint and the second is the local temperature as measured by the temperature sensor on the device. Both reported values are handled by the same event listener implemented in `TuyaThermostatCluster.temperature_change()`. To make code simpler both reports are handled by lines 78:83 and Tuya->ZCL attribute name translation is handled by the `TEMPERATURE_ATTRS` mapping, both values require the same *10 conversion from Tuya to ZCL standard.

zhaquirks/tuya/ts0601_trv_sas.py
blazewicz: There is no need to handle local temperature in `map_attribute` because this value is a read only property of the device. Only the heating setpoint can be changed by the user (HA dashboard) and only this value requires translation from ZCL to Tuya format.

zhaquirks/tuya/ts0601_trv_sas.py
blazewicz: This device has only `Off` and `Heat` modes. Disabling the builtin schedule on each state change assures that it always stays off.

blazewicz:
I feel you buddy, winter has come to Poland as well. I hope you'll be able to enjoy this patch soon, I've been using it ever since I've published this PR and it works fine.

javicalle (approved):
If it has been working for you, I have no objections to the change.
The explanations are convincing, but I think the change may have enough impact that I can't assess by myself.
Simply indicate that if the change presents serious problems for users, it is likely that I will ask you for support to fix it or to go back.

javicalle:
@blazewicz one question, the disable schedule is a 'one time' action, or is needed to call every time?

I was thinking that we can update the attribute in the device binding? That will be enough?

blazewicz:
@javicalle I understand your concerns and of course I'll try to help if any issue arise from my changes.

> I was thinking that we can update the attribute in the device binding? That will be enough?

Thanks for the suggestion, It might be similar to the issue handled in `ParksideTuyaValveManufCluster.bind()`:

https://github.com/zigpy/zha-device-handlers/blob/8d0bcc6e549e3a5c650ec0dadcbff9a35ae5e7e2/zhaquirks/tuya/ts0601_valve.py#L204:L213

```python
    async def bind(self):
        """
        Bind cluster.

        When adding this device tuya gateway issues factory reset,
        we just need to reset the frost lock, because its state is unknown to us.
        """
        result = await super().bind()
        await self.write_attributes({self.attributes_by_name["frost_lock_reset"].id: 0})
        return result
```

 I've moved schedule disable command to `bind()` method, the diff:

```diff
diff --git a/zhaquirks/tuya/ts0601_trv_sas.py b/zhaquirks/tuya/ts0601_trv_sas.py
index a5304a2..4cd5249 100644
--- a/zhaquirks/tuya/ts0601_trv_sas.py
+++ b/zhaquirks/tuya/ts0601_trv_sas.py
@@ -73,6 +73,15 @@ class ManufacturerThermostatCluster(TuyaManufClusterAttributes):
         SASSWEL_LOCAL_TEMP_ATTR: "local_temperature",
     }

+    async def bind(self):
+        """Bind cluster."""
+        # this quirk does not support programmig modes so we force the schedule mode to be always off
+        # more details: https://github.com/zigpy/zha-device-handlers/issues/1815
+        result = await super().bind()
+        _LOGGER.warning("ManufacturerThermostatCluster.bind()")
+        # await self.write_attributes({SASSWEL_SCHEDULE_ENABLE_ATTR: 0})
+        return result
+
     def _update_attribute(self, attrid, value):
         super()._update_attribute(attrid, value)
         if attrid in self.TEMPERATURE_ATTRS:
@@ -131,12 +140,10 @@ class ThermostatCluster(TuyaThermostatCluster):
             return {SASSWEL_HEATING_SETPOINT_ATTR: round(value / 10)}

         if attribute == "system_mode":
-            # this quirk does not support programmig modes so we force the schedule mode to be always off
-            # more details: https://github.com/zigpy/zha-device-handlers/issues/1815
             if value == self.SystemMode.Off:
-                return {SASSWEL_STATE_ATTR: 0, SASSWEL_SCHEDULE_ENABLE_ATTR: 0}
+                return {SASSWEL_STATE_ATTR: 0}
             if value == self.SystemMode.Heat:
-                return {SASSWEL_STATE_ATTR: 1, SASSWEL_SCHEDULE_ENABLE_ATTR: 0}
+                return {SASSWEL_STATE_ATTR: 1}


 class Thermostat_TYST11_c88teujp(TuyaThermostat):
```

Here's my testing procedure:

1. Rollback to vanilla 0.0.85.
2. Reconfigure 3 of my TRV's in ZHA (by adding the same device once again).
3. Confirm that all TRV's are set to the default 16 °C according to the default schedule.
4. Change temperature setpoint to 15 °C.
5. Confirm that setpoint changes back to 16 °C automatically.
6. Apply code changes, restart HA.
7. Repeat points 2 through 4.

I've also added extra debug logs to confirm that request is sent to the device and proper response is received - it is. As strange as it looks, all devices properly report schedule state to be off, but the schedule is still active. Temperature setpoint eventually changes to 16 °C and a clock icon is displayed on the LED display, presumably indicating that current value is set by the schedule.

IMG_20221120_125016

I've tried adding a 3 second delay between call to `.bind()` and sending the request to disable the schedule and I've observed the same result. This is how I've implemented the delay:

```diff
diff --git a/zhaquirks/tuya/ts0601_trv_sas.py b/zhaquirks/tuya/ts0601_trv_sas.py
index 4cd5249..179ad1c 100644
--- a/zhaquirks/tuya/ts0601_trv_sas.py
+++ b/zhaquirks/tuya/ts0601_trv_sas.py
@@ -1,5 +1,6 @@
 """Saswell (Tuya whitelabel) 88teujp thermostat valve quirk."""

+import asyncio
 import logging

 from zigpy.profiles import zha
@@ -77,9 +78,16 @@ class ManufacturerThermostatCluster(TuyaManufClusterAttributes):
         """Bind cluster."""
         # this quirk does not support programmig modes so we force the schedule mode to be always off
         # more details: https://github.com/zigpy/zha-device-handlers/issues/1815
+        _LOGGER.warning("ManufacturerThermostatCluster.bind(): starting bind procedure")
         result = await super().bind()
-        _LOGGER.warning("ManufacturerThermostatCluster.bind()")
-        # await self.write_attributes({SASSWEL_SCHEDULE_ENABLE_ATTR: 0})
+
+        async def disable_schedule():
+            await asyncio.sleep(3)
+            _LOGGER.warning("ManufacturerThermostatCluster.bind(): requesting schedule disable")
+            await self.write_attributes({SASSWEL_SCHEDULE_ENABLE_ATTR: 0})
+
+        asyncio.create_task(disable_schedule())
+
         return result

     def _update_attribute(self, attrid, value):
@@ -100,6 +108,8 @@ class ManufacturerThermostatCluster(TuyaManufClusterAttributes):
             self.endpoint.device.battery_bus.listener_event(
                 "battery_change", 0 if value == 1 else 100
             )
+        elif attrid == SASSWEL_SCHEDULE_ENABLE_ATTR:
+            _LOGGER.warning("Reported schedule state change: %s", value)


 class ThermostatCluster(TuyaThermostatCluster):
```

Then I've rolled back to the version from this PR, restarted HA (reconfigure is not required) and confirmed schedule is disabled on all TRV's.

IMG_20221120_125640

dmulcahey:
Thanks for the work here!

javicalle:
Thanks for testing.

Hilpas:
> > Just to be sure.
> > Are that change indented to be applied to all the devices in that quirk?
> > That will be all these devices:
> > ```
> >             ("_TZE200_c88teujp", "TS0601"),
> >             ("_TZE200_azqp6ssj", "TS0601"),
> >             ("_TZE200_yw7cahqs", "TS0601"),
> >             ("_TZE200_9gvruqf5", "TS0601"),
> >             ("_TZE200_zuhszj9s", "TS0601"),
> >             ("_TZE200_2ekuz3dz", "TS0601"),
> >             ("_TYST11_KGbxAXL2", "GbxAXL2"),
> >             ("_TYST11_c88teujp", "88teujp"),
> >             ("_TYST11_azqp6ssj", "zqp6ssj"),
> >             ("_TYST11_yw7cahqs", "w7cahqs"),
> >             ("_TYST11_9gvruqf5", "gvruqf5"),
> >             ("_TYST11_zuhszj9s", "uhszj9s"),
> > ```
> 
> Fair point, I've assumed these were all clones.
> 
> All except these three share the same configuration in Z2M:
> 
> * `_TZE200_2ekuz3dz`
> * `_TYST11_azqp6ssj`
> * `_TYST11_9gvruqf5`
> 
> Here's the list:
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/96563772e92eb459bf52c10e241c3a725968bcab/devices/saswell.js#L11:L22
> 
> The `_TYST11_azqp6ssj` seems to be yet another sibling of `_TZE200_c88teujp` according to https://zigbee.blakadder.com/RTX_ZB-RT1.html.
> 
> I couldn't find any information regarding `_TYST11_9gvruqf5`.
> 
> The `_TZE200_2ekuz3dz` seems to be a completely different device type and I think it should get a separate quirk. More on that device: https://zigbee.blakadder.com/Beok_TGR85-ZB.html
> 
> I have no way of testing them all of course.

Unfortunately 2ekuz3dz is not working at all using this quirk. Not a single entity is available.

blazewicz:
> Unfortunately 2ekuz3dz is not working at all using this quirk. Not a single entity is available.

@Hilpas thats probably because of missing output clusters, like I said, this one should have a separate quirk on its own since it's not a TRV.

Could you provide a list of supported clusters for this device?

Hilpas:
> > Unfortunately 2ekuz3dz is not working at all using this quirk. Not a single entity is available.
> 
> @Hilpas thats probably because of missing output clusters, like I said, this one should have a separate quirk on its own since it's not a TRV.
> 
> Could you provide a list of supported clusters for this device?

This might help, otherwise i could try to provide some Information later.
https://github.com/zigpy/zha-device-handlers/issues/1721

blazewicz:
But #1721 was opened on August 30-th, almost 3 months before this PR was merged in.

PR #1819, opened 2022-10-11
Add GLEDOPTO GL-C-009 quirk

agners:
Remove the color cluster which is not available on this device.


zhaquirks/gledopto/glc009.py
```diff
@@ -0,0 +1,77 @@
+"""GLEDOPTO GL-C-009 device."""
+from zigpy.profiles import zha, zll
```
javicalle, line 2: Just remove the `zha` import.

Issue #1787, opened 2022-09-25
ZHA Blitzwolf SHP 15 energy monitor outlet, on/off switch wont register in Home Assistant


yfands:
Hello,

Purchaised two Blitzwolf SHP 15, all seems to be ok.
But if I turn the device on/off (on the device itself) Home Assistant wont register it. Other way around, switching in Home assistant, does turn the device on/off.

The Blitzwolf SHP 15 shows up as:
model": “TS011F
device_type”: "0x010a

```
“node_descriptor”: “NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)”,
“endpoints”: {
“1”: {
“profile_id”: 260,
“device_type”: “0x010a”,
“in_clusters”: [
“0x0000”,
“0x0003”,
“0x0004”,
“0x0005”,
“0x0006”,
“0x0702”,
“0x0b04”,
“0xe000”,
“0xe001”
],
“out_clusters”: [
“0x000a”,
“0x0019”
]
}
},
“manufacturer”: “_TZ3000_u5u4cakc”,
“model”: “TS011F”,
“class”: “zhaquirks.tuya.ts011f_plug.Plug”
}
```

all my other devices Blitzwolf SHP 13 are working correct and show up as model: TS0121

Can somebody verify my findings? (and a possible solution) thanks in advance.
Regards Frank

javicalle:
It seems to me a pairing issue. Have you tried to repair the device?

It the problem persist, enable the debug logging and attach the relevant info about your device:
* https://www.home-assistant.io/integrations/zha/#debug-logging

yfands:
Hello Javicalle,

Thanks for the reply,
[...]

PR #1833, opened 2022-10-16
Make `TS011F` plugs inherit `EnchantedDevice`

javicalle:
There are reports that some Blitzwolf SHP 15 plugs (depending on its version) don't report the On/Off update from the physical device. The `herdsman-converters` lib have 3 implementations for the `TS011F` plugs, 2 of them 'implements' the EnchantedDevice approach:
* https://github.com/Koenkk/zigbee-herdsman-converters/blob/96563772e92eb459bf52c10e241c3a725968bcab/devices/tuya.js#L2185

My assumption is that the use of the `EnchantedDevice` cannot be harmful. Also I have not located implementations that could be affected if a device starts to behave in a 'standard' way, *but* there are no guarantees on what will happen and I do not know the scope (of devices) that the change can affect.

If the change is considered too risky, I can try to narrow it down to only the devices that have been reported, but I'll need to check how to match 2 compatible signatures and ensure priority is resolved in the desired order:
```python
        MODEL: "TS011F",
vs 
        MODELS_INFO: [("TS011F", "_TZ3000_u5u4cakc")],
```

Fix: #1787

### DIFF
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -35,7 +35,7 @@ from zhaquirks.tuya import (
 from zhaquirks.tuya.mcu import EnchantedDevice
 
 
-class Plug(CustomDevice):
+class Plug(EnchantedDevice):
     """Tuya plug with restore power state support."""
 
     signature = {


MattWestb:
My experience and is adding it to the class and keep the `MODEL` and i think we need doing this on all new tuya devices Zigbee and MCU.

First  out of the TS011F was the LIDL with updated firmware and more is likely cumming.

MattWestb (approved):
👍

yfands:
Does that mean I can delete ts011f_plug.py from my custom_zha_quirks directory? 

Regards Frank

javicalle:
> Does that mean I can delete ts011f_plug.py from my custom_zha_quirks directory?

Not yet.
The PR has been just merged to the code mainstream. It will be included in the next `zha-quirks` tag/version. There is no a ETA for the version, but for sure will be before the next major HA version (v2022.11).

Sooo, in short: no, but it will be inlcuded in the next HA version.

MattWestb:
The updated quirk is in released HA 2022.10.5.

So feel free updating HA and also before deleting the "temporary quirk" before doing the install then the local is higher priority then the device is being loaded.

After update look for the quirk is being loaded if its one "standard" or one local quirk and also pleas reporting back if its working OK.

Issue #1846, opened 2022-10-20
[BUG] Ikea Starkvind air purifier & air purifier table sensors missing


willliamchan:
**Describe the bug**
All sensors are missing, zha pick up the device as fan only.

**To Reproduce**
N/A

**Expected behavior**
Expecting zha to pick up sensor like PM25, status light etc.

**Screenshots**

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4476, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0007",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0202",
        "0xfc57",
        "0xfc7c",
        "0xfc7d"
      ],
      "out_clusters": [
        "0x0019",
        "0x0400",
        "0x042a"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "IKEA of Sweden",
  "model": "STARKVIND Air purifier table",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2022.10.4",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.10.5",
    "docker": true,
    "arch": "aarch64",
    "timezone": "Europe/London",
    "os_name": "Linux",
[...]

PR #1855, opened 2022-10-22
New Ikea starkvind quirk

javicalle:
New quirk version for the Ikea 'STARKVIND Air purifier table'.
Just cloned the previous one with the new signature.

Also some constants have been refactored.

To fix a `mypy` validation a value check has been added `if fan_mode and fan_mode > 1 and fan_mode < 11:`.

Related to: #1846

### DIFF
--- a/zhaquirks/ikea/starkvind.py
+++ b/zhaquirks/ikea/starkvind.py
@@ -27,9 +27,8 @@ from zhaquirks.const import (
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-from zhaquirks.ikea import IKEA, IKEA_CLUSTER_ID
+from zhaquirks.ikea import IKEA, IKEA_CLUSTER_ID, WWAH_CLUSTER_ID
 
-WWAH_CLUSTER_ID = 0xFC57  # decimal = 64599
 _LOGGER = logging.getLogger(__name__)
 
 
--- a/zhaquirks/ikea/starkvind.py
+++ b/zhaquirks/ikea/starkvind.py
@@ -221,24 +221,14 @@ class IkeaSTARKVIND(CustomDevice):
     }
 
 
-class IkeaSTARKVIND_v2(CustomDevice):
+class IkeaSTARKVIND_v2(IkeaSTARKVIND):
     """STARKVIND Air purifier by IKEA of Sweden."""
-
-    def __init__(self, *args, **kwargs):
-        """Init."""
-        self.pm25_bus = Bus()
-        self.change_fan_mode_bus = Bus()
-        self.change_fan_mode_ha_bus = Bus()
-        super().__init__(*args, **kwargs)
 
     signature = {
         # <SimpleDescriptor endpoint=1 profile=260 device_type=7 (0x0007)
         # device_version=0
         # input_clusters=[0, 3, 4, 5, 514, 64599, 64637] output_clusters=[25, 1024, 1066]>
-        MODELS_INFO: [
-            (IKEA, "STARKVIND Air purifier"),
-            (IKEA, "STARKVIND Air purifier table"),
-        ],
+        MODELS_INFO: IkeaSTARKVIND.signature[MODELS_INFO].copy(),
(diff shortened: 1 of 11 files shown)


MattWestb:
Very nice work done !!

The second IKEA cluster is new and was first seen in Styrbar N Remote and we was not having any name of it so it was called `WWAH_CLUSTER_ID`  (you your imagination for interpreting it).

Then more 3rd gen device is using it i think it better giving it one more "serious" name like `IKEA_CLUSTER2_ID` or somthing else (perhaps `IKEA_CLUSTER1_ID` and `IKEA_CLUSTER2_ID` for `0xFC7C` and `0xFC57`).
I dont have any strong preference but i like the code look more serious and all is better then crappy naming without any meaning in them.

Also is the MFC cluster being used in ZHA for patching some things and can being broken is changing the names ?

Then looking on the Hex it looks it can being one bit error that IKEA have done `0xFC7C` / `0xFC57` ( the second hex is having the 2 bit sett and not set > + - 2).
That is not relevant for us then the devices is on the market and we can changing the data setting in them.

javicalle:
> Also is the MFC cluster being used in ZHA for patching some things and can being broken is changing the names ?

I don't think that the name is used, but I can check it.

I would left the cluster rename for a new PR.

TheJulianJES:
> The second IKEA cluster is new and was first seen in Styrbar N Remote and we was not having any name of it so it was called `WWAH_CLUSTER_ID` (you your imagination for interpreting it).

FYI, ``0xFC57`` should be the cluster ID of the "Works with all Hubs" (WWAHu) "standardized" Zigbee cluster which is now mainly used if the device is directly connected to a Alexa. I'm not sure if the IKEA devices actually implement that properly though, as I've never tried it.

zhaquirks/ikea/starkvind.py
```diff
@@ -220,3 +219,90 @@ def __init__(self, *args, **kwargs):
...
+
+class IkeaSTARKVIND_v2(CustomDevice):
+    """STARKVIND Air purifier by IKEA of Sweden."""
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        self.pm25_bus = Bus()
+        self.change_fan_mode_bus = Bus()
+        self.change_fan_mode_ha_bus = Bus()
+        super().__init__(*args, **kwargs)
```
TheJulianJES, line 232: I think it might be a good idea if ``IkeaSTARKVIND_v2`` would extend the original ``IkeaSTARKVIND`` quirk class instead of ``CustomDevice``.
Then, the ``__init__`` method could be removed in the v2 class, so it isn't duplicated in the quirk file.

MattWestb:
Then i think the `"Works with all Hubs"`  /  `0xFC57` shall being removed from IKEA part and moved to constants or some other place then its not manufacture (IKEA) related and can being used for more device / manufactures.

Not urgent for this PR but best fixing so not getting more python pasta slate in the future.

Thanks for explaining **TheJulianJES** and sharing your knowledge !!!

TheJulianJES:
> Then i think the `"Works with all Hubs"` / `0xFC57` shall being removed from IKEA part and moved to constants

Yeah, probably. I'm not sure if any other devices even have that cluster currently.
For now, moving it to the main IKEA file is a lot better already than having them in every quirk individually.

MattWestb:
I agree but its not very urgent like getting devices working that have getting its firmware updated.

We shall putting it on the ToDo list and getting this PR merged so the users is getting full functionality in there devices.

zhaquirks/ikea/starkvind.py
dmulcahey: +1

javicalle: Done

dmulcahey:
> So we feel comfortable keeping the name `WWAH_CLUSTER_ID`?

absolutely. I am looking for the SI document that outlines it so we can implement all of it as well.

javicalle:
Thanks

PR #1857, opened 2022-10-23
XBee: Use dict for response event data

Shulyaka:
This is a breaking change that might require some modifications in user configuration (automations or trigger-based sensors).
But without this change those automations are probably broken anyway for any modern HA release (I think since 2022.4.1).

Since https://github.com/home-assistant/core/pull/69631 ZHA now requires the event data to be dict and does not accepts plain strings (see https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/core/channels/base.py#L405).
This PR fixes XBee quirk to provide that. The data for Serial data event is sent as {"data": ...} and the AT command responses are sent as {"response": ...}.
Also updating some outdated info in readme for PWM.

### DIFF
--- a/tests/test_xbee.py
+++ b/tests/test_xbee.py
@@ -347,7 +347,7 @@ async def test_remote_at_non_native(
     ].add_listener(listener)
 
     def mock_at_response(*args, **kwargs):
-        """Simulate remote AT comand response from device."""
+        """Simulate remote AT command response from device."""
         xbee3_device.handle_message(
             XBEE_PROFILE_ID,
             XBEE_AT_RESPONSE_CLUSTER,
@@ -472,7 +472,7 @@ async def test_remote_at_tx_failure(zigpy_device_from_quirk):
     ].add_listener(listener)
 
     def mock_at_response(*args, **kwargs):
-        """Simulate remote AT comand response from device."""
+        """Simulate remote AT command response from device."""
         xbee3_device.handle_message(
             XBEE_PROFILE_ID,
             XBEE_AT_RESPONSE_CLUSTER,
--- a/tests/test_xbee.py
+++ b/tests/test_xbee.py
@@ -28,7 +28,9 @@ async def test_basic_cluster(zigpy_device_from_quirk):
     """Test Basic cluster."""
 
     xbee3_device = zigpy_device_from_quirk(XBee3Sensor)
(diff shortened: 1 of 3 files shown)


dmulcahey:
the referenced PR restored list functionality. This should not be necessary. What is the issue you are trying to fix. Please include debug logs from your automation that isn't working.

Shulyaka:
We are trying to send a plain number or a string, not list.
Here is an example log, I think it can explain better than me:
```
DEBUG [zigpy_xbee.uart] Frame received: b'\x97\xe4\x00\x13\xa2\x00A\xa0~\x1aH\x7fTP\x00\x005'
DEBUG [zigpy_xbee.api] Frame received: remote_at_response
DEBUG [zigpy_xbee.api] Remote AT command response from: (228, 00:13:a2:00:41:a0:7e:1a, 0x487f, b'TP', 0, b'\x005')
WARNING [zigpy.util] Error calling listener <bound method ZigbeeChannel.zha_send_event of <homeassistant.components.zha.core.channels.general.LevelControlClientChannel object at 0xacf25e68>> with args ('tp_command_response', 53): TypeError("Unexpected zha_send_event 'tp_command_response' argument: 53")
DEBUG [zigpy.util] Error calling listener <bound method ZigbeeChannel.zha_send_event of <homeassistant.components.zha.core.channels.general.LevelControlClientChannel object at 0xacf25e68>> with args ('tp_command_response', 53)
  File "/srv/homeassistant/deps/lib/python3.10/site-packages/zigpy/util.py", line 55, in listener_event
    result.append(method(*args))
  File "/usr/lib/python3.10/site-packages/homeassistant/components/zha/core/channels/base.py", line 405, in zha_send_event
    raise TypeError(f"Unexpected zha_send_event {command!r} argument: {arg!r}")
TypeError: Unexpected zha_send_event 'tp_command_response' argument: 53
```

dmulcahey:
can you add some tests for this? We are trying to get the test coverage up for all custom logic like this.

Shulyaka:
Will try

Shulyaka:
Done, added some tests (and done some other small fixes along the way)

dmulcahey:
> Done, added some tests (and done some other small fixes along the way)

Thanks!

dmulcahey (approved):

PR #1859, opened 2022-10-23
Fix a few command() return objects

javicalle:
I believe that there are a few incorerct return objects for the `async def command()` method.

Maybe related to:
- #1790
- #1768


javicalle:
Maybe also related:
* https://github.com/zigpy/zha-device-handlers/blob/f2822031a464bfdf766dfccebf667a502622ebbb/zhaquirks/xbee/__init__.py#L297

dmulcahey (approved):

javicalle:
Thanks David.

PR #1864, opened 2022-10-25
Debug `LevelControl` kwargs parameter

javicalle:
Add the `kwargs` parameter from the command() method to the debug log.


dmulcahey (approved):

javicalle:
Thanks David

PR #1871, opened 2022-10-29
Pass `**kwargs` to super call in `F000LevelControlCluster`

javicalle:
Probably not needed because the call is not used but...


dmulcahey (approved):

javicalle:
Thanks David

Issue #1877, opened 2022-11-01
[Device Support Request] TS0601 - _TZE200_bkkmqmyo - Hiking DDS238-2 DIN Circuit Breaker with Energy Monitoring


flaktivated:
**Is your feature request related to a problem? Please describe.**
When I attempt to Pair the Hiking DDS238-2 Zigbee Metering Circuit Breaker to Home Assistant via ZHA it is unable to identify the device and cannot use its features.

**Describe the solution you'd like**
I would appreciate if I could get some assistance to get the new DIN 2-Pole Circuit Breaker with Power Meter onboarded to be supported in ZHA.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_bkkmqmyo",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant Container",
    "version": "2022.10.2",
    "dev": false,
    "hassio": false,
    "virtualenv": false,
    "python_version": "3.10.5",
    "docker": true,
    "arch": "x86_64",
    "timezone": "Australia/Sydney",
    "os_name": "Linux",
[...]

Issue #1700, opened 2022-08-21
[Device Support Request] Tuya Zigbee smart energy meter DDS238-2 Zigbee with a new manufacturer variant (_TZE200_bkkmqmyo)


mmaccaroni:
Hi, I purchased this device which was markes as working with zha, unfortunatrly I got one with a new manufacturer, and, upon pairing, no entities appear.
This is the signature

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_bkkmqmyo",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

Tried to adapt other quirks that were shared in the past  with no success

darckly:
Hello @mmaccaroni ,

I assume the new device is the single phase power metter, correct?
What you tried to change in the quirk, and how installed it?

mmaccaroni:
> Hello @mmaccaroni ,
> 
> I assume the new device is the single phase power metter, correct? What you tried to change in the quirk, and how installed it?

Hi darckly, thanks for your reply. Yes it is the single phase.
I set up a folder for custom quirks and pointed to in my configuration.yaml.
The change I made iwas the manufacturer string as per signature, but nothing happened. I truly have no idea what am I doing wrong.

darckly:
[...]

PR #1878, opened 2022-11-01
TS0601 - _TZE200_bkkmqmyo - Full Support excluding Alarm and its Configuration (#1877)(#1700)

flaktivated:
Added Support for On-Off Switch.


flaktivated:
Linking PR to existing support issue raised:
https://github.com/zigpy/zha-device-handlers/issues/1700

Closed duplicate Issue:
https://github.com/zigpy/zha-device-handlers/issues/1877

flaktivated:
Testing Evidence: 
image
image

zhaquirks/tuya/ts0601_din_power.py
```diff
@@ -150,3 +233,60 @@ def __init__(self, *args, **kwargs):
...
+                OUTPUT_CLUSTERS: [Time.cluster_id, Ota.cluster_id],
+            },
+            16: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.SMART_PLUG,
+                INPUT_CLUSTERS: [
+                    TuyaOnOff,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
```
javicalle, line 290: Why setting the `TuyaOnOff` cluster in another endpoint? Since the `DEVICE_TYPE` is the same, couldn't it go to endpoint 1?

flaktivated: Originally it did live under endpoint 1, but the switch did not toggle. I looked into the `TuyaOnOff` class and found the following code:
```python
            cmd_payload.command_id = TUYA_CMD_BASE + self.endpoint.endpoint_id
```
https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L519

`TUYA_CMD_BASE` has the value of `0x0100`
https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L110

My device has the On/Off Switch accessible under attribute: `0x0110`
So the only way to get  `cmd_payload.command_id` to be equal to my switch's attribute was put it under endpoint 16 (`0x0010`)

javicalle: 🤦🏻‍♂️ That makes me wonder why I didn't migrate the quirk to the MCU classes at the time...

Well, the implementation can be left as is, but you could also tune it here:
```python
         if attrid == HIKING_DIN_SWITCH_ATTR:
             self.endpoint.device.switch_bus.listener_event(SWITCH_EVENT, 0x0101, value)
```

flaktivated: Is there anything that I need to do to get this PR merged?

dmulcahey:
@javicalle is this good to go in your opinion?

javicalle:
Not too familiar with all the `Metering` and `ElectricalMeasurement` but the code looks good to me.

dmulcahey:
@flaktivated thanks for the contribution!! @javicalle thanks as always for all of the help and support! The community owes you a debt of gratitude for the amount of time you have spent cleaning up after Tuya!!

Issue #1884, opened 2022-11-02
[Device Support Request] Configurable Z-Stack Router firmware radio `transmit_power`


ruimarinho:
**Is your feature request related to a problem? Please describe.**
I'm not entirely sure this requires a quirk, but Koenkk has just released a version of the Z-Stack router firmware with support for remotely setting the CC2652P radio amplifier gain (https://github.com/Koenkk/Z-Stack-firmware/commit/3d43a50acdc09d6ff12605de4a5d104389c69e85#diff-f4872ebd913c23b0dc0d04319df88a7d1d21179f1ef005e0b9d5500f890fef0fR186-R201). It is part of the basic cluster.

**Describe the solution you'd like**
It would be great to be able to set the transmit power of the radio via a slider input.

TheJulianJES:
This should be available in the December HA 2022.12.0 release through the "Manage Zigbee Device" -> "Clusters" UI.
It's possible that a config entity in ZHA could also be added, but there are some things to consider with older firmwares.

If you need it now, you should be able to add the ``zhaquirks/texasinstruments/router.py`` file from https://github.com/zigpy/zha-device-handlers/pull/1886 as a custom quirk. But please remove it when the change is available in HA (but that'll still take some time). I'll try to update this issue when it gets merged into HA.

Hedda:
FYI, the premium version of PTVO Zigbee Configurable Firmware has the same function but via its ”RF signal level” sensor -> https://ptvo.info/zigbee-configurable-firmware-features/rf-signal-level-control/

> [RF signal level control](https://ptvo.info/zigbee-configurable-firmware-features/rf-signal-level-control/)
> 
> The firmware can control RF signal strength for all chips with a power amplifier. This feature allows you to find tune your mesh network and exclude situations when a device constantly jumps from one router to another because a signal level from both router is almost identical.
> 
> The firmware also allows you to control signal level remotely in run-time.
> 
> ⚠️ THIS FEATURE IS AVAILABLE IN THE [PREMIUM VERSION](https://ptvo.info/zigbee-configurable-firmware-features/premium/)
> 
> To control signal level, you should add the “RF signal level” sensor on any output, using any [virtual pin](https://ptvo.info/zigbee-configurable-firmware-features/gpio/#virtual). The “Signal strength level” parameter defines an initial value after flashing or resetting, where 0 is the lowest signal level and 20 is the highest signal level. Please note, this number is not a precise number of signal power in dBm.
> 
[...]

PR #1886, opened 2022-11-03
Allow configuring transmit power on TI Z-Stack routers

TheJulianJES:
### Changes
Z-Stack router firmware ``20221102`` and above adds a ``transmit_power`` attribute on the ``Basic cluster``.
This quirk makes that attribute accessible through ZHA.

~~I also added ``TubesZB``/``tubeszb.router`` to the signature, even though it doesn't support the configurable attribute yet, but I'm sure he'll add it in the future. cc @tube0013 (still on the Z-Stack ``dev`` branch as of right now though)~~
Edit: It's possible that no new updates will be made for "TubesZB" custom router firmware and as current versions don't support the configurable attribute, I removed it from the signature entirely.

(A config entity in ZHA could also be added. For older versions, a ~~warning~~ debug message will be printed on startup saying that the attribute isn't supported for routers running older firmware.)
(PR for ZHA config entity: https://github.com/home-assistant/core/pull/81520)

Closes https://github.com/zigpy/zha-device-handlers/issues/1884
Z2M converter commit: https://github.com/Koenkk/zigbee-herdsman-converters/commit/8fb1ba4c232e451af1e9953b8c476c2f4b2ef561
Z-Stack ``firmware.patch`` commit: https://github.com/Koenkk/Z-Stack-firmware/commit/3d43a50acdc09d6ff12605de4a5d104389c69e85
Z-Stack issue: https://github.com/Koenkk/Z-Stack-firmware/issues/341

The quirk is tested and works.

### Testing
This quirk will likely be available in Home Assistant 2022.12.0 or later.
If you want to test it now, download the [router.py](https://raw.githubusercontent.com/zigpy/zha-device-handlers/55802e2b67af1877189186683451fdfb9429309f/zhaquirks/texasinstruments/router.py) file and put it in your configured ``custom_zha_quirks`` directory. Then, restart Home Assistant.
For now, you can change the transmit power using the "Manage Zigbee Device" UI -> "Clusters".
Select ``BasicCluster`` and ``transmit_power``. Then, you can read the attribute. If your firmware supports it, it'll return a number (like ``5`` or ``9``). If your firmware is still unsupported, it'll return ``None`` and you'll need to update your router firmware.
To change the transmit power, simply replace the number in the upper text field (leave the lower text field empty) and then click "Set Zigbee attribute". You can re-read the attribute to verify that the change worked.

Also, make sure to remove the custom quirk as soon as it's merged into Home Assistant.

### DIFF
--- a/zhaquirks/texasinstruments/__init__.py
+++ b/zhaquirks/texasinstruments/__init__.py
@@ -1,0 +1,1 @@
+"""Texas Instruments devices."""
--- a/zhaquirks/texasinstruments/router.py
+++ b/zhaquirks/texasinstruments/router.py
@@ -25,7 +25,7 @@ class TiRouter(CustomDevice):
     """Texas Instruments Z-Stack router device."""
 
     signature = {
-        MODELS_INFO: [("TexasInstruments", "ti.router"), ("TubesZB", "tubeszb.router")],
+        MODELS_INFO: [("TexasInstruments", "ti.router")],
         ENDPOINTS: {
             8: {
                 PROFILE_ID: zha.PROFILE_ID,
--- a/zhaquirks/texasinstruments/router.py
+++ b/zhaquirks/texasinstruments/router.py
@@ -1,0 +1,69 @@
+"""Texas Instruments Z-Stack router device."""
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
+from zigpy.zcl.clusters.general import Basic, GreenPowerProxy, Identify
+
+from zhaquirks import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+
+
+class BasicCluster(CustomCluster, Basic):
+    """Texas Instruments Basic cluster."""
+
+    attributes = Basic.attributes.copy()
+    attributes[0x1337] = ("transmit_power", t.int8s, False)
+
+
+class TiRouter(CustomDevice):
+    """Texas Instruments Z-Stack router device."""
+
+    signature = {
+        MODELS_INFO: [("TexasInstruments", "ti.router"), ("TubesZB", "tubeszb.router")],
+        ENDPOINTS: {
+            8: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: 0x00FF,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                ],
+            },
+            242: {
+                PROFILE_ID: 41440,
+                DEVICE_TYPE: 97,
+                INPUT_CLUSTERS: [],
+                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
+            },
+        },
+    }
+    replacement = {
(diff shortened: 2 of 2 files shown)


tube0013:
NIce, I actually just flashed a router a few minutes ago to look at this.  Honestly not sure I will be continuing to compile my own fw.

TheJulianJES:
Should I remove your model ID from the signature for now? (can also be added back later of course)

tube0013:
Yeah, remove it for now. thanks!

tube0013:
Quirk works for me🎉

Hedda:
FYI, the premium version of PTVO Zigbee Configurable Firmware has the same function when built Zigbee Router firmware for CC2652P and CC1352P with it and enable its ”RF signal level” sensor -> https://ptvo.info/zigbee-configurable-firmware-features/rf-signal-level-control/

> ## RF signal level control
> 
> The firmware can control RF signal strength for all chips with a power amplifier. This feature allows you to find tune your mesh network and exclude situations when a device constantly jumps from one router to another because a signal level from both router is almost identical.
> 
> The firmware also allows you to control signal level remotely in run-time.
> 
> ⚠️ THIS FEATURE IS AVAILABLE IN THE [PREMIUM VERSION](https://ptvo.info/zigbee-configurable-firmware-features/premium/)
> 
> To control signal level, you should add the “RF signal level” sensor on any output, using any [virtual pin](https://ptvo.info/zigbee-configurable-firmware-features/gpio/#virtual). The “Signal strength level” parameter defines an initial value after flashing or resetting, where 0 is the lowest signal level and 20 is the highest signal level. Please note, this number is not a precise number of signal power in dBm.
> 
> RF signal level configuration
> To control signal strength remotely, you should write the corresponding level value to the configured endpoint (e.g., output 5 will work on endpoint 5).
> 
> ⚠️ Please be careful because this action has an immediate effect, and your device may become unresponsive if a signal strength is too small.
> 
> You can also create a [custom converter](https://ptvo.info/zigbee-configurable-firmware-features/expert/) for Z2M on the “Expert” tab that adds the corresponding option field to exposes.
> 
> The firmware does not send the configured signal strength value with periodic reports because this value is constant. If you need to get this value, you may use the “Refresh” button in exposes or the “get” via MQTT.
> 
> Zigbee2MQTT commands
```
read – immediately reads the signal strength value

Topic: zigbee2mqtt/[friedly_name]/get/[channel]
Channel: l1, l2, l3 … l16
Payload: 1
Returns the channel state: {"[channel]": value}
write – writes the signal strength value

Topic: zigbee2mqtt/[friedly_name]/set
Payload: {"[channel]": value}
Channel: l1, l2, l3 … l16
```

ErikApption:
I have HA 2023.4.4 but I don't see transmit_power for my Sonoff CC2652P - is this expected?

TheJulianJES:
@ErikApption Make sure you're running a somewhat recent Z-Stack router release on that stick.
You might need to re-pair it after updating.

ErikApption:
> @ErikApption Make sure you're running a somewhat recent Z-Stack router release on that stick. You might need to re-pair it after updating.

I do have last week's firmware installed but I did install the firmware after configuring the radio. How would I re-pair? Would this be a migrate radio thing?

TheJulianJES:
Configuring transmit power is for Z-Stack **routers**, not coordinators.

If you want to change transmit power for your coordinator, then have a look at the zigpy repo.
It can be done via YAML.

TheJulianJES:
Check this: https://github.com/zigpy/zigpy-znp#configuration

Only paste in the first couple of rows that you actually need.

Most of the times, you shouldn't really increase transmit power though, as it'll only "shout louder", but won't hear other devices better.

Issue #1724, opened 2022-08-30
[Device Support Request] centechia (and others?) Temperatur/Humidity sensor with display


bsafh:
**Is your feature request related to a problem? Please describe.**
The device pairs, and battery, temperature, humidity entities show up as well, but battery level is "unknown", both temperature and humidity are always zero (on several devices of this type)

**Describe the solution you'd like**
I would like to use this sensor for temperature and humidity readings.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0302",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0402",
        "0x0405"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_qoy0ekbd",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant Supervised",
    "version": "2022.8.7",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.10.5",
    "docker": true,
    "arch": "x86_64",
    "timezone": "Europe/Berlin",
    "os_name": "Linux",
    "os_version": "5.10.0-17-amd64",
    "supervisor": "2022.08.5",
    "host_os": "Debian GNU/Linux 11 (bullseye)",
    "docker_version": "20.10.17",
    "chassis": "vm",
    "run_as_root": true
  },
[...]

PR #1887, opened 2022-11-04
New ts0601_sensor square

javicalle:
Add the new ts0601 sensor definition `_TZE200_qoy0ekbd`.
To avoid duplicate code, I have moved the Tuya -> Zigbee value conversion from the `dp_to_attribute` to the `TuyaRelativeHumidity` cluster. Not happy with the solution but I don't find any better one to avoid duplicate the `TemperatureHumidityManufCluster`.

Now the device can define a `RH_MULTIPLIER` to apply to value. If not defined, defaults to `100`.

Fixes: #1670 
Fixes: #1724


dmulcahey (approved):

javicalle:
Thanks David

PR #1890, opened 2022-11-04
Update Inovelli blue series quirk

dmulcahey:
This PR moves the Inovelli animation type enumerations into quirks since they may change frequently and it adds a new parameter to the Inovelli mfg cluster.

### DIFF
--- a/zhaquirks/inovelli/types.py
+++ b/zhaquirks/inovelli/types.py
@@ -24,8 +24,8 @@ class AllLEDEffectType(types.enum8):
     Medium_Blink = 0x0F
     Slow_Chase = 0x10
     Fast_Chase = 0x11
-    Slow_Siren = 0x12
-    Fast_Siren = 0x13
+    Fast_Siren = 0x12
+    Slow_Siren = 0x13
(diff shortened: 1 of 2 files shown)


zhaquirks/inovelli/types.py
```diff
@@ -0,0 +1,44 @@
...
+    Medium_Falling = 0x0A
+    Fast_Falling = 0x0B
+    Slow_Rising = 0x0C
+    Medium_Rising = 0x0D
+    Fast_Rising = 0x0E
+    Medium_Blink = 0x0F
+    Slow_Chase = 0x10
+    Fast_Chase = 0x11
+    Slow_Siren = 0x12
+    Fast_Siren = 0x13
```
codyhackw, line 28: Fast and Slow ended up swapped, so will be:
Fast_Siren = 0x12
Slow_Siren = 0x13

dmulcahey: Just for Siren?

codyhackw: Yep, all other effects line up correctly.

PR #1895, opened 2022-11-05
Add basic support for Moes ZSS-QY-HP

blakadder:
Add support for `_TZE200_wukb7rhc` which is the model I own and could verify as working

https://zigbee.blakadder.com/Moes_ZSS-QY-HP.html


dmulcahey (approved):

dmulcahey:
Thanks!

PR #1899, opened 2022-11-06
Add support for new 3 gang Tuya switch

bramdan:


zhaquirks/tuya/ts001x.py
```diff
@@ -308,6 +308,104 @@ class TuyaTripleNoNeutralSwitch(TuyaSwitch):
...
+        # maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264,
+        # maximum_outgoing_transfer_size=82, descriptor_capability_field=0)
+        # "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4098,
+        # maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264,
+        # maximum_outgoing_transfer_size=82, descriptor_capability_field=0)"
+        MODEL: "TS0013",
+        ENDPOINTS: {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=100
+            # device_version=1
+            # input_clusters=[0, 4, 5, 6]
```
javicalle, line 325: `# input_clusters=[0, 3, 4, 5, 6]`

zhaquirks/tuya/ts001x.py
```diff
@@ -308,6 +308,104 @@ class TuyaTripleNoNeutralSwitch(TuyaSwitch):
     }
 
 
+class TuyaTripleNoNeutralSwitchV2(EnchantedDevice, TuyaSwitch):
+    """Tuya 3 gang no neutral light switch (v2)."""
```
javicalle, line 312: I want to propose call as variants (opposite to v2 that seems to be a newer version).
```python
class TuyaTripleNoNeutralSwitchVar03(EnchantedDevice, TuyaSwitch):
    """Tuya 3 gang no neutral light switch (variant 03)."""
```

javicalle (changes requested):
There are coming new versions (or variants) of TS001X devices

zhaquirks/tuya/ts001x.py
bramdan: nice catch

zhaquirks/tuya/ts001x.py
bramdan: makes sense, I took it a step further, what do you think?

zhaquirks/tuya/ts0013.py
```diff
@@ -0,0 +1,326 @@
...
+                    Scenes.cluster_id,
+                    TuyaZBOnOffAttributeCluster,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
+        },
+    }
+
+
+class TuyaTripleNoNeutralSwitchVar(TuyaSwitch):
```
javicalle, line 218: Maybe TuyaTripleNoNeutralSwitchVar03(TuyaSwitch)

javicalle:
Mixed feelings about splitting the file. Tuya folder already has a lot of files, but in the other hand file is growing up unordered... 
I believe that I would prefer keeping all in the same file, but I don't have a strong opinion.

dmulcahey:
Let's not split the file for now.

dmulcahey:
@bramdan can we put these back in the other file for now please? We can reassess splitting at a later date.

dmulcahey:
> @bramdan can we put these back in the other file for now please? We can reassess splitting at a later date.

@bramdan any chance you can do this?

ufanders:
Hi, is this quirk usable as-is? I just received a bunch of these switches and all relays fire at once! Would love to tweak whatever is necessary locally if I can. Thanks!

MattWestb:
If your device is having the same signature it is very likely fixing your problem with the device then its casting the tuya magic that shall fixing it.

TheJulianJES:
@bramdan @ufanders Can either of you provide the device signature (or diagnostic information) for these unsupported Tuya switches?

TheJulianJES:
Rebased to fix merge conflict. Also merged all Tuya switches into one file again.
Fixed the comment for clusters and renamed the class (although the "naming scheme" completely varies in that file already, so I'm not sure what I chose is the best option).

Should be ready for another review now.

javicalle:
Hi Julian, 
if I haven't missed anything, `TuyaTripleNoNeutralSwitchVar2` is exactly the same as `TuyaTripleNoNeutralSwitch_2` (added in #2091?)

TheJulianJES:
Yep, thanks for catching that! We should probably have a test for duplicate quirks

Closing as this is already implemented then.

zhaquirks/tuya/ts0013.py
chuksum, line 1: This fixed my issue with the BSEED 3 GANG no neutral wire light switch.

PR #1919, opened 2022-11-11
Xbee: Code refactoring

Shulyaka:
A little refactoring: make cluster classes non-internal for clarity and move types to a separate file.


Shulyaka:
Ok, I've decided to separate the refactoring and the new functionality into two separate PRs.
This is for the refactoring only, no new functionality.

PR #1929, opened 2022-11-17
Update .pre-commit-config.yaml

javicalle:
Seems that flake8 has been moved to github, but I can't tell for sure.

In the github repo they state that it is a licit change:
* https://github.com/PyCQA/flake8/issues/1737


dmulcahey:
Thanks

Issue #1925, opened 2022-11-15
[Device Support Request] Legrand Wireless Dimmer Switch (battery percentage = unknown)


stevenwfoley:
**Is your feature request related to a problem? Please describe.**
Device does not display remaining battery percentage.

**Describe the solution you'd like**
I would like to see the remaining battery percentage displayed for the device.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=1, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4129, maximum_buffer_size=89, maximum_incoming_transfer_size=63, server_mask=10752, maximum_outgoing_transfer_size=63, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0104",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x000f",
        "0x0020",
        "0xfc01"
      ],
      "out_clusters": [
        "0x0000",
        "0x0003",
        "0x0006",
        "0x0008",
        "0x0019",
        "0xfc01"
      ]
    }
  },
  "manufacturer": " Legrand",
  "model": " Remote dimmer switch",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2022.11.2",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.10.7",
    "docker": true,
    "arch": "x86_64",
    "timezone": "America/Detroit",
    "os_name": "Linux",
    "os_version": "5.15.74",
    "supervisor": "2022.10.2",
    "host_os": "Home Assistant OS 9.3",
    "docker_version": "20.10.18",
    "chassis": "vm",
    "run_as_root": true
  },
  "custom_components": {
[...]

PR #1934, opened 2022-11-18
Add Legrand remote dimmer

javicalle:
New quirk for battery conversion of Legrand ` Remote dimmer switch`.
Also applies the `LegrandCluster` but not sure if usefull here.

Fixes: #1925

### DIFF
--- a/tests/test_legrand.py
+++ b/tests/test_legrand.py
@@ -1,0 +1,28 @@
+"""Tests for Legrand."""
+import pytest
+
+import zhaquirks
+
+zhaquirks.setup()
+
+
+@pytest.mark.parametrize(
+    "voltage, bpr",
+    (
+        (32.00, 200),  # over the max
+        (30.00, 200),  # max
+        (29.0, 160),
+        (28.0, 120),
+        (27.50, 100),  # 50%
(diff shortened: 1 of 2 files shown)


MattWestb:
Then getting more Legrand quirks i think the `LegrandCluster` shall be moved to the init and cleaning the current quirk that is using different naming / methods for defining it but is not needed for this update. 👍

javicalle:
I thought about the same thing, but for now I'm going to leave this way.


> cleaning the current quirk that is using different naming / methods for defining

I'm not sure which naming / methods you mean. Maybe the `LegrandCluster` attributes? The truth is that I have not looked at its function...

MattWestb:
> naming / methods

The current quirk is using:
Sign: `MANUFACTURER_SPECIFIC_CLUSTER_ID,`
Replace: `LegrandCluster,
`

Updated is using:
Sign: `LegrandCluster.cluster_id,`
Replace: `LegrandCluster,`

Then later perhaps making the definition in the INIT then making all using your naming that i like more then the current one :-))

Its no problem, get it merged and the user is getting there deice working OK.

BTW. Good work done !!

Issue #1944, opened 2022-11-20
[Device Support Request] Tuya TZE200 yjjdcqsq


tho071088:
**Cheap Tuya ZigBee Temperature Sensor.**
Device is detected but no sensors are fond.
[Link to device on Aliexpress ](https://a.aliexpress.com/_EuvXvTP) 

**Temperature, Humidity and Battery Status**
The three sensors are displayed on HA.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_yjjdcqsq",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
          "id": 81
        },
        "profile_id": 260,
        "in_clusters": {
          "0x0004": {
            "endpoint_attribute": "groups",
            "attributes": {},
            "unsupported_attributes": {}
          },
          "0x0005": {
            "endpoint_attribute": "scenes",
            "attributes": {},
            "unsupported_attributes": {}
          },
          "0xef00": {
            "endpoint_attribute": null,
            "attributes": {},
            "unsupported_attributes": {}
          },
          "0x0000": {
            "endpoint_attribute": "basic",
            "attributes": {
              "0x0001": {
                "attribute_name": "app_version",
                "value": 72
              },
[...]

PR #1951, opened 2022-11-21
Add `_TZE200_yjjdcqsq` Tuya sensor

javicalle:
Add new device signature for the `_TZE200_yjjdcqsq` sensor

Fix: #1944


javicalle:
Thanks David.

sopelt:
Does this device work properly? I paired two ... but they both are stuck with the red LED on and with unknown states.

sopelt:
Note to self and other impatient people: just give the device 15-30 min ... the LED will go out and values will start reporting :D

PR #1961, opened 2022-11-23
New Zonnsmart TS0601 TRV implementation, add scheduling 

gmsoft-tuxicoman:
**This PR does the following :**

 - FIx OnOff clusters when you have multiple TRV (only the last one added was working)
 - Use updates() for the ZONNSMARTManufCluster attributes
 - Fix local_temperature_calibration read back from the TRV
 - Implement the holiday_start_stop attribute
 - Implement scheduling support in the ZONNSMARTManufCluster
 - Enable time setting on the TRV

**Scheduling details :**

Schedules can be read from the schedule_<day> attributes. They cannot be queried from the TRV directly, you must first use the *Online* button to force the TRV to send the values or set the schedule using the schedule_set attribute. Until you do so, the schedules will appear empty.

The default schedule for a day on my TRV is the following : 
`06:00/17.0 12:00/21.0 14:00/17.0 17:00/21.0 24:00/17.0`
The temperature will vary in the following way : 
 - 00:00 -> 06:00 : 17C
 - 06:00 -> 12:00 : 21C
 - 12:00 -> 14:00 : 17C
 - 14:00 -> 17:00 : 21C
 - 17:00 -> 24:00 : 17C

You can set the schedule for a single day or multiple day with the following syntax : 
`<day> <period> <period> ...`
For example, writing the value `mon 09:00/17 21:00/21.5 24:00/17` to the schedule_set attribute will set the temperature in the morning and evening to 17C and 21.5C between 9AM and 9PM.

`day` can be any off `mon`, `tue`, `wed`, `thu`, `fri`, `sat`, `sun`, `mon-fri`, `sat-sun`, `mon-sun`. The last 3 will set the schedule for multiple days at once.

**Holiday start - stop :**
The holiday_start_stop will not be populated until you use the *Online* button.
The syntax is the following : `2022/12/01 01:01 | 2022/12/31 12:00`.
In the above example the whole month of December 2022 will be set as holiday and the TRV will switch automatically to the holiday temperature set in the holiday_temperature attribute.
Once the end date is reached, it will switch back to the scheduled/auto mode.

HA PR: https://github.com/home-assistant/core/pull/87546

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -470,7 +470,7 @@ async def test_zonnsmart_state_report(zigpy_device_from_quirk, quirk):
         tuya_cluster.handle_message(hdr, args)
 
     assert len(thermostat_listener.cluster_commands) == 0
-    assert len(thermostat_listener.attribute_updates) == 11
+    assert len(thermostat_listener.attribute_updates) == 13
     assert thermostat_listener.attribute_updates[0] == (0x0000, 2110)  # TEMP
     assert thermostat_listener.attribute_updates[1] == (0x0012, 2050)  # TARGET
     assert thermostat_listener.attribute_updates[4] == (0x0014, 1700)  # HOLIDAY
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -472,14 +472,14 @@ async def test_zonnsmart_state_report(zigpy_device_from_quirk, quirk):
     assert len(thermostat_listener.cluster_commands) == 0
     assert len(thermostat_listener.attribute_updates) == 13
     assert thermostat_listener.attribute_updates[0] == (0x0000, 2110)  # TEMP
-    assert thermostat_listener.attribute_updates[1] == (0x0012, 2050)  # TARGET
-    assert thermostat_listener.attribute_updates[4] == (0x0014, 1700)  # HOLIDAY
-    assert thermostat_listener.attribute_updates[5] == (0x0010, 110)  # OFFSET
-    assert thermostat_listener.attribute_updates[6] == (0x0025, 0)  # MANUAL
-    assert thermostat_listener.attribute_updates[7] == (0x4002, 1)
-    assert thermostat_listener.attribute_updates[8] == (0x0025, 1)  # SCHEDULE
-    assert thermostat_listener.attribute_updates[9] == (0x4002, 0)
-    assert thermostat_listener.attribute_updates[10] == (0x001C, 4)  # HEAT ON
+    assert thermostat_listener.attribute_updates[3] == (0x0012, 2050)  # TARGET
+    assert thermostat_listener.attribute_updates[6] == (0x0014, 1700)  # HOLIDAY
+    assert thermostat_listener.attribute_updates[7] == (0x0010, 110)  # OFFSET
+    assert thermostat_listener.attribute_updates[8] == (0x0025, 0)  # MANUAL
+    assert thermostat_listener.attribute_updates[9] == (0x4002, 1)
+    assert thermostat_listener.attribute_updates[10] == (0x0025, 1)  # SCHEDULE
+    assert thermostat_listener.attribute_updates[11] == (0x4002, 0)
+    assert thermostat_listener.attribute_updates[12] == (0x001C, 4)  # HEAT ON
 
     assert len(tuya_listener.cluster_commands) == 9
     assert len(tuya_listener.attribute_updates) == 9
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -82,6 +82,10 @@ ZCL_TUYA_VALVE_ZONNSMART_TEMP_OFFSET = (
 ZCL_TUYA_VALVE_ZONNSMART_MODE_MANUAL = b"\t2\x01\x03\x04\x02\x04\x00\x01\x01"
 ZCL_TUYA_VALVE_ZONNSMART_MODE_SCHEDULE = b"\t2\x01\x03\x04\x02\x04\x00\x01\x00"
 ZCL_TUYA_VALVE_ZONNSMART_HEAT_STOP = b"\t2\x01\x03\x04\x6b\x01\x00\x01\x00"
+ZCL_TUYA_VALVE_ZONNSMART_SCHEDULE_MONDAY = b"\tY\x01\x00\x85l\x00\x00\x1e\x1e\x00\xc8*\x00\xd2Z\x00\xaa~\x00\xdc\x90\x00\xc8\x90\x00\xd2\x90\x00\xaa\x90\x00\xd2\x90\x00\xaa\x90\x00\xd2"
+ZCL_TUYA_VALVE_ZONNSMART_HOLIDAY_DATES = (
+    b"\tT\x01\x00&.\x03\x00\x18202101010101202303231305"
+)
 
 ZCL_TUYA_EHEAT_TEMPERATURE = b"\tp\x02\x00\x02\x18\x02\x00\x04\x00\x00\x00\xb3"
 ZCL_TUYA_EHEAT_TARGET_TEMP = b"\t3\x01\x03\x05\x10\x02\x00\x04\x00\x00\x00\x15"
@@ -447,6 +451,7 @@ async def test_zonnsmart_state_report(zigpy_device_from_quirk, quirk):
     valve_dev = zigpy_device_from_quirk(quirk)
     tuya_cluster = valve_dev.endpoints[1].tuya_manufacturer
 
+    tuya_listener = ClusterListener(tuya_cluster)
     thermostat_listener = ClusterListener(valve_dev.endpoints[1].thermostat)
 
     frames = (
@@ -457,6 +462,8 @@ async def test_zonnsmart_state_report(zigpy_device_from_quirk, quirk):
         ZCL_TUYA_VALVE_ZONNSMART_MODE_MANUAL,
         ZCL_TUYA_VALVE_ZONNSMART_MODE_SCHEDULE,
         ZCL_TUYA_VALVE_ZONNSMART_HEAT_STOP,
+        ZCL_TUYA_VALVE_ZONNSMART_SCHEDULE_MONDAY,
+        ZCL_TUYA_VALVE_ZONNSMART_HOLIDAY_DATES,
     )
     for frame in frames:
         hdr, args = tuya_cluster.deserialize(frame)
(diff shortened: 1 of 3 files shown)


MattWestb:
tuya TRV schedule is one large mess and i think all different types is using different implementing of it so its not possible making one nice GUI in ZHA for all TRV.

Great work done !!

LennardPlay:
Can you help me understand how this works? I do understand how the shedule is formated, but I don't understand where to write it. I found the ZONNSMARTManufCluster but there's no schedule attribute there.

gmsoft-tuxicoman:
Have you applied this PR to your installation ?  If so you should have "shedule_set" attribute in the ZONNSMARTManufCluster.

LennardPlay:
I don't know how to do this (sorry, I'm a newbe). Can you please explain?

gmsoft-tuxicoman:
https://community.home-assistant.io/t/how-to-setup-local-zha-quirks/341226
Copy the patched ts0601_trv.py  in zustom_zha_quirks and restart.

gmsoft-tuxicoman:
Would someone be able to review this please ?
@dmulcahey or @Adminiuga maybe ?

dmulcahey:
@javicalle this look good to you?

javicalle:
The whole implementation of TRVs is not familiar to me, and to me it's a huge monster that I hope I don't have to deal with.

I was going to argue against the implementation of the `OnOff` clusters but I see that it is consistent with the rest of the code so it would be fine to me.

As for the rest, it seems to me that the general 'disorder' of the TRVs is increasing, but I don't have a better proposal (or time available).

gmsoft-tuxicoman:
@dmulcahey please ?

TheJulianJES:
Can you rebase this? (So newer tests run)

TheJulianJES:
Hmm, the `OnOff` stuff is definitely hacky, but that's already in at the moment and I guess it's one of the few ways of doing it with the current quirks/ZHA implementation.
Fixing it so that it actually works with multiple TRVs is a needed improvement.

This PR looks like it will conflict with:
- https://github.com/zigpy/zha-device-handlers/pull/1192

but that PR seems somewhat stale anyways and it seems like @jacekk015 made a custom repo for TRV quirks: https://github.com/jacekk015/zha_quirks

These already seem to contain some fixes (i. e. the `OnOff` stuff for multiple devices if I'm not mistaken).
Obviously, it would be best to also get these in officially eventually.

@jacekk015 If you have a minute, do you have any opinions on this PR?

Nothing overly wrong sticks out to me in the added schedule stuff, but it's quite a bit of code. In general, the TRV implementation seems to be quite cluttered and should probably be split up *eventually*.
@gmsoft-tuxicoman I assume these changes still work correctly?

gmsoft-tuxicoman:
@TheJulianJES Yes, it still works. I have 14 of those TRVs and I have no issue at all.

jacekk015:
Keep so many different TRVs under one file is madness - in my opinion.
Since my quirks didn't complete tests I've abandoned any hope for any official PR.
I've taken quirks code in the state they were and extended/corrected them to users need.
Newest quirks were written according to my simplest approach, but older ones weren't rewritten since they just work.
For quirk rewrite you need to have a user that will test it - otherwise that's just a waste of time.
When user gets a working quirk he is loosing any free time to spare, mostly.
So you will end up with you quirk rewritten on the disk or Github cloud... waiting.
Don't have so much spare time to waste.

I've started to check if TuyaNewManufCluster is usable for TRV's but that's a switch from Attributes to DP's
For now I didn't touch it.

OnOff is a hack, but you can't even set a name/description for OnOff, so when user get two switches without a proper name they are lost already.
So again, it is like it is.

Looking at the latest quirk created
https://github.com/jacekk015/zha_quirks/blob/main/ts0601_trv_etop.py
you have to write over 300 lines of code to make it supported somehow.
I didn't find any TRV quirk example created form top to bottom using newest, proper, approach.
Most of them are glued together pieces of code found in different places.

Will that change? I don't think so - latest 2 years[winters] showed me that the code 'disorder' exist till today.

MattWestb:
@javicalle is trying getting one better way getting ZHA climate being matched from device class and attribute so not updating climate component then adding one new device is the quirk. If its working its easy making `tuyaTRV1` `tuyaTRV2` and yo in and in ZHA climate and the quirk is setting the name and attribute and ZHA is loading the right patches for the device.

I have the first PRed TRV and its badly working at all with ZHAs quirk and with the extended its have all in but is being broken 2 time in the year then its not on the master.

Rewriting all TRV to good working code is possible but its need time from devs and users that can doing all the testing and in the long run is very needed.

I think we need work being done in ZHA for making its easier and better integrating this devices in ZHA like functions for temperature calibrating and switches with fix function like open window and valve detect but it can being possible doing but its needs more testing and help from the "ZHA" side and not only saying its not possible.
So if the ZHA climate can matching functions we can doing it with all other 2 in one clean easy way.

Knowledge i knowing have but not the interest fixing it in some places is not there.

Pleas dont do one "deCONZ" (self killing) of ZHA by doing it worse, try doing it best and we can doing it (we have nearly all tuya TRVs working with extended functions and its bad for many users must installing code they cant maintaining).

gmsoft-tuxicoman:
If I understand correctly what @MattWestb is saying, we probably need ZHA to have better API for TRVs.

I do agree with that, right now there is no proper way to implement a TRV unless you use those hacks.
The way forward is probably to get a better API for the TRVs and then migrate each model to this new API.

Until then, I don't see a better solution than using the hacky code as it is.

gmsoft-tuxicoman:
@javicalle After looking some more at this code and how it's handled in HA, I've found the proper way to deal with all those OnOff clusters.

I've removed all that cruft and created a corresponding HA PR : https://github.com/home-assistant/core/pull/87546

I guess those should be merged at around the same time or at least the HA one first so that users have some time to update their code.

tests/test_tuya.py
```diff
@@ -462,25 +462,25 @@ async def test_zonnsmart_state_report(zigpy_device_from_quirk, quirk):
         tuya_cluster.handle_message(hdr, args)
 
     assert len(thermostat_listener.cluster_commands) == 0
-    assert len(thermostat_listener.attribute_updates) == 11
+    assert len(thermostat_listener.attribute_updates) == 9
```
TheJulianJES, line 465: So two attributes no longer update when the TRV reports its state? Or what is the reason this number changes?

gmsoft-tuxicoman: Good catch, I found the issue and fixed the code instead :)

TheJulianJES:
If possible, it would be nice if you could implement some (basic) tests

gmsoft-tuxicoman:
I'm working on unit tests. It appears that recent changes in tuya clusters broke this PR so I need to fix it too.

gmsoft-tuxicoman:
Codecov seems broken or disabled. I've added unit test but it didn't catch those :(
Can someone please do a manual review (or fix codecov) ?

Mikaka27:
Hi, is it possible to test this by using only the zha quirk? I'm asking since you've made some changes to home assistant core as well, and I'm wondering if it works alone now?

MattWestb:
Hey @puddly can you rerunning the test then it was not working after last fixes and its being great if we can getting this working on ZHA so we can using the same principes for other devices.

Thanks in advance !!

gmsoft-tuxicoman:
I rebased, it should help for the tests.

Mikaka27:
From looking at this we need to add more negative tests, since those code branches are not covered in codecov.

Mikaka27:
There is another issue, you cannot set temperature greater than 25.5, because you multiply them by 10, and then treat them as uint8_t, so you can fit 255 value as max.

Here is what happens if I try to set 26 degress in schedule:
```
2023-05-07 13:35:39.088 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [139659890865680] 260 is not an unsigned 8 bit integer
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 199, in handle_call_service
    await hass.services.async_call(
  File "/usr/src/homeassistant/homeassistant/core.py", line 1849, in async_call
    task.result()
  File "/usr/src/homeassistant/homeassistant/core.py", line 1889, in _execute_service
    await cast(Callable[[ServiceCall], Awaitable[None]], handler.job.target)(
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 874, in admin_handler
    await result
  File "/usr/src/homeassistant/homeassistant/components/zha/websocket_api.py", line 1297, in set_zigbee_cluster_attributes
    response = await zha_device.write_zigbee_attribute(
  File "/usr/src/homeassistant/homeassistant/components/zha/core/device.py", line 748, in write_zigbee_attribute
    response = await cluster.write_attributes(
  File "/usr/local/lib/python3.10/site-packages/zhaquirks/tuya/__init__.py", line 519, in write_attributes
    await super().command(
  File "/usr/local/lib/python3.10/site-packages/zigpy/quirks/__init__.py", line 200, in command
    return await self.request(
  File "/usr/local/lib/python3.10/site-packages/zigpy/zcl/__init__.py", line 311, in request
    hdr, request = self._create_request(
  File "/usr/local/lib/python3.10/site-packages/zigpy/zcl/__init__.py", line 273, in _create_request
    request.serialize()  # Throw an error before generating a new TSN
  File "/usr/local/lib/python3.10/site-packages/zigpy/types/struct.py", line 257, in serialize
    chunks.append(value.serialize())
  File "/usr/local/lib/python3.10/site-packages/zigpy/types/struct.py", line 257, in serialize
    chunks.append(value.serialize())
  File "/usr/local/lib/python3.10/site-packages/zhaquirks/tuya/__init__.py", line 271, in serialize
    return b"".join([self._item_type(i).serialize() for i in self[:]])
  File "/usr/local/lib/python3.10/site-packages/zhaquirks/tuya/__init__.py", line 271, in <listcomp>
    return b"".join([self._item_type(i).serialize() for i in self[:]])
  File "/usr/local/lib/python3.10/site-packages/zigpy/types/basic.py", line 94, in __new__
    raise ValueError(
ValueError: 260 is not an unsigned 8 bit integer
```

gmsoft-tuxicoman:
@Mikaka27 thanks for the patch and reporting the 26C issue. I think I know how to fix it. I'll update the PR shortly.

TheJulianJES:
I looked at this PR and the [HA PR](https://github.com/home-assistant/core/pull/87546) and they mostly seem to be fine. At least nothing bad that caught my eye. I've added some comments in the HA PR.

For this PR, it would be nice (although not required I guess) if tests would cover the cases where a `ValueError` is thrown.
The complete PR is a bit hard to fully review, as I don't completely understand Tuya's protocol. But again, it seems to be fine.

IMO, if the last comments are addressed, we could try to move this forward for HA 2023.11.0. (2023.10.0b0 is out already and since this is a breaking change, we can't and shouldn't add this in the beta phase.)
We need to make sure to list this as a breaking change for the zha-quirks dependency bump at some point then.

zhaquirks/tuya/ts0601_trv.py
```diff
@@ -931,46 +921,288 @@ async def command(
...
+        elif day == "saturday" or day == "sat":
+            b += [0x20]
+        elif day == "sunday" or day == "sun":
+            b += [0x40]
+        elif day == "work" or day == "mon-fri":
+            b += [0x1F]
+        elif day == "weekend" or day == "sat-sun":
+            b += [0x60]
+        elif day == "week" or day == "mon-sun":
+            b += [0x7F]
```
javicalle: All these values are bitmap positions for weekdays:
monday= 00000001
saturday=00100000
work=00011111
weekend=01100000
...
The code works but guessing if there are some nice way to represent it. Maybe binary values would be more easy to guess.

javicalle (approved):
I don't have an hard opinion on all the TRV implementation.
Migrating from fake clusters to manufacturer attributes is fine to me but not sure how much 'breaking change' that would be.

TheJulianJES:
Since the [quirk ID stuff](https://github.com/home-assistant/core/pull/102482) for the ZHA PR hasn't made it into HA yet, I don't see how this can make 2023.11.0. (2023.11.0 beta already releases tomorrow)

Hopefully, the "quirk ID" implementation is ready before 2023.12.0, so we can get it into that release.

EDIT: It's in now.

---

Btw., these two review comments are still valid IMO:
- https://github.com/zigpy/zha-device-handlers/pull/1961#discussion_r1342130757
- https://github.com/zigpy/zha-device-handlers/pull/1961#issuecomment-1737942490 (cover last remaining lines in tests)

gmsoft-tuxicoman:
This should be good to go again. I've added the quirk ID for the zonnsmart TRV and also fixed state_change function when no value is cached.

gmsoft-tuxicoman:
@TheJulianJES could you please review this ? it should fairly quick, not much changed since your last review :)

Thanks !

TheJulianJES:
We'll try to target this for HA 2024.3.0. Merge this soon, do the quirks bump in the next few weeks and then the corresponding HA PR can be rebased.
Ideally, we should get both a quirks bump and the HA PR in before a nightly build is made, but it's also not the end of the world if the entities are missing there I think.

TheJulianJES (approved):
LGTM. Thanks!
Fine to merge only for the next quirks release targeting HA Core 2024.3.0

TheJulianJES:
Also, it might be worth rebasing this once more (or merging changes from the current branch) to re-trigger CI with the latest zigpy version, as TSN handling changed there slightly, so the tests might need to be slightly adjusted.

Similar to:
- https://github.com/zigpy/zha-device-handlers/pull/2885

zhaquirks/tuya/ts0601_trv.py
```diff
@@ -958,7 +1140,16 @@ def __init__(self, *args, **kwargs):
                 t.int32s,
                 True,
             ),
-            ZONNSMART_WEEK_FORMAT_ATTR: ("week_format", t.uint8_t, True),
+            ZONNSMART_SCHEDULE_TYPE_ATTR: (
+                "schedule_type",
+                ZONNSMARTScheduleType,
+                True,
+            ),
```
TheJulianJES, line 1149: To generally comment on the attribute defintion:
Ideally, we already want to use the new zigpy-style `AttributeDefs` definitions, like this:

Existing attribute defs can be extended by just inheriting that class (although that requires the base class to also use the new-style):
https://github.com/zigpy/zha-device-handlers/blob/2a39c07e48a2f3cd1956fffac009ec0c64b0f783/zhaquirks/schneiderelectric/dimmers.py#L44-L47

We should/need to convert those attributes/command definitions, but I guess it might not have to be in this PR(?), since it's already open very long. (would be nice though)

(Tuya base classes might need to be converted first for the commands. `TuyaManufClusterAttributes` doesn't have any attributes at the moment, so those could already be converted I think.)
(The advantage of the new `AttributeDefs` is also that the name/id can be accessed via `Cluster.AttributeDefs.attribute_name.id` or `.name` (if in class, self.AttributeDefs.attribute_name.id` is also enough)

zhaquirks/tuya/ts0601_trv.py
```diff
@@ -931,19 +922,210 @@ async def command(
...
+
+    def __init__(self, value, *args, **kwargs):
+        """Parse the schedule string and serialize it."""
+
+        super().__init__(*args, **kwargs)
+        if value is None:
+            return
+
+        if not isinstance(value, str):
+            raise ValueError("Only string value allowed")  # noqa: TRY004
```
TheJulianJES, line 1012: So, all errors aren't covered. Would it be possible to add tests for those? It should just be this per possible error:
```python
    with pytest.raises(
        ValueError,
        match="Only string value allowed",
    ):
        # call code here to produce the error
```

TheJulianJES:
Also linking this comment regarding quirks v2 and "exposing entities in quirks" here:
https://github.com/home-assistant/core/pull/87546#issuecomment-1955688910

zhaquirks/tuya/ts0601_trv.py
gmsoft-tuxicoman: I added tests for all those but I encountered a bug in zigpy while doing so.

Please review https://github.com/zigpy/zigpy/pull/1353 to make the tests work.

Thanks

gmsoft-tuxicoman:
https://github.com/zigpy/zigpy/pull/1353 was merged, hopefully the tests will work once current head will be used in this repo.

PR #1962, opened 2022-11-23
Add power configuration cluster to ZLinky_TIC (requires firmware v12.0+)

pdecat:
For reference:
* https://github.com/fairecasoimeme/Zlinky_TIC/releases/tag/v12.0
* https://github.com/fairecasoimeme/Zlinky_TIC#powerconfiguration-cluster-0x0001----from-v12

Marked as draft as I'm unsure how this affects devices with pre-v12.0 firmware.

### DIFF
--- a/zhaquirks/lixee/zlinky.py
+++ b/zhaquirks/lixee/zlinky.py
@@ -206,8 +206,4 @@ class ZLinkyTICFWV12(ZLinkyTIC):
     signature[MODELS_INFO] = [(LIXEE, "ZLinky_TIC_FW_V12")]
 
     # Insert PowerConfiguration cluster in signature for devices with firmware v12.0+
-    signature[ENDPOINTS][1][INPUT_CLUSTERS] = (
-        signature[ENDPOINTS][1][INPUT_CLUSTERS][:1]
-        + [PowerConfiguration.cluster_id]
-        + signature[ENDPOINTS][1][INPUT_CLUSTERS][1:]
-    )
+    signature[ENDPOINTS][1][INPUT_CLUSTERS].insert(1, PowerConfiguration.cluster_id)
--- a/zhaquirks/lixee/zlinky.py
+++ b/zhaquirks/lixee/zlinky.py
@@ -1,4 +1,6 @@
 """Quirk for ZLinky_TIC."""
+from copy import deepcopy
+
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
 import zigpy.types as t
@@ -193,3 +195,19 @@ class ZLinkyTIC(CustomDevice):
             },
         },
     }
+
+
+class ZLinkyTICFWV12(ZLinkyTIC):
+    """ZLinky_TIC from LiXee with firmware v12.0+."""
+
+    signature = deepcopy(ZLinkyTIC.signature)
+
+    # Update model info
+    signature[MODELS_INFO] = [(LIXEE, "ZLinky_TIC_FW_V12")]
+
+    # Insert PowerConfiguration cluster in signature for devices with firmware v12.0+
+    signature[ENDPOINTS][1][INPUT_CLUSTERS] = (
+        signature[ENDPOINTS][1][INPUT_CLUSTERS][:1]
+        + [PowerConfiguration.cluster_id]
+        + signature[ENDPOINTS][1][INPUT_CLUSTERS][1:]
+    )
--- a/zhaquirks/lixee/zlinky.py
+++ b/zhaquirks/lixee/zlinky.py
@@ -2,7 +2,13 @@
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
 import zigpy.types as t
-from zigpy.zcl.clusters.general import Basic, GreenPowerProxy, Identify, Ota
+from zigpy.zcl.clusters.general import (
+    Basic,
+    GreenPowerProxy,
+    Identify,
+    Ota,
+    PowerConfiguration,
+)
 from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement, MeterIdentification
 from zigpy.zcl.clusters.manufacturer_specific import ManufacturerSpecificCluster
 from zigpy.zcl.clusters.smartenergy import Metering
@@ -147,6 +153,7 @@ class ZLinkyTIC(CustomDevice):
                 DEVICE_TYPE: zha.DeviceType.METER_INTERFACE,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
                     Identify.cluster_id,
                     Metering.cluster_id,
                     MeterIdentification.cluster_id,
@@ -170,6 +177,7 @@ class ZLinkyTIC(CustomDevice):
                 DEVICE_TYPE: zha.DeviceType.METER_INTERFACE,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
                     Identify.cluster_id,
                     ZLinkyTICMetering,
                     MeterIdentification.cluster_id,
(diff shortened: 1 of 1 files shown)


pdecat:
I guess it won't cause issues with older firmware as the new cluster is (currently) not in the signature.

image

image

image

pdecat:
Marked as draft again as the new cluster is supposed to be discoverable: https://github.com/fairecasoimeme/Zlinky_TIC/issues/140#issuecomment-1326123687

MattWestb:
Make one new device class with the new EP/Cluster setting and it shall working OK.
Also then doing OTA and changing the EP/Cluster setting it can being needed deleting the device and reading it for the system saving the new settings so the quirk can being loaded OK.

pdecat:
> Make one new device class with the new EP/Cluster setting and it shall working OK.

Thanks!

> Also then doing OTA and changing the EP/Cluster setting it can being needed deleting the device and reading it for the system saving the new settings so the quirk can being loaded OK.

Is the signature cached somewhere?

pdecat:
Looks like the cached _in clusters_ information may come from this table:

```
# sqlite3 config/zigbee.db "SELECT * FROM in_clusters_v10 WHERE ieee = '00:15:8d:00:ff:ff:ff:ff';"
00:15:8d:00:ff:ff:ff:ff|1|0
00:15:8d:00:ff:ff:ff:ff|1|3
00:15:8d:00:ff:ff:ff:ff|1|1794
00:15:8d:00:ff:ff:ff:ff|1|2817
00:15:8d:00:ff:ff:ff:ff|1|2820
00:15:8d:00:ff:ff:ff:ff|1|65382
00:15:8d:00:ff:ff:ff:ff|242|33
```

MattWestb:
Dont forget reversing the original device class then that firmware for it dont have power config cluster.

All devices is saved in the `zigbee.db` in the configure directory and is being used the init devices and applying quirks for it then joining and also the starting the system.

One good example is updating one old IKEA controller or light that was released with ZLL firmware and getting one Zigbee 3 firmware can working OK but some devices must being deleted and rejoined for working OK.

I think to be on the safe side is to recommending deleting the device after OTA update and waiting one minute (ZHA must doing the DB work finished) and doing one new joining of the device so the cluster setting is being OK.

pdecat:
Was about to open an in zigpy, but looks like this was already discussed here: https://github.com/zigpy/zigpy/issues/912#issuecomment-1035487693

MattWestb:
By the way nice IEEE of your device :-))))

pdecat:
> By the way nice IEEE of your device :-))))

I redacted it :)

pdecat:
I've added a ZLinky_TIC_FW_V12 quirk with power configuration cluster so that both pre-v12 and post-v12 versions are supported:
```
2022-11-24 13:01:32.046 DEBUG (MainThread) [zigpy.quirks.registry] Checking quirks for LiXee ZLinky_TIC (00:15:8d:00:ff:ff:ff:ff)
2022-11-24 13:01:32.046 DEBUG (MainThread) [zigpy.quirks.registry] Considering <class 'zhaquirks.lixee.zlinky.ZLinkyTICFWV12'>
2022-11-24 13:01:32.046 DEBUG (MainThread) [zigpy.quirks.registry] Fail because input cluster mismatch on at least one endpoint
2022-11-24 13:01:32.046 DEBUG (MainThread) [zigpy.quirks.registry] Considering <class 'zhaquirks.lixee.zlinky.ZLinkyTIC'>
2022-11-24 13:01:32.046 DEBUG (MainThread) [zigpy.quirks.registry] Found custom device replacement for 00:15:8d:00:ff:ff:ff:ff: <class 'zhaquirks.lixee.zlinky.ZLinkyTIC'>
```

zhaquirks/lixee/zlinky.py
```diff
@@ -170,6 +178,7 @@ class ZLinkyTIC(CustomDevice):
                 DEVICE_TYPE: zha.DeviceType.METER_INTERFACE,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
```
pdecat, line 181: Adding the PowerConfiguration cluster for both versions of the quirk allows supporting that cluster for devices that were already existing but upgraded from pre-v12 to v12 without having to delete then re-add the device from ZHA.

zhaquirks/lixee/zlinky.py
```diff
@@ -186,3 +195,19 @@ class ZLinkyTIC(CustomDevice):
...
+    signature = deepcopy(ZLinkyTIC.signature)
+
+    # Update model info
+    signature[MODELS_INFO] = [(LIXEE, "ZLinky_TIC_FW_V12")]
+
+    # Insert PowerConfiguration cluster in signature for devices with firmware v12.0+
+    signature[ENDPOINTS][1][INPUT_CLUSTERS] = (
+        signature[ENDPOINTS][1][INPUT_CLUSTERS][:1]
+        + [PowerConfiguration.cluster_id]
+        + signature[ENDPOINTS][1][INPUT_CLUSTERS][1:]
```
javicalle, line 212: Just a comment, tt is not necessary that the definition of the clusters be ordered.

pdecat: Thanks, I was wondering if it was necessary and/or cleaner to insert it.

Can switch to just appending `PowerConfiguration.cluster_id` to `signature[ENDPOINTS][1][INPUT_CLUSTERS]` if preferred.

javicalle: Just a matter of taste.
I'm not skilled in Python (but getting better thanks to HA!) so I had to make an effort to understand what was going on, once understood is fine to me.

pdecat: Now that you mention it, using `list.insert(i, x)` is probably more readable.

https://docs.python.org/3/tutorial/datastructures.html

pdecat: Simplified code with `list.insert(i, x)`

zhaquirks/lixee/zlinky.py
```diff
@@ -186,3 +195,15 @@ class ZLinkyTIC(CustomDevice):
             },
         },
     }
+
+
+class ZLinkyTICFWV12(ZLinkyTIC):
+    """ZLinky_TIC from LiXee with firmware v12.0+."""
+
+    signature = deepcopy(ZLinkyTIC.signature)
```
pdecat, line 203: Should I keep the same model name so that the fix in https://github.com/home-assistant/core/pull/82602/files#diff-97adf01229931f17610655838232182c16a7f10291291d780cfd472be187406eR510 continues to work?

zhaquirks/lixee/zlinky.py
```diff
@@ -186,3 +195,15 @@ class ZLinkyTIC(CustomDevice):
...
+
+    # Update model info
+    signature[MODELS_INFO] = [(LIXEE, "ZLinky_TIC_FW_V12")]
```
pdecat, line 206: Should I keep the same model name so that the fix in https://github.com/home-assistant/core/pull/82602/files#diff-97adf01229931f17610655838232182c16a7f10291291d780cfd472be187406eR510 continues to work?

javicalle: But, is `ZLinky_TIC_FW_V12` the model value for that devices, isn't it?
You need to keep the model to be able to match the quirk.
I believe that you can overwrite it in the replacement part, but not sure if would be a good thing.
You can also add the `ZLinky_TIC_FW_V12` model in HA to expose all the entities.

pdecat: I invented it. The quirk works with both strings for the model version. I believe it is only esthetic.

javicalle: But, you're replacing the `MODELS_INFO` in the new quirk. No device could match the signature...
Do you have check if v12 devices are loading the quirk? From what I see them couldn't.

javicalle: (And would be fine having 2 quirks with the same `MODELS_INFO` and diferent signatures)

MattWestb:
If changing the model info in the V12 firmware its being OK and i think it one good way to see that the device is running the updated version but normally it shall being enough reading the software version on the basic cluster for getting that.

zhaquirks/lixee/zlinky.py
pdecat: Ok, got it. I did not actually test to remove then re-add my device ZLinky_TIC with firmware v12, that's why I did not face any issue.
`model` attribute of `basic` cluster returns `ZLinky_TIC`, so I'll revert that change.

PR #1965, opened 2022-11-25
Foo

aweisberg:
Foo

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# ZHA Device Handlers For Home Assistant
+# ZHA Device Handlers For Home Assistant 
 
 ![CI](https://github.com/zigpy/zha-device-handlers/workflows/CI/badge.svg?branch=dev)
 [![Coverage Status](https://coveralls.io/repos/github/zigpy/zha-device-handlers/badge.svg)](https://coveralls.io/github/zigpy/zha-device-handlers)


aweisberg:
It works!
image

MattWestb:
Have you trying updating the device firmware then this "bug" is pretty old and all devices (not the old motion sensor) have getting updated and this its one left from the early upgrades from ZLL to Zigbee 3 ?

aweisberg:
I updated the firmware, removed it, moved the quirk directory, restarted HA, and then added it back and it worked.

It would still be nice if it worked without a firmware update. It took quite some doing just to figure out what was wrong and get a working quirk. Figuring out that a firmware updated would fix it is similarly complicated.

If it was adding a whole new quirk and there was a policy of minimizing adding quirks then sure, but it's just one line to an existing quirk so why not?

MattWestb:
I finding the quirk is very OK but its best updating the firmware so old bugs in the Zigbee stack is fixed for getting one more stable network.

Also we is having many old quirks for IKEA controllers that can being deleted **if** only running the last firmware (and not very old ZLL ones) but its not good making user sad then there devices is not working OK longer.

aweisberg:
@MattWestb would  you mind reopening this PR? I am unable to update it since I closed it. I forgot to ask permission from my employer to contribute (and by policy I don't think they will) and need to force push to replace the commit.

MattWestb:
Sorry @aweisberg im not one maintainer so i cant ding it but i asking @javicalle if hi can doing it then hi is in our time zone.

javicalle:
No, I'm not able to reopen the PR.

MattWestb:
Thanks @javicalle !!

@aweisberg I think the easiest way is making one new pull request with the same changes and letting the old being closed.

For my its good that also devices that is not updated is working OK for our users so pleas do it if its possible.

Lg

Issue #1629, opened 2022-06-22
[BUG] aqara cube knock on faces 2, 3, 5, 6 report wrong activated face


mouth4war:
**Describe the bug**
Doing a double knock on faces 2, 3, 5 and 6 reports incorrect values for activated face:

Below example is for active face 2:

    "command": "knock",
    "args": {
        "value": 513,
        "description": "aqara logo facing user rotated 90 degrees left",
        "activated_face": 5

**To Reproduce**
1. Enable event listening. 
2. Flip the cube and knock on each face (1-6), check events.
3. Faces 1 and 4 work fine but knocking on faces 2, 3, 5, 6 generates incorrect activated_face values.

**Expected behavior**
Testing results for knock events: 6 and 3 are interchanged so are 2 and 5. 1 and 4 are reporting correctly.

**Screenshots**

<details>

<summary>Device signature</summary>

```yaml
Paste the device signature here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
Paste the diagnostic information here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>
<summary>Additional logs</summary>

```
Paste any additional debug logs here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

**Additional context**

TheJulianJES:
Relevant PR that also contains some discussion: https://github.com/zigpy/zha-device-handlers/pull/1972

alfa43:
Hi, so i got a new Aquqra Cube t1 pro and i got the same issue. How do i resolve this? How do i implement this fix?

PR #1972, opened 2022-11-29
Fix wrongly activated face for Aqara cube

Biotronic:
This should fix issue #1629 . Judging by the diagrams I find online, the bug report is wrong, in that the knock and slide events actually report the correct face, and it's the flip gesture that's wrong.

Should I write tests for this? Also, is this the right place to fix it? The IDs seem awfully ordered, so the issue might be further down the stack.

Also done: added deactivated face for 180 degree flip, and description for flip gesture (used IDs 1000-1005 for this. If that's a bad idea, I'd love to correct my wicked ways).

### DIFF
--- a/zhaquirks/xiaomi/aqara/cube_aqgl01.py
+++ b/zhaquirks/xiaomi/aqara/cube_aqgl01.py
@@ -57,6 +57,14 @@ FLIP_BEGIN = 50
 FLIP_DEGREES = "flip_degrees"
 FLIP_END = 180
 FLIPPED = "device_flipped"
+
+FLIP_1_VALUE = 1000  # aqara skyside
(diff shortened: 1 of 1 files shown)


elupus:
Do we know if this is still valid for the new generations cubes?

Biotronic:
I was waiting to answer this until I could get a T1 Pro and check, but that takes a lot longer than I expected, and I plain forgot. Currently, I have no idea. Would love if someone else who has one, could test.

Biotronic:
Also worth noting: issue #1629 has been closed due to inactivity. Not sure what to make of that.

elupus:
Btw, does this end up as a breaking change for HA? In the event message?

elupus:
Not sure this should be closed

DunklerPhoenix:
Not sure this should be closed²

PR #1987, opened 2022-12-04
Add smoke density attribute for Mijia Smoke Detector

mstefany:
I didn't find any reference for measurement of a kind "smoke density" in Zigbee Cluster Library Specification, so instead I am doing translation of nested attribute `100` in attribute `0xFF01`/`65281` ("aqara attributes") to AnalogInput cluster (as described in tutorial via `Bus()` mechanism), while also setting description to "Smoke Density".

This is basically ZHA implemention of Z2M's basic cluster processing for this smoke sensor, [see](https://github.com/Koenkk/zigbee-herdsman-converters/blob/2636217a818d08ec3a2ec2fe4c84a9c1afa99041/lib/xiaomi.js#L245).

### DIFF
--- a/tests/test_xiaomi.py
+++ b/tests/test_xiaomi.py
@@ -25,6 +25,7 @@ from zhaquirks.xiaomi import (
     CONSUMPTION_REPORTED,
     LUMI,
     POWER_REPORTED,
+    SMOKE_DENSITY_REPORTED,
     VOLTAGE_REPORTED,
     XIAOMI_NODE_DESC,
     BasicCluster,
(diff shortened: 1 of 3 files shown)


pauln:
@mstefany It'd be nice to see this land - I've got some of these smoke alarms, and I remember seeing the smoke density reading in the Xiaomi Home app when I had the alarms paired to the Aqara hub, so my data-hungry brain would love to be able to see it again.  Is there something still needing to be done to progress this PR from draft stage, and if so, is it something that I might be able to help with?

mstefany:
@pauln I think the PR is basically ready, at least from my limited point of view.
The reason I didn't take it further is that the value is actually integer reading from 0 to 255, and sort of feels like internal value for the sensor itself and cannot be represented into something like ppm or so. Or at least I was not able to find anything like that online (some sort of conversion) - only some video from OpenHAB users that showed this 0 to 255 range.

pauln:
Personally, I'd be happy to see the value as reported.  If a conversion / mapping to a more "meaningful" value such as PPM is later determined, that'd be a bonus - but getting the data in any form seems like a step up from not getting it at all.

dmulcahey:
Is this ready? If so can you mark it ready for review?

TheJulianJES:
ZHA does not create a sensor entity for this at the moment (with the quirk), right?

mstefany:
Nope, I assume I need to write code for ZHA to generate a (diagnostic) sensor for this.

PR #1989, opened 2022-12-05
tuya mcu enhancements

Shulyaka:
I am working on a quirk for Tuya ts0601 _TZE200_hkdl5fmv circuit breaker (https://aliexpress.com/item/1005004749980758.html)
I've found out that some of the features cannot be implemented without some enhancements to tuya base classes. This is the PR for them.

Identified gaps:
1. The support for `TuyaDPType.RAW` is incomplete. If the device sends a DP with RAW type, we are unable to get the payload.
2. A single Tuya datapoint may correspond to several zcl attributes. For example, threshold parameters and alarm settings may come in a single RAW datapoint.
3. A single zcl attribute may correspond to several Tuya datapoints. For example, voltage and current alarm configuration may come in different DPs but correspond to the same `ac_alarms_mask` attribute.
4. When writing attributes on `TuyaAttributesCluster`, the local cache is not updated.

Solution:
1. The `TuyaData.payload` was enhanced to handle `TuyaDPType.RAW` case.
2. `DPToAttributeMapping.attribute_name` can now be a tuple containing several attributes. `DPToAttributeMapping.converter` must also return a tuple with the corresponding attribute values. `DPToAttributeMapping.dp_converter` will receive as many arguments as there are attributes for this DP. For the attribute that is being updated, the new value will be provided, for the rest attributes the cached values will be used.
3. A new class `AttributeWithMask` is introduced. If `DPToAttributeMapping.converter` returns an instance of this class, then only part of the attribute value that corresponds to the mask is updated.
4. `TuyaAttributesCluster.write_attributes()` was enhanced to call `super().write_attributes()`, and the `LocalDataCluster` does the rest.

Example usage:

```python
class LeakageParameters(t.Struct):
    self_test_auto_days: t.uint8_t
    self_test_auto_hours: t.uint8_t
    self_test_auto: t.Bool
    over_leakage_current_threshold: BigEndianInt16
    over_leakage_current_trip: t.Bool
    over_leakage_current_alarm: t.Bool
    self_test: SelfTest


class TemperatureSetting(t.Struct):
    over_temperature_threshold: t.int8s
    over_temperature_trip: t.Bool
    over_temperature_alarm: t.Bool


class TuyaRCBOElectricalMeasurement(ElectricalMeasurement, TuyaAttributesCluster):
    ...


class TuyaRCBODeviceTemperature(DeviceTemperature, TuyaAttributesCluster):
    ...


class TuyaRCBOManufCluster(TuyaMCUCluster):
        ...
        109: DPToAttributeMapping(
            TuyaRCBOElectricalMeasurement.ep_attribute,
            (
                "self_test_auto_days",
                "self_test_auto_hours",
                "self_test_auto",
                "over_leakage_current_threshold",
                "over_leakage_current_trip",
                "over_leakage_current_alarm",
                "self_test",
            ),
            TuyaDPType.RAW,
            lambda x: (x[0], x[1], x[2], x[4] | x[3] << 8, x[5], x[6], SelfTest(x[7])),
            lambda *fields: LeakageParameters(*fields),
        ),
        112: DPToAttributeMapping(
            TuyaRCBODeviceTemperature.ep_attribute,
            ("high_temp_thres", "over_temp_trip", "dev_temp_alarm_mask"),
            TuyaDPType.RAW,
            lambda x: (x[0], x[1], AttributeWithMask(x[2] << 1, 0x02)),
            lambda x, y, z: TemperatureSetting(x, y, z & 0x02),
        ),
    }
```

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -200,6 +200,8 @@ class TuyaData(t.Struct):
                 return bitmaps[len(self.raw)].deserialize(self.raw)[0]
             except KeyError as exc:
                 raise ValueError(f"Wrong bitmap length: {len(self.raw)}") from exc
+        elif self.dp_type == TuyaDPType.RAW:
+            return self.raw[::-1]
 
         raise ValueError(f"Unknown {self.dp_type} datapoint type")
 
@@ -1325,7 +1327,7 @@ class DPToAttributeMapping:
     """Container for datapoint to cluster attribute update mapping."""
 
     ep_attribute: str
-    attribute_name: str
+    attribute_name: str | tuple
     converter: Optional[
         Callable[
             [
@@ -1335,6 +1337,14 @@ class DPToAttributeMapping:
         ]
     ] = None
     endpoint_id: Optional[int] = None
+
+
+@dataclasses.dataclass
+class AttributeWithMask:
+    """Container for the attribute and its mask."""
+
+    value: Any
+    mask: int
 
 
 class TuyaNewManufCluster(CustomCluster):
@@ -1466,4 +1476,14 @@ class TuyaNewManufCluster(CustomCluster):
         if dp_map.converter:
             value = dp_map.converter(value)
 
-        cluster.update_attribute(dp_map.attribute_name, value)
+        if isinstance(dp_map.attribute_name, tuple):
+            for k, v in zip(dp_map.attribute_name, value):
+                if isinstance(v, AttributeWithMask):
+                    v = cluster.get(k, 0) & (~v.mask) | v.value
+                cluster.update_attribute(k, v)
+        else:
+            if isinstance(value, AttributeWithMask):
+                value = (
+                    cluster.get(dp_map.attribute_name, 0) & (~value.mask) | value.value
+                )
+            cluster.update_attribute(dp_map.attribute_name, value)
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -179,7 +179,12 @@ class TuyaData(t.Struct):
         res.dp_type, data = TuyaDPType.deserialize(data)
         res.function, data = t.uint8_t.deserialize(data)
         res.raw, data = t.LVBytes.deserialize(data)
-        if res.dp_type not in (TuyaDPType.BITMAP, TuyaDPType.STRING, TuyaDPType.ENUM):
+        if res.dp_type not in (
+            TuyaDPType.BITMAP,
+            TuyaDPType.STRING,
+            TuyaDPType.ENUM,
+            TuyaDPType.RAW,
+        ):
             res.raw = res.raw[::-1]
         return res, data
 
@@ -201,7 +206,7 @@ class TuyaData(t.Struct):
             except KeyError as exc:
                 raise ValueError(f"Wrong bitmap length: {len(self.raw)}") from exc
         elif self.dp_type == TuyaDPType.RAW:
-            return self.raw[::-1]
+            return self.raw
 
         raise ValueError(f"Unknown {self.dp_type} datapoint type")
 
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -1327,7 +1327,7 @@ class DPToAttributeMapping:
     """Container for datapoint to cluster attribute update mapping."""
 
     ep_attribute: str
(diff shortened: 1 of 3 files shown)


zhaquirks/tuya/__init__.py
```diff
@@ -1466,4 +1476,14 @@ def _dp_2_attr_update(self, datapoint: TuyaDatapointData) -> None:
         if dp_map.converter:
             value = dp_map.converter(value)
 
-        cluster.update_attribute(dp_map.attribute_name, value)
+        if isinstance(dp_map.attribute_name, tuple):
```
dmulcahey, line 1479: The tests aren’t hitting this case FYI

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -206,7 +221,19 @@ def from_cluster_data(self, data: TuyaClusterData) -> Optional[TuyaCommand]:
             datapoint_type = mapping.dp_type
             val = data.attr_value
             if mapping.dp_converter:
-                val = mapping.dp_converter(val)
+                args = []
+                if isinstance(mapping.attribute_name, tuple):
```
dmulcahey, line 225: It appears the tests aren’t hitting this case FYI

Shulyaka: They will, once I submit the PR for _TZE200_hkdl5fmv

dmulcahey:
@javicalle can you look this over when you have a moment please

Shulyaka:
There is one thing missing: when we write an attribute that corresponds to several datapoints, only one datapoint is updated. I couldn't find a good way to do that which doesn't break backwards compatibility. If I find one, I will either update this PR or raise a separate one.

javicalle:
Hi @Shulyaka, first of all: thank you for your contribution and make it clear that my notions in Python are quite basic, so I may not know how to interpret the entire implementation or that basic concepts are not familiar to me. 
I apologize in advance and hope you will be patient.

Your PR is pretty lengthy, so let me focus on a couple of the changes first before tackling the main reason.

I was going to tell you to rebase to take my latest changes into account, but I just saw that you found them.

1. There is something that feels strange to me with the new RAW tractament in the `payload()` function:
```python
        elif self.dp_type == TuyaDPType.RAW:
            return self.raw[::-1]
```

That is already done in the `deserialize()` part, so that would be duplicated in somewhere:
* https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L182-L184


2. Also I have the fealing that the new `TuyaReverseStruct` don't would be needed. All the MCU <--> Zigbee payload is getting reversed, if until now it has not be needed I would think that we can go without it.


I'm still checking the code and trying to remember all the logic that I put in it.

PS: I have not finished the review but I see that you are working on code right now, so I just leave that comments here.

Shulyaka:
Hi @javicalle, thank you for reviewing it!
If you want, I can break down this PR into smaller parts.
1. Yes, deserialize() already reverses the data, and I am reversing it back for the payload. I think I'd better add TuyaDPType.RAW into the condition to reverse in deserialize, like this:
```python
    @classmethod
    def deserialize(cls, data: bytes) -> Tuple["TuyaData", bytes]:
        ...
        if res.dp_type not in (TuyaDPType.BITMAP, TuyaDPType.STRING, TuyaDPType.ENUM, TuyaDPType.RAW):
            res.raw = res.raw[::-1]
        return res, data
```
2. No, it is still needed, but for another reason now. The reason it was not used before is that no one dealt with `TuyaDPType.RAW`. What is essentially doing now is switching endianess of each member. An alternative would be to create a bunch of big-endian int types. On the other hand, we did that already in the past (see `tuya.BigEndianInt16` or [xbee/types.py](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/xbee/types.py)). Should I make another request to [zigpy.types](https://github.com/zigpy/zigpy/tree/dev/zigpy/types) to include big-endian numbers?

Shulyaka:
I have removed `TuyaReverseStruct` for now and fixed the `TuyaData` as discussed.
Please review.

javicalle:
I think that we agree the RAW part.
Also agree that `super().write_attributes()`. Nice catch.

For all the pythonic discussion I will accept any explanation you have. In that aspect I cannot comment or raise objections because they are beyond my knowledge. I am sorry not to be helpful.

I'm checking now all the other part. You can do a separate PR for all this part or keep it here as you wish,but for me that part is reviewed and fine.

PR #1991, opened 2022-12-07
Tuya TS0601 _TZE200_hkdl5fmv RCBO

Shulyaka:
Add support for Tuya ts0601 _TZE200_hkdl5fmv circuit breaker (https://aliexpress.com/item/1005004749980758.html)

Special thanks to @danieledwardgeorgehitchcock for https://github.com/Koenkk/zigbee-herdsman-converters/pull/3197

image

### DIFF
--- a/zhaquirks/tuya/ts0601_rcbo.py
+++ b/zhaquirks/tuya/ts0601_rcbo.py
@@ -1,5 +1,5 @@
 """Tuya Din RCBO Circuit Breaker."""
-from typing import Dict, Optional, Union
+from typing import Any, Dict, Optional, Union
 
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
@@ -256,6 +256,25 @@ class TuyaRCBOElectricalMeasurement(ElectricalMeasurement, TuyaAttributesCluster
             0xF770: ("rms_historical_current", t.uint16_t),
         }
     )
+
+    def update_attribute(self, attr_name: str, value: Any) -> None:
+        """Calculate active current and power factor."""
+
+        super().update_attribute(attr_name, value)
+
+        if attr_name == "rms_current":
+            rms_voltage = self.get("rms_voltage")
+            if rms_voltage:
+                apparent_power = value * rms_voltage / 1000
+                super().update_attribute("apparent_power", int(apparent_power))
+
+        if attr_name == "active_power":
+            apparent_power = self.get("apparent_power")
+            if apparent_power:
+                power_factor = value / apparent_power * 1000
+                if power_factor > 1000:
+                    power_factor = 1000
+                super().update_attribute("power_factor", int(power_factor))
 
 
 class TuyaRCBODeviceTemperature(DeviceTemperature, TuyaAttributesCluster):
--- a/zhaquirks/tuya/ts0601_rcbo.py
+++ b/zhaquirks/tuya/ts0601_rcbo.py
@@ -167,9 +167,9 @@ class TuyaRCBOOnOff(TuyaOnOff, TuyaAttributesCluster):
     attributes = TuyaOnOff.attributes.copy()
     attributes.update(
         {
-            0xF090: ("countdown_timer", t.uint32_t),
+            0x8000: ("child_lock", t.Bool),
             0x8002: ("power_on_state", PowerOnState),
-            0xF1D0: ("child_lock", t.Bool),
+            0xF090: ("countdown_timer", t.uint32_t),
             0xF740: ("trip", Locking),
         }
     )
@@ -340,12 +340,12 @@ class TuyaRCBOManufCluster(TuyaMCUCluster):
 
     dp_to_attribute: Dict[int, DPToAttributeMapping] = {
         TUYA_DP_STATE: DPToAttributeMapping(
-            TuyaOnOff.ep_attribute,
+            TuyaRCBOOnOff.ep_attribute,
             "on_off",
             TuyaDPType.BOOL,
         ),
         TUYA_DP_COUNTDOWN_TIMER: DPToAttributeMapping(
-            TuyaOnOff.ep_attribute,
+            TuyaRCBOOnOff.ep_attribute,
             "countdown_timer",
             TuyaDPType.VALUE,
         ),
--- a/zhaquirks/tuya/ts0601_rcbo.py
+++ b/zhaquirks/tuya/ts0601_rcbo.py
@@ -174,7 +174,7 @@ class TuyaRCBOOnOff(TuyaOnOff, TuyaAttributesCluster):
         }
     )
 
-    server_commands = Metering.server_commands.copy()
+    server_commands = TuyaOnOff.server_commands.copy()
     server_commands.update(
         {
             0x74: foundation.ZCLCommandDef("clear_locking", {}, False),
--- a/zhaquirks/tuya/ts0601_rcbo.py
+++ b/zhaquirks/tuya/ts0601_rcbo.py
@@ -1,0 +1,586 @@
+"""Tuya Din RCBO Circuit Breaker."""
+from typing import Dict, Optional, Union
+
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.general import (
+    Basic,
+    DeviceTemperature,
+    Groups,
+    Ota,
+    Scenes,
+    Time,
+)
+from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
+from zigpy.zcl.clusters.smartenergy import Metering
+
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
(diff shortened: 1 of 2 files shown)


dmulcahey:
Are apparent power, instantaneous demand and power factor actually supported?

Shulyaka:
They are not provided by the device, but we can calculate apparent power by multiplying the voltage by the current, and then we can calculate the power factor by dividing the active power by the apparent power. Should I do it?
And I have no idea what instantaneous demand is...

dmulcahey:
ID is the current power draw from what is connected. (How much power is being used in near real time). I am more interested in why they show up if they don’t have values. It means we have something screwy somewhere either in quirks or in ZHA itself.

Shulyaka:
I fail to see the difference:

> 4.9.2.2.6.10 ActivePower
> Represents the single phase or Phase A, current demand of active power delivered or received at the premises, in Watts (W). Positive values indicate power delivered to the premises where negative values indicate power received from the premises.

> 10.4.2.2.5.1 InstantaneousDemand Attribute
> InstantaneousDemand represents the current Demand of Energy, Gas, or Water delivered or received at the premises. Positive values indicate demand delivered to the premises where negative values indicate demand received from the premises. InstantaneousDemand is updated continuously as new measurements are made.

Shulyaka:
I think that's something in ZHA, like if certain cluster is present, the sensors are added regardless if the respective attributes are set.

Shulyaka:
I've just discovered the tuya.PowerOnState class, give me a moment to rewrite the quirk to use it instead of a new class,,,

zhaquirks/tuya/ts0601_rcbo.py
```diff
@@ -0,0 +1,586 @@
...
+            foundation.GeneralCommand.Default_Response
+        ].schema(command_id=command_id, status=foundation.Status.UNSUP_CLUSTER_COMMAND)
+
+
+class TuyaRCBOManufCluster(TuyaMCUCluster):
+    """Tuya with power measurement data points."""
+
+    dp_to_attribute: Dict[int, DPToAttributeMapping] = {
+        TUYA_DP_STATE: DPToAttributeMapping(
+            TuyaOnOff.ep_attribute,
```
javicalle, line 346: Use `TuyaRCBOOnOff` here.

zhaquirks/tuya/ts0601_rcbo.py
```diff
@@ -0,0 +1,586 @@
...
+            "on_off",
+            TuyaDPType.BOOL,
+        ),
+        TUYA_DP_COUNTDOWN_TIMER: DPToAttributeMapping(
+            TuyaOnOff.ep_attribute,
```
javicalle, line 351: Use `TuyaRCBOOnOff` here.

javicalle (changes requested):
Use the same cluster reference in all the mappings.

zhaquirks/tuya/ts0601_rcbo.py
Shulyaka: Fixed

Shulyaka:
Don't merge yet, I want to get the GUI "power on state" selector working (I just found out that it exists!)

Shulyaka:
Ah, there is a hard coded white list in ZHA code for that
```python
@CONFIG_DIAGNOSTIC_MATCH(
    channel_names=CHANNEL_ON_OFF,
    models={"TS011F", "TS0121", "TS0001", "TS0002", "TS0003", "TS0004"},
)
@CONFIG_DIAGNOSTIC_MATCH(
    channel_names="tuya_manufacturer",
    manufacturers={
        "_TZE200_7tdtqgwv",
        "_TZE200_amp6tsvy",
        "_TZE200_oisqyl4o",
        "_TZE200_vhy3iakz",
        "_TZ3000_uim07oem",
        "_TZE200_wfxuhoea",
        "_TZE200_tviaymwx",
        "_TZE200_g1ib5ldv",
        "_TZE200_wunufsil",
        "_TZE200_7deq70b8",
        "_TZE200_tz32mtza",
        "_TZE200_2hf7x9n3",
        "_TZE200_aqnazj70",
        "_TZE200_1ozguk6x",
        "_TZE200_k6jhsr0q",
        "_TZE200_9mahtqtg",
    },
)
class TuyaPowerOnStateSelectEntity(ZCLEnumSelectEntity, id_suffix="power_on_state"):
    """Representation of a ZHA power on state select entity."""

    _select_attr = "power_on_state"
    _enum = TuyaPowerOnState
    _attr_name = "Power on state"
```

Ok, will raise a separate request to ZHA. This PR is ready from my side.

Shulyaka:
@dmulcahey Shall I include apparent power and power factor calculation into the quirk?

изображение

Shulyaka:
I decided to include it, because it's just cool:)

dmulcahey:
@Shulyaka I merged this into the current release and all the tests are now failing... Can you take a peek?

Shulyaka:
Something to do with the byte order

dmulcahey:
yeah https://github.com/zigpy/zha-device-handlers/pull/1999/files

dmulcahey:
image

dmulcahey:
Will this break your prior PR too?

javicalle:
I  have reverted a PR that mess with the TuyaDPType.VALUE, maybe related?

javicalle:
Can I help in some way?

dmulcahey:
> Yes, they are failing because of #1999, which was not merged when I submitted this PR.
> I can create a workaround, then after we reapply #1984 we can revert it as well.

do I need to kill the HA PR w/ the release and do another?

Shulyaka:
It's up to you. But only this device is affected.

Shulyaka:
Do you plan to reapply #1984 some day when the side effects are sorted out?

Shulyaka:
My understanding is the following:
1. Tuya uses big-endian int types.
2. Because zigbee types are little-endian, some tuya quirks made a workaround to reverse byte order. But not all of them.
3. @javicalle made a fix (#1984) to reverse the byte order for all quirks, but that broke those quirks with the workaround, because the values became double-reversed.
4. The fix was reverted (#1999) which in turn broke this PR.

Shulyaka:
By the way there are now big endian types to zigpy: https://github.com/zigpy/zigpy/pull/1124

dmulcahey:
> Do you plan to reapply #1984 some day when the side effects are sorted out?

This is a question for @javicalle I try to stay out of the Tuya stuff...

javicalle:
> 1. Tuya uses big-endian int types.
> 2. Because zigbee types are little-endian, some tuya quirks made a workaround to reverse byte order. But not all of them.

I have to admit that I'm not familar with all that endians, but my understanding is that **all** the quirks using the `TuyaNewManufCluster` (and descendants) made the reverse byte order for the `TuyaDPType.VALUE` (and the BOOL, but that doesn't matter)


> 3. @javicalle made a fix ([Fix TuyaDPType.VALUE payload #1984](https://github.com/zigpy/zha-device-handlers/pull/1984)) to reverse the byte order for all quirks, but that broke those quirks with the workaround, because the values became double-reversed.

I was trying to figure out what the problem was with #1554 and in my analysis I have detected a problem with the reverse of TuyaDPType.VALUE working only in one direction:

1. this code deserialize the Tuya message and reverse all TuyaDPType except the `TuyaDPType.VALUE`:
https://github.com/zigpy/zha-device-handlers/blob/1d7b87e516cd4e3a1505cf979e5d4e42837cd01a/zhaquirks/tuya/__init__.py#L181-L188

2. at this part, the `TuyaData` generation inverts all the bytes for every TuyaDPType (included VALUE): 
https://github.com/zigpy/zha-device-handlers/blob/1d7b87e516cd4e3a1505cf979e5d4e42837cd01a/zhaquirks/tuya/mcu/__init__.py#L243

I have to clarify that not all the quirks use these classes and not all the ones that use the classes have the problem. 

> 4. The fix was reverted ([Revert PR #1984 #1999](https://github.com/zigpy/zha-device-handlers/pull/1999)) which in turn broke this PR.

That's correct.
Your PR was already affected for #1984 and you have to fix your code because of that. I can't find it but I remember causing the problem.



I believe that most of the test fails because the expected value on test, but I try to review your code again to search for problems in the DPType.

Shulyaka:
@javicalle please check #2009

javicalle:
> Do you plan to reapply #1984 some day when the side effects are sorted out?

I have a new version, but I haven't found any user able to test.
But once your PR is applied that would be my problem 😉 



> @javicalle please check #2009

Already doing.

Shulyaka:
> I have a new version, but I haven't found any user able to test.

Please tag me when it is ready, I would like to check it as well if you don't mind. By the way we also reverse the endianess in [Data.from_value()](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L224).

javicalle:
Done, @Shulyaka no objections from my side

javicalle:
> By the way we also reverse the endianess in [Data.from_value()](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L224).

I had missed that call. Thanks for pointing me.
The truth is that the whole implementation of the `from_cluster_data` is quite disastrous (my fault) 😞

PR #1998, opened 2022-12-08
Extra signature for Elko Super TR on firmware 1.2.0R

heikkih:
My version of Elko Super TR was not picked up by the quirk, so I've added the specific signature for this firmware/model to make it work fully.


MattWestb:
If it one updated version you must doing one new device class for not braking "old" devices using the original device class.

heikkih:
> If it one updated version you must doing one new device class for not braking "old" devices using the original device class.

Hmm... you are correct. I tried putting my signature first in the list and it was not recognized, so it would have broken the others. Thanks for catching this.

My bad for misunderstanding the architecture. Seemed just a little bit excessive to copy-paste into a new device class. 

I'll close this and add a new device class in a new PR.

Issue #2005, opened 2022-12-10
Support for TUYA ZigBee Smart Circuit Breaker [TS011F _TZ3000_qeuvnohg]


zleba:
The smart circuit breaker https://www.aliexpress.com/item/1005004747066832.html (with metering) is using "zigpy.device.Device" class.
There are two main problems:
1) Total power consumption in kWh is 100times higher, i.e. the "divider" id=0x0302 is not applied
2) The current in A is 1000times higher, i.e. it's not divided by "ac_current_divisor" id=0x0603

These are addressed in https://github.com/zigpy/zha-device-handlers/pull/2008
The remaining issue is the dummy "temperature" and dummy "power factor" displayed in Home assistant.
See also
https://zigbee.blakadder.com/Tongou_TO-Q-SY1-JZT.html

<details>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x010a",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0402",
        "0x0702",
        "0x0b04",
        "0xe000",
        "0xe001"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZ3000_qeuvnohg",
  "model": "TS011F",
  "class": "zigpy.device.Device"
}

[...]

PR #2008, opened 2022-12-11
Adding support for Tongou circuit breaker (#2005)

zleba:
Related to https://github.com/zigpy/zha-device-handlers/issues/2005
The device is the following one
https://zigbee.blakadder.com/Tongou_TO-Q-SY1-JZT.html

### DIFF
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -153,6 +153,12 @@ class CircuitBreaker(EnchantedDevice):
                 ],
                 OUTPUT_CLUSTERS: [Time.cluster_id, Ota.cluster_id],
             },
+            242: {
+                PROFILE_ID: 41440,
+                DEVICE_TYPE: 97,
+                INPUT_CLUSTERS: [],
+                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
+            },
         },
     }
 
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -95,6 +95,68 @@ class Plug(EnchantedDevice):
     }
 
 
+class CircuitBreaker(EnchantedDevice):
+    """Tuya smart circuit breaker with metering (_TZ3000_qeuvnohg). https://www.aliexpress.com/item/1005004747066832.html """
+
+    signature = {
+        MODEL: "TS011F",
+        ENDPOINTS: {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=266
+            # device_version=1
+            # input_clusters=[0, 3, 4, 5, 6, 1026, 1794, 2820, 57344, 57345]
+            # output_clusters=[10, 25]>
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_PLUG_IN_UNIT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    0x0402,
+                    Metering.cluster_id,
+                    ElectricalMeasurement.cluster_id,
+                    TuyaZBE000Cluster.cluster_id,
+                    TuyaZBExternalSwitchTypeCluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [Time.cluster_id, Ota.cluster_id],
+            },
+            # <SimpleDescriptor endpoint=242 profile=41440 device_type=97
+            # device_version=1
+            # input_clusters=[]
+            # output_clusters=[33]>
+            242: {
+                PROFILE_ID: 41440,
+                DEVICE_TYPE: 97,
+                INPUT_CLUSTERS: [],
+                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
+            },
+        },
+    }
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_PLUG_IN_UNIT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    TuyaZBOnOffAttributeCluster,
+                    0x0402,
+                    TuyaZBMeteringCluster,
+                    TuyaZBElectricalMeasurement,
+                    TuyaZBE000Cluster,
+                    TuyaZBExternalSwitchTypeCluster,
+                ],
+                OUTPUT_CLUSTERS: [Time.cluster_id, Ota.cluster_id],
+            },
+        },
+    }
+
+
 class Plug_1AC(CustomDevice):
     """Tuya plug without metering with restore power state support."""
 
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts011f_plug.py
```diff
@@ -95,6 +95,68 @@ class Plug(EnchantedDevice):
...
+            },
+            # <SimpleDescriptor endpoint=242 profile=41440 device_type=97
+            # device_version=1
+            # input_clusters=[]
+            # output_clusters=[33]>
+            242: {
+                PROFILE_ID: 41440,
+                DEVICE_TYPE: 97,
+                INPUT_CLUSTERS: [],
+                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
```
MattWestb, line 133: Dont deleting the Zigbee GP endpoint in the replacement is one part of the device function.

MattWestb (changes requested):

zhaquirks/tuya/ts011f_plug.py
zleba: Yes, I will keep the 242 endpoint, what confused me was that the "Plug" class does not keep it.

zleba: I am not sure, what is the meaning of the 0x402 cluster, it reports as "temperature" but the value is constant and always "-630" so it has likely some other meaning...

zleba:
"Temperature" seems to have other meaning in reality and there is also a problem with the "Power factor" which is a typical problem for the Tuya plugs.
The "Power factor" can be calculated as
"Active power" / ("RMS current" * "RMS voltage"),
where all 3 numbers are available. 

[](url)
Screenshot from 2022-12-11 22-56-54

javicalle:
`0x0402` is the `TemperatureMeasurement` cluster:
https://github.com/zigpy/zigpy/blob/a5ae12efc6510df9c487e775bef7af8937ad85f5/zigpy/zcl/clusters/measurement.py#L72

Add the import at the top of the file and use it

zleba:
@javicalle , for me it looks like it doesn't measure the temperature, we really don't have -6.3C all the time.
Also in the user guide, there is no mention of the temperature.
Maybe the manufacturer just used some non-standard labelling of the cluster.

Is there a way to remove this cluster until we understand what is stored inside?

2022-12-12 08:26:10.365 DEBUG (MainThread) [zigpy.zcl] [0x9C67:1:0x0402] Received ZCL frame: b'\x18\x12\n\x00\x00)\x8a\xfd'
2022-12-12 08:26:10.366 DEBUG (MainThread) [zigpy.zcl] [0x9C67:1:0x0402] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl(frame_type=<FrameType.GLOBAL_COMMAND: 0>, is_manufacturer_specific=0, direction=<Direction.Client_to_Server: 1>, disable_default_response=1, reserved=0, *is_cluster=False, *is_general=True, *is_reply=True), tsn=18, command_id=10, *direction=<Direction.Client_to_Server: 1>, *is_reply=True)
2022-12-12 08:26:10.368 DEBUG (MainThread) [zigpy.zcl] [0x9C67:1:0x0402] Decoded ZCL frame: TemperatureMeasurement:Report_Attributes(attribute_reports=[Attribute(attrid=0x0000, value=TypeValue(type=int16s, value=-630))])
2022-12-12 08:26:10.369 DEBUG (MainThread) [zigpy.zcl] [0x9C67:1:0x0402] Received command 0x0A (TSN 18): Report_Attributes(attribute_reports=[Attribute(attrid=0x0000, value=TypeValue(type=int16s, value=-630))])
2022-12-12 08:26:10.372 DEBUG (MainThread) [zigpy.zcl] [0x9C67:1:0x0402] Attribute report received: measured_value=-630

javicalle:
Sure, remove it in the `replacement` section (only here).

MattWestb:
If no one have "found" what tuya is using it for i think omitting its the best for not getting complains from users.
If some have one tuya ZBGW and can look what data they is using and sniffing if its needed we can "finding" what is being used for and implanting it.

zleba:
Ok, can I do something with the "Power factor" to not be displayed as "Unknown"?

zhaquirks/tuya/ts011f_plug.py
```diff
@@ -95,6 +95,73 @@ class Plug(EnchantedDevice):
     }
 
 
+class CircuitBreaker(EnchantedDevice):
+    """Tuya smart circuit breaker with metering (_TZ3000_qeuvnohg) https://www.aliexpress.com/item/1005004747066832.html."""
```
Shulyaka, line 99: I wonder, is it a permanent link or it can become inaccessible earlier than the product itself?

zhaquirks/tuya/ts011f_plug.py
```diff
@@ -95,6 +95,73 @@ class Plug(EnchantedDevice):
...
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_PLUG_IN_UNIT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    0x0402,  # unknown meaning, it is not temperature
```
javicalle, line 117: add the `from zigpy.zcl.clusters.measurement import TemperatureMeasurement` (between the `TemperatureMeasurement` and the `Metering`) and use the `TemperatureMeasurement.cluster_id`

zhaquirks/tuya/ts011f_plug.py
javicalle: I would remove it and also the `_TZ3000_qeuvnohg` reference.

javicalle (changes requested):

Shulyaka:
> Ok, can I do something with the "Power factor" to not be displayed as "Unknown"?

ZHA adds this sensor automatically. A user can disable the entity with UI.

But as you correctly pointed out, you can calculate the value, here is my example:
https://github.com/zigpy/zha-device-handlers/pull/1991/commits/c7df85fe7d6283ebbb1b543310706b3e9f8de9eb#diff-e9e69504176ec01e00193622a35fe1bb09552546852ca23d1648d97655dbf73dR260

Shulyaka:
By the way, ZHA expects this value in permille, which differs from the ZCL specification.

zhaquirks/tuya/ts011f_plug.py
arsenyspb: @zleba thanks for your contribution, maybe point to [Tongou TO-Q-SY1-JZT](https://zigbee.blakadder.com/Tongou_TO-Q-SY1-JZT.html) rather?

zhaquirks/tuya/ts011f_plug.py
arsenyspb: @zleba I confirm it on my batch (all 4 show this, and it's certainly not a temperature), I think you are on the right path to ignore this reading, it's also not in the manual (as you confirmed [here](https://github.com/zigpy/zha-device-handlers/issues/2005) )

arsenyspb:
Hi @zleba thank you this quirk PR to support of ZHA community 🤝 🙇 👍  [Tongou TO-Q-SY1-JZT](https://zigbee.blakadder.com/Tongou_TO-Q-SY1-JZT.html) is popular: there were no mentions of it till Q3/22, but now a spike in issues and asks all over community.

I'm sure you may have seen it that Z2M's got workaround for [division by 100](https://github.com/Koenkk/zigbee-herdsman-converters/blob/348dc94d3daaba5d7041410e1dda274af236b23f/devices/tuya.js#L3428) of summation and A to mA with division by 1000. I'm not sure how this would be solved in ZHA at the moment, but maybe you can add instructions for community in your [original issue](https://github.com/zigpy/zha-device-handlers/issues/2005)?

Really looking forward to seeing this PR merged.

TheJulianJES:
Thanks for your contribution @zleba. The changes (and more) are included in:
- https://github.com/zigpy/zha-device-handlers/pull/225

PR #1999, opened 2022-12-08
Revert PR #1984

javicalle:
Users are reporting bad behavior with the dimming function after PR #1984:
* https://github.com/zigpy/zha-device-handlers/issues/1554#issuecomment-1343399518

Reverting to fix issues.

Fix: #1996
Fix: #1750

PR #2009, opened 2022-12-12
Fix ts0601_rcbo.py after #1999

Shulyaka:
This PR is to fix the quirk for Tuya ts0601 _TZE200_hkdl5fmv circuit breaker for the changes made in #1999
See #1991 for more details.


Shulyaka:
There's a more clean way to do that, please hold on

javicalle:
The code doesn't seem like it's going to break anything implemented, so it looks good to me.

The only thing to comment would be that this part here:
```python
t.uint32_t.deserialize(x.serialize()[::-1])[0]
```

seems to me to be the same as what this code does (with a lot of validations):
```python
t.LVBytes.deserialize(val)[0]
```

So not sure if in some way that would be simplified, but if code works for you is fine to me.

I suggest that we release the current version 'as is' and if you want we can comment on the rest and assess if another PR is necessary.

javicalle (approved):
LGTM

PR #2016, opened 2022-12-17
Added support for RM3500ZB Calypso

claudegel:
Added support for the water heater tank Calypso from Sinopé, RM3500ZB

### DIFF
--- a/zhaquirks/sinope/switch.py
+++ b/zhaquirks/sinope/switch.py
@@ -10,6 +10,7 @@ import zigpy.types as t
 from zigpy.zcl.clusters.general import (
     Basic,
     BinaryInput,
+    DeviceTemperature,
     Groups,
     Identify,
     LevelControl,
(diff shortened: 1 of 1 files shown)


zhaquirks/sinope/switch.py
```diff
@@ -123,7 +142,6 @@ class SinopeTechnologiesLoadController(CustomDevice):
         # output_clusters=[3, 4, 25]>
         MODELS_INFO: [
             (SINOPE, "RM3250ZB"),
-            (SINOPE, "RM3500ZB"),
```
javicalle, line 126: Are you sure that there is no `RM3500ZB` devices with that signature.
Because you have added a new device signature there's no need to remove it from others quirks just because have the same model. While the signatures are different you can keep both of them.

claudegel: Yes I'm sure that there are no RM3500ZB with that signature. It was added there before the devices launch. Now that it is available I was able to test it and compare with the RM3250ZB. Both devices are quite different in there signature with lot of difference in manufacturer cluster specific attributes. So after testing I put them in separate signature. There is only one model for RM3500ZB and none with the signature of the RN3250ZB.

PR #2017, opened 2022-12-17
RFC: Tuya data refactoring (implicit conversion of `tuya.Data`)

Shulyaka:
Ok, I am not sure all this is needed and I need your advice.

This PR adds implicit conversions for `tuya.Data` class.

`tuya.Data.to_value` and `tuya.Data.from_value` were removed. Instead the type conversion function were added.

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -276,12 +276,12 @@ def test_ts0121_signature(assert_signature_matches_quirk):
 
 async def test_tuya_data_conversion():
     """Test tuya conversion from Data to ztype and reverse."""
-    assert Data([4, 0, 0, 1, 39]).to_value(t.uint32_t) == 295
-    assert Data([4, 0, 0, 0, 220]).to_value(t.uint32_t) == 220
-    assert Data([4, 255, 255, 255, 236]).to_value(t.int32s) == -20
-    assert Data.from_value(t.uint32_t(295)) == [4, 0, 0, 1, 39]
-    assert Data.from_value(t.uint32_t(220)) == [4, 0, 0, 0, 220]
-    assert Data.from_value(t.int32s(-20)) == [4, 255, 255, 255, 236]
+    assert t.uint32_t(Data([4, 0, 0, 1, 39])) == 295
+    assert t.uint32_t(Data([4, 0, 0, 0, 220])) == 220
+    assert t.int32s(Data([4, 255, 255, 255, 236])) == -20
+    assert Data(t.uint32_t(295)) == [4, 0, 0, 1, 39]
+    assert Data(t.uint32_t(220)) == [4, 0, 0, 0, 220]
+    assert Data(t.int32s(-20)) == [4, 255, 255, 255, 236]
 
 
 class TuyaTestManufCluster(TuyaManufClusterAttributes):
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -231,22 +231,35 @@ class TuyaData(t.Struct):
 class Data(t.List, item_type=t.uint8_t):
     """list of uint8_t."""
 
-    @classmethod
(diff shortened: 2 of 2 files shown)


javicalle:
Hi @Shulyaka, thanks once again for your contributions.

Firts I will try to put some background info and then I will try to check the code.

When firsts Tuya dimmers come, I noticed that will need some work to make it work. I check the code and looking for a solution I found the [`TuyaNewManufCluster`](https://github.com/zigpy/zha-device-handlers/blob/954745eaf47b205f99534e34e11679b8853d8878/zhaquirks/tuya/__init__.py#L1355). 
Once I understood how it works, I decided to use that approach for new implementations, but (due to my poor knowledge of Python) I tried to avoid affecting existing implementations as much as possible. It is for this reason that most of the implementations in `zhaquirks.tuya.mcu.__init__.py` extend the `zhaquirks.tuya.__init__.py` classes (instead of implementing on top of them).
That approach allowed me to create new implementations without the risk of 'breaking' existing implementations and also segregate the MCU classes from other standard Zigbee implementations.

My goal was also to bring over time all the implementations related to the Tuya MCU devices to the new package as the knowledge and development of the classes allowed me and (once I was confident enough) to migrate all the references and quirks to the new ones classes and delete the old classes so that nobody uses them.

Obviously, my forecast has not been fulfilled but in my head it is still the goal to meet.

I hope all this explanation helps you understand what I did in `zhaquirks.tuya.mcu.__init__.py` and my approach in my review.

Shulyaka:
Totally agree on both approach and future goals. By the way, are all tuya devices MCU? I mean, can they all in theory be migrated to `TuyaMCUCluster`?
Python is not my native language either.
But I think some of these ideas might be worth including. I've seen you were going to do part of it yourself (remove reversing of `TuyaData.raw` on deserializing and do it in `TuyaData.payload` when needed)

MattWestb:
First gen tuya TRVs that is using tuya MCU was having the Zigbee manufacture name `_TYST11_*` and then updated Zigbee module all is having `TS0601` as Zigbee model id.

But i can being "hybrid devices" that looks like normal Zigbee devices but is also sending tuya MCU commands.

javicalle:
> are all tuya devices MCU? I mean, can they all in theory be migrated to `TuyaMCUCluster`?

No, they are some Tuya devices behaving 'standard' zigbee devices.

Tuya devices are like a bizarre forest where you can find any kind of devices.


> But I think some of these ideas might be worth including.

Yes, for sure.


> I've seen you were going to do part of it yourself

Well, I was looking at the `TuyaData.VALUE` but you're right. I'll try to check it over your implementation.

Shulyaka:
I've split this PR into 3, now it has only left a single change.

javicalle:
Thanks for not giving up.

Shulyaka:
I am having fun:)

dmulcahey:
@Shulyaka is this still being worked on? I know you split a bunch into separate PR’s. Just checking on this one. Thanks!

Shulyaka:
Hi! Yes. This is the remaining part.
I've just rebased it.
Sorry for the delay.

javicalle (approved):
Despite that all the `Data` methods are beyond my level I am confident in the tests so...
LGTM

TheJulianJES:
Do we want to merge this for the next zha-quirks release?

TheJulianJES (approved):
LGTM – don't see any obvious issues 😄

javicalle:
> Do we want to merge this for the next zha-quirks release? (so for HA Core 2023.3.0)

Why not? 😉
Let's put it on the beta release.

kukudemajia:
This is a terrible change, after all, many end users are not experts in this area.

javicalle:
> This is a terrible change, after all, many end users are not experts in this area.

Can you elaborate it?

PR #2021, opened 2022-12-18
fix tuya rcbo types

Shulyaka:
Fix some types for rcbo (#1991)

1. countdown_timer should be big-endian, not little-endian
2. trip locking should be boolean, not enum
3. on_off should be boolean, not int

I've decided to raise this part separately from #2017


javicalle (approved):
LGTM

PR #2026, opened 2022-12-21
Don't use `TuyaData.raw` directly

Shulyaka:
I think this is the least controversial part of #2017, I decided to raise it separately to make the review process easier.

`tuya.TuyaData.raw` is no longer affected on deserialization. That allows to remove the overloaded `deserialize()` method. It also fixes a cosmetic issue that payload is wrongly calculated in the logs for outgoing messages. Instead, a setter for the `TuyaData.payload` was added so that `raw` is no longer read or written directly.

Supersedes #1984

### DIFF
--- a/tests/test_tuya_clusters.py
+++ b/tests/test_tuya_clusters.py
@@ -143,7 +143,7 @@ def test_tuya_data_bitmap():
     assert r.raw == b"\x40"
     assert r.payload == 0x40
 
-    r.payload = t.bitmap8(0x82)
+    r.payload = 0x82
     assert r.raw == b"\x82"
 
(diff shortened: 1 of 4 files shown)


Shulyaka:
I will add some tests

javicalle:
One more time: thanks for your efforts.
I'm away from my computer now but i can say that LGTM.


> It also fixes a cosmetic issue that payload is wrongly calculated in the logs for outgoing messages.

THANKS for that. Was driving me nuts every time I step again in an issue.

javicalle (approved):
LGTM
If no objection I would like to put in the next HA beta release (and not in a 2022.12.X).

Issue #858, opened 2021-04-19
[Device Support Request] Philips Hue Wall Switch Module


aleex1848:
**Describe the solution you'd like**
There should be a zha_event triggered when the switch is pushed.

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be 

Log of the device joining the network:
```python
Device 0x2ab8 (00:17:88:01:0b:00:12:18) joined the network
[0x2ab8:zdo] ZDO request ZDOCmd.Device_annce: [0x2AB8, 00:17:88:01:0b:00:12:18, 128]
[0x2ab8] Requesting 'Node Descriptor'
Tries remaining: 2
[0x2ab8] Extending timeout for 0x52 request
[0x2ab8] Node Descriptor: NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4107, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0, *allocate_address=True, *complex_descriptor_available=False, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False, *is_valid=True, *logical_type=<LogicalType.EndDevice: 2>, *user_descriptor_available=False)
[0x2ab8] Discovering endpoints
Tries remaining: 3
[0x2ab8] Extending timeout for 0x53 request
[0x2ab8] Discovered endpoints: [1]
[0x2ab8:1] Discovering endpoint information
Tries remaining: 3
[0x2ab8] Extending timeout for 0x54 request
[0x2ab8:1] Discovered endpoint information: SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=2080, device_version=1, input_clusters=[0, 1, 3, 64512], output_clusters=[3, 4, 6, 8, 25])
[0x2ab8] Extending timeout for 0x55 request
light.osram_classic_b40_tw_lightify_e287a900_level_light_color_on_off: polling current state
[0x5a9a:3:0x0006] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=GLOBAL_COMMAND manufacturer_specific=False is_reply=True disable_default_response=True> manufacturer=None tsn=86 command_id=Command.Read_Attributes_rsp>
[...]

PR #2027, opened 2022-12-21
RDM001: add automation triggers for right switch

silviokennecke:
The Philips Hue Wall Switch Module (RDM001) supports up to two switches. If the right switch is enabled, can be controlled by the PhilipsBasicCluster attribute `mode` (`0x0034`). Apparently, the only device events which currently get emitted are the ones for the left switch.

This PR adds device triggers for the commands of the right switch. The triggers for the left switch are also affected by this PR to not have them named differently. Therefore, this PR is a **breaking change**.

Related issues: #858

### DIFF
--- a/zhaquirks/philips/rdm001.py
+++ b/zhaquirks/philips/rdm001.py
@@ -25,7 +25,6 @@ from zhaquirks.const import (
     DOUBLE_PRESS,
     ENDPOINTS,
     INPUT_CLUSTERS,
-    LEFT,
     LONG_PRESS,
     LONG_RELEASE,
     MODELS_INFO,
@@ -38,6 +37,7 @@ from zhaquirks.const import (
     SHORT_PRESS,
     SHORT_RELEASE,
     TRIPLE_PRESS,
+    TURN_ON,
     ZHA_SEND_EVENT,
 )
 from zhaquirks.philips import PHILIPS, SIGNIFY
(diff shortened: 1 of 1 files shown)


TheJulianJES:
Generally, I think this is the right approach.

Although for users who only have one switch attached, the triggers with left/right could be a bit confusing, but we can't really switch the automations depending on the mode.

Also, there might be quite a few users with these Hue wall-switch modules already, so it would be best to avoid a breaking change.

(It's not clean at all, but should we maybe consider just adding the "right" triggers in addition to the the TURN_ON triggers for the left button? We could still change it later (with a breaking change for the "left" button users, the right button triggers would stay unaffected.))

silviokennecke:
Thought about that, too. I chose to make the PR as it is, as I thought to better have it clean than having old code to carry around. But I'll change it

silviokennecke:
I assume it would not work, but is it possible to define a command twice?
Then we could mark the old `TURN_ON` events as deprecated and allow users to already use the new triggers.

What do you think @TheJulianJES? I think you know the home-assistant/core code better than me.

```
(SHORT_PRESS, TURN_ON): {COMMAND: "left_press"},
(SHORT_PRESS, LEFT): {COMMAND: "left_press"},
```

MattWestb:
I dont knowing if its possible or not allowed but i have using `button` and `button 1` for one device that have only one button and can being rotated and have 2 different modes.
Its working well but is not so nice and you can look for the DA here https://github.com/zigpy/zha-device-handlers/blob/9b09ea5f2c99671fb6621ff42d552532526b85f4/zhaquirks/tuya/ts004f.py#L133

PR #2028, opened 2022-12-22
DA Support for Styrbar Version 2.4.5

MattWestb:
The OTA is released but have lower version in the metadata (Thanks our Puddly for verifying it !!) so ZHA and Z2M is not updating it and its looks Dirigera also not updating them.
But the old trådfri gateway is doing it OK and ZHA is showing the new version of the firmware OK so we can getting user using them.

IKEA have changing dim down to using COMMAND_STOP instead of COMMAND_STOP_ON_OFF that was used for both dim up and down > Dim down was trigger DA dim up. The updated is dim down still triggering dim up DA for old firmware  but the updated is doing it OK.

Tested with both firmware and have using all DA for doing automatons !!!

Also some small cluster setting we is getting on the way.

Have fixing some wrong  spelling in the signature.

PS: for user having problem with right / left long pres: The device is sending one reset cycle for syncing all light bounded and in it is  also ON command sent and we cant do anything about it so ON DA is also triggered  and the release is sending random numbers so we cant using it in DA.
The device is one light controller if like getting mega clicks use one HA controllers that is sending "clicks" !!!

Edit: The OTA updating problem is documented here https://github.com/zigpy/zigpy/discussions/660#discussioncomment-4460743.

Edit 2: Group binding is gone 👎 👎 👎

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -615,7 +615,7 @@ KNOWN_DUPLICATE_TRIGGERS = {
             (zhaquirks.aurora.aurora_dimmer.COLOR_DOWN, const.LEFT),
         ],
     ],
-    zhaquirks.ikea.fourbtnremote.IkeaTradfriRemote: [
+    zhaquirks.ikea.fourbtnremote.IkeaTradfriRemoteV1: [
         [
             (const.LONG_RELEASE, const.DIM_UP),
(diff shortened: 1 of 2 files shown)


MattWestb:
Fixed test that was failing then renaming the original device class and i like renaming it for looking better and making it more future proof.

MattWestb:
Thanks **D**  !!

PR #2031, opened 2022-12-22
Fixed Nodon module

altmenorg:
Added ``__init__.py`` and fixed syntax for the module to actually work.

This makes https://github.com/zigpy/zha-device-handlers/pull/1626 actually work.

### DIFF
--- a/zhaquirks/nodon/sin-4-2-20.py
+++ b/zhaquirks/nodon/sin-4-2-20.py
@@ -1,4 +1,6 @@
 """NodOn on/off switch two channels."""
+import logging
+
 from zigpy.profiles import zha
 from zigpy.quirks import CustomDevice
 from zigpy.zcl.clusters.general import (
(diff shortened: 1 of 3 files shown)


zhaquirks/nodon/sin-4-2-20.py
```diff
@@ -23,6 +25,8 @@
     PROFILE_ID,
 )
 
+_LOGGER = logging.getLogger(__name__)
```
TheJulianJES, line 28: I think the logging part isn't used and should be removed.

zhaquirks/nodon/sin-4-2-20.py
```diff
@@ -1,4 +1,6 @@
 """NodOn on/off switch two channels."""
+import logging
+
```
TheJulianJES, line 3: I think the logging part isn't used and should be removed.

TheJulianJES (approved):
Rebased the PR. I also renamed the file, so it can be correctly imported and removed the logger.

LGTM now

Issue #2025, opened 2022-12-20
[Device Support Request] Support temperature offset for Tuya Saswell TRV _TZE200_c88teujp


jclsn:
**Is your feature request related to a problem? Please describe.**
I would like to set a temperature offset for the Saswell TRVs. It seems there needs to be a data point mapping implemented. The model is listed here 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/927f352c9dc848f716c23abf9d2a0a56c13a9f67/devices/saswell.js#L11

and here 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/927f352c9dc848f716c23abf9d2a0a56c13a9f67/lib/tuya.js#L516

but I have no idea how to implement a mapping here 

https://github.com/zigpy/zha-device-handlers/blob/c2b2893d027fdcd3a225996fdf382d2ac72ff799/zhaquirks/tuya/ts0601_valve.py#L161

@javicalle dmulcahey said to mention you here ;)

**Describe the solution you'd like**
Implement a data point mapping for a given model

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0004",
        "0x0005",
        "0x0201",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_c88teujp",
  "model": "TS0601",
[...]

PR #2039, opened 2022-12-29
Add temperature calibration for Tuya TZE200 c88teujp 

jclsn:
This adds temperature calibration for the Tuya TZE200 c88teujp. The temperature can be calibrated by -6 to 6°C/F in steps of one. The step size is a bit large, but the device seems to translate data point integers directly to full degrees and not decidegrees as it does for the local temperature or heating setpoint.

Fixes https://github.com/zigpy/zha-device-handlers/issues/2025


zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -128,15 +153,15 @@ def map_attribute(self, attribute, value):
...
             # more details: https://github.com/zigpy/zha-device-handlers/issues/1815
             if value == self.SystemMode.Off:
-                return {SASSWEL_STATE_ATTR: 0, SASSWEL_SCHEDULE_ENABLE_ATTR: 0}
+                return {SASWELL_STATE_ATTR: 0, SASWELL_SCHEDULE_ENABLE_ATTR: 0}
             if value == self.SystemMode.Heat:
-                return {SASSWEL_STATE_ATTR: 1, SASSWEL_SCHEDULE_ENABLE_ATTR: 0}
+                return {SASWELL_STATE_ATTR: 1, SASWELL_SCHEDULE_ENABLE_ATTR: 0}
 
 
 class Thermostat_TYST11_c88teujp(TuyaThermostat):
```
MattWestb, line 167: This device class is using the same tuya MCU but one older Zigbee module so it shall working the same way
Pleas put the calibration in it so users that having both the new and old version is getting the same functionality.

MattWestb (changes requested):

MattWestb:
Only the bus is miss place and need being in the first position:
```
-from zhaquirks import LocalDataCluster, Bus
+from zhaquirks import Bus, LocalDataCluster
```

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -192,9 +217,54 @@ class Thermostat_TYST11_c88teujp(TuyaThermostat):
     }
 
 
+class Thermostat_TZE200_c88teujp_TemperatureOffset(LocalDataCluster, AnalogOutput):
```
javicalle, line 220: Remove manufacturer name from class name:
```python
class TemperatureOffset(LocalDataCluster, AnalogOutput):
```

(or maybe `TuyaTemperatureOffset`)

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -192,9 +217,54 @@ class Thermostat_TYST11_c88teujp(TuyaThermostat):
...
+            if isinstance(attrid, str):
+                attrid = self.attributes_by_name[attrid].id
+            if attrid not in self.attributes:
+                self.error("%d is not a valid attribute id", attrid)
+                continue
+            self._update_attribute(attrid, value)
+
+            await self.endpoint.tuya_manufacturer.write_attributes(
+                {SASWELL_TEMPERATURE_CALIBRATION_ATTR: value}, manufacturer=None
+            )
```
javicalle, line 256: That will update `SASWELL_TEMPERATURE_CALIBRATION_ATTR` for any cluster attribute being updated.
Validate `attriid == 0x0055` (`present_value`)

javicalle (changes requested):

zhaquirks/tuya/ts0601_trv_sas.py
jclsn: I don't understand. The name used to be like this before.

zhaquirks/tuya/ts0601_trv_sas.py
jclsn: I changed the name to TuyaTemperatureOffset

zhaquirks/tuya/ts0601_trv_sas.py
jclsn: If I understand you right, this will overwrite ``attriid = 0x0055`` with  ``attriid = 0x38`` from the quirk? I wouldn't know how to address this. Delete the statement?

zhaquirks/tuya/ts0601_trv_sas.py
MattWestb: Yes putting in the same as you have doing in old line `195 class Thermostat_TZE200_c88teujp(TuyaThermostat):` so both device classes is getting the same functionality.

zhaquirks/tuya/ts0601_trv_sas.py
javicalle: The call is fine, the problem would be when is it called.

The `TuyaTemperatureOffset` would have multiple attributes (all the ones from [AnalogInput](https://github.com/zigpy/zigpy/blob/8af92dee9182ce54966621da1d0125e67b8c72f5/zigpy/zcl/clusters/general.py#L1167)).

But you only wants to call this part if the updated attribute is the `present_value`:
```python
            await self.endpoint.tuya_manufacturer.write_attributes(
                {SASWELL_TEMPERATURE_CALIBRATION_ATTR: value}, manufacturer=None
            )
```
and not for any other attribute.

Usually you would wrap the code with a validation like:
```python
        if attrid == 0x0055:  # `present_value`
                await self.endpoint.tuya_manufacturer.write_attributes(
                    {SASWELL_TEMPERATURE_CALIBRATION_ATTR: value}, manufacturer=None
                )
```

zhaquirks/tuya/ts0601_trv_sas.py
javicalle: > I added the calibration to that list. Is that correct?

Not, is not needed.

@MattWestb is asking to you to implement the same code in the `Thermostat_TYST11_c88teujp` class:
https://github.com/zigpy/zha-device-handlers/blob/f66d4335e8f0a95abbc154543054a870bfde0ad9/zhaquirks/tuya/ts0601_trv_sas.py#L167

Just add to it:
* temperature_calibration_bus 
* TuyaTemperatureOffset

You would need to put the new classes upper than the `Thermostat_TYST11_c88teujp` definition

jclsn: Isn't that related to the fix tor the schedule mode that has been implemented lately? I did not do that, but I should include it in the PR whilst were at it?

zhaquirks/tuya/ts0601_trv_sas.py
jclsn: I added your suggested validation.

zhaquirks/tuya/ts0601_trv_sas.py
jclsn: Should be fixed now.

zhaquirks/tuya/ts0601_trv_sas.py
```diff
@@ -81,17 +97,26 @@ def _update_attribute(self, attrid, value):
...
             )
 
+        if attrid == (SASWELL_TEMPERATURE_CALIBRATION_ATTR):
+            self.endpoint.device.temperature_calibration_bus.listener_event(
+                "set_value", value
+            )
+        elif attrid in (SASWELL_LOCAL_TEMP_ATTR, SASWELL_HEATING_SETPOINT_ATTR):
+            self.endpoint.device.thermostat_bus.listener_event(
+                "state_temp_change", attrid, value
+            )
```
javicalle, line 118: What is the intention of this? Where it come from?

The `thermostat_bus` doesn't have an `state_temp_change` event listener:
* https://github.com/zigpy/zha-device-handlers/blob/c2b2893d027fdcd3a225996fdf382d2ac72ff799/zhaquirks/tuya/__init__.py#L594

zhaquirks/tuya/ts0601_trv_sas.py.orig
```diff
@@ -0,0 +1,382 @@
+"""Saswell (Tuya whitelabel) 88teujp thermostat valve quirk."""
```
javicalle, line 1: Remove that file from PR.

javicalle (changes requested):

zhaquirks/tuya/ts0601_trv_sas.py.orig
jclsn: Oh sorry

zhaquirks/tuya/ts0601_trv_sas.py
jclsn: No idea, I just copied the stuff over from the file you suggested until I got something working. I don't really understand all the building blocks yet. Is that not needed?

zhaquirks/tuya/ts0601_trv_sas.py.orig
jclsn: Done

zhaquirks/tuya/ts0601_trv_sas.py
jclsn: Yeah guess it doesn't do anything really. I removed it.

zhaquirks/tuya/ts0601_trv_sas.py.orig
javicalle: Nope.
The file is still here 🤷🏻‍♂️

dmulcahey:
I deleted the extra file. @javicalle is everything else here looking good?

javicalle:
Yes, it is.
Thanks David.

javicalle (approved):
LGTM

PR #2040, opened 2022-12-29
Ensure all ZCL command definitions use dictionaries

puddly:
A single Tuya device is still using the old ZCL command format (which would have caused runtime errors).


MattWestb:
_The never ending story_ "ZCL" and very fast is coming the same for Matter devices :-))

Great catch !!!

PR #2043, opened 2022-12-30
Set TuyaData.dp_type automatically

Shulyaka:
This is also a part separated from #2017 

This PR adds the `tuya.TuyaData.__init__()` constructor that automatically maps TuyaData.dp_type from zigpy type. That allows to remove the `dp_type` field from the `tuya.mcu.DPToAttributeMapping` mapping which may simplify the things.

This is up to the developers to decide whether this is needed or not.

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -1,6 +1,7 @@
 """Tuya devices."""
 import dataclasses
 import datetime
+import enum
 import logging
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -225,6 +226,30 @@ class TuyaData(t.Struct):
             self.raw = value.serialize()
         else:
             raise ValueError(f"Unknown {self.dp_type} datapoint type")
+
+    def __new__(cls, *args, **kwargs):
+        """Disable copy constrctor."""
+        return super().__new__(cls)
+
+    def __init__(self, value=None, function=0, *args, **kwargs):
+        """Convert from a zigpy typed value to a tuya data payload."""
+        if value is None:
+            return
+        elif isinstance(value, (t.bitmap8, t.bitmap16, t.bitmap32)):
+            self.dp_type = TuyaDPType.BITMAP
+        elif isinstance(value, (bool, t.Bool)):
+            self.dp_type = TuyaDPType.BOOL
+        elif isinstance(value, enum.Enum):
+            self.dp_type = TuyaDPType.ENUM
+        elif isinstance(value, int):
+            self.dp_type = TuyaDPType.VALUE
+        elif isinstance(value, str):
+            self.dp_type = TuyaDPType.STRING
+        elif isinstance(value, t.Struct):
+            self.dp_type = TuyaDPType.RAW
+
+        self.function = function
+        self.payload = value
 
 
 class Data(t.List, item_type=t.uint8_t):
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -247,7 +247,7 @@ class TuyaData(t.Struct):
             self.dp_type = TuyaDPType.VALUE
         elif isinstance(value, str):
             self.dp_type = TuyaDPType.STRING
-        elif isinstance(value, t.Struct):
+        else:
             self.dp_type = TuyaDPType.RAW
 
         self.payload = value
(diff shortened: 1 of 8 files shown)


javicalle:
I like the idea very much, but I'm not sure if will work.
I thought I had done tests to validate the `TuyaData.dp_type`, but I see that there aren't any, so not sure.

My concern is that the implementation is based on the type of the data:
```python
isinstance(value, whatever)
```

And the instances of `TuyaDatapointData` will be created here:
* https://github.com/zigpy/zha-device-handlers/blob/a6556fea6ab6400984094dcc88d7bd772e322e97/zhaquirks/tuya/mcu/__init__.py#L182

Upper in code, the method will be called from here:
* https://github.com/zigpy/zha-device-handlers/blob/a6556fea6ab6400984094dcc88d7bd772e322e97/zhaquirks/tuya/mcu/__init__.py#L230

And data will come in a `TuyaClusterData` object that is defined as:
* https://github.com/zigpy/zha-device-handlers/blob/a6556fea6ab6400984094dcc88d7bd772e322e97/zhaquirks/tuya/mcu/__init__.py#L59-L67

Do you see my point?

I'm pretty sure that must be a way to avoid the type instantiation as you want but I haven't found it.
In origin, we know the cluster attribute to update and its value, but not sure if a direct mapping Zigbee attribute type <--> MCU DP type will be possible.

Shulyaka:
Fixed annotation for `TuyaClusterData.attr_value`.

The type is set when we assign the attribute, here is an example for bool:

https://github.com/zigpy/zha-device-handlers/blob/507edfcc6fa6e00c42fad28d5b084751ce8a8557/zhaquirks/tuya/mcu/__init__.py#L337

Shulyaka:
Hm, that fails for python 3.8 and 3.9, checking.

javicalle:
Ok, I see...
What about this?
https://github.com/zigpy/zha-device-handlers/blob/a6556fea6ab6400984094dcc88d7bd772e322e97/zhaquirks/tuya/mcu/__init__.py#L109

The method is invoked when you update one cluster attribute from the 'manage cluster' view (and maybe other places). 🤔

Shulyaka:
> What about this?

In this case we are using the type of the cluster attribute.
See
https://github.com/zigpy/zigpy/blob/master/zigpy/zcl/__init__.py#L547

Shulyaka:
> Hm, that fails for python 3.8 and 3.9, checking.

Ok, I removed the annotation fix because I couldn't easily make it work for python below 3.10. `zigpy.types.Struct` doesn't want to accept `typing.Union`.

Regardless, it looks like TuyaClusterData.attr_value can have types other than int, I don't know if it is a zigpy bug or not.

dmulcahey:
Are we good with this?

dmulcahey:
@Shulyaka @javicalle please let me know when this is mergable

Shulyaka:
From my perspective, it is ready. @javicalle just needs to decide if the idea worth it.

dmulcahey:
@javicalle thoughts?

javicalle:
I meant to do some testing and validate this change but all my installation is broken and I'm having trouble spending time to fix it.
Please let me check it out over the weekend.

javicalle:
First of all, I apologize to @Shulyaka for not having been able to review the topic before 🙇🏻‍♂️ 

Apart from a minor comment, let's go for it!
@dmulcahey  would it be possible to not include the change until the 01/25 beta release?
@Shulyaka anything I can help you with, just tell me.

zhaquirks/tuya/__init__.py
```diff
@@ -226,6 +227,31 @@ def payload(self, value):
...
+            self.dp_type = TuyaDPType.BOOL
+        elif isinstance(value, enum.Enum):
+            self.dp_type = TuyaDPType.ENUM
+        elif isinstance(value, int):
+            self.dp_type = TuyaDPType.VALUE
+        elif isinstance(value, str):
+            self.dp_type = TuyaDPType.STRING
+        elif isinstance(value, t.Struct):
+            self.dp_type = TuyaDPType.RAW
+
```
javicalle, line 252: Maybe a safe check here to be sure to assign the `dp_type`?

dmulcahey:
@Shulyaka can you rebase this and see if any of the recently merged tuya quirks need changes? Once that’s done I’ll merge this and get a release produced for the beta.

Shulyaka:
Done.
Sorry for the delay, I had a flu

zhaquirks/tuya/__init__.py
Shulyaka: I have a different idea, to set it default to RAW

Shulyaka: Done

PR #2046, opened 2022-12-30
Merger TS004Fs to EnchantedDevice.

MattWestb:
Merger all device classes of **TS004f** to `EnchantedDevice` and adding one more found DS. All function is tested after deleting and reset and removing the battery and adding them they is getting the magic function then doing one new joining. All commands looks OK in the log but have not testing the DA then its not changed.

Its 2 more new devices that we have working but the user have not confirming they is working OK so not implanted (LIDL pad and the press knob).

### DIFF
--- a/zhaquirks/tuya/ts004f.py
+++ b/zhaquirks/tuya/ts004f.py
@@ -1,7 +1,6 @@
 """Tuya TS004F devices."""
 from __future__ import annotations
 
-import asyncio
 import logging
 
 from zigpy.profiles import zha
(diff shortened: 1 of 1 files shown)


MattWestb:
@javicalle  Now the most tricky "normal" Zigbee device is migrated and looks working great with your implementation.

**feliz año nuevo** !!

javicalle (approved):
LGTM

javicalle:
> **feliz año nuevo** !!

Happy new year to you too!!!

MattWestb:
Thanks **D**(&)**J**

PR #2049, opened 2022-12-31
Merger all TS001X device classes to "EnchantedDevice"

MattWestb:
Use `EnchantedDevice` as some new devices is needing it and it shall not harming old devices wot we knowing.

Fixes: #2042


javicalle (approved):
LGTM

PR #2057, opened 2023-01-02
Fix lumi plug maus01 energy monitoring and configuration

dmulcahey:
This PR removes the skip configuration flag and adds 2 missing clusters so that energy monitoring works correctly. There are several other functions supported by the device that I'll add in a subsequent PR. 

image


ownaginatious:
Hey, thanks for adding support for this thing! I just bought one of these US plugs and it works right out of the box :+1:

> There are several other functions supported by the device that I'll add in a subsequent PR.

Does this include the "power outage memory" functionality that the EU variant has? Just curious because if it can't be supported due to hardware limitations, I'll return this for a different plug.

TheJulianJES:
@ownaginatious Please create a new issue. I might be able to throw together a custom quirk later and post it in that issue for you to test.

PR #2060, opened 2023-01-03
Fix reported values for Tuya Air sensor

adrianschroeter:
CO2 and Formaldehyd values were mixed up. Also the formaldehyd value always reported wrong values, I can prove it by being still alive :)

The formaldehyd value is now in the 0.00X ppm range normaly. That means we need also the increase the decimals in HA zha/sensor.py from 0 to 3 or the value will always be zero.  Shall I create also a pull request for HA?

@cloudbr34k84 you seem to had the same problem in https://github.com/zigpy/zha-device-handlers/issues/1313 , could you verify that this is solving the issue for you as well?

### DIFF
--- a/zhaquirks/tuya/air/__init__.py
+++ b/zhaquirks/tuya/air/__init__.py
@@ -52,9 +52,9 @@ class TuyaCO2ManufCluster(TuyaNewManufCluster):
 
     dp_to_attribute: Dict[int, DPToAttributeMapping] = {
         2: DPToAttributeMapping(
-            TuyaAirQualityCO2.ep_attribute,
+            TuyaAirQualityFormaldehyde.ep_attribute,
             "measured_value",
-            lambda x: x * 1e-6,
+            lambda x: x * 1e-9,
         ),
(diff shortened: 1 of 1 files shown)


javicalle:
I'm pretty sure that there are 2 diferent devices/firmware/implementations for these devices.
I would suggest to duplicate the `TuyaCO2ManufCluster` implementation and use the new one just for the afected device also duplicating the `TuyaCO2ManufCluster`.

(That way also will prevent the tests from failing)

adrianschroeter:
okay, but I close this for now, since I do not trust the values at all so much. CO2 stayed around 360 ppm the entire days. (The sensor raises to 500 when breathing against it, so it does react, but having 350-370 the entire day is not really believable).

Need to find out more details of the _TZE200_dwcarsat device first.

(without the patch CO2 is report as 6 ppm and formaldehyde jumps to 40000 though which seems to be complete garbage).

javicalle:
For future reference:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/3982

Issue #1773, opened 2022-09-20
[Device Support Request] Tint E14 RGB CCT


nicufarmache:
**Is your feature request related to a problem? Please describe.**
My Tint E14 light doesn't correctly report the color capabilities (I can only change the color temperature in HA)
[https://zigbee.blakadder.com/Muller_Licht_404019.html](https://zigbee.blakadder.com/Muller_Licht_404019.html)

**Describe the solution you'd like**
Add a quirk for it

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4635, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x010d",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0300",
        "0x1000",
        "0x100f"
      ],
      "out_clusters": [
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "MLI",
  "model": "tint-ExtendedColor",
  "class": "zhaquirks.mli.tintE14rgbcct.TintRGBCCTLight"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant Container",
    "version": "2022.9.5",
    "dev": false,
    "hassio": false,
    "virtualenv": false,
    "python_version": "3.10.5",
[...]

PR #2071, opened 2023-01-06
Add color support for Tint E14 RGB CCT lights

nicufarmache:
Related to #1773 

Add color capabilities to Tint E14 lights

the device is: https://zigbee.blakadder.com/Muller_Licht_404019.html

### DIFF
--- a/zhaquirks/mli/tintE14rgbcct.py
+++ b/zhaquirks/mli/tintE14rgbcct.py
@@ -1,0 +1,113 @@
+"""Tint E14 RGB CCT."""
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster, CustomDevice
+from zigpy.zcl.clusters.general import (
+    Basic,
+    GreenPowerProxy,
+    Groups,
+    Identify,
+    LevelControl,
+    OnOff,
+    Ota,
+    Scenes,
+)
+from zigpy.zcl.clusters.lighting import Color
+from zigpy.zcl.clusters.lightlink import LightLink
+
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+
+
+class TintRGBCCTColorCluster(CustomCluster, Color):
+    """Tint RGB+CCT Lighting custom cluster."""
+
+    # Set correct capabilities to ct, xy, hs
+    # Tint bulbs do not correctly report this attribute
+    _CONSTANT_ATTRIBUTES = {0x400A: 0b11111}
+
+
+class TintRGBCCTLight(CustomDevice):
+    """Tint E14 RGB+CCT Lighting device."""
+
+    signature = {
+        MODELS_INFO: [("MLI", "tint-ExtendedColor")],
+        ENDPOINTS: {
+            1: {
+                #   "profile_id": 260,
+                #   "device_type": "0x010d",
+                #   "in_clusters": [
+                #     "0x0000",
+                #     "0x0003",
+                #     "0x0004",
+                #     "0x0005",
+                #     "0x0006",
+                #     "0x0008",
+                #     "0x0300",
+                #     "0x1000",
+                #     "0x100f"
+                #   ],
+                #   "out_clusters": [
+                #     "0x0019"
+                #   ]
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.EXTENDED_COLOR_LIGHT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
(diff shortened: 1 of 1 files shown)


zhaquirks/mli/tintE14rgbcct.py
```diff
@@ -0,0 +1,115 @@
...
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+
+MANUFACTURER_SPECIFIC_CLUSTER_ID = 0x100F  # decimal = 4111
```
TheJulianJES, line 26: Thanks for the PR! Everything generally looks good, but:

The reason the tests are failing is that manufacturer specific clusters should only exists between ``0xFC00`` and ``0xFFFF`` (``64512`` to ``65535``).

``0x100F`` is well below that and doesn't match any existing Zigbee clusters (as far as I know).

To "fix" this, a new cluster needs to be created. Something like the following could work:
```python
class TintManufacturerSpecificCluster(CustomCluster):
    """Tint manufacturer specific cluster."""
    cluster_id = 0x1000  # not inside the manufacturer specific cluster range
```

For ``INPUT_CLUSTERS``, ``TintManufacturerSpecificCluster.cluster_id`` can still be used, but for the ``OUTPUT_CLUSTERS``, just ``TintManufacturerSpecificCluster`` needs to be used.
This *should* resolve the tests failing.

zhaquirks/mli/tintE14rgbcct.py
```diff
@@ -0,0 +1,115 @@
...
+                #     "0x0005",
+                #     "0x0006",
+                #     "0x0008",
+                #     "0x0300",
+                #     "0x1000",
+                #     "0x100f"
+                #   ],
+                #   "out_clusters": [
+                #     "0x0019"
+                #   ]
```
TheJulianJES, line 59: Also – this is not wrong – but I think these redundant comments can just be removed, as they don't add any new information.

zhaquirks/mli/tintE14rgbcct.py
```diff
@@ -0,0 +1,115 @@
...
+                ],
+                OUTPUT_CLUSTERS: [Ota.cluster_id],
+            },
+            242: {
+                #   "profile_id": 41440,
+                #   "device_type": "0x0061",
+                #   "in_clusters": [],
+                #   "out_clusters": [
+                #     "0x0021"
+                #   ]
```
TheJulianJES, line 81: Same here (redundant comment).

zhaquirks/mli/tintE14rgbcct.py
nicufarmache: Thanks for the tips. I'll make the changes and commit them after I test on my instance.

zhaquirks/mli/tintE14rgbcct.py
```diff
@@ -0,0 +1,97 @@
...
+
+    cluster_id = 0x100F  # not inside the manufacturer specific cluster range
+
+
+class TintRGBCCTColorCluster(CustomCluster, Color):
+    """Tint RGB+CCT Lighting custom cluster."""
+
+    # Set correct capabilities to ct, xy, hs
+    # Tint bulbs do not correctly report this attribute
+    _CONSTANT_ATTRIBUTES = {0x400A: 0b11111}
```
TheJulianJES, line 38: One more thing: did you check if the bulb actually supports Hue/Saturation, enhanced Hue, color loop, (XY, and color temp)?

You can test this by turning **off** the "Always prefer XY color mode" option in the ZHA config panel:

This will make ZHA use "enhanced Hue" commands if supported, otherwise it'll use normal HS commands, or fallback to XY.
Because of this of quirk, ZHA should use "enhanced Hue" commands. If that doesn't work, change this to ``0b11110``. If that also doesn't work, change it to ``0b11100``.
Please also test if the "color loop" effect works with this light.


Judging [from this](https://github.com/zigpy/zigpy/blob/297a2b0f364fe7a31de3b4966d70d1a67d432358/zigpy/zcl/clusters/lighting.py#L25-L30), it should be:
``0b11000`` if only XY and color temp are supported
``0b11100`` if only XY, color temp, and color loop are supported
``0b11110`` if only XY, color temp, color loop, and HS are supported (but not "enhanced Hue")
... and so on

Other than that, the PR looks ready to me!

nicufarmache: I'll check all these options and update the capabilities

nicufarmache: After turning off "Always prefer XY color mode":
with `0b11111` the light would turn red when I tried to set any color
with `0b11110` it worked properly

Also I tested color loop and it works. 

So I decided to change it to `0b11110`

TheJulianJES (approved):
LGTM. Thanks!

PR #2073, opened 2023-01-06
Add yet another version of Tuya air sensor.

rwalker777:
This one is square and black.

Device ID:
TS0601
_TZE200_c2fmom5z


TheJulianJES:
Nice! Did you test if the sensor works correctly with the quirk?

rwalker777:
Yes, 2.5pm is not showing up. Others are there.  Really need to figure out
a way to turn the reporting frequency or at least throttle the logging rate.

Issue #1699, opened 2022-08-21
Lonsonho_TS0043 par _TZ3000_uyjmm0et


FabzX11:
**Is your feature request related to a problem? Please describe.**
The __TZ3000_uyjmm0etof the TS0043 isn't supported by Home Assistant. Device can pair but no ZHA events fire and buttons do not respond to automation.

Switches are bought at Aliexpress (https://aliexpress.com/item/4001095272467.html)

**Describe the solution you'd like**
It would be great if someone could find a solution to my problem.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0006",
        "0xe000"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    },
    "3": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    },
    "4": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    }
  },
  "manufacturer": "_TZ3000_uyjmm0et",
  "model": "TS0043",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

[...]

Issue #1683, opened 2022-08-07
[Device Support Request] New TS0043 variant _TZ3000_kkahwiyu 3 button switch/remote 


julianjwong:
**Is your feature request related to a problem? Please describe.**
The _TZ3000_kkahwiyu variant of the TS0043 isn't supported by Home Assistant. Device can pair but no ZHA events fire and buttons do not respond to automation.

**Describe the solution you'd like**
Would be great if someone could update the quirk to handle this model. I am happy to test custom quirks.

<details>

<summary>Device signature</summary>

```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0006",
        "0xe000"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    },
    "3": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    },
    "4": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    }
  },
  "manufacturer": "_TZ3000_kkahwiyu",
  "model": "TS0043",
  "class": "zigpy.device.Device"
}

```

</details>

<details>

<summary>Diagnostic information</summary>

```
[...]

Issue #1557, opened 2022-05-11
[Device Support Request] TS0041 needs quirk from TS0044


OlivierHens:
Tried adding a switch very similar to the Tuya 1 button switch (TS0041) but failed. Quirk does not get loaded due to endpoints mismatch.  link to switch on Ali : https://nl.aliexpress.com/item/1005002858374901.html

However the endpoints are matching those present in TS0044.py
Seems that these buttons are created in mass with 4 buttons and depending on housing and soldering only 1, 2, 3 or 4 are active.

<details>
<summary>Device signature </summary>

```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0006",
        "0xe000"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    },
    "3": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    },
    "4": {
      "profile_id": 260,
      "device_type": "0x0000",
      "in_clusters": [
        "0x0001",
        "0x0006"
      ],
      "out_clusters": []
    }
  },
  "manufacturer": "_TZ3000_x7mej5oc",
  "model": "TS0041",
  "class": "zigpy.device.Device"
```

</details>

<details>
[...]

Issue #1990, opened 2022-12-06
[Device Support Request] Aliexpress Wall Switch and Remote


sid12345678910:
**Is your feature request related to a problem? Please describe.**
I bought a 3 gang switch that has 6 buttons, 4 relays and 2 buttons used for scene control. I also purchased a 4 button scene controller. while all 4 relays show up in home assistant the 2 buttons show up as switches and when pressed give a unkown cluster command when pressed. this also happens for the 4 button scene controller. The 4 buttons show up as switches and give the same error.

**Describe the solution you'd like**
For the 6 button switch apart from getting the buttons to work, it need it to be able to set a switch into a button if needed. it has 4 relays, I am only using 3.

For the 4 button i just need them to show up as buttons in Home Assistant.

<details>

<summary>Device signature</summary>

Device sig of the 6 button wall switch:
```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0004",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0xe000",
        "0xe001"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0004",
      "in_clusters": [
        "0x0004",
        "0x0005",
        "0x0006",
        "0xe001"
      ],
      "out_clusters": []
    },
    "3": {
      "profile_id": 260,
      "device_type": "0x0004",
      "in_clusters": [
        "0x0004",
[...]

PR #2074, opened 2023-01-07
Add yet another TS0043 variant

fingolfin:
This one reports four on/off switches , just like the `TuyaSmartRemote0042TOPlusA` does, and also in general the profile looks just like that (except with one more physical button). Thus I named this quirk `TuyaSmartRemote0043TOPlusA`.

However, I did deviate in the modelling of the replacement endpoints, and instead of copying them from the TS0042 variant, I decided to model them more closely on the other TS0043 variants. (I have no clue if that's a good or bad idea, advice is highly welcome, I am new to this).

I got it [from AliExpress](https://de.aliexpress.com/item/1005004775066234.html?spm=a2g0o.order_list.order_list_main.47.42b45c5fBbHuxw&gatewayAdapt=glo2deu). 

Tuya-ZigBee-Wireless-Wand-Schalter-1-2-3Gang-Wifi-Automatisierung-Push-Taste-Batterie-Powered-Zigbee-Gateway jpg_640x640

The profile looks like what was reported on issue #1699 so perhaps this PR fixes that, too, although there we have `"manufacturer": "_TZ3000_uyjmm0et",`  while for mine the id is `_TZ3000_w3c7ouru`.

Likewise, issue #1683 with `_TZ3000_kkahwiyu` probably is resolved by this.

Finally, I think issue #1557 is related, and a similar quirk ought to be added for TS0041, but I don't have such a device to test.

### DIFF
--- a/zhaquirks/tuya/ts0043.py
+++ b/zhaquirks/tuya/ts0043.py
@@ -192,6 +192,103 @@ class TuyaSmartRemote0043TO(CustomDevice):
     }
 
 
+class TuyaSmartRemote0043TOPlusA(CustomDevice):
+    """Tuya 3-button remote device with time on out cluster."""
+
+    signature = {
+        # SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=0, device_version=1, input_clusters=[0, 1, 6, 57344], output_clusters=[10, 25]))
+        # SizePrefixedSimpleDescriptor(endpoint=2, profile=260, device_type=0, device_version=1, input_clusters=[1, 6], output_clusters=[])
+        # SizePrefixedSimpleDescriptor(endpoint=3, profile=260, device_type=0, device_version=1, input_clusters=[1, 6], output_clusters=[])
+        # SizePrefixedSimpleDescriptor(endpoint=4, profile=260, device_type=0, device_version=1, input_clusters=[1, 6], output_clusters=[])
+        MODEL: "TS0043",
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
+                    OnOff.cluster_id,
+                    TuyaZBE000Cluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [Time.cluster_id, Ota.cluster_id],
+            },
+            2: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
+                INPUT_CLUSTERS: [
+                    PowerConfiguration.cluster_id,
+                    OnOff.cluster_id,
+                ],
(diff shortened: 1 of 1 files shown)


fingolfin:
Oh and also #1990 (or at least the second part of it, it seems to talk about two devices?))

zhaquirks/tuya/ts0043.py
```diff
@@ -192,6 +192,99 @@ class TuyaSmartRemote0043TO(CustomDevice):
...
+            },
+            4: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
+                INPUT_CLUSTERS: [
+                    PowerConfiguration.cluster_id,
+                    OnOff.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
```
TheJulianJES, line 238: So the device only has three switches but four endpoints? (Hence you removed the 4th endpoint in the replacement?)

The same signature is used by a four button ``TS0044`` device:
https://github.com/zigpy/zha-device-handlers/blob/961567488e9456c59691340814af817776b85dce/zhaquirks/tuya/ts0044.py#L232-L281
So I guess Tuya just lazily renamed the model whilst keeping the fourth endpoint.

I think this should be good to go though, as it's a different model name (so we shouldn't be worried that a four button switch is matched accidentally).

TheJulianJES (approved):
LGTM! Thanks for your contribution.

MattWestb:
Missing the `SimpleDescriptor` for all endpoints that making it easier for devs and user finding the right quirk then testing new devices.

TheJulianJES:
> Missing the `SimpleDescriptor` for all endpoints that making it easier for devs and user finding the right quirk then testing new devices.

I've never used that comment tbh. It's also not there for many other Tuya remotes or quirks in general.
But you're correct that it is present for all other remotes in that file.

There's an experimental quirk generator/matcher here: https://github.com/zigpy/zha-device-handlers/tree/dm/quirk-stub-generator/quirk_generator (that doesn't need comments)

@fingolfin Maybe add the comment with the SimpleDescriptor. It should be the following if I'm correct (add that before ``model`` – see other quirks in that same class for reference):
```python
        # SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=0, device_version=1, input_clusters=[0, 1, 6, 57344], output_clusters=[10, 25]))
        # SizePrefixedSimpleDescriptor(endpoint=2, profile=260, device_type=0, device_version=1, input_clusters=[1, 6], output_clusters=[])
        # SizePrefixedSimpleDescriptor(endpoint=3, profile=260, device_type=0, device_version=1, input_clusters=[1, 6], output_clusters=[])
        # SizePrefixedSimpleDescriptor(endpoint=4, profile=260, device_type=0, device_version=1, input_clusters=[1, 6], output_clusters=[])
```

MattWestb:
Its good then its in place and is right but some have only copy the last device class and making changes on the endpoints and not updating the simple discretion and that is very bad then trying !fast patching" testing new devices for users (still must recalculating Dec / Hex).

fingolfin:
I'll look into it later today

fingolfin:
Added the `SizePrefixedSimpleDescriptor` now.

TheJulianJES (approved):
Still looks good! Should be good to go

PR #2080, opened 2023-01-10
Add quirk for extended on/off cluster of TS011F _TYZB01_hlla45kx

simonaldrich:
Sold in the UK as the [Scolmore ClickSmart+ 30036](https://zigbee.blakadder.com/ClickSmart_CMA30036.html) Double Socket this device [supports](https://github.com/Koenkk/zigbee2mqtt/discussions/10831#discussioncomment-2137035) a child-lock / key-guard function which is enabled by using the Tuya extended on/off cluster.

### DIFF
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -1,7 +1,8 @@
 """TS011F plug."""
 
 from zigpy.profiles import zha
-from zigpy.quirks import CustomDevice
+from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
 from zigpy.zcl.clusters.general import (
     Basic,
     GreenPowerProxy,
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts011f_plug.py
```diff
@@ -1022,3 +1022,68 @@ class Plug_4AC_2USB_Metering(EnchantedDevice):
...
+            },
+            2: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.MAIN_POWER_OUTLET,
+                INPUT_CLUSTERS: [
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            }
```
TheJulianJES, line 1087: Looks like there's a comma missing at the end here (black formatter)

fingolfin, line 1087:
```suggestion
            },
```

simonaldrich:
Further testing of the device has shown that this device **only** supports the child-lock attribute from the extended on/off cluster. Will make further commits to reflect this.

simonaldrich:
Device now shows up correct in Home Assistant with only the Child lock configuration present
ha

MattWestb:
Then reducing the general function of one general device class (all TS011F with the same EP and cluster) i think its better using `MODELS_INFO:` instead of  `MODEL:` so only this device is getting the wanted reduced function.

By the way its looks nice in the GUI !!

MattWestb:
Nice !!
Its also possible for user hidden not used functions in the GUI but i like it clean like you have doing and its not interferes with other device :-)))

Its one tuya jungle out there !!!

dmulcahey:
needs a rebase

simonaldrich:
Rebased as [PR2085](https://github.com/zigpy/zha-device-handlers/pull/2085)

PR #673, opened 2020-12-27
Refactor Tuya Remote quirks

Adminiuga:
Refactor Tuya smart remote quirks. Maintains compatibility with the original quirk and device triggers.
The original quirk manipulated raw messages by overwriting the command ID based on command argument. While clever, I don't think this was the correct approach in zigpy terms and was causing some extra noise in the logs (data remains). 

This PR defines proper "cluster specific" command for press types and fire events depending on the press type.

Add signature for Benexmart Tuya remotes: [2 gang](https://www.aliexpress.com/item/4001230762751.html) and [3 gang](https://www.aliexpress.com/item/4001230762751.html)

@dmulcahey let me know if you prefer this PR to be split in two

PR #2082, opened 2023-01-10
Add missing long-press action to one TS0043 variant

fingolfin:
While I don't have access to this variant, the missing long press actions seem
like an accidental omission (the PR #673 introducing this variant makes no
hint about it, and it links to an AliExpress description of the switch which
mentions long presses explicitly). Adding them seems harmless enough: the
worst case scenario is that these devices really don't generate long press
events, but then no real harm is done. Perhaps some users will be disappointed
to be shown this event as an option but not being able to actually use it. In
that case, my suggestion would be to wait for an issue report on this; then
these events can be removed again, but this time with a comment explicitly
stating this omission is intentional. In contrast, in the best case some users
will be happy to finally be able to use long press actions on their switches.


MattWestb:
### Great catch !!!
Then tuya is masters of reusing (bad ?) code (and hardware) is more then likely all device is having long press and only adding more or lesser endpoints in the firmware and populating micro switches on the PCB (TS0044 PCB can being populated with all combinations of switches).

If you is interesting its possible making the device trigger working for all device classes by putting  it before the device class and linking it to it like being done in the TS0044 quirk:
https://github.com/zigpy/zha-device-handlers/blob/9b09ea5f2c99671fb6621ff42d552532526b85f4/zhaquirks/tuya/ts0044.py#L27-L47
Its no must but its making the code less duplicated and clean.

MattWestb (approved):

TheJulianJES (approved):
LGTM. Seems reasonable to add them

PR #2085, opened 2023-01-10
Add quirk for extended on/off cluster of TS011F _TYZB01_hlla45kx

simonaldrich:
Rebased [PR#2080](https://github.com/zigpy/zha-device-handlers/pull/2080)

Sold in the UK as the [Scolmore ClickSmart+ 30036](https://zigbee.blakadder.com/ClickSmart_CMA30036.html) Double Socket this device [supports](https://github.com/Koenkk/zigbee2mqtt/discussions/10831#discussioncomment-2137035) a child-lock / key-guard function which is enabled by using a Tuya extended on/off cluster.

### DIFF
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -1071,3 +1071,69 @@ class Plug_v2(EnchantedDevice):
             },
         },
     }
+
+
+class Plug_2AC_hlla45kx(CustomDevice):
+    """Tuya 2 outlet with child lock support."""
+
+    signature = {
+        MODEL: "TS011F",
+        ENDPOINTS: {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=9
+            # device_version=0
+            # input_clusters=[0, 10, 4, 5, 6]
+            # output_clusters=[25]>
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.MAIN_POWER_OUTLET,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Time.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id
+                ],
+                OUTPUT_CLUSTERS: [Ota.cluster_id],
+            },
+            # <SimpleDescriptor endpoint=2 profile=260 device_type=9
+            # device_version=0
+            # input_clusters=[4, 5, 6]
+            # output_clusters=[]>
+            2: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.MAIN_POWER_OUTLET,
+                INPUT_CLUSTERS: [
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            }
+        },
+    }
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.MAIN_POWER_OUTLET,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Time.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    TuyaZBOnOffAttributeCluster
+                ],
+                OUTPUT_CLUSTERS: [Ota.cluster_id],
+            },
+            2: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.MAIN_POWER_OUTLET,
+                INPUT_CLUSTERS: [
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    TuyaZBOnOffAttributeCluster
+                ],
+                OUTPUT_CLUSTERS: [],
+            }
+        },
+    }
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -1,7 +1,8 @@
 """TS011F plug."""
 
 from zigpy.profiles import zha
-from zigpy.quirks import CustomDevice
+from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
 from zigpy.zcl.clusters.general import (
     Basic,
     GreenPowerProxy,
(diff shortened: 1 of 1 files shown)


TheJulianJES:
Do the other custom Tuya attributes `power_on_state`/`backlight_mode` both return `None` (/unsupported attribute) when read using the HA cluster UI?

simonaldrich:
I retested this last night. The other extended attributes _can_ be written and read with the HA interface.

However, at least for the `backlight_mode` attribute it has no effect on the actual device. Please see [this video](https://user-images.githubusercontent.com/17856915/212289795-359805b6-10e2-4007-9b04-5d0f6b625344.mp4)  (apologies for the somewhat shaky camera work) which shows:
  * Switching both outlets on and off
  * Engaging `child-lock`: physical keys are disabled (lights flash three times when key pressed)
  * Removing `child-lock`: physical keys are enabled
  * Switch `backlight_mode` through different modes: no visible effect

I appreciate this makes life somewhat difficult. I would prefer that if the unit does not actually support the functionality of the custom attributes (which the manufacturer makes no claim to support) that they were not exposed at all. However, I realise that this makes a special case for this device which is always annoying from a code maintenance point-of-view.

I am open to the consensus of the community.

MattWestb:
Have you testing power on state ?
I agree its irritating having not supported functions in the GUI and on the other side its bad not have the function that the device is supporting and the user is being sad.
Adding custom tweaks in ZHA is OK for my but its not so easy mange for the tuya jungle with new devices every week.

simonaldrich:
I've not tested power-on state (cutting the power to all the outlets last night would not have made me popular :slightly_smiling_face:). I've got a spare outlet which I can wire up to a switched line and test this attribute tonight.

I agree with you about not blanket hiding attributes which might work on some devices. It's probably worth pointing out that (per your suggestion) this quirk is now limited to only [this device](https://github.com/zigpy/zha-device-handlers/pull/2085/commits/d87271650b91fd7f695f7a6b17c4503c96f459ad#) (using `MODEL_INFO`)

zhaquirks/tuya/ts011f_plug.py
```diff
@@ -36,6 +37,13 @@
 from zhaquirks.tuya.mcu import EnchantedDevice
 
 
+class TuyaZBOnOffChildLockAttributeCluster(CustomCluster, OnOff):
```
javicalle, line 40: Maybe `TuyaZBOnOffChildLockCluster`?

zhaquirks/tuya/ts011f_plug.py
```diff
@@ -1071,3 +1079,69 @@ class Plug_v2(EnchantedDevice):
             },
         },
     }
+
+
+class Plug_2AC_hlla45kx(CustomDevice):
```
javicalle, line 1084: I would avoid the `hlla45kx` reference.
How about `Plug_2AC_var03` (variation 3)?

javicalle (changes requested):
A couple of comments.

simonaldrich:
I was able to wire up the spare outlet to an independently switchable power supply this evening and have confirmed that the extended `power_on_state` attribute _is_ actually supported by the device.

As a result I've made the following changes:
  * Renamed the custom device class to `Plug_2AC_var03` as suggested by @javicalle
  * Renamed the custom cluster to `TuyaZBOnOffVariantCluster` to indicate that it is a variant of the `TuyaZBOnOffAttributeCluster` and moved it next to where it is first used.

I contemplated making the custom cluster a nested class of the custom device but I hope to shortly add a quirk for the [ClickSmart FCU](https://click-smart.com/products/smart-fcu) device which, I suspect, will have exactly the same cluster functionality.

zhaquirks/tuya/ts011f_plug.py
simonaldrich: Agreed

zhaquirks/tuya/ts011f_plug.py
simonaldrich: Have renamed to `TuyaZBOnOffVariantCluster` since it turns out that the `power_state` attribute actually is supported. See comment [below](https://github.com/zigpy/zha-device-handlers/pull/2085#issuecomment-1382569133) for further detail.

javicalle:
In Z2M, device also implements the `backlight` but with a low/medium/high intensity:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/8328b8f8b73b5f675407873de8fb64f1ff8c1731/devices/tuya.js#L926-L941

Could you check if it's the case?

simonaldrich:
I did already see the Z2M backlight brightness support. The underlying values it sends (`0`, `1` and `2` I think) are the same as the `TuyaZBOnOffAttributeCluster`. I can confirm it makes no difference to the led brightness.

javicalle (approved):
LGTM

TheJulianJES:
Also looks good to me.

Just noting that leaving out ``backlight_mode`` will probably cause a warning in HA, as that attribute isn't found (when initialized for the config entities). It's fine if it returns "unsupported", but it should still be there for all ``TS011F`` models. We can't really match both manufacturer and model on the ZHA side if I'm correct.

Maybe we should consider leaving backlight mode in? So just using the normal ``TuyaZBOnOffAttributeCluster`` cluster (with ``backlight_mode`` support, even though the hardware for this specific Tuya plug is broken and doesn't respect that setting.)
Then, we could also remove the specific manufacturer again and match all ``TS011F`` models with that "cluster signature".
I suspect there'll be more Tuya plugs that would match then (and even ones where the backlight mode is actually respected).

The only downside would be that changing the backlight mode config entity in ZHA will not do anything for this specific Tuya plug, but they literally implemented the attribute and made it writeable (hence Z2M also supports it). And it's Tuya, sooo...

javicalle:
> We can't really match both manufacturer and model on the ZHA side if I'm correct.

It can be done (but I hate depply that approach)
https://github.com/home-assistant/core/blob/03d9d30eca44feb11315a2ecb5402a8cf150a280/homeassistant/components/zha/core/channels/manufacturerspecific.py#L62-L95


> The only downside would be that changing the backlight mode config entity in ZHA will not do anything for this specific Tuya plug

...and someone will create an issue for this (but I can live with that). 
But I agree also with that approach. The more standard the implementations are, the easier to maintain. Making exceptions for some devices generates duplicate code.

In any case, I am not going to oppose the PR as it is

simonaldrich:
What I really want is for the device to actually implement the backlight mode attribute, it would be really useful! 

In lieu of that, I think it would cause fewer issues for the project if I just change the device to use the `TuyaZBOnOffAttributeCluster` since it does actually support reading/writing the attribute. I'll just live with disabling the backlight entity in HA.

I'll make another commit to this effect later this evening.

TheJulianJES (approved):
LGTM. I think this is the "best/cleanest approach" possible.

dmulcahey:
> Also looks good to me.
> 
> Just noting that leaving out `backlight_mode` will probably cause a warning in HA, as that attribute isn't found (when initialized for the config entities). It's fine if it returns "unsupported", but it should still be there for all `TS011F` models. We can't really match both manufacturer and model on the ZHA side if I'm correct.
> 
> Maybe we should consider leaving backlight mode in? So just using the normal `TuyaZBOnOffAttributeCluster` cluster (with `backlight_mode` support, even though the hardware for this specific Tuya plug is broken and doesn't respect that setting.) Then, we could also remove the specific manufacturer again and match all `TS011F` models with that "cluster signature". I suspect there'll be more Tuya plugs that would match then (and even ones where the backlight mode is actually respected).
> 
> The only downside would be that changing the backlight mode config entity in ZHA will not do anything for this specific Tuya plug, but they literally implemented the attribute and made it writeable (hence Z2M also supports it). And it's Tuya, sooo...

we can match both mfg and model

Issue #2090, opened 2023-01-13
[Device Support Request] TS0011 and TS0013


amiraizat124:
**Is your feature request related to a problem? Please describe.**
I cannot assign the proper quirk while paring the TS0011 and TS0013

I already tried using local quirk and without a local quirk.
I attached info for TS0011.

<details>

<summary>Device signature</summary>

```yaml
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZ3000_5ftkaulg",
  "model": "TS0011",
  "class": "zigpy.device.Device"
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
zha:
  enable_quirks: true
  custom_quirks_path: /config/custom_zha_quirks/
```

</details>

<details>
<summary>Additional logs</summary>

```
[0x2aba] Discovered basic device information for <Device model='TS0011' manuf='_TZ3000_5ftkaulg' nwk=0x2ABA ieee=a4:c1:38:af:e8:e0:74:67 is_initialized=True>
Device is initialized <Device model='TS0011' manuf='_TZ3000_5ftkaulg' nwk=0x2ABA ieee=a4:c1:38:af:e8:e0:74:67 is_initialized=True>
Checking quirks for _TZ3000_5ftkaulg TS0011 (a4:c1:38:af:e8:e0:74:67)
[...]

PR #2091, opened 2023-01-13
Adding new / old TS0011 and TS0013 version.

MattWestb:
Putting in comments for 2 family and adding TS0011 and TS0013 to the current TS0012 version.

Looks like have making changing of other classes but i have moving then together in family:

```
# NoNeutralSwitch family = without tuya cluster and Identify.
# Switch_2 family = Normal NoNeutralSwitch (without tuya cluster) and with Identify.

```
Other with tuya clusters is not touched.

### DIFF
--- a/zhaquirks/tuya/ts001x.py
+++ b/zhaquirks/tuya/ts001x.py
@@ -21,6 +21,7 @@ from zhaquirks.tuya import (
 from zhaquirks.tuya.mcu import EnchantedDevice
 
 
+# NoNeutralSwitch family = without tuya cluster and Identify.
 class TuyaSingleNoNeutralSwitch(EnchantedDevice, TuyaSwitch):
     """Tuya 1 gang no neutral light switch."""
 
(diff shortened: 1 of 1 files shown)


MattWestb:
User feedback from the current version in the PR https://github.com/zigpy/zha-device-handlers/issues/2090#issuecomment-1381752161.

zhaquirks/tuya/ts001x.py
```diff
@@ -212,8 +356,8 @@ class TuyaDoubleNoNeutralSwitch_2(EnchantedDevice, TuyaSwitch):
     }
 
 
-class TuyaTripleNoNeutralSwitch(EnchantedDevice, TuyaSwitch):
-    """Tuya 3 gang no neutral light switch."""
+class Tuya_TripleNoNeutralSwitch_2(EnchantedDevice, TuyaSwitch):
```
TheJulianJES, line 359: Maybe remove the underscore between ``Tuya`` and ``TripleNoNeutralSwitch`` here, as the others also don't have one. (The one for ``_2`` is fine)

Also, should ``TuyaTripleNoNeutralSwitch_2`` be moved under ``TuyaTripleNoNeutralSwitch``, as they're really similar? (So "_2" quirks directly under the "main quirk" of each device?)

MattWestb: https://github.com/MattWestb/zha-device-handlers/blob/1fb77ea83c7dfe32cd6beeb3bbcb814eaecaf7b2/zhaquirks/tuya/ts001x.py#L142
is over 
https://github.com/MattWestb/zha-device-handlers/blob/1fb77ea83c7dfe32cd6beeb3bbcb814eaecaf7b2/zhaquirks/tuya/ts001x.py#L359

The rest is one large mess with failing parts of signature and so on so i not touching it for the moment.

Normally new families / version is coming on all 3 types and its good having  them together as was done from the beginning.

TheJulianJES (approved):
LGTM

TheJulianJES:
See https://github.com/zigpy/zha-device-handlers/issues/2090#issuecomment-1383332911

Maybe a small change is still needed?

MattWestb:
Not small its need one new device class then its one new device family (without time cluster).

TheJulianJES:
I see. So a separate PR for that?

MattWestb:
I hate rebasing its only make problem so i preferring putting in this then have getting it working OK in the issue but i can doing on new if you like.

TheJulianJES:
Wasn't sure what you want to do. Just do whatever is best for you 👍

MattWestb:
Thanks D

PR #2092, opened 2023-01-13
New Sinopé switch SP2600ZB SP2610ZB support

claudegel:
This is to add support for new generation of Sinopé switch SP2600ZB and SP2610ZB. They added LightLink cluster which is missing in the first generation.

### DIFF
--- a/zhaquirks/sinope/switch.py
+++ b/zhaquirks/sinope/switch.py
@@ -39,43 +39,20 @@ from zhaquirks.sinope import SINOPE
 SINOPE_MANUFACTURER_CLUSTER_ID = 0xFF01
 
 
-class SinopeMultiControllerManufacturerCluster(CustomCluster):
-    """SinopeMultiControllerManufacturerCluster manufacturer cluster."""
+class SinopeManufacturerCluster(CustomCluster):
+    """SinopeManufacturerCluster manufacturer cluster."""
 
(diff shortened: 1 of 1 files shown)


javicalle (approved):
LGTM

javicalle:
I would suggest to put all the attributes in just one `SinopeManufacturerCluster` and use it everywhere. If attributes are exposed in HA in some way, would require some tuning.
But that is beyond of the scope of the PR.

claudegel:
If I put everything in one SinopeManufacturerCluster, the old device that don't have the 0x1000 cluster in there signature won't be detected properly.

javicalle:
> If I put everything in one SinopeManufacturerCluster, the old device that don't have the 0x1000 cluster in there signature won't be detected properly.

Not the `CustomDevice`s implementations, only the 3 `CustomCluster`:
* `SinopeMultiControllerManufacturerCluster`
* `SinopeLoadControllerManufacturerCluster`
* `SinopeCalypsoControllerManufacturerCluster`

And would use it in all the Sinope `CustomDevice`s

claudegel:
Ok I gonna merge the three CustomCluster into one and add it to this PR

javicalle:
Thanks for your contribution.

javicalle (approved):
LGTM

Issue #2093, opened 2023-01-14
[Device Support Request] - TuYa TS0601 Soil Sensor


dgaust:
Please add support for the Tuya TS0601 Soil Sensor.

**Describe the solution you'd like**
Inclusion of a quirk to support data provided by the sensor.

<details>
<summary>Device signature</summary>

```{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_myd45weu",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2023.1.4",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.10.7",
    "docker": true,
    "arch": "x86_64",
    "timezone": "Australia/Sydney",
    "os_name": "Linux",
    "os_version": "5.15.80",
    "supervisor": "2022.12.1",
    "host_os": "Home Assistant OS 9.4",
    "docker_version": "20.10.19",
    "chassis": "embedded",
    "run_as_root": true
  },
  "custom_components": {
    "alexa_media": {
      "version": "4.5.0",
      "requirements": [
        "alexapy==1.26.4",
        "packaging>=20.3",
        "wrapt>=1.12.1"
      ]
    },
[...]

PR #2102, opened 2023-01-16
Add TS0601 Soil Moisture Sensor

dgaust:
Fixes: #2093

ZHA Quirk to support the Tuya based Soil Moisture sensor. Based off the quirks ts0601_sensor.py and the zigbee-herdsman converter

https://github.com/zigpy/zha-device-handlers/issues/2093

Temperature value is multiplied by 100 to get degrees centigrade Humidty is based on the standard TuyaRelativeHumidity value Battery is multiplied by two to get battery % value


javicalle:
Thanks for your contribution and **good job** despite 'not 100% sure what I'm doing' 😉 

A few comments:

Edit your first post and put a `Fixes: #2093`. This way, issue will be closed once merged the PR.

I'm not sure if we need to create a new file for this devices. Until now is just another temp/hum sensor.
I would suggest to put the new code in the existing `ts0601_sensor.py`:
* `class SoilManufCluster(TuyaMCUCluster):`  <-- renamed from yours `TemperatureHumidityManufCluster`
* `class TuyaSoilSensor(CustomDevice):`

HA have specific definitions for 'moisture' entities. I will try to check if we can follow the implementations.

It seems that device is reporting also the temp UOM (probably also allows to set it). That would be a improvement, but can go in another PR.

javicalle:
It seems that most of the ZHA magic is already implemented. I would suggest to implement the device like:
* create a new `TuyaSoilCluster`
```python
class TuyaSoilMoisture(SoilMoisture, TuyaLocalCluster):
    """Tuya local SoilMoisture cluster."""

```

You will need to import the `SoilMoisture` class:
```python
from zigpy.zcl.clusters.measurement import (
    RelativeHumidity,
    SoilMoisture,
    TemperatureMeasurement,
)
```

* use the new class in your device (`dp_to_attribute` and `replacement`)

I believe that HA will create the new entity. And probably will be pretty similar to the old one but IMHO 'this is the way'.

dgaust:
> I was thinking about the same, but keep merged for now. Let's see how it grows first.

Ok, I've made those changes to the file, bringing the standalone file back into the sensor.py version. Have also added the soil moisture class rather than using humidity. It did create a new sensor in Home Assistant and all seems to be working fine

image

javicalle:
You have the imports duplicated. Keep the 'new' format.

The temp OUM will not be an easy fix (ZHA assumes that temp units will be Celsius always). I believe that we can go 'as is'.

dgaust:
> You have the imports duplicated. Keep the 'new' format.
> 
> The temp OUM will not be an easy fix (ZHA assumes that temp units will be Celsius always). I believe that we can go 'as is'.

Yep, I'd noticed that once it failed and have made the adjustments.

javicalle:
You can check the isort report for more info:
https://github.com/zigpy/zha-device-handlers/actions/runs/3934353893/jobs/6729028066

The check complains about the extra break lines (the `-` lines). Delete it.

dgaust:
> You can check the isort report for more info: https://github.com/zigpy/zha-device-handlers/actions/runs/3934353893/jobs/6729028066
> 
> The check complains about the extra break lines (the `-` lines). Delete it.

Thansk, I didn't know how to interpret what the issue was :D

All fixed and tests passed.

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -24,6 +28,20 @@ class TuyaTemperatureMeasurement(TemperatureMeasurement, TuyaLocalCluster):
...
+        """Apply a correction factor to value."""
+        if attr_name == "measured_value":
+            value = value * (
+                self.endpoint.device.RH_MULTIPLIER
+                if hasattr(self.endpoint.device, "RH_MULTIPLIER")
+                else 100
+            )
+        return super().update_attribute(attr_name, value)
+
+
```
javicalle, line 44: Remove the `update_attribute` method here. The `RH_MULTIPLIER` sates for RelativeHumidity. Keep just the class definition like:
```python
class TuyaSoilMoisture(SoilMoisture, TuyaLocalCluster):
    """Tuya local SoilMoisture cluster."""


```
(2 break lines before next class `TuyaRelativeHumidity`)

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -207,3 +224,77 @@ class TuyaTempHumiditySensorVar03(CustomDevice):
...
+        5: DPToAttributeMapping(
+            TuyaTemperatureMeasurement.ep_attribute,
+            "measured_value",
+            dp_type=TuyaDPType.VALUE,
+            converter=lambda x: x * 100,
+        ),
+        3: DPToAttributeMapping(
+            TuyaSoilMoisture.ep_attribute,
+            "measured_value",
+            dp_type=TuyaDPType.VALUE,
```
javicalle, line 242: If removed from method, we need to add here the value conversion like:
```python
            dp_type=TuyaDPType.VALUE,
            converter=lambda x: x * 100,
```

javicalle (changes requested):
A little fix to remove the `update_attribute` method.
If needed in a future we can go back to it.

javicalle:
With the last changes the PR will be fine to me.
But can you check yhe device logs for any warning about unsupported handlers or similar. There must be a DP=9 for the UOM

javicalle (approved):
LGTM

Issue #2047, opened 2022-12-30
Thermostate Danfoss Ally Quirk - Integrate new model eTRV0103


michib-xx:
https://github.com/zigpy/zha-device-handlers/blob/23a166902e50d5c32019e0d4d32a968715041bb6/zhaquirks/danfoss/thermostat.py#L114

Hello,

I just installed my new Danfoss Ally, which is has this signature:
```
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4678, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x000a",
        "0x0020",
        "0x0201",
        "0x0204",
        "0x0b05"
      ],
      "out_clusters": [
        "0x0000",
        "0x0019"
      ]
    }
  },
  "manufacturer": "Danfoss",
  "model": "eTRV0103",
  "class": "danfoss.thermostat.DanfossThermostat"
}
```

Obviously, this is a different model than the default quirk is made for. Hence, the quirk is not applied.
I assume/hope that attributes and all are the same. 

Could you integrate the new model in the default quirk?

Thanks a lot.

Micha

javicalle:
But your device signature is saying that the quirk is loading.
Maybe do you have enabled your local quirk?

michib-xx:
Yes, you are right. I locally added the model to the list and it loaded. Actually, I just recognized this after I wrote this issue.
Nevertheless, the model should be integrated into the default file.

michib-xx:
[...]

PR #2108, opened 2023-01-16
Adding hardware updated Ally

MattWestb:
Fixes: #2047


TheJulianJES:
Seems to fix https://github.com/zigpy/zha-device-handlers/issues/2047?
Were you able to test this?

MattWestb:
Nop i was seen papers from Danfoss with the hardware info and saying only the device name was changed.
I looking if its also cluster changes.

PR #2123, opened 2023-01-21
Add "long release event" for IKEA 5-button remotes

MatMaul:
Adds "long release triggers" for up and down buttons.

~~Also do a refactor to avoid copy pasting triggers definition, since they are mostly the same.~~
-> Refactor done in https://github.com/zigpy/zha-device-handlers/pull/2155

This is tested with the very first version of the remote, the one defined in `fivebtnremote.py`.

### DIFF
--- a/zhaquirks/ikea/fivebtnremote.py
+++ b/zhaquirks/ikea/fivebtnremote.py
@@ -27,6 +27,8 @@ from zhaquirks.const import (
     COMMAND_RELEASE,
     COMMAND_STEP,
     COMMAND_STEP_ON_OFF,
+    COMMAND_STOP,
+    COMMAND_STOP_ON_OFF,
     COMMAND_TOGGLE,
     DEVICE_TYPE,
     DIM_DOWN,
@@ -36,6 +38,7 @@ from zhaquirks.const import (
     INPUT_CLUSTERS,
     LEFT,
     LONG_PRESS,
+    LONG_RELEASE,
     MODELS_INFO,
     OUTPUT_CLUSTERS,
     PARAMS,
@@ -137,6 +140,11 @@ class IkeaTradfriRemote1(CustomDevice):
             ENDPOINT_ID: 1,
             PARAMS: {"move_mode": 0},
         },
+        (LONG_RELEASE, DIM_UP): {
+            COMMAND: COMMAND_STOP_ON_OFF,
+            CLUSTER_ID: 8,
+            ENDPOINT_ID: 1,
+        },
         (SHORT_PRESS, DIM_DOWN): {
             COMMAND: COMMAND_STEP,
             CLUSTER_ID: 8,
@@ -148,6 +156,11 @@ class IkeaTradfriRemote1(CustomDevice):
             CLUSTER_ID: 8,
             ENDPOINT_ID: 1,
             PARAMS: {"move_mode": 1},
+        },
+        (LONG_RELEASE, DIM_DOWN): {
+            COMMAND: COMMAND_STOP,
+            CLUSTER_ID: 8,
+            ENDPOINT_ID: 1,
         },
         (SHORT_PRESS, LEFT): {
             COMMAND: COMMAND_PRESS,


MatMaul:
This is tested with the very first version of the remote, the one defined in `fivebtnremote.py`.

MattWestb:
I like using  the same DA and only making the changed DA in the device class.
The problem is testing if all is working OK.
The `fivebtnremote.py` is being used for the old ZLL firmware and most users have updating there firmware for getting more functions and also fix some bugs that making the controller losing the network setting then battery is out and also battery draining problems. 
The last firmware is loosing group bindings and if fixing multiple commands being received on some coordinators (some have filtering for broadcast command with same frame number EZSP not).
Reversing firmware for testing you need braking the remote (its very well glued inside !!) for getting access to the SWD pads and flashing it with one SWD probe.
I only having the last firmware on my 5BT so i cant testing the old version easy put possible.

First look it that some DA is being added for versions that is not supporting it but i have not testing it on my devices.

@dmulcahey Is it possible "updating" / overriding one linked DA like long release OnOff from the original and putting one updated one in the V2 class ?

Im not one code warrior so im luck of knowledge in the (Pyhton) Spaghetti.

MatMaul:
I pulled the battery out after waiting in case of, and indeed my remote is now recognised as `zhaquirks.ikea.fivebtnremotezha.IkeaTradfriRemote3` and it works too :tada: 

So at least `zhaquirks.ikea.fivebtnremote.IkeaTradfriRemote` and `zhaquirks.ikea.fivebtnremotezha.IkeaTradfriRemote3` work with my PR.

MattWestb:
Here and i have seen it for some months and it was also wrong in our tuya TRV matrix.
https://github.com/jacekk015/zha_quirks/pull/2#issuecomment-1405611096

Look in Z2M is also saying its one Thermostat and not one TRV.

I think the error is over 2 year old but is fixed by the thermostat quirk was added and was loading and not the TRV version.

MattWestb:
My 5BT have trying updating atlas 2 time in the night and its rolling back to 2.3.080 :-(((
I have deleting, resetting and taking out the battery and repairing and trying updating it one more time (if the battery is still living so long).

So its looks the problem with OTA is rolling back is still there and if the last try is not working i doing one firmware downgrade to last working groups with SWD and looking if its working.

It shall being good to knowing if the Variant 3 compatible or not for having the DA working OK for our users.

zhaquirks/ikea/fivebtnremotezha.py
```diff
@@ -111,77 +100,13 @@ class IkeaTradfriRemote1(CustomDevice):
...
-                "param1": 3328,
-                "param2": 0,
+        **IkeaTradfriRemote.device_automation_triggers,
+        **{
+            (LONG_PRESS, TURN_ON): {
+                COMMAND: COMMAND_RELEASE,
+                CLUSTER_ID: 5,
+                ENDPOINT_ID: 1,
+                PARAMS: {"param1": 537},
             },
```
TheJulianJES, line 110: Hmm, so only that "firmware" uses a different parameter for LONG_PRESS/ON?
The code itself looks good to me (and copies previous behavior). Is there any thread on this or can someone confirm that one "version" actually uses this parameter? 

instead of:
 
```python
        (LONG_PRESS, TURN_ON): {
            COMMAND: COMMAND_RELEASE,
            CLUSTER_ID: 5,
            ENDPOINT_ID: 1,
            PARAMS: {"param1": 0},
        },
```

MattWestb: I was getting problem with my remote and must doing one flash erase and writing one new bootloader main flash and user data and its have booted and is trying updating the firmware but its taking some hours.

The different parms is real for HA firmware is was 2 or more version and perhaps one with one mixed. The 2.3.080 was doing all the same and it was working OK.
Also the stop_with_on_off / stop is only working on 2.3.080 all earlier is using the same commands for doe both stop.

Then OTA updating firmware and the device is changing clusters on the endpoints is deleting and readding needed or the quirk is not working OK or not loaded.

Testing ZLL firmware i need finding one very old one (version 1.X) and SWD flashing it after extracting it with zigpy but i dont knowing if its making some sense then its old buggy one and its only updating the firmware and its working for the user.

I was thinking deleting all old version and only having the 2.3.080 with group binding and the latest with device binding but i think the maintainers dont like it.

Only is that the 2.3.080 is the last with working group binding for user that is using Zigbee Light Groups.

TheJulianJES: So in your opinion, the changes should be fine, right?

> I was thinking deleting all old version and only having the 2.3.080 with group binding and the latest with device binding but i think the maintainers dont like it.

Yeah, we shouldn't really do this, as (1) many users don't update their devices and (2) ZHA might still have an outdated signature cached.

> Only is that the 2.3.080 is the last with working group binding for user that is using Zigbee Light Groups.

Hmm, I remember you posted a table with the corresponding firmware versions for group binding (for every device). Can't find it now, can you link me that again?

https://www.zigbee2mqtt.io/devices/E1524_E1810.html mentions 2.3.075 and later only supports device binding.

MatMaul: > Hmm, so only that "firmware" uses a different parameter for LONG_PRESS/ON?

I think this is probably a mistake: The `release` event is also used for the arrows release, and in this case `param1` holds the time it stayed pressed. Someone probably saw this event and put it as `LONG_PRESS/ON` ?

 Also I think we should remove `LONG_PRESS/ON` since it is not really working: you first get the `toggle` event anyway, and there is no way to tell if it's still holding or not. Last version (2.4.45) also seems to have remove it altogether.
More info regarding that [here](https://www.zigbee2mqtt.io/devices/E1524_E1810.html#toggle-button).

Instead, I think we should provide a generic event for release, whatever is inside `param1`. Unfortunately there is no param to discriminate between a release of left or right arrow. I am not sure what is the best way to do so however.

MatMaul: After having a quick look at the rest of the repo perhaps `(LONG_RELEASE, LONG_RELEASE)` or `(LONG_RELEASE, BUTTON)` would make sense as a generic `release` event for both arrows ?

I tried both and `(LONG_RELEASE, BUTTON)` looks less ugly in the UI for me.

MattWestb:
Firmware v24.4.5 is now working OK in PR https://github.com/zigpy/zha-device-handlers/pull/2140 but have not your long release from hold up and down.
Also Long release for left / right hold is also possible but little tricky. Read the comments in the linked PR.

I shall testing SWD flashing older firmware for testing but its taking some time then its looks like its problems with the bootloader version miss match.

@TheJulianJES and @dmulcahey shall we putting all DA in the old ZLL quirk and importing it and making short cuts but can braking automatons for users ?
One way is doing it only for the HA version and leaving the old ZLL untouched for not braking things.
I like the change but is not easy testing the all IKEA controllers is blocked for firmware downgrading in the firmware and need SWD flashing of the right firmware (and need finding it 2) and then extracting the EBL file from it and flashing with SWD and hope its working (now i can doing OK without corrupting the flash after getting the reset pin connected).

PS: I have not testing this updated quirk then having doing the tests for the  v24.4.5 version.

TheJulianJES:
> shall we putting all DA in the old ZLL quirk and importing it and making short cuts but can braking automatons for users

Just importing device automation triggers from another file won't break any automations, as long as they stay the same.

I'm still unsure if we really want the ``fivebtnremotezha.py`` quirks to depend on ``fivebtnremote.py``.
Since it's the same device, we could even consider just putting them all in the same file.

I do think this cleanup for the device automation triggers is needed.

zhaquirks/ikea/fivebtnremotezha.py
TheJulianJES: > I think this is probably a mistake: The release event is also used for the arrows release, and in this case param1 holds the time it stayed pressed

I'd think we should just remove that "special case", as it seems to invalid.

> Also I think we should remove LONG_PRESS/ON since it is not really working

We would break existing automations if we completely remove a device automation trigger though. Let's just leave it in for now and have this PR be (1) refactoring + (2) adding the new UP/DOWN long release triggers + (3) removing the invalid LONG_PRESS/ON with "param 537".
For more changes, a new PR should be opened.

MattWestb:
I think merger them to one quirk is the best but looking very carefully for not braking the DAs is the best way.
Then only adding new DAs for versions that we have testing so not braking old firmware devices.
Also if user like have more function upgrade to the latest OTA with working group or device binding.
On the other hand if adding DAs that  the device is not supporting is making bad user experience but "upgrade your device firmware" is the standard fix.

TheJulianJES:
Yeah, let's just merge both five button files into one. It's the same device and the device automation triggers also all seem to be the same(?) regardless of firmware

stickpin:
@TheJulianJES should I merge two quirks into one and also to PRs into one?

TheJulianJES:
Rebased the PR on top of the refactoring changes from https://github.com/zigpy/zha-device-handlers/pull/2155

MattWestb:
It was on bug in GSDK that corrupted the flash:
`1081342 Corrected an issue where initial SWO data sent immediately after initializing SWO would become corrupted. `
From 4.2.1

MatMaul:
Thanks all for all the work on this remote and the rebase !

I guess we can go ahead and merge this one too ?

MatMaul:
@TheJulianJES little bump, would be nice to get that in for the next release :) 

Thanks !

TheJulianJES:
I didn't have the time to test this yet, so just for completeness: Can you attach a screenshot of how the new device triggers look in the UI?

TheJulianJES (approved):
The new actions seem to work correctly (tested using the latest firmware).
Thanks for your contribution!

TheJulianJES:
The new actions show as following in the UI:
- "Dim up" released after long press
- "Dim down" released after long press

Regarding the long release on left/right buttons, the issue was that there's no easy way (other than to cache previous events) to differentiate the buttons, correct?

MattWestb:
The Open/Close have implanting it and its possible doing for more devices like OnOff Dimmer switch and so on.
https://github.com/zigpy/zha-device-handlers/blob/31913069b2b8442a5507228db4f9f6b3ed2e47a2/zhaquirks/ikea/opencloseremote.py#L46

Issue #2121, opened 2023-01-20
Develco SPLZB-131: Wrong device temperature (divisor)


eriknn:
Hi,
it seems that device temperature is reported "as is" from the device, but then a divisor of 100 is applied, giving a wrong result.
If I'm not mistaken the 100 divisor is default for device temperature, so this is probably a bug in the device.
Is this something that could be solved with a quirk?

TheJulianJES:
You did not fill out the issue template.

Please provide the device signature and diagnostic information for the affected device.

eriknn:
**Is your feature request related to a problem? Please describe.**
Yes, seems to be a bug in the device

**Describe the solution you'd like**
Use divisor = 1 instead of divisor = 100 which I believe is default

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4117, maximum_buffer_size=72, maximum_incoming_transfer_size=128, server_mask=0, maximum_outgoing_transfer_size=128, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 49353,
      "device_type": "0x0001",
      "in_clusters": [
        "0x0005",
        "0x0006"
      ],
      "out_clusters": []
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0002",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0009",
        "0x0702",
        "0x0b04"
      ],
      "out_clusters": [
        "0x0000",
        "0x0003",
        "0x000a",
        "0x0019",
        "0x0406"
      ]
    }
  },
  "manufacturer": "Develco Products A/S",
  "model": "SPLZB-131",
[...]

PR #2132, opened 2023-01-25
Fix Develco SPLZB-131 plug device temperature divisor

eriknn:
Adds quirk for SPLZB-131!

Fixes https://github.com/zigpy/zha-device-handlers/issues/2121

### DIFF
--- a/zhaquirks/develco/power_plug.py
+++ b/zhaquirks/develco/power_plug.py
@@ -1,4 +1,4 @@
-"""Smart Plugs."""
+"""Develco smart plugs."""
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
 from zigpy.zcl.clusters.general import (
@@ -30,7 +30,7 @@ DEV_TEMP_ID = DeviceTemperature.attributes_by_name["current_temperature"].id
 
 
 class DevelcoDeviceTemperature(CustomCluster, DeviceTemperature):
-    """Device Temperature. Modify divisor."""
+    """Custom device temperature cluster to multiply the temperature by 100."""
 
     def _update_attribute(self, attrid, value):
         if attrid == DEV_TEMP_ID and value is not None:
--- a/zhaquirks/develco/power_plug.py
+++ b/zhaquirks/develco/power_plug.py
@@ -1,0 +1,110 @@
+"""Smart Plugs."""
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.general import (
+    Alarms,
+    Basic,
+    DeviceTemperature,
+    Groups,
+    Identify,
+    OnOff,
+    Ota,
+    Scenes,
+    Time,
+)
+from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
+from zigpy.zcl.clusters.measurement import OccupancySensing
+from zigpy.zcl.clusters.smartenergy import Metering
+
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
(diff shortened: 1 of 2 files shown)


zhaquirks/develco/power_plug.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class DeviceTemperatureCluster(CustomCluster, DeviceTemperature):
+    """Device Temperature. Modify divisor."""
+
+    cluster_id = DeviceTemperature.cluster_id
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        super().__init__(*args, **kwargs)
```
TheJulianJES, line 37: This can be removed

zhaquirks/develco/power_plug.py
```diff
@@ -0,0 +1,110 @@
...
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+from zhaquirks.develco import DEVELCO
+
+
+class DeviceTemperatureCluster(CustomCluster, DeviceTemperature):
+    """Device Temperature. Modify divisor."""
+
+    cluster_id = DeviceTemperature.cluster_id
```
TheJulianJES, line 33: This can be removed

TheJulianJES:
Thanks for the PR! I've added two small comments. Please try and see if the quirk still works after making those changes.

eriknn:
Thanks for the feedback! This is my first commit to a community project, so it's a steep "github learning curve" for me here.
Not sure if I understood your comments correctly. Did you just mean the _init function?

TheJulianJES:
All good! Thanks for doing the PR!

I think the explicit cluster_id definition can also be removed, as it should be inherited from the `DeviceTemperatureCluster` already.

zhaquirks/develco/power_plug.py
```diff
@@ -0,0 +1,141 @@
...
+        if voltage > 0 and current > 0:
+            voltage = voltage / 100
+            current = current / 1000
+            pf = 100 * power / (voltage * current)
+
+            if pf > 100:
+                pf = 100
+            elif pf < 0:
+                pf = 0
+        self._update_attribute(self.POWERFACTOR_ID, pf)
```
TheJulianJES, line 73: Can we leave this manual power factor calculation out for now?
I know ZHA creates a "power factor" sensor by default (even though the device does not support it). This happens for a lot of devices and needs to be fixed in ZHA.
We just don't do this manual calculation for any other devices and I wouldn't do start with it for now.

eriknn: Sure, new pull request created!

zhaquirks/develco/power_plug.py
```diff
@@ -0,0 +1,106 @@
...
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: 0xC0C9,
+                DEVICE_TYPE: 1,
+                INPUT_CLUSTERS: [
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
```
TheJulianJES, line 55: I'm guessing the removed endpoint 1 in the replacement did not have a function at all?

zhaquirks/develco/power_plug.py
```diff
@@ -0,0 +1,106 @@
...
+from zhaquirks.develco import DEVELCO
+
+DEV_TEMP_ID = DeviceTemperature.attributes_by_name["current_temperature"].id
+
+
+class DevelcoDeviceTemperature(CustomCluster, DeviceTemperature):
+    """Device Temperature. Modify divisor."""
+
+    def _update_attribute(self, attrid, value):
+        if attrid == DEV_TEMP_ID and value is not None:
```
TheJulianJES, line 36: Did you ever get `None` as a value for `_update_attribute`?

eriknn: No, that check might be unnecessary.....

TheJulianJES: Removed it for now.

TheJulianJES:
Everything should be good now. I've added a small test for the multiplication too.

Thanks for the PR!

PR #2138, opened 2023-01-26
Deleting _TZE200_2ekuz3dz

MattWestb:
This ID is wrong its one Thermostat and have one working quirk. Then the thermostat is loaded before the TRV is also matching the thermostat and is not the TRV so its not braking any device onlyy fixing one old error.

https://github.com/zigpy/zha-device-handlers/blob/6b8cd2cd03804842765b37519622849d7c5ddb80/zhaquirks/tuya/ts0601_electric_heating.py#L144


TheJulianJES:
Interesting. Was this issue reported anywhere?

MattWestb:
Here and i have seen it for some months and it was also wrong in our tuya TRV matrix.
https://github.com/jacekk015/zha_quirks/pull/2#issuecomment-1405611096

Look in Z2M is also saying its one Thermostat and not one TRV.

I think the error is over 2 year old but is fixed by the thermostat quirk was added and was loading and not the TRV version.

TheJulianJES (approved):
Thanks!

Issue #2115, opened 2023-01-19
[BUG] Danfoss Ally - Active quirk leads to malfunction


michib-xx:
**Describe the bug**
I have a new Danfoss Ally TRV eTRV0103, where I used the updated quirk (#2108, as custom) to tune the control performance via the Control algorithm scale factor ( Attribute ID = 0x4020 in the thermostate cluster 513). To decrease overshots, I have set the value to 26 (bit4 = 16 + 10), and 22 and 6 later on. Using the GUI (based on the quirk) I was able to set the attribute and then read it back. However, it did not have any affect. But when I disabled the quirk, it worked exactly as desired. 

**To Reproduce**
1. Change control performance via the attribute `Control algorithm scale factor`.
2. Trigger a setpoint step and record/monitor control performance.
3. Disable quirk.
4. Repeat 2.

**Expected behavior**
An active quirk should not change the devices behavior.

**Screenshots**
This screenshot shows the thermostate quantities setpoint and local temperature as well as the control action (`pi_heat_demand`).
The temperature steps where the quirk was active are marked with a red Q. One can clearly see the agressive control action and the resulting overshot.
![grafik](https://user-images.githubusercontent.com/3462787/213552615-78dd3164-8f33-4363-adb4-c18173d86f7d.png)

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4678, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x000a",
        "0x0020",
        "0x0201",
        "0x0204",
[...]

Issue #1726, opened 2022-08-30
[Device Support Request] Danfoss Ally eTRV0100 - other features than just temperature and setpoint


OZ1SEJ:
**Is your feature request related to a problem? Please describe.**
Currently, the Danfoss Ally integration supports a temperature sensor (what's the current temperature at the valve?) and a temperature control (setpoint).

**Describe the solution you'd like**
According to https://zigbee.blakadder.com/Danfoss_014G2461.html the device supports

- % opening
- child lock
- external temperature sensor
- local temperature battery
- open window detection
- ota
- thermostat

...of which only "local temperature battery" (I suppose) and "thermostat" are currently supported in ZHA. The other parameters seem to be available by creating template sensors, but I believe these ought to be directly available.

EDIT: Also, the current implementation allows to change HVAC mode between "Heat" and "Off", which actually _isn't_ supported by the device. So I believe this should be removed.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4678, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x000a",
        "0x0020",
        "0x0201",
        "0x0204",
        "0x0b05"
      ],
      "out_clusters": [
        "0x0000",
        "0x0019"
      ]
    }
  },
  "manufacturer": "Danfoss",
  "model": "eTRV0100",
  "class": "zhaquirks.danfoss.thermostat.DanfossThermostat"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
[...]

Issue #717, opened 2021-01-11
[Device Support Request] Danfoss Ally thermostat: specific device settings (i.e. display Viewing direction)


SimpleUserHA:
**Is your feature request related to a problem? Please describe.**
Looking for support to configure the viewing direction of the LCD.

Not sure at which level additional device specific properties and their potential values are being defined to allow configuring them in other systems and user interfaces using zigpy.

**Describe the solution you'd like**
The thermostat supports configuring the viewing direction, in case it it mounted upside down.

```
= Cluster = 0x0204 Thermostat UI Configuration =
  * Attribute ID = 0x4000 Viewing Direction
     Range: 0 to 1
     0x00 = viewing direction 1
     0x01 = viewing direction 2
     Default is 0, but overwritten to value from production configuration at Init
```

Is it the reponsibility of zigpy or zha or homeassistant to  provide this cluster and attribute to the user?

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
```
{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4678, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x000a",
        "0x0020",
        "0x0201",
        "0x0204",
        "0x0b05"
      ],
      "out_clusters": [
        "0x0000",
        "0x0019"
      ]
    }
  },
  "manufacturer": "Danfoss",
  "model": "eTRV0100",
  "class": "zhaquirks.danfoss.thermostat.DanfossThermostat"
}
```

**Additional context**
[Thread at HomeAssistant forum](https://community.home-assistant.io/t/danfoss-ally-thermostat-firmware-update-ota-v1-08-open-window-external-temp-zha-deconz/261951)
[full spec about the Danfoss thermostat](https://github.com/SimpleUserHA/DanfosseTRV/raw/main/Danfoss%20Ally%20Radiator%20Thermostat%201.08%20Zigbee%20Cluster%20Specifications%2016122020.pdf)

PR #2150, opened 2023-01-29
Clean up Danfoss TRV custom attributes definitions and commands

Caius-Bonus:
This PR refactors and improves the quirk to make it possible to add binary_sensors, sensors, switches, numbers and selects in the ZHA integration.

~~Main changes:~~
~~1. It splits up the custom clusters into proprietary clusters with ids. This makes it so that ZHA doesn't need to know which models and vendors are supported. It only needs to know the cluster names.~~

Miscellaneous:
1. It fixes up the proprietary attribute access and types.
2. It renames some of the badly named proprietary attributes.
    This might break some users using ZHA clusters to address the TRV
3. It adds a preheat_command.
4. It also adds the Popp eT093WRG and Hive TRV003.
5. It maps System Mode Off to change setpoint to 5 C
6. It fixes Danfoss not requesting the time at connection by writing it on bind.

- [X] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [X] There is no commented out code in this PR.
- [x] The code has been formatted using Black

In ZHA I added this complementary PR: https://github.com/home-assistant/core/pull/86907

This PR will have an effect on the following issues:
https://github.com/zigpy/zha-device-handlers/issues/2115 (I don't know what could cause this, but it might change)
https://github.com/zigpy/zha-device-handlers/issues/1726 (It will now have this)
https://github.com/zigpy/zha-device-handlers/issues/717 (It will now have this)
https://github.com/zigpy/zha-device-handlers/issues/2728 (It will now have this)

## sources (I appreciate the documentation and updates):

1. The initial version is based on the following documentation: https://web.archive.org/web/20231215170505/https://assets.danfoss.com/documents/202524/AU417130778872en-000101.pdf (old revision), https://assets.danfoss.com/documents/367874/AU417130778872en-000102.pdf (new revision)

2. Other documentation used is: https://assets.danfoss.com/documents/284715/AM375549618098en-000103.pdf

3. Source for information regarding the new revision in the firmware package: https://www.danfoss.com/nl-nl/products/dhs/smart-heating/smart-heating/danfoss-ally/danfoss-ally-support/#tab-software

This documentation contradicts information regarding programming_operation_mode. Because the 2nd documentation seems more precise and supports both the old and the new G-serial TRVs, this version will be used


zhaquirks/danfoss/thermostat.py
```diff
@@ -26,47 +26,171 @@
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-from zhaquirks.danfoss import D5X84YU, DANFOSS
+
+from . import DANFOSS, HIVE, POPP
```
TheJulianJES, line 30: For making it easier to use custom quirks, please use absolute imports instead of relative imports.

zhaquirks/danfoss/thermostat.py
```diff
@@ -26,47 +26,171 @@
...
+    attributes = ManufacturerSpecificCluster.attributes.copy()
+    attributes.update(danfoss_thermostat_attr)
+
+    async def write_attributes(self, attributes, manufacturer=None):
+        """Write attributes to thermostat cluster."""
+        return await self.endpoint.thermostat.write_attributes(attributes, manufacturer)
+
+    async def bind(self):
+        """Read attributes before ZHA binds, this makes sure the entity is created."""
+        await read_attributes(self, self.endpoint.thermostat, danfoss_thermostat_attr)
```
TheJulianJES, line 121: Hmm. I think it would be better if we have the attributes that need to be read when pairing in ``ZCL_INIT_ATTRS`` here: https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/core/channels/manufacturerspecific.py

zhaquirks/danfoss/thermostat.py
Caius-Bonus: Forgot about that one. I'll fix that.

zhaquirks/danfoss/thermostat.py
Caius-Bonus: Fixed now. Thanks a lot

Caius-Bonus:
I fixed the merge-conflicts cause by the automatic checks.

Caius-Bonus:
Converting to a draft, because the code changed a lot and needs to be retested with hardware.

Caius-Bonus:
Tested and working. Ready for review

TheJulianJES:
> It splits up the custom clusters into proprietary clusters with ids. This makes it so that ZHA doesn't need to know which models and vendors are supported. It only needs to know the cluster names.

Hmm, I see why you want to do that. The current HA/ZHA + quirks implementation isn't great for custom stuff like this.
However, I'm not sure that artificially splitting up clusters is a good idea.

In HA, we should be able to match on a quirk (and cluster) via something like this:
```python
@MULTI_MATCH(
    cluster_handler_names="thermostat",
    quirk_classes={"thermostat.DanfossThermostat"},
)
```
Would this help?

Caius-Bonus:
@TheJulianJES How would this work for configuration in manufacturerspecific.py? I need to match on clusters there too.

Caius-Bonus:
I took a look at manufacturerspecific.py and came up with a way to match the quirk without requiring the custom clusters. Any comments?

TheJulianJES:
Just looked at it and [left a comment in the HA PR](https://github.com/home-assistant/core/pull/86907#discussion_r1341866443). It's not an ideal way though.

Caius-Bonus:
Tested and fully working again

TheJulianJES:
Since the [quirk ID stuff](https://github.com/home-assistant/core/pull/102482) (and your ID PR) for the ZHA PR haven't made it into HA yet, this won't make the 2023.11.0 release, as the 2023.11.0 beta is cut tomorrow.

Hopefully, the "quirk ID" implementation is ready before 2023.12.0, so we can get it into that release.
Thanks again for all the work on your PRs!

Caius-Bonus:
Thanks. It would be nice to get it merged before then.

patrykwoj91:
Any chance it will get merged anytime soon?

zhaquirks/danfoss/thermostat.py
```diff
@@ -26,113 +54,350 @@
...
-        }
-    )
+from zhaquirks.quirk_ids import DANFOSS_ALLY_THERMOSTAT
+
+DANFOSS = "Danfoss"
+HIVE = DANFOSS
+POPP = "D5X84YU"
+
+OCCUPIED_HEATING_SETPOINT_NAME = "occupied_heating_setpoint"
+SYSTEM_MODE_NAME = "system_mode"
```
TheJulianJES: For getting ZCL attribute ids or names, you can use: `Thermostat.AttributeDefs.system_mode.name`
(or `.id` for the id)

Caius-Bonus: Thanks, that tidies the constants up nicely.

TheJulianJES:
> Any chance it will get merged anytime soon?

Yeah, we'll try to target it for HA 2024.3.0. Similar to: https://github.com/zigpy/zha-device-handlers/pull/1961#issuecomment-1917878266

Although I was under the impression that the breaking changes are more "breaking" in HA. 
(Must have confused it with the Tuya TRV one a bit, since that gets rid of entities via the `AnalogInput` cluster and replaces it with proper custom entities)

---

If you're around at the moment btw., the HA 2023.2.0 beta is tomorrow.
You might be able to still get in this PR:
-  https://github.com/home-assistant/core/pull/106563

TheJulianJES:
Also, thanks again for all your PRs and sorry for the long wait.
At least all PRs with quirk IDs, your custom cluster handler matching, and so on are merged.

Caius-Bonus:
@TheJulianJES Thanks and no problem. I replied to all feedback and could resolve all but one.

zhaquirks/danfoss/thermostat.py
```diff
@@ -26,113 +55,442 @@
...
+            if len(result) == 1:
+                success_global.extend(result[0])
+            elif len(result) == 2:
+                success_global.extend(result[0])
+                failure_global.extend(result[1])
+
+        if failure_global:
+            return [success_global, failure_global]
+        else:
+            return [success_global]
```
TheJulianJES, line 183: Not completely required, but it might be nice if this line could also be covered in the tests.
(assuming codecov is correct and this isn't covered)

TheJulianJES:
To give another small update, I was finally able to get around to reviewing the HA PR a bit more and left some comments there now.

Also linking this comment regarding quirks v2 and "exposing entities in quirks" here: https://github.com/home-assistant/core/pull/87546#issuecomment-1955688910
This should (hopefully) come very soon. The text in that comment is basically also applies to your HA PR.

zhaquirks/danfoss/thermostat.py
Caius-Bonus: Done.

Caius-Bonus:
Thanks, I just responded to all feedback of the ZHA PR and this PR is ready.

TheJulianJES:
@Caius-Bonus Could this technically be merged without breaking any current user-exposed functionality even if the ZHA PR doesn't make it for the .3 release for some reason?
(Custom attributes are not user-exposed IMO. Only the entities)

zhaquirks/danfoss/thermostat.py
```diff
@@ -152,16 +510,19 @@ class DanfossThermostat(CustomDevice):
...
                     DanfossDiagnosticCluster,
                 ],
-                OUTPUT_CLUSTERS: [Basic, Ota],
+                OUTPUT_CLUSTERS: [Basic.cluster_id, Ota.cluster_id],
             }
         }
     }
+
+
+z = DanfossAdaptationRunStatusBitmap
```
TheJulianJES: Also, for what is this used?

Caius-Bonus:
> @Caius-Bonus Could this technically be merged without breaking any current user-exposed functionality even if the ZHA PR doesn't make it for the .3 release for some reason? (Custom attributes are not user-exposed IMO. Only the entities)

It shouldn't break anything, because there are no entities in ZHA currently depending on this quirk. The original quirk just changed the way setpoint changes were done and that behaviour will be exactly identical.

zhaquirks/danfoss/thermostat.py
Caius-Bonus: Oops, that was not supposed to be committed.

Caius-Bonus: Removed it.

Caius-Bonus:
@TheJulianJES, there were a few issues with the checks not running correctly, but after 2 retries it works again. Ready for merge.

Mikaka27:
@TheJulianJES do you know if this will make it into 2024.04?? I'm wondering whether to buy this device. Thanks :)

Edit: Or 2024.05 :)

zhaquirks/danfoss/thermostat.py
```diff
@@ -26,113 +55,442 @@
...
+            id=0x4022, type=types.uint16_t, access="rw", is_manufacturer_specific=True
+        )  # non-configurable reporting
+
+
+class DanfossTimeCluster(CustomizedStandardCluster, Time):
+    """Danfoss cluster for fixing the time."""
+
+    async def write_time(self):
+        epoch = datetime(2000, 1, 1, 0, 0, 0, 0)
+        current_time = (datetime.utcnow() - epoch).total_seconds()
```
TheJulianJES: `utcnow()` is deprecated and replaced by `datetime.datetime.now(datetime.UTC)`
(There might then be issues with the subtraction because of one datetime having a timezone and the other not.)

zhaquirks/danfoss/thermostat.py
```diff
@@ -26,113 +55,442 @@
...
+
+        time_zone = (
+            datetime.fromtimestamp(86400) - datetime.utcfromtimestamp(86400)
```
TheJulianJES: `utcfromtimestamp(t)` is deprecated and replaced by `datetime.datetime.fromtimestamp(t, datetime.UTC)`
(There might then be issues with the subtraction because of one datetime having a timezone and the other not.)

tests/test_danfoss.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+
+    patch_danfoss_trv_bind = mock.patch.object(
+        Time,
+        "bind",
+        mock.AsyncMock(side_effect=mock_wildcard),
+    )
+
+    with patch_danfoss_trv_bind:
+        with patch_danfoss_trv_write:
```
TheJulianJES: This can be put into one line:
```python
    with patch_danfoss_trv_bind, patch_danfoss_trv_write:
```

tests/test_danfoss.py
```diff
@@ -0,0 +1,241 @@
...
+    patch_danfoss_trv_write = mock.patch.object(
+        danfoss_time_cluster,
+        "_write_attributes",
+        mock.AsyncMock(side_effect=mock_write),
+    )
+
+    patch_danfoss_trv_bind = mock.patch.object(
+        Time,
+        "bind",
+        mock.AsyncMock(side_effect=mock_wildcard),
```
TheJulianJES: We could also add this as an annotation to the test instead:
```python
@mock.patch("zigpy.zcl.Cluster.bind", mock.AsyncMock())
```

tests/test_danfoss.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def mock_write(attributes, manufacturer=None):
+        records = [
+            WriteAttributesStatusRecord(foundation.Status.SUCCESS)
+            for attr in attributes
+        ]
+        return [records, []]
+
+    def mock_wildcard(*args, **kwargs):
+        return
```
TheJulianJES: This can be removed if we use an annotation on the test to mock zigpy bind().

tests/test_danfoss.py
```diff
@@ -0,0 +1,241 @@
...
+
+    assert_signature_matches_quirk(
+        zhaquirks.danfoss.thermostat.DanfossThermostat, signature
+    )
+
+
+async def test_danfoss_time_bind(zigpy_device_from_quirk):
+    device = zigpy_device_from_quirk(zhaquirks.danfoss.thermostat.DanfossThermostat)
+
+    danfoss_time_cluster = device.endpoints[1].in_clusters[Time.cluster_id]
```
TheJulianJES: We should be able to use the `ep_attribute` here for accessing that cluster:
```suggestion
    danfoss_time_cluster = device.endpoints[1].time
```

tests/test_danfoss.py
```diff
@@ -0,0 +1,241 @@
...
+
+            assert 0x0000 in danfoss_time_cluster._attr_cache
+            assert 0x0001 in danfoss_time_cluster._attr_cache
+            assert 0x0002 in danfoss_time_cluster._attr_cache
+
+
+async def test_danfoss_thermostat_write_attributes(zigpy_device_from_quirk):
+    device = zigpy_device_from_quirk(zhaquirks.danfoss.thermostat.DanfossThermostat)
+
+    danfoss_thermostat_cluster = device.endpoints[1].in_clusters[Thermostat.cluster_id]
```
TheJulianJES: We should be able to use the `ep_attribute` here for accessing that cluster:
```suggestion
    danfoss_thermostat_cluster = device.endpoints[1].thermostat
```

zhaquirks/danfoss/thermostat.py
```diff
@@ -26,113 +55,442 @@
...
 
+    async def bind(self):
+        """According to the documentation of Zigbee2MQTT there is a bug in the Danfoss firmware with the time.
+        It doesn't request it, so it has to be fed the correct time.
+        """
+        await self.endpoint.time.write_time()
 
-class DanfossUserInterfaceCluster(CustomCluster, UserInterface):
-    """Danfoss custom cluster."""
+        return await super().bind()
```
TheJulianJES: Since we're calling the `bind()` method for the Time cluster in the tests and the `bind()` method for the Thermostat cluster isn't covered yet, you can just call that method in the `test_danfoss_time_bind` test.

tests/test_danfoss.py
```diff
@@ -0,0 +1,241 @@
...
+
+    patch_danfoss_trv_bind = mock.patch.object(
+        Time,
+        "bind",
+        mock.AsyncMock(side_effect=mock_wildcard),
+    )
+
+    with patch_danfoss_trv_bind:
+        with patch_danfoss_trv_write:
+            await danfoss_time_cluster.bind()
```
TheJulianJES: So, calling `await thermostat_cluster.bind()` here (instead of directly `bind()` on the Time cluster), we can also cover the `bind()` method in the Thermostat cluster.

(This also needs the `Thermostat` `bind()` method to be mocked, but using the "zigpy bind annotation mock" also does that.)

TheJulianJES (commented):
Trying to get this wrapped up (finally 😄 ). Few minor comments.

Also, can you rebase this once more? There were some ruff changes and I just want to make sure this PR still passes then.

I'll merge this then. For the custom functionality in ZHA, we should ideally try to move this into a v2 quirk afterwards. The [docs](https://github.com/zigpy/zha-device-handlers/blob/dm/quirks-v2-documentation/quirks_v2.md) aren't ready yet, but it's not that difficult. Most of the code (clusters and so on) stay the same. It's just the `CustomDevice` that gets modified. But again, that's for another PR, not this one.
It would allow us to move all/most of the functionality from your HA/ZHA PR into quirks directly which also makes future changes way easier, as it's all in quirks.

zhaquirks/danfoss/thermostat.py
Caius-Bonus: Fixed.

zhaquirks/danfoss/thermostat.py
Caius-Bonus: Fixed. This specific subtraction should result in a difference based on the timezone, so that is correct.

tests/test_danfoss.py
Caius-Bonus: Fixed.

tests/test_danfoss.py
Caius-Bonus: Fixed.

tests/test_danfoss.py
Caius-Bonus: Fixed.

zhaquirks/danfoss/thermostat.py
Caius-Bonus: You're right.

tests/test_danfoss.py
Caius-Bonus: Fixed.

zhaquirks/danfoss/thermostat.py
Hoffmann77: If you merely need the utc offset in seconds you could maybe use something like:
```
time_zone = datetime.now().astimezone().utcoffset().total_seconds()
```
This should avoid the failing test.

Caius-Bonus:
@TheJulianJES Ready

Caius-Bonus:
@Mikaka27 Keep in mind that even when this is merged, https://github.com/home-assistant/core/pull/86907 needs to be merged before this is useful in home assistant.

zhaquirks/danfoss/thermostat.py
```diff
@@ -26,113 +55,456 @@
...
+    async def bind(self):
+        """According to the documentation of Zigbee2MQTT there is a bug in the Danfoss firmware with the time.
+
+        It doesn't request it, so it has to be fed the correct time.
+        """
+        result = await super().bind()
+
+        await self.write_time()
+
+        return result
```
TheJulianJES: Last thing, we can remove these two empty lines:
```suggestion
        result = await super().bind()
        await self.write_time()
        return result
```

TheJulianJES (approved):
LGTM, thanks!
The timezone stuff also looks cleaner now.

TheJulianJES:
Codecov is still broken, but three lines are not covered by tests. I'm guessing this is another case of the "write time" stuff with `bind()`. As that's simple enough, that can be addressed in a future PR. This one is good to go IMO.

`Name                                               Stmts   Miss  Cover`
`zhaquirks/danfoss/thermostat.py                      162      3    98%`

PR #2152, opened 2023-01-30
Merge PR #2123 and PR #2140 + fivebtnremote files merge

stickpin:
- Merge PR #2123 and PR #2140
- fivebtnremote.py and fivebtnremotezha.py are now combined together

FYI @MatMaul, @TheJulianJES, @MattWestb

PR #2140, opened 2023-01-27
TRADFRI remote control - add fw support v24.4.5

stickpin:
TRADFRI remote control  - Add support for firmware v24.4.5

More details in this bug report: https://github.com/zigpy/zha-device-handlers/issues/2139

PR #2155, opened 2023-01-31
Refactor IKEA five button remote quirks

TheJulianJES:
This combines the quirks from `fivebtnremotezha.py` and `fivebtnremote.py`, as they are one device with different firmware versions and the device automation triggers all seem to be the same (they were somewhat duplicated previously).

One parameter was changed to `537` in https://github.com/zigpy/zha-device-handlers/pull/1511, but [after some discussion](https://github.com/zigpy/zha-device-handlers/pull/2123#discussion_r1089477679), this seems to invalid and we can likely just replace it with the same device automation triggers for the other automations (which is done in this PR).

Also tested the changes with two different firmware versions of the IKEA five button remote.

--- 

Note: The automation triggers are all the same now (and not missing): all quirks inherit from `IkeaTradfriRemote1`.

@MattWestb @stickpin for review

This takes out the refactoring part from:
- https://github.com/zigpy/zha-device-handlers/pull/2123
- https://github.com/zigpy/zha-device-handlers/pull/2152

When merged, I'll rebase the following PRs:
- https://github.com/zigpy/zha-device-handlers/pull/2123 (to only add the two new triggers)
- https://github.com/zigpy/zha-device-handlers/pull/2140 (to add the new signature with existing triggers)

(Having separate PRs for the (1) refactoring, (2) added triggers and (3) for the new signature makes it so we don't have one huge commit on the repo that changes a bunch of things)

(I've also done the refactoring here in multiple commits (will be squashed when merging), so it's easier to see the exact steps/changes here).


stickpin (approved):

MattWestb (approved):
I only having v24.4.5 in my test system and production system is no use testing the quirk (with old ver 3 quirk) if its working OK for your remotes with it.
Lets get it merged and we can testing the v24.4.5 in the next PR.

Issue #2113, opened 2023-01-18
[Device Support Request] TS0601 _TZE200_qyflbnbj Tuya outdoor temperature and humidity sensor


m0nty81:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Sensor connection

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_qyflbnbj",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
Paste the diagnostic information here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>
<summary>Additional logs</summary>

```
[0xC241:1:0xef00] Unknown cluster command 1 b'\x00u\x01\x02\x00\x04\x00\x00\x00\x00'
[0xC241:1:0xef00] Unknown cluster command 1 b'\x00u\x01\x02\x00\x04\x00\x00\x00\x00'
[0xC241:1:0xef00] Unknown cluster command 1 b'\x00u\x01\x02\x00\x04\x00\x00\x00\x00'
[0xC241:1:0xef00] Unknown cluster command 1 b'\x00u\x01\x02\x00\x04\x00\x00\x00\x00'
[...]

PR #2175, opened 2023-02-04
tuya: fix negative values for some devices

### LINKED ISSUE #1976: [BUG] Negative temperatures wrap around to 6000°C with Tuya Smart Air Detector 6 in 1
**Describe the bug**

[Tuya Smart Air Detector 6 in 1](https://zigbee.blakadder.com/Tuya_DCR-KQG.html) has working temperature range 0-60°C by specs.  
But handling of temps out of specs leads to wacky statistics.

Issue here is that apparently either raw reported values or their source representation uses unsigned 16 bit numbers and negative values written as [2's complement](https://en.wikipedia.org/wiki/Two%27s_complement) signed integer, which makes this numbers in range of ~65500-ish.

This leads to reported temperatures and temperatures displayed in HA to be above 6000°C which by any means doesn't make any sense and makes charts unusable since all other sensors will have temperatures around 0°C at this scale.

However, despite temperatures being in unsupported range, I think it will be more convenient to convert them back from 2's complement to signed form. For example, it could be done like so:
```(((x&0xFFFF)^0x8000) - 0x8000)```

`x` here is 16 bit _integer_ value. Hence if it is, let's say, 65528, then we'll get -8 as the result.  
Which converts to "real" temperature value of -0.8°C instead of current 6552.8°C.

It brings results back to normal range and, despite them not being inside supported value range, still makes them usable.

Actually, values themselves, after being converted, are quite accurate.  
For example, when sensor reports temperature 6549.9°C, it gives -3.7°C after conversion using formula above. And sensor rated for negative temps gives -3.5°C (I added this as 3rd chart for the reference).

**To Reproduce**
1. Go to place with negative ambient temps
2. Plug sensor there
3. Connect it to HA
4. See high temperatures

**Expected behavior**
Don't see temperatures outside of normal ambient temperatures range

**Screenshots**
Charts

Bad temp

Real temp

<details>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0x0402",
        "0x0405",
        "0x040d",
        "0x042a",
        "0x042b",
        "0x042e",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZE200_dwcarsat",
  "model": "TS0601",
  "class": "ts0601_air_quality.TuyaCO2Sensor2GPP"
}
```

</details>

<details>

```yaml
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>

```
```

</details>

**Additional context**

Shulyaka:
When using tuya datapoint of type `TuyaDPType.VALUE` we are currently assuming it to be of unsigned type. This seems to be an incorrect assumption.
This PR interpret them as a signed type.
Probably fixes #1976 and the negative temperature issue from #2113
Fixes https://github.com/zigpy/zha-device-handlers/issues/2113

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -153,7 +153,7 @@ class TuyaData(t.Struct):
     def payload(
         self,
     ) -> Union[
-        t.uint32_t_be,
+        t.int32s_be,
         t.Bool,
         t.CharacterString,
         t.enum8,
@@ -164,7 +164,7 @@ class TuyaData(t.Struct):
     ]:
         """Payload accordingly to data point type."""
         if self.dp_type == TuyaDPType.VALUE:
-            return t.uint32_t_be.deserialize(self.raw)[0]
+            return t.int32s_be.deserialize(self.raw)[0]
         elif self.dp_type == TuyaDPType.BOOL:
             return t.Bool.deserialize(self.raw)[0]
         elif self.dp_type == TuyaDPType.STRING:
@@ -186,7 +186,7 @@ class TuyaData(t.Struct):
     def payload(self, value):
         """Set payload accordingly to data point type."""
         if self.dp_type == TuyaDPType.VALUE:
-            self.raw = t.uint32_t_be(value).serialize()
+            self.raw = t.int32s_be(value).serialize()
         elif self.dp_type == TuyaDPType.BOOL:
             self.raw = t.Bool(value).serialize()
         elif self.dp_type == TuyaDPType.STRING:
--- a/zhaquirks/xiaomi/aqara/feeder_acn001.py
+++ b/zhaquirks/xiaomi/aqara/feeder_acn001.py
@@ -145,8 +145,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
 
     def _parse_feeder_attribute(self, value: bytes) -> None:
         """Parse the feeder attribute."""
-        # attribute is big endian, so we need to reverse the bytes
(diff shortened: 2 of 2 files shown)


javicalle (approved):
LGTM

javicalle:
Thanks Denis 🙇🏻

Shulyaka:
It was your idea :)

javicalle:
And I agree with the #1976 fix.  Also (not tagged) #2113.

PR #2176, opened 2023-02-04
Tuya: fix negative temperatures for some devices

Shulyaka:
Some custom clusters that are supposed to store the temperature, are of unsigned type.
When these devices are exposed to negative temperatures, the value they store becomes too big because of the integer overflow.
This PR fixes the types at least for some of them.

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -40,7 +40,7 @@ ZCL_TUYA_SET_TIME_REQUEST = b"\tp\x24\x00\00"
 ZCL_TUYA_MOTION = b"\tL\x01\x00\x05\x03\x04\x00\x01\x02"
 ZCL_TUYA_SWITCH_ON = b"\tQ\x02\x006\x01\x01\x00\x01\x01"
 ZCL_TUYA_SWITCH_OFF = b"\tQ\x02\x006\x01\x01\x00\x01\x00"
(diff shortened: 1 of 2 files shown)


tests/test_tuya.py
```diff
@@ -40,7 +40,7 @@
 ZCL_TUYA_MOTION = b"\tL\x01\x00\x05\x03\x04\x00\x01\x02"
 ZCL_TUYA_SWITCH_ON = b"\tQ\x02\x006\x01\x01\x00\x01\x01"
 ZCL_TUYA_SWITCH_OFF = b"\tQ\x02\x006\x01\x01\x00\x01\x00"
-ZCL_TUYA_ATTRIBUTE_617_TO_179 = b"\tp\x02\x00\x02i\x02\x00\x04\x00\x00\x00\xb3"
+ZCL_TUYA_ATTRIBUTE_617_TO_179 = b"\tp\x02\x00\x02i\x02\x00\x04\xff\xff\xff\xe3"
```
javicalle, line 43: Just to be sure: 
You have updated the test data to be able to validate against negative values, is it correct?
That values comes from a real report or is just the 'matching' ones?

I'm not judging, I'm just pretending to understand.

Shulyaka: It's not a real report, it's just made up because I don't have the actual device. But the representation of negative values is similar in other user's logs.
I've updated the attribute type, then updated the test to use the negative values.

PR #2180, opened 2023-02-06
int16s_temp_workaround

Shulyaka:
Fixes https://github.com/zigpy/zha-device-handlers/issues/1976

TODO: tests

PR #2188, opened 2023-02-10
Fix Tuya signed 16 bit temperature values

javicalle:
Another approach to fix the 16 bits reports from some devices.
Supersed #2180
Fixes: #1976
(And will not break any sensor while temp < 4095 ºC)


javicalle:
@Shulyaka I was also figuring out how can be done the same that yours PR.
Because I don't feel confortable with all the bitwise operations I searched for another approach.
I don't like all the `serialize` / `deserialize` part but I haven't found a more elegant a way to implement that.

I don't think there is one version better than another, it's more a matter of preference.

You can take the tests for your PR.

Shulyaka:
@javicalle I like your approach more

Shulyaka:
@TheJulianJES No, https://github.com/zigpy/zha-device-handlers/pull/2176 is not related.

javicalle:
Any idea how can be done to parse the value directly?

What I would have is:
* a 'CustomTemperature' definition like the one defined:
```python
class CustomTemperature(t.Struct):
    """Custom temperature wrapper."""

    field_1: t.int16s_be
    temperature: t.int16s_be
```
* define the attribute type with that class:
```python
            0xEF12: ("custom_temperature", CustomTemperature, False),
```
* handle the value directly without conversions:
```python
    def update_attribute(self, attr_name: str, value: Any) -> None:
        """Calculate the current temperature."""

        self._update_attribute(self.attributes_by_name[attr_name].id, value)

        if attr_name == "custom_temperature":
            custom_temperature = self.attributes_by_name["custom_temperature"]  # that would be an CustomTemperature instance
            if custom_temperature:
                temp = custom_temperature.temperature * 10
                super().update_attribute("measured_value", temp)
```

I have tried all the data types that I can but I didn't get any clean solution.

javicalle:
Ready to review now.
Not sure if better to maintain (and extend). 
Thanks @Shulyaka!

zhaquirks/tuya/air/__init__.py
```diff
@@ -31,9 +31,42 @@ class TuyaAirQualityVOC(TuyaLocalCluster):
...
+
+    def update_attribute(self, attr_name: str, value: Any) -> None:
+        """Calculate the current temperature."""
+
+        super().update_attribute(attr_name, value)
+
+        if attr_name == "custom_temperature":
+            custom_temperature = self._attr_cache.get(
+                self.attributes_by_name[attr_name].id,
+            )
```
TheJulianJES, line 65: Not that it matters, but what's the benefit of getting `custom_temperature` from the attribute cache again when it should also be `value` here?

javicalle: You are right!!! 
I was expecting that value was the raw value (so no `value.temperature`) and we have to recover from class 😅 
Just checked and working like a charm

Thanks Julian!

TheJulianJES (approved):
LGTM

Issue #2181, opened 2023-02-06
[Device Support Request] consumer_connected binary_sensor for lumi.plug.mmeu01


myhomeiot:
Please add ```consumer_connected``` binary_sensor for lumi.plug.mmeu01.
Thanks in advance.

I add it as attribute into OppleCluster class (plug_eu.py) and test it using Manage Zigbee Device from Home Assistant - it's report correctly if something physically connected into the socket (even without power consumption):
```
    attributes = {
        0x0009: ("mode", types.uint8_t, True),
        0x0201: ("power_outage_memory", types.Bool, True),
        0x0207: ("consumer_connected", types.Bool, True),
    }
```

javicalle:
You can do it yourself by editing the file from the browser:
* https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/xiaomi/aqara/plug_eu.py

Save the changes and follow the prompts to generate the PR.

That way other users can benefit from the change thanks to you.

myhomeiot:
> Save the changes and follow the prompts to generate the PR.

Unfortunately it's requires also the changes in Home Assistant Core, so PRs must be coordinated.

javicalle:
Well, it can be added in the quirk at any time.
In HA it would be needed that the `zha-quirk` version is released to be able to create the new entities.

PR #2198, opened 2023-02-15
Add "consumer connected" attribute to Xiaomi EU plugs

TheJulianJES:
For zha-quirks, this fixes https://github.com/zigpy/zha-device-handlers/issues/2181
Follow-up PR for ZHA to create the binary sensor:
- https://github.com/home-assistant/core/pull/88194

Tested using a Xiaomi `MAEU01` plug on v32 and v41 firmware. 
https://github.com/zigpy/zha-device-handlers/issues/2181 mentions it also works on an `MMEU01` plug (so the other Xiaomi EU plug).

### DIFF
--- a/zhaquirks/xiaomi/aqara/plug_eu.py
+++ b/zhaquirks/xiaomi/aqara/plug_eu.py
@@ -68,6 +68,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
     attributes = {
         0x0009: ("mode", types.uint8_t, True),
         0x0201: ("power_outage_memory", types.Bool, True),
+        0x0207: ("consumer_connected", types.Bool, True),
     }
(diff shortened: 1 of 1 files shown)


javicalle (approved):
LGTM

myhomeiot:
@TheJulianJES Maybe it's will be good idea to make this sensor as `device_class: plug`?
PS: Sorry I doesn't mention this before, I just found that HA has this device class :(

image

TheJulianJES:
Yep, I also just saw that a couple of days ago 😄
If I don't forget, I'll PR that change in the next few days or so

TheJulianJES:
@myhomeiot 

PR'ed:
- https://github.com/home-assistant/core/pull/89476

Btw., you don't need to use `customize.yaml`. You can change it using the entity editor:
image

PR #2208, opened 2023-02-18
Support double click and long press for Sinopé light and dimmer

claudegel:
- Added support for reporting on double click on / off and long press on / off for Sinope light and dimmer. This allow to do some automation when somebody double click or long press on device button.
- Added many new attributes for manufacturer cluster on Sinopé thermostat, light, switch and sensor.
- Added description and value for attributes of type t.enum8 on Sinopé thermostat, light and switch.


zhaquirks/sinope/light.py
```diff
@@ -29,29 +30,55 @@
...
         0x0050: ("onLedColor", t.uint24_t, True),
         0x0051: ("offLedColor", t.uint24_t, True),
         0x0052: ("onLedIntensity", t.uint8_t, True),
         0x0053: ("offLedIntensity", t.uint8_t, True),
+        0x0054: ("actionReport", action, True),
         0x0055: ("minIntensity", t.uint16_t, True),
         0x00A0: ("Timer", t.uint32_t, True),
         0x0119: ("ConnectedLoad", t.uint16_t, True),
+        0x0200: ("Unknown", t.bitmap32, True),
+        0xFFFD: ("cluster_revision", t.uint16_t, True),
```
TheJulianJES, line 74: This applies to all Sinope quirks (and not just this section):
Generally, zigpy and zha-quirks should define attributes in `snake_case` (for the strings/names).

Would it be possible to change this or could that break existing stuff? (I see it's already the case in the currently merged quirks)

zhaquirks/sinope/light.py
```diff
@@ -29,29 +30,55 @@
...
-from zhaquirks.sinope import SINOPE
+from zhaquirks.sinope import LIGHT_DEVICE_TRIGGERS, SINOPE
 
 SINOPE_MANUFACTURER_CLUSTER_ID = 0xFF01
 
 
 class SinopeTechnologiesManufacturerCluster(CustomCluster):
     """SinopeTechnologiesManufacturerCluster manufacturer cluster."""
 
+    class keypadLock(t.enum8):
```
TheJulianJES, line 41: Class names should be `CamelCase`. So this would be `KeypadLock` for example (same for all other places).

zhaquirks/sinope/light.py
```diff
@@ -29,29 +30,55 @@
...
+        """actionReport values."""
+
+        single_on = 0x01
+        single_release_on = 0x02
+        long_on = 0x03
+        double_on = 0x04
+        single_off = 0x11
+        single_release_off = 0x12
+        long_off = 0x13
+        double_off = 0x14
```
TheJulianJES, line 57: Zigpy uses a mix of `snake_case` and `CamelCase` for these. This should be: `Single_On`, `Single_Release_On`, ...

Although zigpy also isn't completely consistent in upper-casing the following letters. So I guess `Single_on` and so on would also be fine according to those conventions. (See [zigpy code as an example](https://github.com/zigpy/zigpy/blob/2cbb79f4335947a71c039a37ff26e87940dbd7a7/zigpy/zcl/clusters/hvac.py#L142-L161))

TheJulianJES:
Thanks! Generally looks good. If possible, it would be great if the naming scheme of classes and attributes could be adapted to conventions.

claudegel:
Thanks, let me do the changes

zhaquirks/sinope/light.py
claudegel: so keypadLockout should be keypad_lockout and so on. There is no problem to change this.

zhaquirks/sinope/light.py
claudegel: Ok I'll switch  to Single_on ...

claudegel:
I have a question if you can help. In thermostat.py in two subclass as follow:
```
    class SensorType(t.enum8):
        """temp_sensor_type values."""

        K10 = 0x00
        K12 = 0x01

    class TimeFormat(t.enum8):
        """time_format values."""

        H24 = 0x00
        H12 = 0x01
```
How can I change H24 to 24h or 24_h and K10 to 10k or 10_k
Loook like it throw and error if it start with a number instead of a letter

TheJulianJES:
> How can I change H24 to 24h or 24_h and K10 to 10k or 10_k
Loook like it throw and error if it start with a number instead of a letter

Yeah, you can't. Other possibilities I can think of are:  `Format_24h`, `Format_24_H`, `H_24`, `F_24h`, and similar.
(Maybe the first one makes the most sense?)

For example, zigpy also seems to just prefix the "type" again:
https://github.com/zigpy/zigpy/blob/2cbb79f4335947a71c039a37ff26e87940dbd7a7/zigpy/zcl/clusters/hvac.py#L104-L108

claudegel:
Thanks I've changed the time and sensor format description

TheJulianJES:
Looks pretty good now. I assume everything works as expected with the devices?

One thing I noticed (although it wasn't done in this PR) is this:
https://github.com/zigpy/zha-device-handlers/blob/ae2ed2dc0d4a85ba9f74decdceccab2db482c981/zhaquirks/sinope/thermostat.py#L185-L204

and this:
https://github.com/zigpy/zha-device-handlers/blob/ae2ed2dc0d4a85ba9f74decdceccab2db482c981/zhaquirks/sinope/thermostat.py#L240-L258

All clusters are referenced directly, although the "unmodified" clusters could just be referenced by their `cluster_id` (like done everywhere else).
Maybe that can also be changed in this PR or another PR?

claudegel:
Yes I'll make those changes in this PR today. Better clean everything at once

claudegel:
Finish cleaning, thank you

TheJulianJES (approved):
LGTM. Thanks!

Hoping to get a second review, as it's quite a few changes, but since it's "only" attribute definitions and device triggers, it should be fine.

PR #2215, opened 2023-02-23
Fix long release automation trigger for Aqara Mini button

ndpeter:
Corrected LONG_RELEASE in b1acn01 device to COMMAND_RELEASE.  Previously both long press and release would fire simultaneously as they were both tied to COMMAND_HOLD. 

I've checked this with the b1acn01 button device. but have not confirmed with the AQ3 switch.


TheJulianJES:
Looks like the device automation triggers were always wrong. Looking at Z2M, I think https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/xiaomi/aqara/sensor_swit.py should also be changed.
Can you also do that in this PR?

javicalle:
Maybe also in `thirdreality`?:
https://github.com/zigpy/zha-device-handlers/blob/f6ab87e1c48ccb6db262e8732ca6b63d67fb6252/zhaquirks/thirdreality/button.py#L113

TheJulianJES (approved):
Thanks!

TheJulianJES:
I'll address the other Aqara mini button and ThirdReality buttons in a separate PR.
I suspect the ThirdReality one was just copy-pasted from the Aqara one.

PR #2217, opened 2023-02-23
Remove color cluster for GLEDOPTO GL-C-009P (mini) dimmer

TheJulianJES:
Some versions of the GL-C-009P dimmer seem to have a Color cluster for some reason.
The quirk removes this cluster.


javicalle (approved):
LGTM

PR #2218, opened 2023-02-23
Clean up IKEA quirks

TheJulianJES:
- use constant in the shortcut button quirk

This mostly cleans up the IKEA TRADFRI plug quirk:
- use constants instead of strings
- add model name (so we're not only matching on manufacturer and cluster signature)

I think this quirk is only used for older TRADFRI plug firmware versions, but it would be good to also have the newer signature, as that still exposes the `LevelControl` cluster IIRC (and we should probably remove that, as ZHA creates a bunch of config entities for that).

### DIFF
--- a/zhaquirks/ikea/shortcutbtn.py
+++ b/zhaquirks/ikea/shortcutbtn.py
@@ -37,7 +37,12 @@ from zhaquirks.const import (
     SHORT_PRESS,
     TURN_ON,
 )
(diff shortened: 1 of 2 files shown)


MattWestb:
I  think we shall implanting handles for ZGPP cluster on the endpoint so can importing it nice and its looks the same  on all devices.

javicalle (approved):
LGTM

PR #2219, opened 2023-02-23
Initial Aqara Smoke Detector implementation

TheJulianJES:
### WIP

This adds all known attributes that the Aqara Smoke Detector exposes and properly updates the IAS entity that ZHA creates when `smoke` is detected.

Addresses part of: https://github.com/zigpy/zha-device-handlers/issues/1828 (more information in that issue)

### DIFF
--- a/zhaquirks/xiaomi/__init__.py
+++ b/zhaquirks/xiaomi/__init__.py
@@ -86,7 +86,13 @@ CONSUMPTION_REPORTED = "consumption_reported"
 VOLTAGE_REPORTED = "voltage_reported"
 ILLUMINANCE_MEASUREMENT = "illuminance_measurement"
 ILLUMINANCE_REPORTED = "illuminance_reported"
-SMOKE = "smoke"
+# SMOKE = "smoke"
+# SMOKE_DENSITY = "smoke_density"
+# TEST = "test"
+# BUZZER_MANUAL_MUTE = "buzzer_manual_mute"
+# HEARTBEAT_INDICATOR = "heartbeat_indicator"
+# LINKAGE_ALARM = "linkage_alarm"
+# LINKAGE_ALARM_STATE = "linkage_alarm_state"
 XIAOMI_AQARA_ATTRIBUTE = 0xFF01
 XIAOMI_AQARA_ATTRIBUTE_E1 = 0x00F7
 XIAOMI_ATTR_3 = "X-attrib-3"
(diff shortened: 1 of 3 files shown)


zhaquirks/xiaomi/__init__.py
```diff
@@ -302,14 +308,14 @@ def _update_attribute(self, attrid, value):
                 "update_battery_percentage",
                 attributes[BATTERY_PERCENTAGE_REMAINING_ATTRIBUTE],
             )
-        if SMOKE in attributes:
+        if "smoke" in attributes:
```
johnboiles, line 311: I'm curious, why not use the `SMOKE` constant here?

TheJulianJES: It's temporary. Z2M has a bunch of duplicated stuff (you'll see this further below in the code) that I now duplicated and I didn't want to make a bunch of constants yet not knowing which ones we keep in the end.

TheJulianJES:
This quirks PR should be ready for now.

TheJulianJES:
@javicalle If you have some time, can you look at this (and maybe https://github.com/zigpy/zha-device-handlers/pull/2238) to see if you notice any glaring mistakes?

zhaquirks/xiaomi/aqara/smoke.py
```diff
@@ -0,0 +1,131 @@
...
+        SMOKE_DENSITY_DBM: ("smoke_density_dbm", types.Single, True),
+    }
+
+    def _update_attribute(self, attrid: int, value: Any) -> None:
+        """Pass attribute update to another cluster if necessary."""
+        super()._update_attribute(attrid, value)
+        if attrid == SMOKE:
+            self.endpoint.ias_zone.update_attribute(ZONE_STATUS, value)
+        elif attrid == SMOKE_DENSITY:
+            self.update_attribute(SMOKE_DENSITY_DBM, SMOKE_DENSITY_DBM_MAP[value])
```
javicalle, line 73: Do we prefer an error here if `value` not in `SMOKE_DENSITY_DBM_MAP` or silently return something else?
Not an strong opinion from me, just thinking loud.

javicalle (approved):
LGTM
Just a comment.

zhaquirks/xiaomi/aqara/smoke.py
TheJulianJES: Hmm. The sensor seems to only report a value from 0 to 10 for the smoke density (in my testing and Z2M does this in a very similar way: https://github.com/Koenkk/zigbee-herdsman-converters/blob/ac73a48e057ab5a0b7d8d55bbd09a6dda258be15/lib/xiaomi.js#L389-L390)

I think if it reports another value, we should get an exception, so some user hopefully reports this 😄

PR #2220, opened 2023-02-24
Add Smartwings WM25L-Z quirk

tube0013:
Fixes: #1689 

Initial support for Smartwings blinds/shades - they all seem to use the same zigbee motor/variant. In some or all cases the the blinds need to be inverted with the included physical RF remote as per the instructions included with the blinds/shades.

Thanks to @warllo54 and @haywiremk for the actual work here, this was pulled for #1689 and it works for me 😀

### DIFF
--- a/zhaquirks/smartwings/wm25lz.py
+++ b/zhaquirks/smartwings/wm25lz.py
@@ -38,6 +38,7 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
         *args,
         manufacturer=None,
         expect_reply=True,
+        tries=1,
         tsn=None,
         **kwargs
     ):
@@ -53,6 +54,7 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
             *args,
             manufacturer=manufacturer,
             expect_reply=expect_reply,
+            tries=tries,
             tsn=tsn,
             **kwargs,
         )
--- a/zhaquirks/smartwings/wm25lz.py
+++ b/zhaquirks/smartwings/wm25lz.py
@@ -30,11 +30,17 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
     """
 
     cluster_id = WindowCovering.cluster_id
-    CMD_GO_UP = 0x0001 #reported as down_close in ZHA commands
-    CMD_GO_DOWN = 0x0000 #reported as up_open in ZHA commands
-    
+    CMD_GO_UP = 0x0001  # reported as down_close in ZHA commands
+    CMD_GO_DOWN = 0x0000  # reported as up_open in ZHA commands
+
     async def command(
-        self, command_id, *args, manufacturer=None, expect_reply=True, tsn=None, **kwargs
+        self,
+        command_id,
+        *args,
+        manufacturer=None,
+        expect_reply=True,
+        tsn=None,
+        **kwargs
     ):
         """Override default commands for up and down. they are backwards."""
 
@@ -50,6 +56,7 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
             tsn=tsn,
             **kwargs,
         )
+
 
 class WM25LBlinds(CustomDevice):
     """Custom device representing Smartwings WM25LZ blinds."""
--- a/zhaquirks/smartwings/wm25lz.py
+++ b/zhaquirks/smartwings/wm25lz.py
@@ -1,4 +1,4 @@
-"""Device handler for Smartwings WM25L-Z blinds."""
+"""Device handler for Smartwings blinds."""
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
 from zigpy.zcl.clusters.closures import WindowCovering
@@ -58,7 +58,7 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
 
 
 class WM25LBlinds(CustomDevice):
-    """Custom device representing Smartwings WM25/L-Z roller blinds."""
+    """Custom device representing Smartwings WM25LZ blinds."""
 
     signature = {
         # <SimpleDescriptor endpoint=1 profile=260 device_type=514
--- a/zhaquirks/smartwings/wm25lz.py
+++ b/zhaquirks/smartwings/wm25lz.py
@@ -30,11 +30,11 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
     """
 
     cluster_id = WindowCovering.cluster_id
-    CMD_GO_UP = 0x0001  # reported as down_close in ZHA commands
-    CMD_GO_DOWN = 0x0000  # reported as up_open in ZHA commands
-
+    CMD_GO_UP = 0x0001 #reported as down_close in ZHA commands
+    CMD_GO_DOWN = 0x0000 #reported as up_open in ZHA commands
+    
     async def command(
-        self, command_id, *args, manufacturer=None, expect_reply=True, tsn=None
+        self, command_id, *args, manufacturer=None, expect_reply=True, tsn=None, **kwargs
     ):
         """Override default commands for up and down. they are backwards."""
 
@@ -47,9 +47,9 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
             *args,
             manufacturer=manufacturer,
             expect_reply=expect_reply,
-            tsn=tsn
+            tsn=tsn,
+            **kwargs,
         )
-
 
 class WM25LBlinds(CustomDevice):
     """Custom device representing Smartwings WM25LZ blinds."""
--- a/zhaquirks/smartwings/wm25lz.py
+++ b/zhaquirks/smartwings/wm25lz.py
@@ -1,6 +1,10 @@
 """Device handler for Smartwings blinds."""
+from typing import Any
+
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
+from zigpy.zcl import foundation
 from zigpy.zcl.clusters.closures import WindowCovering
 from zigpy.zcl.clusters.general import (
     Basic,
@@ -30,13 +34,13 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
 
     async def command(
         self,
-        command_id,
+        command_id: foundation.GeneralCommand | int | t.uint8_t,
         *args,
-        manufacturer=None,
-        expect_reply=True,
-        tries=1,
-        tsn=None,
-        **kwargs
+        manufacturer: int | t.uint16_t | None = None,
+        expect_reply: bool = True,
+        tries: int = 1,
+        tsn: int | t.uint8_t | None = None,
+        **kwargs: Any,
     ):
         """Override default commands for up and down. They need to be backwards."""
         # swap up and down commands
--- a/zhaquirks/smartwings/wm25lz.py
+++ b/zhaquirks/smartwings/wm25lz.py
@@ -23,9 +23,11 @@ from zhaquirks.const import (
 
 
 class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
-    """WindowCovering cluster implementation.
+    """
+    WindowCovering cluster implementation.
     This implementation inverts the reported covering percent for non standard
-    devices that don't follow the reporting spec."""
+    devices that don't follow the reporting spec.
+    """
 
     cluster_id = WindowCovering.cluster_id
     CURRENT_POSITION_LIFT_PERCENTAGE = 0x0008
--- a/zhaquirks/smartwings/wm25lz.py
+++ b/zhaquirks/smartwings/wm25lz.py
@@ -1,4 +1,6 @@
 """Device handler for Smartwings blinds."""
+from __future__ import annotations
+
 from typing import Any
 
 from zigpy.profiles import zha
--- a/zhaquirks/smartwings/wm25lz.py
+++ b/zhaquirks/smartwings/wm25lz.py
@@ -29,8 +29,8 @@ class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
     devices that don't follow the reporting spec.
     """
 
-    CMD_GO_UP = WindowCovering.commands_by_name["down_close"].id
-    CMD_GO_DOWN = WindowCovering.commands_by_name["up_open"].id
+    CMD_UP_OPEN = WindowCovering.commands_by_name["up_open"].id
+    CMD_DOWN_CLOSE = WindowCovering.commands_by_name["down_close"].id
 
     async def command(
         self,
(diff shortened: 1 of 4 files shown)


zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+    CMD_GO_TO_LIFT_PERCENTAGE = 0x0005
+
+    def _update_attribute(self, attrid, value):
+        if attrid == self.CURRENT_POSITION_LIFT_PERCENTAGE:
+            value = 100 - value
+        super()._update_attribute(attrid, value)
+
+    async def command(
+        self, command_id, *args, manufacturer=None, expect_reply=True, tsn=None
+    ):
```
javicalle, line 43: Add the `**kwargs` to the method:
```python
    async def command(
        self, command_id, *args, manufacturer=None, expect_reply=True, tsn=None, **kwargs
    ):
```

* https://github.com/zigpy/zigpy/blob/be5b2914a12ac6a19af80c5a83dab40d7957aa01/zigpy/quirks/__init__.py#L163-L172

zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+            percent = 100 - percent
+            v = (percent,)
+            return await super().command(command_id, *v)
+        return await super().command(
+            command_id,
+            *args,
+            manufacturer=manufacturer,
+            expect_reply=expect_reply,
+            tsn=tsn
+        )
```
javicalle, line 58: Same here:
```python
        return await super().command(
            command_id,
            *args,
            manufacturer=manufacturer,
            expect_reply=expect_reply,
            tsn=tsn,
            **kwargs,
        )
```

javicalle (changes requested):

zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+
+
+class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
+    """WindowCovering cluster implementation.
+
+    This implementation inverts the reported covering percent for non standard
+    devices that don't follow the reporting spec.
```
TheJulianJES, line 29: I guess this comment should be changed to only mention the inverted commands, as the reporting seems to be fine(?)

zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+    cluster_id = WindowCovering.cluster_id
+    CMD_GO_UP = 0x0001  # reported as down_close in ZHA commands
+    CMD_GO_DOWN = 0x0000  # reported as up_open in ZHA commands
```
TheJulianJES, line 34: Hmm, per ZCL the commands are the other way around (like mentioned in the comments): [WindowCovering cluster zigpy link](https://github.com/zigpy/zigpy/blob/be5b2914a12ac6a19af80c5a83dab40d7957aa01/zigpy/zcl/clusters/closures.py#L684-L685)
Maybe it makes more sense to name them to what they should be here according to ZCL?

(and not really important, but I think you could also use ``WindowCovering.commands_by_name["up_open"].id`` and same for "down_close")

zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
+    """WindowCovering cluster implementation.
+
+    This implementation inverts the reported covering percent for non standard
+    devices that don't follow the reporting spec.
+    """
+
+    cluster_id = WindowCovering.cluster_id
```
TheJulianJES, line 32: I think this line can be removed, as the ``InvertedWindowCoveringCluster`` class is already inheriting from ``WindowCovering``.

zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+        expect_reply=True,
+        tsn=None,
+        **kwargs
+    ):
+        """Override default commands for up and down. they are backwards."""
+
+        if command_id == self.CMD_GO_UP:
+            return await super().command(self.CMD_GO_DOWN, *args)
+        if command_id == self.CMD_GO_DOWN:
+            return await super().command(self.CMD_GO_UP, *args)
```
TheJulianJES, line 50: I guess it already works like that, but do we want to keep the rest of the parameters?
For example:
```python
return await super().command(
    self.CMD_GO_UP,
    *args,
    manufacturer=manufacturer,
    expect_reply=expect_reply,
    tsn=tsn,
    **kwargs,
)
```

or to avoid copying that three times, maybe something like this could be done to just override the ``command_id``:
```python
if command_id == self.CMD_GO_UP:
    command_id = self.CMD_GO_DOWN
elif command_id == self.CMD_GO_DOWN:
    command_id = self.CMD_GO_UP

return await super().command(
    command_id,
    *args,
    manufacturer=manufacturer,
    expect_reply=expect_reply,
    tsn=tsn,
    **kwargs,
)
```
I think that might be a bit better/cleaner

zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+
+
+class WM25LBlinds(CustomDevice):
+    """Custom device representing Smartwings WM25LZ blinds."""
+
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=514
+        # device_version=1
+        # input_clusters=[0, 1, 3, 4, 5, 32, 258]
```
TheJulianJES, line 67: I think the ``32`` should be removed from this comment, as there's no ``PollControl`` cluster present.
```suggestion
        # input_clusters=[0, 1, 3, 4, 5, 258]
```

zhaquirks/smartwings/wm25lz.py
tube0013: Do you mean changing this to:
```
up_open = WindowCovering.commands_by_name["up_down"].id
down_close = WindowCovering.commands_by_name["up_open"].id
```

TheJulianJES: Yeah. It's not really needed, but might be nice not to hard-code cluster attributes/commands further.
I'd still keep them as constants though (as they're each referenced to twice in the code)
```python
CMD_GO_UP = WindowCovering.commands_by_name["up_down"].id
CMD_GO_DOWN = WindowCovering.commands_by_name["up_open"].id
```
There, the constants would be named after ZCL/so what ZHA reports.

tube0013:
Hopefully good to go.

zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+
+class InvertedWindowCoveringCluster(CustomCluster, WindowCovering):
+    """WindowCovering cluster implementation.
+
+    This implementation inverts the reported covering percent for non standard
+    devices that don't follow the reporting spec.
+    """
+
+    CMD_GO_UP = WindowCovering.commands_by_name["down_close"].id
+    CMD_GO_DOWN = WindowCovering.commands_by_name["up_open"].id
```
TheJulianJES, line 33: Looking from the ZCL side, I'd have probably named the variables `CMD_GO_UP` = `up_open` (and vice versa).
But I guess yours is fine too when looking from the "blind side".

TheJulianJES (approved):

zhaquirks/smartwings/wm25lz.py
```diff
@@ -0,0 +1,105 @@
...
+        expect_reply=True,
+        tsn=None,
+        **kwargs
+    ):
+        """Override default commands for up and down. they are backwards."""
+
+        if command_id == self.CMD_GO_UP:
+            command_id = self.CMD_GO_DOWN
+        elif command_id == self.CMD_GO_DOWN:
+            command_id = self.CMD_GO_UP
```
javicalle, line 49: That code would be a little tricky:
```python
        if command_id == self.CMD_GO_UP:      # "down_close"
            command_id = self.CMD_GO_DOWN     # "up_open"
        elif command_id == self.CMD_GO_DOWN:  # "up_open"
            command_id = self.CMD_GO_UP       # "down_close"
```

javicalle (approved):
If that works it LGTM

Probaby that should be a generic class class with inverted controls in `zhaquirks`, but that is another issue.

tube0013:
thanks @javicalle 

I've reached out to them before about group support (as they don't seem to like groups or scenes) but not the direction.

I currently have 2 shades from last year. I will try their work around.

I have another 5 on their way which sound like they will have the updated firmware.

javicalle:
I doubt I can achieve it (even if it's feasible) but in HA the blinds should have a switch to reverse the operation mode and probably another to activate the 'maintenance mode'.
The problem is that I don't know if there is a standard Zigbee command/procedure or if each manufacturer implements it in their own way.

tube0013:
the odd thing about these devices is, this is what the window_covering_mode attribute reads on default settings:

```
WindowCoveringMode.16|Motor_in_maintenance_mode
```

And changing this to reverse motor etc, has no affect on the motion. I'lm curious to see what the remote work around does, it seems very similar to the method to reverse the motor as documented, with holding the 2 buttons a bit longer and hitting the back p2 button an additional time. 

The remote is not zigbee, it is what ever their "proprietary" RF signal is.

javicalle:
I have just reviewed the linked issue and I see that that path has been already walked so I don't expect that remote procedure could fix anything.

tube0013:
> I have just reviewed the linked issue and I see that that path has been already walked so I don't expect that remote procedure could fix anything.

yeah I couldn't get anything to change with the remote. things are going to get complicated if the new fw works and has the same signature.

javicalle:
I was thinking the same but you already have some of the new ones, isn't?

tube0013:
New ones are being delivered today or tomorrow.

I pushed their support a little, and they have offered to send me new zigbee modules. I still pushed back on getting an OTA file. As I asked if it was different hardware.

`We have update the software  without changing the hardware. The zigbee update need a hub. Sorry.We did not develop it. `

MattWestb:
Do you knowing witch chip they is using ?
If Silabs you can flashing then with commander over SWD.

Also if they having implanted working OTA in the old firmware it shall working OK with ZHA and local OTA files if they not have making strange things.

javicalle:
> I still pushed back on getting an OTA file

I have also asked for that file but still not answered me.
Not sure if the problem would be the OTA file or if the device implements it. I'll update if get some feedback.

tube0013:
Pulled the zigbee module out on one of the new ones.

image

image

MattWestb:
MG21 :-)))
I hope they have not locking the chip down !!

TheJulianJES:
Is it already clear if the new devices also need their command inverted?

tube0013:
just joined one with no quirk enabled and it's behaving the same., and quirk seems to work as with the others.

The battery reading still seems to be off by 1/2, as a fully charged one is showing 50%, I though I had this working, am I missing something there?

I have 4 more to try lol

TheJulianJES:
> The battery reading still seems to be off by 1/2, as a fully charged one is showing 50%

That's weird. `DoublingPowerConfigurationCluster` seems to be used. Just to make sure, can you remove the quirk and then read the `BATTERY_PERCENTAGE_REMAINING = 0x0021` attribute on the `PowerConfiguration` cluster.
(If it's 25, the quirk worked correctly. If it's also 50 without the quirk, it didn't work.)

tube0013:
with no quirk it that attribute is reading 100:

image


Battery still shows 50% even after reading that attribute.

image

TheJulianJES:
Ok, so that's correct at least. (ZCL `battery_percentage_remaining is of 200%)
And what happens if you read the attribute with the quirk?

tube0013:
get 100 also, but this time the battery updated to 100. so quirk is working I guess I've been going back and forth from quirk to no quirk so much reports were not coming in frequently enough so the no quirked value was showing.

javicalle:
They confirm that it is not possible to update via OTA the old devices:
```
I checked with our R&D department and it is impossible to update the firmware version.
We are so sorry for our old customers can not solve this issue.
We don't know why this issues did not happened on the SmartThings Platform.
The link of firmware for you check:
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-window-shade-battery.src/zigbee-window-shade-battery.groovy
```
I don't think it's worth explaining to them that SmartThings do the same thing we do: reverse the commands for their blinds.

They have also offered to send me a device (without specifying which one). I don't think we can get more technical information from them, but if someone finds it interesting to have the device to sniff traffic, they can be asked.

tube0013:
What I'm confused about is the new blind I tested last night theoretically should have the newer module/fw but it doesn't.

I have an Amazon Echo out for delivery, I'm going to pair one to it for a day or 2 and sniff the traffic to see if there is anything interesting. they ship a amazon set up guide with the blinds, but they also don't have the WWAH cluster so my understanding is they should work by the spec for amazon support.

TheJulianJES:
Added a few small changes (`tries` parameter, typing, comments, changed variable names to what ZCL commands means).

I'm really curious to see if the Amazon Echo *just works* with it.
But since the new shades still have this issue, I feel like we can merge this safely.

tube0013:
appreciate it. so I dropped the echo off at UPS store for a refund today. after messing with it last night it was clear no FW update would come and the overall experience was rather poor as far as control, only command specifying a percentage worked.

TheJulianJES:
@javicalle Wanna have another look to see if everything is still good?

javicalle (approved):
LGTM.

Probably `InvertedWindowCoveringCluster` is candidate to be a generic zhaquirk class (like `MotionWithReset`, `MotionOnEvent`, ...).

TheJulianJES:
Yeah, if another quirks need this, we can pull it out (and change the test).

tube0013:
Thanks @TheJulianJES !

tube0013:
commenting here to close the loop on this. New modules showed up today. I was able to dump the mainflash. with puddly's help and discovered these are tuya modules. 

image

The appear to be exact same FW, behave the same, and have the same current file version under OTA: 2 

I guess can be lucky that these generally work with standard commands, but limited with no real group and scene support.

tube0013:
It's an silabs chip. I did dump the mainflash region to file but did not include the userdata. I'll open a discussion later to not add on to this PR any more. I did want to pull an existing module and dump that to file to see if the hashes matched as a way to absolutely confirm it's the same fw.

The module has a four wire rj jack and connects to a wire from the the motor. So it's using 2 wires for control. The motor works alone with a 433.92 remote and has a physical button on it for control/setup. SmartWings sells a variety of blinds with options for zigbee/z-wave/Wi-Fi/HomeKit connections so I am sure they just swap this module. The question I have is does the motor run off a tuya mcu or is it custom and they just use a tuya module to interface with it which I think is possible. 🤷🏼‍♂️

MattWestb:
IKEA blind is using one normal comport and 9600 as baud rate and i think many other is doing the same or doing as tuya using there MCU / DP commands for talking to it.

For knowing if its using "normal" commands or tuya MCU we need sniffing the comport but i think its very likely is do and can easy changing the radio to WiFi, BT, Z-awe and so on.
But normally tuya devices is using MCU tunneling Zigbee firmware but is also possible using one MCU and the Zigbee module is sending MCU commands to it converted to / from ZCL commands so we dont need do the hacking.

If the motor is running tuya MCU it can being possible connecting one tuya Zigbee tunneling module and its reading the MCU ID and using it as Zigbee manufacture (have getting one dump from one tuya TRV with user data and tuya Zigbee tunneling module is having one factory manufacture that is using until have getting the real manufacture name  from the MCU and start using that instead).

MattWestb:
From one CLI with commander for Billy its 
```
 .\commander.exe readmem --range 0xfe00000:0xfe00800 -d EFR32MG1P233F256GM48 -o E2022UD.s37
```
You only need changing the device type and output format if you like have it in Bin, hex and so on (bin is easy looking in with hex editor but have no CRC check) by changing the out file extension.
The range is the same for MG1 and 2 devices for user data so it is working OK and i have dumping the IKEA MG 21 modules with it :-))

Segger Jlink Flasher is also working OK but its depends what you normally using.

Issue #2225, opened 2023-02-25
[BUG] Since 2023.2.0, Linkind motion sensors reset during motion


jasongabler:
**Describe the bug**
My Linkind PIR Zigbee motion sensors stopped working as expected since 2023.2.0. HA will no longer recognize continued motion from these sensors and seems generally confused until I at least wait for the sensor to go idle. I have concluded that the sensors are not at fault because, when I down grade to 2023.1.7, they perform as expected.  I use ZHA with a Conbee II. I have a 2 year old Zigbee mesh that has worked flawlessly until now.

It was recommended I post here from the following official FB group post:  https://www.facebook.com/groups/HomeAssistant/permalink/3449772355294110/
Note that the post comments have a number of concerns about ZHA since HA 2023.2.0 was released.

Using a simple, native HA automation that has worked flawlessly (Lights on with motion; Lights off after 2min of idle)
1. Walk into sensor zone, lights turn on
2. Keep moving in sensors zone for over 2min, lights will turn off. I've tested this up to 4 minutes
3. Leave zone for 1min, walking back into zone and it _may_ turn on. How long I have to wait for HA to recognize idle seems to vary

Again, this not a matter of poor automation configuration. The same, exact automations are in pay for 2023.1.7 and 2023.2.0. I also watch the motion entity and it fails to detect idle or motion according to the above experience with the automation.

**Expected behavior**
* Walk into idle zone with lights off,  have entity UI show motion and lights go on
* Remain in motion within zone and entity should continue to show motion and lights should stay on indefinitely, until 2min after motion stops.

**Screenshots**
N/A

<details>
I don't really have much debug details because I really don't even know how to begin debugging this. I will update with more information as I continue to test and investigate.

<summary>Device signature</summary>
The only Zigbee related info I see on the device is as follows:
Zigbee info
IEEE: cc:86:ec:ff:fe:cd:54:35
Nwk: 0x4e5b
Device Type: EndDevice
LQI: 127
RSSI: -78
Last Seen: 2023-02-25T15:04:06
Power Source: Battery or Unknown

```yaml
Paste the device signature here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
Paste the diagnostic information here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>
<summary>Additional logs</summary>

```
Paste any additional debug logs here.
[...]

PR #2231, opened 2023-02-26
Remove updating `ZONE_STATE` on motion for `MotionWithReset` cluster

TheJulianJES:
It looks like ``ZONE_STATE`` (0x00) and ``ZONE_STATUS`` (0x02) were swapped for the "MotionWithReset" quirks.
This didn't really make a difference, as the "Zone Status Change Notification" command was still properly sent (0x00) and ZHA used that to update the state of the binary sensor.

However, it seems like updating the ``ZONE_STATE`` attribute is still wrong.
Not sure when this is actually called (if at all), but [here](https://github.com/home-assistant/core/blob/0f018665080c6eaf53153b5794fad5e8b9d4a046/homeassistant/components/zha/binary_sensor.py#L170-L175), ZHA checks (polls) for ``ZONE_STATUS``. For devices which use the "MotionWithReset" stuff, this could produce wrong results.
Furthermore, the ``SENSOR_ATTR`` of that class is also ``ZONE_STATUS`` (and not ``ZONE_STATE``).

Updating ``ZONE_STATE`` is now completely removed, as sending the "zone status update command" is enough and according to ZCL/spec, the ``ZONE_STATUS`` command doesn't support attribute reporting.
The attribute cache not being updated when ZHA gets the command will be fixed with: https://github.com/home-assistant/core/pull/89481 

For information, see: https://github.com/zigpy/zha-device-handlers/issues/2225#issuecomment-1445453807
Also see: https://github.com/home-assistant/core/pull/89481 (not required, but will properly update attribute cache when ZHA gets a "zone status change command")

A new constant was also introduced for the "Zone Status Change Notification" command id. This happens to be the same id that the ``ZONE_STATE`` attribute uses, but sharing a constant for two different things probably isn't ideal.
We could probably also get the id for ``ZONE_STATUS`` and the "Zone Status Change Notification" command via name from zigpy. (Although zha-quirks rarely seems to do this anywhere?)

(See comment diff separately for an easier look on both changes)

cc @javicalle


zhaquirks/const.py
```diff
@@ -110,6 +110,7 @@
 UNKNOWN = "Unknown"
 VALUE = "value"
 ZHA_SEND_EVENT = "zha_send_event"
+ZONE_STATUS_CHANGE_COMMAND = 0
```
javicalle, line 113: Despite being the same, I would prefer the `0x00` value from the command declaration:
https://github.com/zigpy/zigpy/blob/be5b2914a12ac6a19af80c5a83dab40d7957aa01/zigpy/zcl/clusters/security.py#L103

zhaquirks/__init__.py
```diff
@@ -242,8 +243,10 @@ def __init__(self, *args, **kwargs):
     def _turn_off(self):
         self._timer_handle = None
         _LOGGER.debug("%s - Resetting motion sensor", self.endpoint.device.ieee)
-        self.listener_event(CLUSTER_COMMAND, 253, ZONE_STATE, [OFF, 0, 0, 0])
-        self._update_attribute(ZONE_STATE, OFF)
+        self.listener_event(
+            CLUSTER_COMMAND, 253, ZONE_STATUS_CHANGE_COMMAND, [OFF, 0, 0, 0]
```
javicalle, line 247: That `253` magic number seems to be the tsn value:
https://github.com/zigpy/zigpy/blob/be5b2914a12ac6a19af80c5a83dab40d7957aa01/zigpy/zcl/__init__.py#L374

I believe that we can use `None` here.

zhaquirks/__init__.py
```diff
@@ -284,8 +287,10 @@ def __init__(self, *args, **kwargs):
 
     def motion_event(self):
         """Motion event."""
-        super().listener_event(CLUSTER_COMMAND, 254, ZONE_STATE, [ON, 0, 0, 0])
-        self._update_attribute(ZONE_STATE, ON)
+        super().listener_event(
+            CLUSTER_COMMAND, 254, ZONE_STATUS_CHANGE_COMMAND, [ON, 0, 0, 0]
```
javicalle, line 291: Same that above with the `253`

zhaquirks/__init__.py
```diff
@@ -264,7 +267,7 @@ def handle_cluster_request(
         ] = None,
     ):
         """Handle the cluster command."""
-        if hdr.command_id == ZONE_STATE:
+        if hdr.command_id == ZONE_STATUS_CHANGE_COMMAND:
             if self._timer_handle:
```
javicalle, line 271: Can we at least add some log here?
```python
            _LOGGER.debug("ZONE_STATUS_CHANGE_COMMAND with args: %s", args)
```

javicalle (commented):
My thoughts

zhaquirks/__init__.py
```diff
@@ -264,7 +267,7 @@ def handle_cluster_request(
...
                 self._timer_handle.cancel()
             self._timer_handle = self._loop.call_later(self.reset_s, self._turn_off)
```
javicalle, line 273: That must only be performed if the `zone_status` is one of `Alarm_1` or `Alarm_2`. For any other values, the alarm status must be reset. 
(I'm thinking in a `Tamper` or `Battery` status)

zhaquirks/__init__.py
TheJulianJES: > That 253 magic number seems to be the tsn value

Yeah, not sure why ``253``/``254`` are used here in the first place? Seems to be unused. They are basically the upper limit though, so maybe there's a reason?

I'm not sure if we really want to use `None` though, as that's invalid and might break future stuff that (for some reason) uses the TSN?

zhaquirks/const.py
TheJulianJES: Used `0x0000` now (https://github.com/zigpy/zha-device-handlers/pull/2231/commits/d0861d8fa7e899d4f0f58fbc2b62e75ef9a750e7) to fit the style of `ZONE_TYPE`/`ZONE_STATUS`, as they're both 4 digits.

zhaquirks/__init__.py
TheJulianJES: Not sure I've ever had a device that actually uses the other ones, but you're right. See https://github.com/zigpy/zha-device-handlers/pull/2231/commits/7ca79f64b6fa07010849d2bfb4cc3bdc7d92cf44

zhaquirks/__init__.py
TheJulianJES: See https://github.com/zigpy/zha-device-handlers/pull/2231/commits/c16e581404eb1204d9ead1abe788ea5e10a06273
This prints something ``Motion detected: ZoneStatus.Alarm_1``.

I don't think we need to print the complete command for this:
```python
status_change_notification(zone_status=<ZoneStatus.Alarm_1: 1>, extended_status=<bitmap8.0: 0>, zone_id=0, delay=0)
```

TheJulianJES:
Tested the changes with an Aqara P1 sensor (which uses ``MotionOnEvent`` from the occupancy stuff) and a Konke motion sensor (which uses ``MotionWithReset``) and they seem to work as expected.

zhaquirks/__init__.py
TheJulianJES: Actually, HA already logs something, so this seems to somewhat duplicate it:

```python
2023-02-27 00:04:53.504 DEBUG (MainThread) [zigpy.zcl] [0xF31E:1:0x0500] <IEEE> - Motion detected: ZoneStatus.Alarm_1
2023-02-27 00:04:53.505 DEBUG (MainThread) [homeassistant.components.zha.core.channels.base] [0xF31E:1:0x0500]: Updated alarm state: ZoneStatus.Alarm_1
```

javicalle: It could be interesting to know about other values but inside of `if hdr.command_id == ZONE_STATUS_CHANGE_COMMAND and args[0] & 3:` we can't see it.
I would suggest move it outside the 'if' or remove it.
I haven't checked if ZHA would log other values than alarm1 & 2.

TheJulianJES: ZHA should log everything all possible alarm states: https://github.com/home-assistant/core/blob/0f018665080c6eaf53153b5794fad5e8b9d4a046/homeassistant/components/zha/core/channels/security.py#L344-L349

javicalle: Thanks for checking (I'm on the phone now)
Then I believe that can be removed.

javicalle:
No devices to testing myself but changes LGTM. 
Thanks Julian.

javicalle:
BTW bo you know if the devices with the reset quirk are reporting any 'clear' state? (I guess that I already know the answer)

zhaquirks/__init__.py
TheJulianJES: Hmm, never mind. ZHA does *not* log all possible alarm states (sets `state = args[0] & 3` first and then only prints `state`).
That code seems to somewhat override the other bits then... But I guess it's fine because ZHA doesn't do anything with the other bits (tamper, battery, ...)?

TheJulianJES:
> BTW bo you know if the devices with the reset quirk are reporting any 'clear' state?

At least the Konke sensors don't seem to do that (ever).

zhaquirks/__init__.py
MattWestb: Implanting sensor for tamper is asked for but its out of scope for this issue.

TheJulianJES: I've removed the log message for now. It would only log for sensors that use this quirk.
If other bits of that `zone_status` are needed, ZHA (HA Core) should log it.

zhaquirks/__init__.py
TheJulianJES: I'd probably leave the TSN values like this now, but @dmulcahey, do you know why those values were chosen for sending the "zone status change notification" command (from quirks to ZHA)?

javicalle: Just for future reference, waxman is using `None` as `tsn` (and `ZONE_STATE`):
https://github.com/zigpy/zha-device-handlers/blob/dda8d8a0e99692813f34d7e0518eecfdd7a02ece/zhaquirks/waxman/leaksmart.py#L56-L58

TheJulianJES:
Another thing to consider is if we even need/want to send the command (or just update the ``ZONE_STATUS`` attribute only):

[``HA-Core/IasZone/cluster_command``](https://github.com/home-assistant/core/blob/1d9e8c873fb31d27f392a8a3eda5f72886220256/homeassistant/components/zha/core/channels/security.py#L389-L399) and [``HA-Core/IasZone/attribute_updated``](https://github.com/home-assistant/core/blob/1d9e8c873fb31d27f392a8a3eda5f72886220256/homeassistant/components/zha/core/channels/security.py#L341-L349) both do the same really.

javicalle:
> Another thing to consider is if we even need/want to send the command (or just update the `ZONE_STATUS` attribute only)

That is the `CLUSTER_COMMAND` part, isn't it?:
```python
        self.listener_event(
            CLUSTER_COMMAND, 253, ZONE_STATUS_CHANGE_COMMAND, [ON/OFF, 0, 0, 0]
        )
```

My only concern would be if the command is triggering ZHA events and the `_update_attribute` isn't. To me would be fine with just the `_update_attribute` part but maybe is a breaking change somewhere 😅

TheJulianJES:
https://github.com/home-assistant/core/pull/89481 is now approved.

With that PR, it's now enough to just either update the `ZONE_STATUS` attribute directly OR send the cluster command. Both effectively do the same now.
Per ZCL/spec, the attributes aren't reportable, so we should probably just send the command (ZHA then updates attribute cache).

TheJulianJES:
Updating ``ZONE_STATE`` is now completely removed, as sending the "zone status update command" is enough and according to ZCL/spec, the ``ZONE_STATUS`` command doesn't support attribute reporting.

So what this PR now changes is:
- no longer updating a wrong attribute on motion (`ZONE_STATE`) (which was unused/ignored previously anyway)
- introducing a new constant for the "zone status change command" instead of abusing the `ZONE_STATE` constant

The attribute cache not being updated when ZHA gets the command will be fixed with:
- https://github.com/home-assistant/core/pull/89481 (which can be merged independently of this PR/quirks release)

zhaquirks/const.py
```diff
@@ -118,6 +118,7 @@
 UNKNOWN = "Unknown"
 VALUE = "value"
 ZHA_SEND_EVENT = "zha_send_event"
-ZONE_STATE = 0
+ZONE_STATUS_CHANGE_COMMAND = 0x0000
```
javicalle, line 121: Maybe `0x00`?
https://github.com/zigpy/zigpy/blob/1f81b162f84328eb96db18add3104d889fa6c923/zigpy/zcl/clusters/security.py#L104

javicalle (approved):
LGTM

zhaquirks/const.py
TheJulianJES: Other constants are also four digits:
```python
ZONE_STATE = 0x0000
ZONE_TYPE = 0x0001
ZONE_STATUS = 0x0002
```
Doesn't really matter. We should migrate everything to grab the IDs from `zigpy` though probably.

javicalle: These others are cluster attributes that are defined this way (4 digits):
* https://github.com/zigpy/zigpy/blob/1f81b162f84328eb96db18add3104d889fa6c923/zigpy/zcl/clusters/security.py#L69-L75

While cluster commands are 2 digits.

javicalle: And yes, it really doesn't matter. (sorry to be picky)

PR #2232, opened 2023-02-27
Adding missing TOGGLE = "Toggle"

MattWestb:
Command  `COMMAND_TOGGLE` is implanted but not one toggle button that making the command then pressing it.

I need it for Symfonsk 2 or must we implanting `Play`, `Next` and `Privies ` instead of Toggle, Right and Left button. pressed.

Device request with picture of the device  https://github.com/zigpy/zha-device-handlers/issues/2223

### DIFF
--- a/zhaquirks/const.py
+++ b/zhaquirks/const.py
@@ -104,6 +104,7 @@ SHORT_PRESS = "remote_button_short_press"
 ALT_SHORT_PRESS = "remote_button_alt_short_press"
 SKIP_CONFIGURATION = SIG_SKIP_CONFIG
 SHORT_RELEASE = "remote_button_short_release"
+TOGGLE = "Toggle"
 TRIPLE_PRESS = "remote_button_triple_press"
 TURN_OFF = "turn_off"
 TURN_ON = "turn_on"


javicalle:
Ummm, I'm not sure what the need is here.

A toggle command already exists in the constants:
https://github.com/zigpy/zha-device-handlers/blob/f6ab87e1c48ccb6db262e8732ca6b63d67fb6252/zhaquirks/const.py#L63

And as a button I would expect to work with the press events (short, long, double, ...).
And map the events to the related commands (COMMAND_TOGGLE, or whatever...)

Does it make sense to you?

MattWestb:
The normal DA looks like this for IKEA 5 button:
```
        (SHORT_PRESS, TURN_ON): {COMMAND: COMMAND_TOGGLE, CLUSTER_ID: 6, ENDPOINT_ID: 1},
```
And like having "short press toggle button" from the DA.
The button is not one ON button if you look on the picture in the issue.

For devices with one on and one off button its working OK but toggle button is not one on or off button.

javicalle:
I see.
IMHO from a user point of view I think it would be more intuitive the `short press play|next|previous button` but I must admit that I'm not familiar with all the DA aspects/implementations, so it's not a very elaborate opinion either 🤷🏻‍♂️

MattWestb:
I agree with that shall being best. The device is sending light commands and can controlling them direly with normal light commands as all IKEA controllers is doing.
So i finding it bitter using toggle then its the same function in IKEA I Sonnos app Play / Pause is one togging function 2.

The 2 extra buttons is sending manufacture specific commands and i shall using button_1 and 2 and shirt,  double and long press on then = very logic.

MattWestb:
If toggle is not OK then i think (Z)HA shall implanting `0x0506 Media Playback` and command `0x00` to `0x0b` and we is having all OK.

HA shall have or getting the command then its one part of the Matter Application Cluster
Specification Version 1.0. 
6.10. Media Playback Cluster
```
6.10.4. Commands
Id Name Direction Response Access Conformance
0x00 Play Client ⇒ Server PlaybackResponse
O M
0x01 Pause Client ⇒ Server PlaybackResponse
O M
0x02 Stop Client ⇒ Server PlaybackResponse
O M
0x03 StartOver Client ⇒ Server PlaybackResponse
O O
0x04 Previous Client ⇒ Server PlaybackResponse
O O
0x05 Next Client ⇒ Server PlaybackResponse
O O
0x06 Rewind Client ⇒ Server PlaybackResponse
O VS
0x07 FastForward Client ⇒ Server PlaybackResponse
O VS
0x08 SkipForward Client ⇒ Server PlaybackResponse
O O
0x09 SkipBackward Client ⇒ Server PlaybackResponse
O O
0x0a PlaybackResponse
Server ⇒ Client N M
0x0b Seek Client ⇒ Server PlaybackResponse
O AS
```
The Symfononisk is very likely made so IKEA can "piping" the cluster from it to Matter and dont need converting commands then Dirigera is getting Matter support.

PR #2234, opened 2023-02-27
Changing battery to ZVL standard.

MattWestb:
The remote is using ZCL standard as all other 23.X firmware controllers but have one bug that is making it sending 255 = not known or not valid reading.  Its is better getting 127% battery as 255% and hope IKEA is fixing the bug sooner or later.


TheJulianJES:
The signature for `IkeaTradfriRemoteV2` only matches the new Styrbar v2.4.5 firmware, right?

MattWestb:
Last added https://github.com/zigpy/zha-device-handlers/pull/2028 and only Version 2.4.5.
I was not changing the battery then was not knowing IKEA is changing to ZVL standard and that the firmware have the bug.
Z2M is having the same problems with it.

TheJulianJES (approved):

PR #2235, opened 2023-02-27
Merge Aqara Mini button quirks

TheJulianJES:
Somewhat of a follow-up to https://github.com/zigpy/zha-device-handlers/pull/2215

This merges two quirks for the Aqara Mini button.
The device automation triggers are now also removed from the "known duplicate triggers", as they're removed/in one class now).

Untested, but it should work, as the merged quirks were basically identical.


javicalle (approved):
LGTM

Issue #2243, opened 2023-03-02
[BUG] Tuya ZHA quirk: data type missing


SanchYESS:
https://github.com/zigpy/zha-device-handlers/blob/dda8d8a0e99692813f34d7e0518eecfdd7a02ece/zhaquirks/tuya/__init__.py#L252

Got an error message after updating core from 2023.2.5 to 2023.3.0:
```
Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/zigpy_znp/zigbee/application.py", line 592, in on_af_message
    self.packet_received(
  File "/usr/local/lib/python3.10/site-packages/zigpy/application.py", line 875, in packet_received
    self.handle_message(
  File "/usr/local/lib/python3.10/site-packages/zigpy/application.py", line 381, in handle_message
    return sender.handle_message(
  File "/usr/local/lib/python3.10/site-packages/zigpy/device.py", line 370, in handle_message
    self.endpoints[src_ep].handle_message(
  File "/usr/local/lib/python3.10/site-packages/zigpy/endpoint.py", line 224, in handle_message
    handler(hdr, args, dst_addressing=dst_addressing)
  File "/usr/local/lib/python3.10/site-packages/zigpy/zcl/__init__.py", line 373, in handle_message
    self.handle_cluster_request(hdr, args, dst_addressing=dst_addressing)
  File "/usr/local/lib/python3.10/site-packages/zhaquirks/tuya/__init__.py", line 485, in handle_cluster_request
    zvalue = ztype(tuya_data)
  File "/usr/local/lib/python3.10/site-packages/zigpy/types/basic.py", line 93, in __new__
    n = super().__new__(cls, *args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/zhaquirks/tuya/__init__.py", line 253, in __int__
    return ints[self[0]].deserialize(bytes(reversed(self[1:])))[0]
KeyError: 8
```
Seems like my power meter (_TZE200_lsanae15, [custom quirk](https://github.com/zigpy/zha-device-handlers/issues/1768)) sends 8 bytes of data but its type is missing in the __int__() function.
I guess it must be something like `ints = {1: t.int8s, 2: t.int16s, 3: t.int24s, 4: t.int32s, 8: t.int64s}` at least and maybe with other types as well.

javicalle:
@SanchYESS could you attach a log from that kind of messages, please?
[...]

PR #2244, opened 2023-03-02
Enable `ints` up to 8 bytes

javicalle:
Update the `Data` `__int__` to enable up to 8 bytes

Fixes: #2243


javicalle:
@Shulyaka could you check please?

javicalle:
Thanks Julian!

Issue #2264, opened 2023-03-09
[BUG] TS0501 Fan Controller changes behaviour once paired


Brazen00:
**Describe the bug**
I have installed 7x Mercator Ikuu TS0501 Fan Controllers in my home. Before they are paired, the physical "fan speed" button on the controller will cycle the fan between low, medium and high speeds. Once paired to ZHA, the physical "fan speed" button on the controller will only cycle between low and high speeds, skipping the medium speed. If I unpair the controller, the default behaviour returns. On Home Assistant itself, once paired I can change the fan speed between high, medium and low. The incorrect behaviour only occurs on the physical button.

The fix is likely to be to set ```fan_mode_sequence``` to ```FanModeSequence.Low_Med_High``` in a quirk.

This was previously raised [here](https://github.com/zigpy/zha-device-handlers/issues/1577), and was fixed in Z2M [here](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5454).

**Expected behavior**
I expect the physical button to cycle through the 3 speeds.

**Screenshots**

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0202",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZ3210_lzqq3u4r",
[...]

PR #2265, opened 2023-03-09
Add Tuya TS0501 fan controller quirk

TheJulianJES:
Fixes https://github.com/zigpy/zha-device-handlers/issues/2264
(Confirmed working in that issue)

This changes the `fan_mode_sequence` to `Low_Med_High` on pairing, so the physical switch still cycles through all modes properly.

(Not sure what use the Tuya cluster has, but had to pick one `CustomCluster` with the Tuya ID, as the Tuya cluster ID is outside of the range designated for manufacturers (so tests checking for a valid cluster range wouldn't pass before).
Is this fine?)


javicalle (approved):
LGTM

PR #2273, opened 2023-03-14
Add IEKA Symfonisk Gen2 v1/v2 support

MattWestb:
Add DA for Symfonisk Gen 2 with shipping firmware (V1) and first OTA updated via Dirigera (V2).

The V1 is using raw commands on EP1.
V2 is using Matter Switch ZCL commands on 2 extra EP but with IKEA manufacture cluster then the Matter cluster is not in ZCL R8. V2 also need the 2 extra EP being bounded and one PR is made in ZHA for doing it nice but this version is using OnOff Custer as one temporary solution until the ZHA PR is merged.

Also V1 is using IKEA old battery reporting (100%) and V2 is using Zigbee standard (200%) so have adding new updated power cluster class and also for upcoming OnOff, OpenCloce and ShortCut button Gen 2 that is in the pipe,
Adding more constants for getting Matter Switch functionality working OK.

Toggle have not hold as left and right and only sending one command. Up and down is sending single and repeating the command then being held.
V1 is only sending ShortCut command single, double and held.
V2 is sending initial press for all commands and then single pressed and release double is sending initial multi press multiple reseal with arg 2 (2 presses) and held is sending held and and hold release.

Fixes: #2223
Depends on: https://github.com/home-assistant/core/pull/89623


MattWestb:
#### Great thanks to @javicalle and @TheJulianJES for making the code part working and nice implemented and our brave users for testing and feed back !!

Next up is is the Gen 2 of OnOff, OpenClose and ShortCut buttons then IKEA is releasing them !!

zhaquirks/ikea/__init__.py
```diff
@@ -73,6 +82,94 @@ class ScenesCluster(CustomCluster, Scenes):
...
+    name = "ShortCutV1EventsCluster"
+    cluster_id = IKEA_SHORTC_CLUSTER_V1_ID
+
+    server_commands = {
+        0x01: foundation.ZCLCommandDef(
+            "ShortCutV1Events",
+            {
+                "Shortcut_Button": t.int8s,
+                "Shortcut_Event": t.int8s,
+            },
```
TheJulianJES, line 97: I think the convention here is to use `snake_case`(?)

zhaquirks/ikea/__init__.py
```diff
@@ -73,6 +82,94 @@ class ScenesCluster(CustomCluster, Scenes):
...
+            False,
+            is_manufacturer_specific=True,
+        ),
+    }
+
+
+class ShortcutV2Cluster(EventableCluster):
+    # class ShortcutV2Cluster(OnOff, EventableCluster):
```
TheJulianJES, line 105: I guess this can be removed (HA PR)

zhaquirks/ikea/__init__.py
```diff
@@ -87,6 +184,34 @@ class PowerConfiguration2AAACluster(DoublingPowerConfigurationCluster):
     }
 
 
+class PowerConfigurationND2AAACluster(PowerConfigurationCluster):
```
TheJulianJES, line 187: Power configuration needs to be cleaned up.

I'll have a look at this later, but I think we could have "no doubling" classes like this as a base and then have the "doubling classes" just inherit from that **and** `DoublingPowerConfigurationCluster`.

zhaquirks/ikea/symfonisk2.py
```diff
@@ -0,0 +1,347 @@
...
+        (SHORT_PRESS, LEFT): {
+            COMMAND: COMMAND_STEP,
+            CLUSTER_ID: 8,
+            ENDPOINT_ID: 1,
+            PARAMS: {"step_mode": 1},
+        },
+    }
+
+
+class IkeaSYMFONISKgen2V1(CustomDevice):
```
TheJulianJES, line 109: I think I'd just call it something like `IkeaSymfoniskGen2v1`

zhaquirks/ikea/symfonisk2.py
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    device_automation_triggers = (
+        Symfonisk2CommonTriggers.device_automation_triggers.copy()
+    )
+    device_automation_triggers.update(
+        {
+            (SHORT_PRESS, BUTTON_1): {
+                COMMAND: COMMAND_SHORTCUTV1,
+                PARAMS: {"Shortcut_Button": 1, "Shortcut_Event": 1},
```
TheJulianJES, line 177: (Note: If the upper/lower-casing changes for the `server_commands`, it also needs to be changed here)

zhaquirks/ikea/symfonisk2.py
```diff
@@ -0,0 +1,347 @@
...
+            },
+            (LONG_PRESS, BUTTON_2): {
+                COMMAND: COMMAND_SHORTCUTV1,
+                PARAMS: {"Shortcut_Button": 2, "Shortcut_Event": 3},
+            },
+        },
+    )
+
+
+class IkeaSYMFONISKgen2V2(CustomDevice):
```
TheJulianJES, line 203: Same here: maybe `IkeaSymfoniskGen2v2`

zhaquirks/ikea/__init__.py
```diff
@@ -73,6 +82,94 @@ class ScenesCluster(CustomCluster, Scenes):
...
+                "PreviousPosition": t.int8s,
+            },
+            False,
+            is_manufacturer_specific=True,
+        ),
+        0x05: foundation.ZCLCommandDef(
+            "multi_press_ongoing",
+            {
+                "NewPosition": t.int8s,
+                #                 "CurrentNumberOfPressesCounted": t.int8s, # Not implanted by IKEA for the moment.
```
TheJulianJES, line 156: I'm guessing this is something Matter-related that IKEA didn't implement?
I think the comment is fine, but remove the extra spaces after `#`

zhaquirks/ikea/__init__.py
```diff
@@ -7,13 +7,22 @@
...
 
 IKEA = "IKEA of Sweden"
-IKEA_CLUSTER_ID = 0xFC7C  # decimal = 64636
 WWAH_CLUSTER_ID = 0xFC57  # decimal = 64599 ('Works with all Hubs' cluster)
+IKEA_CLUSTER_ID = 0xFC7C  # decimal = 64636
+IKEA_SHORTC_CLUSTER_V1_ID = 0xFC7F  # decimal = 64639 Shortcut V1 commands.
+IKEA_M_SWITCH_CLUSTER_ID = (
+    0xFC80  # decimal = 64640 Shortcut V2 / Matter Switch commands.
+)
+COMMAND_SHORTCUTV1 = "ShortCutV1Events"
```
TheJulianJES, line 25: In general, I wouldn't upper-case the `C` in Short**C**ut, so just use `Shortcut`.
However, I think the convention might be to use `snake_case` here(?) Have to double check though

zhaquirks/ikea/__init__.py
```diff
@@ -7,13 +7,22 @@
...
 _LOGGER = logging.getLogger(__name__)
 
 IKEA = "IKEA of Sweden"
-IKEA_CLUSTER_ID = 0xFC7C  # decimal = 64636
 WWAH_CLUSTER_ID = 0xFC57  # decimal = 64599 ('Works with all Hubs' cluster)
+IKEA_CLUSTER_ID = 0xFC7C  # decimal = 64636
+IKEA_SHORTC_CLUSTER_V1_ID = 0xFC7F  # decimal = 64639 Shortcut V1 commands.
+IKEA_M_SWITCH_CLUSTER_ID = (
+    0xFC80  # decimal = 64640 Shortcut V2 / Matter Switch commands.
+)
```
TheJulianJES, line 24: Hmm, it would be best if we could shorten this comment a bit, so black doesn't split it up (and maybe also remove the dots at the end of the two comments)

zhaquirks/ikea/__init__.py
```diff
@@ -73,6 +82,94 @@ class ScenesCluster(CustomCluster, Scenes):
     )
 
 
+class ShortcutV1Cluster(EventableCluster):
+    """Ikea Shortcut Button Cluster Variant 1."""
+
+    name = "ShortCutV1EventsCluster"
```
TheJulianJES, line 88: Is this name used for anything?

(also -> `Shortcut`)
(and does it need to differ from the class name? I think the clusters UI always uses the class name at the moment?)

TheJulianJES (commented):
Added a few notes.
Hopefully I'll be able to look at the power configuration stuff in the coming days. I'm pretty sure that can be cleaned up, so we don't duplicate it everywhere.
(We might also want to use the non-doubling clusters for other IKEA devices then?)

MattWestb:
I have making the no braking things and little more that was failing between the changes with the snaking.
The power part i think its better waiting refactoring and see if IKEA is rolling out the V24.X firmware of TF then its more used and not only some few users that have updating there devices with Dirigera and if they is fixing the not good working one (Styrbar and 255 = not valid reading).

The class naming need being fixed then of 4 named IKEA controllers is 2 different naming used (Starkvind, Shurtcut, Symfonisk) and IKEA is always using upper case for product names (I think its copyright / trademarks thins behind that).

The corde working changes i have not doing then i dont knowing if its working or not and i cant diag it so better you is doing it and i can testing it.

Was finding one more not snaked that have being added in constants:
https://github.com/zigpy/zha-device-handlers/blob/fa60b9d23a3e94b5a164e6aad1f24f4134544daa/zhaquirks/const.py#L62
that i think you shall putting on the tOOdOO lIST.

zhaquirks/ikea/__init__.py
MattWestb: K.a.
I was copy it from the original V1 PR with all its tricks was working OK so i have not changing it.

MattWestb:
Have updating my test system and changing the 2 local import (IKEA and constants) and testing all DA for both versions.

V1 13 DA all working OK = I is getting the right `zha_events` the automatons shall doing.
V2 17 DA all working OK = I is getting the right `zha_events` the automatons shall doing.

zhaquirks/const.py
```diff
@@ -89,6 +96,7 @@
 OUTPUT_CLUSTERS = SIG_EP_OUTPUT
 PARAMS = "params"
 PRESS_TYPE = "press_type"
+PRESSED = "initiali_switch_press"
```
javicalle, line 99: initiali?

MattWestb: Way not **half** _Italiano_ WiTh CaMeLe CaSe ?
Fixed.

MattWestb: Is OK now ?

javicalle: Yes, ILGTM

TheJulianJES:
I've split up the battery clusters part into: https://github.com/zigpy/zha-device-handlers/pull/2284 and tried to make it a bit cleaner, so the constant attributes aren't copy-/pasted.

Note: It's still untested if that even works, but if I'm thinking correctly, *it should*. I'll try to test that in the next few days.

I will rebase this PR once the battery one gets merged, so leave this PR open. (No need for you to do anything)

zhaquirks/ikea/__init__.py
```diff
@@ -87,6 +181,34 @@ class PowerConfiguration2AAACluster(DoublingPowerConfigurationCluster):
     }
 
 
+class PowerConfigurationND2AAACluster(PowerConfigurationCluster):
```
TheJulianJES, line 184: Just noticed whilst working on the [other PR](https://github.com/zigpy/zha-device-handlers/pull/2284):
You're extending the custom [`PowerConfigurationCluster`](https://github.com/zigpy/zha-device-handlers/blob/fa60b9d23a3e94b5a164e6aad1f24f4134544daa/zhaquirks/__init__.py#L195-L229) instead of `PowerConfiguration` here.
IMO, that cluster is named badly, as it doesn't mention what it's actually doing.
It seems like that cluster is used for devices which only report their battery voltage (and not percentage).
All IKEA devices do also report percentage, so I'm guessing that's supposed to be just `PowerConfiguration`?

MattWestb: Its coming from the original implementation and i think the the naming was the same as was doing with the LightLinkCluster.
Change if you like.

TheJulianJES: Wasn't talking about the naming, but there's a difference between `PowerConfiguration` and `PowerConfigurationCluster`.

The previous implementation for IKEA devices always used `DoublingPowerConfigurationCluster` which inherited from the ZCL `PowerConfiguration` (and not the remapping `PowerConfigurationCluster`).

So there's a difference in behavior, as your current code would ignore the battery percentage value, as it's remapped from the battery voltage. I don't think that's wanted for these IKEA devices, as they properly report percentage, right?

MattWestb: I think you is being right then i was copy the ground for the quirk from Styrbar (4btn) remote that was changed from doubling to no doubling and then its is also having the wrong class for the battery.
Can you fixing to the right class after you have doing the large rework of the IKEA INIT ??

I need my first KAFFEE for the day ;-)

TheJulianJES: Yep, will rebase this PR and fix that minor issue once the battery PR is merged.

TheJulianJES:
Alright, I've rebased the PR and made some small changes (that hopefully aren't breaking).
I think everything looks good now. Can you test this "new version"?

MattWestb:
Updating the INIT and the quirk to the latest commit and patching the imports for using the local quirk ones and 4 not working restarts ZHA was starting OK and loading the automatons OK (if getting delay HA is not loading them and cant getting them loaded without restart of HA).

Have testing all V1 and V2 DA and all is working as expected !!

So all is looking OK also battery reporting  then reading it from the device:
V1 `Clusters DoublingPowerConfig2AAACluster (Endpoint id: 1, Id: 0x0001, Type: in)` 100 and in ZHA 100%
V2 `Clusters PowerConfig2AAACluster (Endpoint id: 1, Id: 0x0001, Type: in)` 168 and in ZHA 81%


Thanks for helping not code warier getting the devices supported !

MattWestb:
You can committing it then its working 100% for V1 and all DAs is working for V2 but no events for EP 2 and 3 until ZHA is fixed and its better then not working at all and its not braking thing and user can doing binding with ZHA toolkit if they like until all is merged and having full working devices in there system.

Also one great thanks to @MichelBrodatzki for the initial V1 code that was making its easier doing  the V2 working well and our ZHA devs for helping with the fixing the problems on the road.

MattWestb:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/5583#event-8828863892
I still wondering how was helping them with V2 and the firmware for updating it to V2 ??

But i understand that other have not getting all in place for both versions and was throng it in without understanding how the V1 and 2 is working and dont have any firmware for upgrading to V2 without Dirigera https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6803

MattWestb:
Thanks JES !!

PR #2284, opened 2023-03-20
Add non-doubling IKEA battery info clusters

TheJulianJES:
Originates from: https://github.com/zigpy/zha-device-handlers/pull/2273#pullrequestreview-1346844044
cc @MattWestb 

Newer IKEA devices no longer need a doubling of the reported battery percentage. This adds the new clusters that still set the battery information but no longer double the value.
I've split this up from the the Symfonisk PR, as it's a lot of "changes" (without them actual being changes). So the other PR doesn't get cluttered. I will rebase the Symfonisk PR on these changes when this gets merged (no need for you to do anything MattWestb – if this works as expected).

It also renames all doubling clusters to include `Doubling`.

I've also changed the "updated IKEA five button remote" to use the (normal/compliant/non-doubling) `PowerConfig1CRCluster`: https://github.com/zigpy/zha-device-handlers/commit/68510d03ff711884dcaff43df78709d67dc4f4c0

The changes are easier to see when going through the commits one by one (although there are a lot of commits now).

### DIFF
--- a/zhaquirks/ikea/__init__.py
+++ b/zhaquirks/ikea/__init__.py
@@ -16,10 +16,12 @@ IKEA_CLUSTER_ID = 0xFC7C  # decimal = 64636
 WWAH_CLUSTER_ID = 0xFC57  # decimal = 64599 ('Works with all Hubs' cluster)
 
 # PowerConfiguration cluster attributes
-BATTERY_VOLTAGE = PowerConfiguration.attributes_by_name["battery_voltage"]
-BATTERY_SIZE = PowerConfiguration.attributes_by_name["battery_size"]
-BATTERY_QUANTITY = PowerConfiguration.attributes_by_name["battery_quantity"]
-BATTERY_RATED_VOLTAGE = PowerConfiguration.attributes_by_name["battery_rated_voltage"]
+BATTERY_VOLTAGE = PowerConfiguration.attributes_by_name["battery_voltage"].id
+BATTERY_SIZE = PowerConfiguration.attributes_by_name["battery_size"].id
+BATTERY_QUANTITY = PowerConfiguration.attributes_by_name["battery_quantity"].id
+BATTERY_RATED_VOLTAGE = PowerConfiguration.attributes_by_name[
+    "battery_rated_voltage"
+].id
 
 
 class LightLinkCluster(CustomCluster, LightLink):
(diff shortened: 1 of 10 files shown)


TheJulianJES:
> but then looking on the device with manager zigbee device you can normally saying to one user "select the power configure cluster and read the attribute batter renaming". [...]

The clusters menu is, similar to custom quirks, only an advanced/developer tool. We shouldn't have to make choices about how to name quirk classes depending on how we want to show them in some developer UI.
Also, I think if a user is tasked to select the `PowerConfiguration` cluster and the list only contains `DoublingPowerConfigurationCluster`, they'll select that.
A bunch of other manufacturer quirks (Aqara, Tuya, ...) all have the custom cluster classes named with varying "naming schemes" (often starting with `Tuya` for example).
So as long as the name is clear for people working on `zha-quirks`, it's a good name IMO.

For the "doubling cluster", we should have a clear name (and not just `D` in the name) where you can immediately recognize the "doubling" one vs the normal IKEA power configuration cluster. I think having `Doubling` as a prefix works for that.
(This is to prevent a confusion/mix-up for anyone looking/working at the IKEA quirks, so they can see the different implementations used by the same device with different firmware.)

I'm not sure if `Normal` is the best prefix for the "compliant cluster" though, but it would be better to have a different name for this than what the doubling clusters were previously called (hence I added the `Normal` prefix for now).
This is to prevent not recognizing the updated clusters in outdated or custom quirks.

Also, reading `battery_percentage_remaining` through the clusters UI should show the original value from the device, as we're only updating attribute cache (and not overriding any "read attribute" methods in the quirk).

TheJulianJES:
> Normal is not needed then its using Zigbee standard reporting and only adding the battery info so better leaving it without it in the name

Previously the "doubling clusters" didn't have anything in the name (no "doubling" or whatever). To prevent confusion about these two "cluster types", we should probably have some kind of prefix (normal, compliant, ...) for now.

I'd also keep `Normal`/`Doubling` in the front, as that's the most important thing the clusters do and it's also what we (should) do with all clusters. (I think the Tuya ones all do this? The main `DoublingPowerConfigurationCluster` also does it for example, so we should follow that.)

IMO, `LightLinkCluster` for the IKEA clusters is a bad choice, as it doesn't imply that it's a custom cluster or what that custom cluster does at all. It's hard to come up with a proper name for it though, but it should probably be changed, as it can also easily be confused with a normal `LightLink` cluster.

I've shortened the name a bit now.

TheJulianJES:
Alright, I've made some more changes (also see [this comment](https://github.com/zigpy/zha-device-handlers/pull/2273#discussion_r1145672521) that's now fixed here).
I've dropped the `Normal` prefix too. Let me know if this looks good to you. So far, the changes seem to work as expected.

TheJulianJES:
Yep, tested all the changes and everything seems to work correctly now.

Issue #1448, opened 2022-03-30
[BUG] / Request send default response in wrong direction from quirk.


MattWestb:
**Describe the bug**
Our beloved TS004F Dimmer switch is re sending commands also after sending default response to it.
Its described in this PR https://github.com/zigpy/zha-device-handlers/pull/1437.

**To Reproduce**
Have sniffing tuya ZBGW and they is doing the same 

**Expected behavior**
Default response shall working OK but the device is doing all possible things wrong and this is one firmware bug.
Reading little more i was finding that Smart things is sending default response for EP2-4 like this:
```
Frame Control Field: Profile-wide (0x00)
    .... ..00 = Frame Type: Profile-wide (0x0)
    .... .0.. = Manufacturer Specific: False
    .... 0... = Direction: Client to Server
    ...0 .... = Disable Default Response: False
```
And Zigbee standard shall being like this and tuya ZBGW and ZHA is doing it so:
```
Frame Control Field: Profile-wide (0x18)
    .... ..00 = Frame Type: Profile-wide (0x0)
    .... .0.. = Manufacturer Specific: False
    .... 1... = Direction: Server to Client
    ...1 .... = Disable Default Response: True
```
ST is doing one thing wrong you is **not** sending one default response for one default response but there is sending it server to client and the device is very happy and is not re sending the command and is sending the next without delay.

**Screenshots**

**Additional context**

Z2M look have implanting it and looks working OK but they is sending 4 DR.

I like getting hep implanting sending "revers" DR from quirks (force  Client to Server and perhaps in the future the opposite if needed) so we can getting the device responding for multiple commands without delay (is now 3 seconds between commands).

One funny but not unexpected thing is that the device is triggering reporting battery then pressing the button for EP2 and its start reporting battery on EP1 and getting one normal DR and its OK and then sending the "on" command from EP2 without delay but is being repeated then the device like having it in the reversed direction (Client to Server and not the standard Server to Client).

Sniff of attribute reported the pressing button 2 and reporting battery attribute and the command:
```
EP2 Event:
First its reporting battery from EP1:

ZigBee Network Layer Data, Dst: 0x0000, Src: 0xc29e
ZigBee Application Support Layer Data, Dst Endpt: 1, Src Endpt: 1
ZigBee Cluster Library Frame, Command: Report Attributes, Seq: 73
    Frame Control Field: Profile-wide (0x08)
        .... ..00 = Frame Type: Profile-wide (0x0)
[...]

Issue #2271, opened 2023-03-13
Improve Tuya `EnchantedDevice` implementation


TheJulianJES:
### Description
The implementation for the "Tuya magic spell" is currently implemented in `EnchantedDevice` like this:
https://github.com/zigpy/zha-device-handlers/blob/eed839c4f312a33da0087229a9b2f67185d5d557/zhaquirks/tuya/mcu/__init__.py#L666-L678
Starting to read attribute reads from ``__init__`` isn't ideal, as this theoretically happens every time as ZHA reloads, but [those attribute reads seemingly do not actually make it to the network](https://github.com/zigpy/zha-device-handlers/issues/2227#issuecomment-1465280921).
We also can't really "re-cast" the magic spell (from a reconfiguration for example).

It would be best if this implementation could be improved. One possibility might be to call it from a `bind()` function from some cluster (that's registered as a channel in ZHA, so the function actually gets called). This makes sure it's called "from a proper place" and also re-casts the device when doing a reconfiguration.
Maybe something like a `BasicTuyaSpell` cluster that casts then spell would work?
Edit: See discussion: https://github.com/zigpy/zha-device-handlers/pull/1427
Apparently `bind()` is not called on the Basic cluster ([disabled here](https://github.com/home-assistant/core/blob/459ea048ba12c42f4abe1edcc46cb3f5bd8a748a/homeassistant/components/zha/core/channels/general.py#L144)).

Some comments from https://github.com/zigpy/zha-device-handlers/issues/2227 include more information on this.

I'll try to look into this soon-ish.

TheJulianJES:
Apparently we also have a duplicate version of this spell (with two different implementations) for a LIDL/Tuya plug:
https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/lidl/ts011f_plug.py

TheJulianJES:
@MattWestb I've read some discussions on the LIDL power strip that apparently requires the magic spell.
It's mentioned that the device only exposes one endpoint if the spell is not cast. This seems to be true, as the signature provided in that quirk only has one endpoint: https://github.com/zigpy/zha-device-handlers/blob/eed839c4f312a33da0087229a9b2f67185d5d557/zhaquirks/lidl/ts011f_plug.py#L135-L141
[...]

PR #2287, opened 2023-03-21
Use custom `OnOff` clusters for casting Tuya spell

TheJulianJES:
This addresses a part of https://github.com/zigpy/zha-device-handlers/issues/2271
This is not a final solution though.

IMO, the current "solution" is far worse, as it creates a task to read attributes in a class constructor. So the only time this happens when it's useful is when the device is first added and the quirk is initialized.
Re-pairing or re-configuring do not cast the spell again (as the quirk class is already loaded).
Reading attributes during restarting (when the device was added previously) is also broken with the current implementation.
Some tests are also left in a bad state with the current implementation.

This new implementation currently requires all Tuya devices that need the spell to implement `TuyaEnchantableOnOffCluster` (or a subclass of it) **and** set the `TUYA_SPELL` attribute to `True` or inherit `EnchantedDevice` (which just sets `TUYA_SPELL` to `True). This is somewhat to provide backwards compatibility with custom quirks (although those quirks need to implement a subclass of `TuyaEnchantableOnOffCluster` which should be given for most "enchantable" Tuya devices though).

This all needs to be checked and tested still, but from what I can see, all devices that are currently an `EnchantedDevice` already implement a subclass of `TuyaEnchantableOnOffCluster`. So the spell will work properly for them now.


TheJulianJES:
At the moment, there's no way to do this properly and that early in the pairing process.
I'd expect that Z2M also doesn't do it that early? And the current code (reading in the constructor of the quirk) also doesn't do it that early? Does the binding happen way later than that?

What happens if the spell is cast later? Does the device still "unlock" with the only difference being that the signature might look differently?

MattWestb:
It is possible doing with EZSP then tuya is doing it with standard firmware in there ZBGW but that is out of this question then its only Bellows and we must doing it with all radios.
If the timing is not OK = fast enough and the device have updating the TC-Link key OK and doing device accouterments its have doing the INIT and most devices is also sending one permit join with time 180 = over and out.

With the TS004F is only sending commands from EP1 and nothing from EP 2-4 if being too late and the device have not getting one spell after being retested `and` the battery was out (only one its keeping the spell) and the user is not getting the device working until the spell is casted then the device its in INIT mode. 

Many other devices is not need it in INIT mode (only devices that  doing "hardware changes" like adding EPs and so on must having it in the INIT mode) and can being done every time but then sniffing tuya ZBGW they is configuring the device network parameters and then sending one leave with rejoin flag sett and the device is rejoining and is being in INIT mode and the ZBGW is **first** casting the spell and then doing normal other configuration of it that was not done in the first phase.
I was testing getting sending leave with rejoin flag from quirk but was not getting it working (the flags was not implemented in the code) so was testing with doing the casting first in the quirk looks working 100% in current (Z)HA but can being broken in the future with code changes.

TheJulianJES:
> With the TS004F is only sending commands from EP1 and nothing from EP 2-4 if being too late

So this PR would *cast the spell* again when reconfiguring the device through the UI. Without re-pairing, would reconfiguring (so casting the spell after it's in the network) make that device *magically* work? (so sending commands on the other endpoints)

MattWestb:
If its casting the spell then doing reconfigure its helping may devices that dont need it in INIT mode and user dont need deleting the device and joining it new the device for getting it working.
But for TS004F and some other device its not making any bad things but the code must casting the spell then its joining devices new then the device is in IINIT mode or its not working.

I can testing it but after the IKEA changes that is waiting or one to 2 week now than i cant having 10 different devices is test mode in 3 different test systems (i have getting my windows 10 WSL not working after some `aggressive device cleaning` so i cant using it for fast testing).

Then i need sniffing the device then joining and resetting it and removing the battery and verifying its that it have loosing the spell and then applying the patch and see its applying it and working well and redoing the same at least 10 times for knowing its working well as it doing today so users is not getting problems in the future and we is hunting problems we have implanting in the code and making  some devices not working OK all the time and getting many strange issues from our users.

MattWestb:
I was putting the INITs in the HA container and patching the device class to not use the `EnchantedDevice` and only `CustomDevice` and they was already using the `TuyaSmartRemoteOnOffCluster` in replacment so i think the casting shall working then its getting it from tuya INIT. But with  wireshark is only doing binding and also then doing reconfigure.

Did i making somthing wrong ??

Used files the device quirk in cunstom quirk and the INITs in the HA container.
[Homeassistant_Bellows.zip](https://github.com/zigpy/zha-device-handlers/files/11063668/Homeassistant_Bellows.zip)

TheJulianJES:
Hmm, I'll need to have a look at how to deal with devices that use multiple enchantable clusters (so on multiple endpoints).
It would all be easier if there's something like a "pre-bind" function that always gets called (even on the `Basic` cluster).

MattWestb:
### Hold your horses @TheJulianJES  !!
I have getting it working and its very good only then adding the `TUYA_SPELL = True` but its taking time getting the device in not spell mode and variegating its without and then adding it agen and testing.

For the moment i testing adding it and its getting spell and then resetting the device and  taking out the battery before its rejoining for loosing the spell and then resetting it and letting it rejoining without getting spell and later doing reconfigure.
I have getting 2 times the reconfigure is fixing the device but i need waiting longer time and more test for see if its 100%.

Shall i putting the last commitment or is it OK running the test with the [1d36fb4](https://github.com/zigpy/zha-device-handlers/pull/2287/commits/1d36fb4cade3a22cf41c594057ede620bd7b3325) ?

TheJulianJES:
The [latest commit](https://github.com/zigpy/zha-device-handlers/pull/2287/commits/59ace2f7e3246cee88c91f586b4a7b11655c8067) will only "cast the spell" before actually binding the `OnOff` cluster. (Previously, it first bound the `OnOff` cluster and then *cast*  the spell)
Not sure if it makes any real difference though.

MattWestb:
I have looking and all the times i have getting it working is the spell casted before the device is requesting updating the TC-Link key and it shall being in the safe time window but i testing more for see i can repeating it at last 10 time with successes or i must looking if need more work.
The TS004F DMS is having 4 endpoints with OnOff on all but only one is active before it getting the casted spell and it shall being the worse device getting working OK.

TheJulianJES:
> The TS004F DMS is having 4 endpoints with OnOff on all but only one is active before it getting the casted spell and it shall being the worse device getting working OK.

Yeah. But zha-quirks already replaces the 1 endpoint signature with the 4 endpoint signature, right?
So ZHA will try to read the Tuya attributes four times. (I'm not sure if that's actually an issue though, since it will only happen when pairing and on reconfiguration and re-reading those attributes should be fine?)

MattWestb:
Its have only one EP 1 without the spell and the quirk is adding 3 new one but the sepll is only being casted one time on EP 1 (all is sniffed with wireshark) so no problems.
I think the LIDL power stripes is doing the same but it dont need it INIT mode and is one router so shall not being any problems with it.

Shall looking one more time if reconfigure is casting on all or only EP 1.

MattWestb:
Reconfigure is only doing EP1 and its very good and no binding of EP 2-4 = OK so not getting strange problems with the device (tuya is not doing any bindings at all on this devices).
```
No.	Time	PAN	Protocol	IEEE Src	IEEE Dst	Zigbee Src	Zigbee Dst	ZBN Dst	Group Nr	ZBN Seq	ZBA Seq	ZDP Seq	Nwk Seq	Src EP	Dst EP	Info
9263	17:20:48,560519	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		202	62		247	1	1	ZCL OnOff: Unknown Command, Seq: 13
9269	17:20:48,709428	0x1a62	ZigBee	0xb502	0x0000	0xb502	0x0000	0x0000		203	80		250	1	1	APS: Ack, Dst Endpt: 1, Src Endpt: 1
9270	17:20:48,718708	0x1a62	ZigBee	0xb502	0x0000	0xb502	0x0000	0x0000		204	80		251	1	1	APS: Ack, Dst Endpt: 1, Src Endpt: 1
9473	17:21:31,367527	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		206	63		19	1	1	ZCL OnOff: Unknown Command, Seq: 16
9491	17:21:31,694858	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		208	64	148	28	0	0	Bind Response, Status: Success
9492	17:21:31,705130	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		210	65		29	1	1	ZCL: Default Response, Seq: 150
9493	17:21:31,715313	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		212	66	152	30	0	0	Bind Response, Status: Success
9494	17:21:31,724774	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		214	67	154	31	0	0	Bind Response, Status: Success
9495	17:21:31,734247	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		216	68	156	32	0	0	Bind Response, Status: Success
9496	17:21:31,743780	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		218	69	158	33	0	0	Bind Response, Status: Success
9497	17:21:31,753901	0x1a62	ZigBee	0xb502	0x0000	0xb502	0x0000	0x0000		219	88		34	1	1	APS: Ack, Dst Endpt: 1, Src Endpt: 1
9508	17:21:32,323970	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		221	70		40	1	1	ZCL: Configure Reporting Response, Seq: 161
9509	17:21:32,341197	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		223	71		41	1	1	ZCL: Read Attributes Response, Seq: 163
9512	17:21:33,241892	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		225	72		42	1	1	ZCL: Report Attributes, Seq: 18
9521	17:21:33,457538	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		227	73		47	1	1	ZCL: Default Response, Seq: 165
9522	17:21:33,467750	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		227	73		47	1	1	ZCL: Default Response, Seq: 165
9523	17:21:33,475744	0x1a62	ZigBee	0xb502	0x0000	0xb502	0x0000	0x0000		228	92		48	1	1	APS: Ack, Dst Endpt: 1, Src Endpt: 1
9576	17:21:47,834266	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		230	74		53	3	1	ZCL OnOff: Unknown Command, Seq: 19
9580	17:21:47,998796	0x1a62	ZigBee	0xb502	0x0000	0xb502	0x0000	0x0000		231	93		56	3	3	APS: Ack, Dst Endpt: 3, Src Endpt: 3
9585	17:21:48,651629	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		233	75		59	3	1	ZCL OnOff: Unknown Command, Seq: 19
9590	17:21:49,350619	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		235	76		63	3	1	ZCL OnOff: Unknown Command, Seq: 19
```
And it was not casted and the EP 2-4 was not working and after reconfigure its start using them (the last OnOff command is send from EP 3) :-))))))

MattWestb:
I is filtering only showing commands sent from the device and its looks like this then joining it:
```
No.	Time	PAN	Protocol	IEEE Src	IEEE Dst	Zigbee Src	Zigbee Dst	ZBN Dst	Group Nr	ZBN Seq	ZBA Seq	ZDP Seq	Nwk Seq	Src EP	Dst EP	Info
522	17:41:33,604536	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	Broadcast	0xfffd		247	231	129	213	0	0	Device Announcement, Nwk Addr: 0xb502, Ext Addr: SiliconL_ff:fe:fd:dc:14
523	17:41:33,615583	0x1a62	ZigBee	0xb502	0x0000	0xb502	0x0000	0x0000		248			214			End Device Timeout Request
525	17:41:33,632139	0x1a62	ZigBee ZDP	0x0000	0xffff	0xb502	Broadcast	0xfffd		247	231	129	18	0	0	Device Announcement, Nwk Addr: 0xb502, Ext Addr: SiliconL_ff:fe:fd:dc:14
526	17:41:33,649692	0x1a62	ZigBee ZDP	0xb0c4	0xffff	0xb502	Broadcast	0xfffd		247	231	129	193	0	0	Device Announcement, Nwk Addr: 0xb502, Ext Addr: SiliconL_ff:fe:fd:dc:14
535	17:41:34,019298	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		252	233	98	220	0	0	Node Descriptor Response, Rev: 22, Nwk Addr: 0xb502, Status: Success
541	17:41:34,240813	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		254	234	100	224	0	0	Active Endpoint Response, Nwk Addr: 0xb502, Status: Success
547	17:41:34,845877	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		0	235	102	228	0	0	Simple Descriptor Response, Nwk Addr: 0xb502, Status: Success
554	17:41:35,343695	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		2	236		231	1	1	ZCL: Read Attributes Response, Seq: 104
570	17:41:35,948966	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		4	237	106	239	0	0	Bind Response, Status: Success
571	17:41:35,959772	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		6	238		240	1	1	ZCL: Default Response, Seq: 108
572	17:41:35,969413	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		8	239	110	241	0	0	Bind Response, Status: Success
573	17:41:35,979857	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		10	240	112	242	0	0	Bind Response, Status: Success
574	17:41:35,987977	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		12	241	114	243	0	0	Bind Response, Status: Success
575	17:41:35,998718	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		14	242	116	244	0	0	Bind Response, Status: Success
585	17:41:36,408050	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		16	243		249	1	1	ZCL: Configure Reporting Response, Seq: 118
586	17:41:36,424107	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		18	244		250	1	1	ZCL: Read Attributes Response, Seq: 120
592	17:41:36,868340	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		20	245		254	1	1	ZCL: Default Response, Seq: 122
600	17:41:37,323630	0x1a62	ZigBee ZDP	0xb502	0x0000	0xb502	0x0000	0x0000		22	246	130	2	0	0	Node Descriptor Request, Nwk Addr: 0x0000
601	17:41:37,334269	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		24	247		3	1	1	ZCL: Read Attributes Response, Seq: 124
611	17:41:37,903443	0x1a62	ZigBee	0xb502	0x0000	0xb502	0x0000	0x0000		25	248		9			Request Key
612	17:41:37,912055	0x1a62	ZigBee	0xb502	0x0000	0xb502	0x0000	0x0000		26	137		10	0	0	APS: Ack, Dst Endpt: 0, Src Endpt: 0
613	17:41:37,921732	0x1a62	ZigBee HA	0xb502	0x0000	0xb502	0x0000	0x0000		28	249		11	1	1	ZCL: Read Attributes Response, Seq: 126
```
554 is reading manufacture and model so quirk can being matched and loaded.
586 is the spell being casted and its red before the joiner is requesting node description of the TC so its knowing if it can updating the TC-Link key or not and is doing it in frame 611.
Need doing little more test  but its looks very good !!!!
Also then restarting HA its not casting the spell then the radio is not _on the air_ as the 3 other methods used and no error in the log.

TheJulianJES:
Rebased to fix a conflict.
I've changed the internal structure a bit, so that (theoretically) we're not as dependent on `OnOff` clusters, but we can also use other clusters that ZHA binds(!) in the future.

I've also added a test to make sure that all quirks which use `TUYA_SPELL`/`EnchantedDevice` also have at least one "enchantable cluster" on endpoint 1.

TheJulianJES:
I can double check later, but does ZHA also call `bind()` for `OUTPUT_CLUSTERS` or only `INPUT_CLUSTERS`?

MattWestb:
I think the check on EP 1 is very OK then tuya is always making EP 1 with basic and identify and so on (but they can changing) like many other is not doing,

Im not 100% sure that some tuya devices is using in cluster then some plugs (like LIDL power strip) can have only in clusters but normally controllers shall having out cluster (but we is doing with tuya devices here).

Can you putting `TUYA_SPELL = True` in `TuyaZBOnOffAttributeCluster` and `TuyaSmartRemoteOnOffCluster` classes that all this devices is needing the spell and we have converted most of them  to `TuyaEnchantableCluster` in the quirks and its little double wok doing it in the quirk and i think if we is finding one that dont can handling it we can putting false in one new device class for fixing it.

Then you have getting all tests working OK give my one sign and i installing the lasts commit and redoing most of the testing but its taking time doing it well for being sure and i thing we and @javicalle dont like braking devices for our users so i like being sure its working OK.

TheJulianJES:
> Can you putting TUYA_SPELL = True in TuyaZBOnOffAttributeCluster and TuyaSmartRemoteOnOffCluster classes that all this devices is needing the spell and we have converted most of them to TuyaEnchantableCluster in the quirks and its little double wok doing it in the quirk and i think if we is finding one that dont can handling it we can putting false in one new device class for fixing it.

The `TUYA_SPELL` attribute is used on a device level. The `TuyaEnchantableCluster` is used on a cluster level.
For a device, it's preferred to keep inheriting `EnchantedDevice` (which then sets `TUYA_SPELL`).

As long as the `EnchantedDevice`s have at least one cluster that's a subclass of `TuyaEnchantableCluster`, they should keep getting the Tuya spell.

If a device does not have `TUYA_SPELL` set (so is just a `CustomDevice`, it won't get the spell even if they use an `EnchantableCluster`.
So a device both needs to have an enchantable cluster **and** the device needs to use `TUYA_SPELL`/`EnchantedDevice` for them to get the spell applied.

I'm still working on some stuff and doing some more testing. I'll need to have a look at when ZHA exactly calls `bind()` too.

TheJulianJES:
Uhh, I see why `TuyaSmartRemoteOnOffCluster` is not working for the spell. ZHA doesn't call `bind()`, as that cluster changes the `ep_attribute`.

MattWestb:
You have without spell in https://github.com/zigpy/zha-device-handlers/pull/2287#issuecomment-1483193842 and was only sending command OnOff from EP1 and after the spell its was start using the new EPs.

And i using the `TuyaSmartRemoteOnOffCluster` plus `TUYA_SPELL = True` in the quirk device class and its working.

MattWestb:
Also not all devices that need the spell is having power config cluster like the LIDL power stripe and other plugs.

TheJulianJES:
> Also not all devices that need the spell is having power config cluster like the LIDL power stripe and other plugs.

Yeah, I was specifically talking about the remotes which use `TuyaSmartRemoteOnOffCluster` and that changes the `ep_attribute`, so ZHA doesn't call `bind()`:
https://github.com/zigpy/zha-device-handlers/blob/8e91d98783bd9bfd1c6906eba9f3853a15ac8c89/zhaquirks/tuya/__init__.py#L934-L948

For those remotes, we could make the `PowerConfiguration` cluster enchantable.

> And i using the TuyaSmartRemoteOnOffCluster plus TUYA_SPELL = True in the quirk device class and its working.

I'd recommend to use `EnchantedDevice` instead of `CustomDevice` (that's what quirks will keep doing). (and `EnchantedDevice` only sets `TUYA_SPELL = True`)

MattWestb:
My test quirk class looks like this:
````
class TuyaSmartRemote004FDMS(CustomDevice):
    """Tuya 4 btton dimmer switch / remote device."""

    TUYA_SPELL = True

    signature = {
        # "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=0, *allocate_address=True, *complex_descriptor_available=False, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False, *is_valid=True, *logical_type=<LogicalType.EndDevice: 2>, *user_descriptor_available=False)",
        # SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=260, device_version=1, input_clusters=[0, 1, 3, 4, 6, 4096], output_clusters=[25, 10, 3, 4, 5, 6, 8, 4096])
        MODELS_INFO: [
            ("_TZ3000_xabckq1v", "TS004F"),
            ("_TZ3000_czuyt8lz", "TS004F"),
        ],
        ENDPOINTS: {
            1: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.DIMMER_SWITCH,
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    Identify.cluster_id,
                    Groups.cluster_id,
                    OnOff.cluster_id,
                    LightLink.cluster_id,
                ],
                OUTPUT_CLUSTERS: [
                    Ota.cluster_id,
                    Time.cluster_id,
                    Identify.cluster_id,
                    Groups.cluster_id,
                    Scenes.cluster_id,
                    OnOff.cluster_id,
                    LevelControl.cluster_id,
                    LightLink.cluster_id,
                ],
            },
        },
    }

    replacement = {
        ENDPOINTS: {
            1: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    Identify.cluster_id,
                    Groups.cluster_id,  # Is needed for adding group then binding is not working.
                    LightLink.cluster_id,
                ],
                OUTPUT_CLUSTERS: [
                    Ota.cluster_id,
                    Time.cluster_id,
                    Identify.cluster_id,
                    Groups.cluster_id,
                    Scenes.cluster_id,
                    TuyaSmartRemoteOnOffCluster,
                    LevelControl.cluster_id,
                    Color.cluster_id,
                    LightLink.cluster_id,
                ],
            },
            2: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
                INPUT_CLUSTERS: [
                    TuyaSmartRemoteOnOffCluster,
                ],
                OUTPUT_CLUSTERS: [],
            },
            3: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
                INPUT_CLUSTERS: [
                    TuyaSmartRemoteOnOffCluster,
                ],
                OUTPUT_CLUSTERS: [],
            },
            4: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
                INPUT_CLUSTERS: [
                    TuyaSmartRemoteOnOffCluster,
                ],
                OUTPUT_CLUSTERS: [],
            },
        },
    }
```
and was working OK.

TheJulianJES:
Alright, I think it's *complete*. The tests should also be relatively good and catch invalid `EnchantedDevice`/Tuya spell implementations.
The only thing I don't really like (other than the fact that we have to do this) is the `TuyaNoBindPowerConfigurationCluster` stuff, but we need to do this, as the "NoBind" cluster obviously blocks `super().bind()` calls, so we have to put the `Enchantable` class first, so the end looks like this:
1. EnchantableClass, 2. NoBindClass, 3. PowerConfigurationCluster

IMO, it's fine for now.

@MattWestb If you want, you can test this again. It seems to work well for me and properly re-sends the spell when reconfiguring or re-pairing the device now. I don't have any Tuya devices that actually require the spell to test with though.
But the "python tests" should already make sure that everything still works.

(Heh, @javicalle you also wanna have a look?)

TheJulianJES:
> My test quirk class looks like this:
> and was working OK.

Ah, huh. Just `TuyaSmartRemoteOnOffCluster` shouldn't have worked, as ZHA shouldn't call `bind()` for that, as it has another `ep_attribute` (than the original `OnOff` one). I guess I'll need to have another look..

MattWestb:
So only need putting the 2 updated INITs in the container and then updating the local quirk and it shall working like i was doing the first test.

I doing it tomorrow and see how its working but for my was the first implementation working OK.
Interesting in the test is early enough then joining and every time its doing new join and plus if its doing on rejoicing or reconnect and also looking if its working OK then reconfigure and also clean restart of HA.

TheJulianJES:
Ah hmm, is `bind()` called for mains-powered devices on ZHA startup too?
(I guess if so, that's fine/good)

MattWestb:
Reset the remote and restarting HA and deleting the device and then adding it and its was getting the spell but little late but working.
[home-assistant_2023-03-24T21-19-12.675Z.log](https://github.com/zigpy/zha-device-handlers/files/11066663/home-assistant_2023-03-24T21-19-12.675Z.log)

MattWestb:
It shall but only on out cluster that is sending commands and not like tuya have done on this device that must getting groups added on in cluster for sending commands to one group.

MattWestb:
I have updated the 2 INITs and restarting and adding the device and its not getting the spell.

ZHA log then new paring and one rejoining:
[home-assistant_2023-03-25T07-53-57.037Z.log](https://github.com/zigpy/zha-device-handlers/files/11068342/home-assistant_2023-03-25T07-53-57.037Z.log)

javicalle:
First of all, thanks to both of you for the work. 
I really want to take a deep look at it. 

> Ah, huh. Just `TuyaSmartRemoteOnOffCluster` shouldn't have worked, as ZHA shouldn't call `bind()` for that, as it has another `ep_attribute` (than the original `on_off` one). I guess I'll need to have another look..

I'm pretty sure that the `ep_attribute` rename was a workaround to prevent HA from creating switch entities for the remote.
Some time after, I thinked that putting as an out cluster would do the job, but I never checked that approach.

MattWestb:
Hola @javicalle !!! :-)
>I'm pretty sure that the ep_attribute rename was a workaround to prevent HA from creating switch entities for the remote.
Some time after, I thinked that putting as an out cluster would do the job, but I never checked that approach.

That very true and the TS004F DMS is having it as in for the moment then tuya have making strange things and the default response is need being sent in the wrong way (to / from server / client is wrong) and i cant getting the system doing it so the switch is repeating the command 3 times also have getting the default response and acking it and the quirk is filtering the extras out with TSN (for the tuya scene commands but is working OK with light  commands).

I was yesterday testing both in and out and ZHA is not making any switches in the GUI but its depends of the device hat is working best with the default response.

@TheJulianJES Is your TS004X sending one command or is it repeated every time ?
Also i was ordering and paying for 2 TS0043 switches and DPD was throwing them in some other mailbox and i was loosing the money so i cant testing them. Can you dumping the main flash and the user data on you device ?
Then i can installing it on one LIDL RGB LED strips module (or one GZB-MG21 module) and ding some testing.

javicalle:
Hi Matt,

> tuya have making strange things and the default response is need being sent in the wrong way (to / from server / client is wrong) and i cant getting the system doing it

Can you point me to the expected response from the device? Maybe that info has been already posted but I have no reference for that (despite I can remember the issue).
Nowadays I have some more experience on the subject and if the problem is in the answer from ZHA, and maybe I can take a look again.

> testing both in and out and ZHA is not making any switches in the GUI

This will happen for sure when `ep_attribute = "TS004X_cluster"` but IIRC letting the default value (`ep_attribute = "on_off"`) ZHA will create entities if they are in the in_cluster but not if in the out_cluster.
I can't say if we can remove the overwrite and just move all the `TuyaSmartRemoteOnOffCluster` to the out_cluster. It is used in all the TS004X family and seems a bit risky to get the same result.
I'm in favor of doing 'code cleanup' from time to time but in this case I'm not sure.

MattWestb:
Then @TheJulianJES is having one TS0043 and i have one TS004F DSM and ROK we can doing good testing of most functions that all the TS004X devices is using so its likely we can testing all very good but not 110% we is not braking things then we redoing the core code for this devices.

The Smart things have finding the reason way its doing this and tuya is not fixing it in ther ZBGW then i was sniffing it but Z2M i think was fixing it and smart things for 100%.

My issue that's one year old and closed https://github.com/zigpy/zha-device-handlers/issues/1448.

If we shall digging deeper we must doing much sniffing and validating that's working but  i think it very good getting it working well because its not looking good in the log with 3 events for events command sent by the device.

I  think its super that some is very good in coding in different parts and other more on the hardware things and also the working mode of different device families like Xiaomi / Aqara, IKEA, Philips HUE and so on and in the end we is getting very mush fixed in some way so our devices is working well :-)))

TheJulianJES:
> Reconfigure is failing on the power config cluster:

Huh, it's also not a `TuyaPowerConfigurationClusterEnchantable` (and yes, the names are getting a bit long).
I'll have a look at the logs later, but 2 out of 3 quirks for the TS004F required some small changes (all other quirks did not).
What quirk does your remote match?
(And can you double-check that you installed it correctly? Maybe even replace the entire zhaquirks package temporarily if you're comfortable with that?)

> I'm pretty sure that the ep_attribute rename was a workaround to prevent HA from creating switch entities for the remote.
Some time after, I thinked that putting as an out cluster would do the job, but I never checked that approach.

Yep, I think that *should* also work. Can probably check later. 

> Is your TS004X sending one command or is it repeated every time ?

It's a TS0041 matching the `TuyaSmartRemote0041TO` quirk. I think I've sometimes seen the duplicate TSN removal printing to the logs, but I'll double-check later.
About dumping flash: heh, still don't have a J-Link probe.

MattWestb:
>About dumping flash: heh, still don't have a J-Link probe.

That one shame !!!!!
How can we living without one ;-))
You need one that is not being broken then connecting to Seger J-link flasher that dont being broken then its try updating the probe firmware for handling EFR32MG21 devices.
Also if the hardware is the same (that is very likely) you can flash the TS004F and testing the functions in real.

The 4 button Dimmer Switch that also can sending TS004X "scene" commands is using  the `TuyaSmartRemote004FDMS` and i cant testing the normal rotating knob that is using `TuyaSmartRemote004FROK` but the tricky is the DMS then its very kinky getting all right with the extra EPs is using.

Pleas explaining what i need changing in the quirk for getting it working !!

Im away some hours for being little social with friends but is back later.

TheJulianJES:
Can you post your `ts004f.TuyaSmartRemote004FDMS`? Look at the changes in this PR to that. Make sure you're using `TuyaPowerConfigurationClusterEnchantable`.

If you have a test setup, it might be be easier to replace the `zhaquirks` library with this temporarily: https://github.com/TheJulianJES/zha-device-handlers/archive/refs/heads/tjj/tuya_spell_in_clusters_clean.zip
(and then remove all custom Tuya quirks)

MattWestb:
[tu_ts004f.py.txt](https://github.com/zigpy/zha-device-handlers/files/11069657/tu_ts004f.py.txt)

TheJulianJES:
That's the only class that needs changes, take a look at the diff in this PR or at the [full file](https://github.com/zigpy/zha-device-handlers/blob/c4ebc5df7aa884aac62014f76ac021f82359b297/zhaquirks/tuya/ts004f.py).
You'll need to import `from . import TuyaPowerConfigurationClusterEnchantable` and use that in the replacement for two of the three versions in that file (if you're using custom quirks).
Again, I'd recommend to just temporarily replace the `zhaquirks` package for testing if you can.

However, since the default `PowerConfiguration` cluster already fails to bind, it still will. But it'll properly cast the Tuya spell anyway.
I wonder if that cluster can't/shouldn't be bound. Maybe we should just use `TuyaNoBindPowerConfigurationCluster` for that remote too (should probably do that in another PR though).

MattWestb:
I have deleting the `TUYA_SPELL = True` and using `class TuyaSmartRemote004FDMS(CustomDevice):` and for power config the new `TuyaPowerConfigurationClusterEnchantable`.
The system making many unneeded and not wanted binding but is not casting the spell after testing multiple times.

As most tuya Zigbee devices they is not needing binding them then is always reporting to the coordinator and its highly unwonted to do on this devices.

But without casting  is the device not working for 95% of our users.

TheJulianJES:
`TuyaPowerConfigurationClusterEnchantable`, like others, will only activate if you're using `EnchantedDevice` (or `TUYA_SPELL`). You need both changes (like done in this PR) for this to work.

MattWestb:
This version is not good then its triggering constant MAC pulling of its parent until i was sending commands from it.
And very likely is is the reason that its binding 0x0008, 0x0006, 0x0005, 0x0300 plus reading ZLL group that all is not needed and making the device behaving bad.
Also reconfigure is broke (user term) / not looking good then its trying configure battery reporting that is always failing and its not castng the spell then doing reconfigure and if the device is rejoining with network key or being retested and doing one joining then the system is having the device i the zigbee.db.

For TS004F is your first version excellent but this is not working with this device.

TheJulianJES:
> This version is not good then its triggering constant MAC pulling of its parent until i was sending commands from it.
And very likely is is the reason that its binding 0x0008, 0x0006, 0x0005, 0x0300 plus reading ZLL group that all is not needed and making the device behaving bad.

1. Is all of the above in regards to your `TS004F`?
2. If so, that devices matches the `TuyaSmartRemote004FDMS` quirk, correct?
3. The changes are very minimal and can't really cause anything like that. Are you saying this behavior didn't occur before these changes? (So ZHA didn't try to configure binding and so on?)
4. The first version for TS004F could not work, as the OnOff cluster wasn't bound, so no spell is sent. You likely still used the older `EnchantedDevice`.
5. Did you replace the entire quirks library with my branch and test it then?
6. Does the `TS004F` need battery binding + attribute reporting or should we block it?

MattWestb:
This is little role the dice hwo the system / device is doing the things.
2/3 times its working OK and 1/3 is doing it very strange (i have testing around 15 times).
Then its working OK its sending the spell then doing new joining, being retested and being joined then ZHA have its is Zogbee.DB and the being reconfigured.

Then doing reconfigure its not sending power config to the device but failing after timeout most of the time but also can doing it without errors.

I have only updating tuya INIT and the MCU INIT in the container and is having this quirk in the local folder: [tu_ts004f.py.txt](https://github.com/zigpy/zha-device-handlers/files/11071488/tu_ts004f.py.txt).

My feeling is that all the binding that is not needed that is making the problems with the timing but the good thing is looks like its fixing one device that is not having getting the spell also with reconfigure that was not possible code used before (it can being that redoing the bindings and the device is doing some configure and being in INIT mode but i dont knowing).

Im spitted if its 100% and can going in production then its not doing the same thing all the times and i dont like that. I have not getting the continues MAC pull from the device but i looking  if i see it then i doing more testing and sniffing it.

This device is having the most extreme "personality" i have see !!!

Do you need more information or test being done ?

TheJulianJES:
> 2. If so, that devices matches the TuyaSmartRemote004FDMS quirk, correct?

To confirm, that is the quirk that your device matches, correct?

> Then doing reconfigure its not sending power config to the device but failing after timeout most of the time but also can doing it without errors.

Do you still get "Executing Tuya spell" and "Executed Tuya spell" in the logs? (Those log entries are printed by the "new Tuya spell". In between those lines, you should see if reading the attributes was successful) 

You can also try to use the `TuyaNoBindPowerConfigurationCluster` for your device. This cluster will still do the new Tuya spell (if `TUYA_SPELL = True` or you're using the updated `EnchantedDevice` which just sets that to true), but that cluster will block binding and attribute reporting of the PowerConfiguration cluster.

MattWestb:
I only doing testing with TS004FDMS then its the most problematic one and later i can doing the ROK (Smart Rotating Knob).

If you looking in the attached quirk i is using the `TuyaNoBindPowerConfigurationCluster`.

I shall doing more testing later but i only looking in the online sniffs for see have the commands is getting to and from the device with the timing so i can looking in the log then.

javicalle:
Just a side comment, if the timing is critical while casting the Tuya spell, enabling the debug logs add a significant overhead (writing to file is expensive in time) and _maybe_ can impact in the results.

MattWestb:
Its one small NanoPi Neo running armbian32 RCP without OTBR and only Zigbeed and 1 IKEA VINDSTYRKA that is sitting in karantene for not getting OTA updates so its very small system but if its working OK with some debug logging its very likely not getting problem then taking it away and also in larger system but cant knowing for 100% until tested.

TheJulianJES:
Do you think we can merge this for the next HA release (or at least the beta)?

It won't break any currently working devices and should be lots better than the current implementation IMO.
But I'm not sure if (or how much) timing is affected.

MattWestb:
I was copy your last tuya INIT and mcu-INIT to the container and updated the local quirk but i was not liking the power config binding so i was putting in the `TuyaNoBindPowerConfigurationCluster` so its not doing it.
The device is reporting battery itself then paring so its looks OK the tuya is not doing more an we is not getting battery draning if all is working as expected.

Then testing the ROK and its looks working very well but is not so tricky as the DMS.

If joining one device new its casting the spell OK as long  ZHA is doing its parts OK.
Some time ZHA is reading the device OK but is displaying it as unknown then its not  loading the quirk and so spell is being casted.
The good thing is looks like reconfigure is casting the spell OK and and the device is reacting at it then is reporting the mode change attribute and start using the extra functions. With the timing its also look OK then its normally being sent before the decide is requesting node description of the coordinator and start updating the TC-Link key procedure so its very early and that is good so not being the the last "minute" and can getting problems.

I  think we cant getting it better and i think updating the INITs and the TS004F quirk is OK and its not braking any devices that is working for our users.
Only updating all classes in the quirk so all device is getting it. and use the `TuyaNoBindPowerConfigurationCluster` then its looks working better.

javicalle:
I haven't reviewed all the implementation in detail (and I'm very sorry for that), but I have an idea in my head that I want to expose. Maybe is a dumb idea.

I was guessing, it would posible to cast the spell from the `tuya_manufacturer` cluster?
All the devices that require the spell will have it. We don't have to worry about the `OnOff` or the `PowerConfiguration`.
If I had understand how it works, it would require to identify the cluster in ZHA and make it a 'fisrts class citizen'. This way, ZHA will trigger always the bind for the `tuya_manufacturer` cluster. And would be up to us to identify the device as a `EnchantedDevice` (or `TUYA_SPELL = True`).

What do you think?

I'm sorry to come here in the last minute when all the discuccion is already setted. I'm impressed for all the work made here I don't want to generate noise now that it seems already done.

TheJulianJES:
Yeah, not a bad idea. I've also thought of that at one point. When I looked at it, I realized that not all (enchanted) devices even expose a Tuya manufacturer cluster (we could "fake one", but I don't know if that's the best idea).

I'm not completely sure on this, but I also think that ZHA binds the cluster from id 0 onwards, so the Tuya manufacturer cluster would probably get bound at the end which might "dramatically" delay the spell.
The PowerConfig/OnOff clusters have lower ids and basically all enchanted Tuya devices already use custom Tuya OnOff/PowerConfig ones.

This implementation definitely isn't "a final solution", but I think it's the best(?) we can do for now (until ZHA/zigpy/quirks is changed).

MattWestb:
Hola JC !!
You is **never** to late with your knowledge !!
Most "normal" tuya Zigbee device like all in the  TS004F class is not having the tuya cluster then they is in ground light controller. With the rest of TS004X is little like old without and new have extra tuya clusters.
My concern is how doing it with some device like sensors and MCU devices that is not heaving OnOff cluster that we can using for triggering casting the spell.

And tuya is using _normal_ zigbee chips with normal MAC so we cant doing like Xiami and fixing MAC range for triggering it then we is casting it on over 50% of all devices then Silicon labs is IKEA and new HUE device and many more.

javicalle:
Now I can see a little more of the full picture.
Nothing more to argue from my side.

Thanks to both of you!

MattWestb:
You is having most experience with strange tuya device from the code side and is knowing how to tricking them. Its one mess and we must living with it then tuya have not implanting all things at all of was forgetting reading the book so in very glad i have most IKEA devices but they is not enough for functionality for all my needs.

TheJulianJES:
This isn't really related to this PR, but for the `TS004F`, there are four quirks/devices:
- `TuyaSmartRemote004FROK` (rotating knob)
- `TuyaSmartRemote004FDMS` (four button remote)
- `TuyaSmartRemote004F` (four button remote)

Only the `TuyaSmartRemote004F` seems to use `TuyaZBOnOffAttributeCluster` instead of `TuyaSmartRemoteOnOffCluster`. From what I can see in the Z2M code, they treat them somewhat similar (only difference if the other endpoints are there)(?)
Is there a reason why the one remote uses that other custom cluster?

And Z2M binds the power configuration cluster for all of these remotes. But you're saying we don't need to do that?

MattWestb:
I have sniffing tuya ZBGW setting it up and they is doing no binding or reporting only normal Zigbee 3 joining then kicking the device then its in the network with leave and rejoin flag set and is casting the spell then its coming back in the network and then is doing nearly nothing. And our experience of the rest TS004F is not doing strange things so you not getting problems like battery draining or device leaving the network.

The `TS004F` is one trick then the system is first matching the device with model info and if its not being matched the device with the same cluster / EP is using the last class with only model and is getting the attribute you need changing its working mode.
If the user like have more functionality they need requesting device support for it.

We have 2 more device that is the same type but is not getting in to PR the single knob and the LIDL that i have not getting (they dont had it in AT so i was not getting any one for testing). plus one more LIDL 4 button remote but its one other TS type but is using the same commands.

TheJulianJES:
Are we good to get this in for now? Should be somewhat "well-tested"

MattWestb:
My devices is working OK and have moving them to its normal system now.
If someone have more device for testing it good but i think it cant being braking for all  device that need it and is already paired with ZHA.

javicalle:
Sorry, no exotic devices around here 😅

TheJulianJES:
Thanks for all the testing too!

erikproper:
Dear all,
   I've had some issues regarding Zigbee devices in my apartment, and going through things sequentially in fixing this.

I have several Tuya based remotes, where I have indeed noticed a battery drain and the habit of reporting 100%.
I'm using the quirks that come with the latest (am up-to-date) home assistant release, and see that the following quirks are used (across my "fleet" of Tuya remotes):
 zhaquirks.tuya.ts0042.TuyaSmartRemote0042TO
 zhaquirks.tuya.ts0043.TuyaSmartRemote0043TO
 zhaquirks.tuya.ts0044.TuyaSmartRemote0044TO
The 44 one seems to work well in the sense of showing 84% matter after a few months of use.
The 42 and 43 share the ill-behaviour as reported in this thread:
- Battery level = 100
- Fast draining of battery (but not all of them ... but occurs across both 42 and 43)

So ... can I help in "debugging" solutions?
Or, should I wait for the next home assistant release (due in the next week I guess).

Still have a pile of the needed batteries, so not in a big hurry ;-)

Cheers,
  Erik

TheJulianJES:
You can update to Home Assistant Core 2023.4.0 (beta releases today) and then remove and re-pair all battery draining remotes.

Issue #2227, opened 2023-02-26
[BUG] Tuya TS0041 TS0042 TS0043 - no battery


provinzio:
**Describe the bug**
The Battery value in Home Assistant is always at 100 %. Eventhough I use the switches for over a year now and already had to change the batteries.

**To Reproduce**
Using automatically discorvered quirks e.g. `zhaquirks.tuya.ts0042.TuyaSmartRemote0042TO`

**Expected behavior**
Correct battery value

**Additional context**
Related fixed issue from zigbee2mqtt https://github.com/Koenkk/zigbee2mqtt/issues/6331
Related misbehavior of the buttons with the quirks https://github.com/zigpy/zha-device-handlers/issues/1746

javicalle:
> Related fixed issue from zigbee2mqtt [Koenkk/zigbee2mqtt#6331](https://github.com/Koenkk/zigbee2mqtt/issues/6331)

The issue is not fixed in Z2M.
The fix from your link generated other errors:
* https://github.com/Koenkk/zigbee2mqtt/issues/8072
and the code is not present in the current version:
* https://github.com/Koenkk/zigbee-herdsman-converters/blob/ac121be70a47856fadd25f3ffe04412db5fa6683/devices/tuya.js#L1805-L1822


Actually the same issue is present in Z2M and not fixed:
* https://github.com/Koenkk/zigbee2mqtt/issues/8500

From the ZHA quirk point of view I would suggest to remove (in the replacement part) all the `PowerConfiguration` clusters except from endpoint 1. IE: for the `TuyaSmartRemote0042TO` quirk:
```python
class TuyaSmartRemote0042TO(CustomDevice):
    """Tuya 2-button remote device with time on out cluster."""

    signature = {
        # SizePrefixedSimpleDescriptor(endpoint=1, profile=260, device_type=0, device_version=1, input_clusters=[0, 1, 6], output_clusters=[10, 25]))
        # SizePrefixedSimpleDescriptor(endpoint=2, profile=260, device_type=0, device_version=1, input_clusters=[1, 6], output_clusters=[])
        MODEL: "TS0042",
        ENDPOINTS: {
            1: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    PowerConfiguration.cluster_id,
                    OnOff.cluster_id,
                ],
[...]

PR #2288, opened 2023-03-21
Add `TuyaNoBindPowerConfigurationCluster` for Tuya remotes to stop battery drain

TheJulianJES:
Untested at the moment.

This PR prevents the binding of the `PowerConfiguration` cluster (and setting up attribute reports).
In return, this prevents battery drain on Tuya remotes.

Fixes https://github.com/zigpy/zha-device-handlers/issues/2227 (if it works)

Also, why do some Tuya remotes have multiple PowerConfiguration clusters? Should we "remove" all of them except one?

Note: The `bind()` and `_configure_reporting()` code is taken out of the `LocalDataCluster`

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -832,6 +832,8 @@ class TuyaLocalCluster(LocalDataCluster):
 
 
 class TuyaNoBindPowerConfigurationCluster(PowerConfiguration, CustomCluster):
+    """PowerConfiguration cluster that prevents setting up binding/attribute reports in order to stop battery drain."""
+
     async def bind(self):
         """Prevent bind."""
         return (foundation.Status.SUCCESS,)
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -831,6 +831,16 @@ class TuyaLocalCluster(LocalDataCluster):
         return self._update_attribute(attr.id, value)
 
 
+class TuyaNoBindPowerConfigurationCluster(PowerConfiguration, CustomCluster):
+    async def bind(self):
+        """Prevent bind."""
+        return (foundation.Status.SUCCESS,)
+
+    async def _configure_reporting(self, *args, **kwargs):  # pylint: disable=W0221
+        """Prevent remote configure reporting."""
+        return (foundation.ConfigureReportingResponse.deserialize(b"\x00")[0],)
+
+
 class TuyaPowerConfigurationCluster(PowerConfiguration, TuyaLocalCluster):
     """PowerConfiguration cluster for battery-operated thermostats."""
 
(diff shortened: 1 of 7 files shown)


MattWestb:
About unused / not function cluster https://github.com/zigpy/zha-device-handlers/discussions/973.

TheJulianJES:
Should fix the battery drain issue. Hoping for someone in https://github.com/zigpy/zha-device-handlers/issues/2227 to confirm.

Can already confirm that the remote still works as expected.

nibo2:
I can confirm from logs and tests that `__init__.py` and `ts0044.py` both initialize and work fine. Battery percentage and voltage is reported at startup.

I can not yet confirm if battery drain is gone or not. Actually, battery drain seems to have stopped already with `SKIP_CONFIGURATION` that I have been running for 2 weeks before this latest test. There has been no drop in battery percentage since then, and also no drop with the latest test.

Is there any more tests I/we should do?

zhaquirks/tuya/__init__.py
```diff
@@ -831,6 +831,18 @@ def update_attribute(self, attr_name: str, value: Any) -> None:
...
+class TuyaNoBindPowerConfigurationCluster(PowerConfiguration, CustomCluster):
+    """PowerConfiguration cluster that prevents setting up binding/attribute reports in order to stop battery drain."""
+
+    async def bind(self):
+        """Prevent bind."""
+        return (foundation.Status.SUCCESS,)
+
+    async def _configure_reporting(self, *args, **kwargs):  # pylint: disable=W0221
+        """Prevent remote configure reporting."""
+        return (foundation.ConfigureReportingResponse.deserialize(b"\x00")[0],)
```
javicalle, line 843: Just guessing, would it be the same that `ConfigureReportingResponseRecord(status=Status.SUCCESS).serialize()`?`
https://github.com/zigpy/zigpy/blob/1f81b162f84328eb96db18add3104d889fa6c923/zigpy/zcl/foundation.py#L438

TheJulianJES: Yep, probably. I think we should also change it in `LocalDataCluster` then. Didn't want to do that in this PR though.

I think I'd leave it like this and then change both (and some other similar stuff) in a future PR(?)

javicalle: Yes, for sure. 👍🏻 (just doing a quick issues review, not trying to criticize nobodies work)

TheJulianJES:
Merging now, so we can move forward with https://github.com/zigpy/zha-device-handlers/pull/2287.

I tested it on one of my Zemismart remotes and it doesn't break anything and correctly blocks binding of the power configuration cluster (which is likely enough to stop the excessive battery drain).

stickpin:
Hi @TheJulianJES,

I am back to ZHA once again with SkyConnect... :D 
Is there any reason why we still have multiple `PowerConfiguration` for TS0044?
Should I fix it and submit the PR?

Thanks and Merry Christmas! :)

PR #2289, opened 2023-03-21
Add Tuya LH-961ZB motion sensor support

Tagnard:
This quirk fixes the problem there the iaszone state is never reset.

### DIFF
--- a/zhaquirks/tuya/sm0202_motion.py
+++ b/zhaquirks/tuya/sm0202_motion.py
@@ -1,0 +1,109 @@
+"""Device handler for Tuya LH-961ZB motion sensor."""
+import asyncio
+from typing import Optional, Tuple, Union
+
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.general import Basic, Identify, Ota, PowerConfiguration
+from zigpy.zcl.clusters.security import IasZone
+
+from zhaquirks import _LOGGER
+from zhaquirks.const import (
+    CLUSTER_COMMAND,
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OFF,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+    ZONE_STATE,
+)
+
+
+class _Motion(CustomCluster, IasZone):
+    """Self reset Motion cluster."""
+
+    reset_s: int = 60
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        super().__init__(*args, **kwargs)
+        self._loop = asyncio.get_running_loop()
+        self._timer_handle = None
+
+    def _turn_off(self):
+        self._timer_handle = None
+        _LOGGER.debug("%s - Resetting motion sensor", self.endpoint.device.ieee)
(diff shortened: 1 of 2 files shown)


javicalle (changes requested):
It seems to me a difficult device to classify within the Tuya ecosystem.
In any case the name of the quirk would change it to `SM0202Motion`

zhaquirks/tuya/sm0202_motion.py
```diff
@@ -0,0 +1,109 @@
...
+            Union[t.Addressing.Group, t.Addressing.IEEE, t.Addressing.NWK]
+        ] = None
+    ):
+        """Handle cluster request."""
+        arg = args[0]
+        if arg & IasZone.ZoneStatus.Alarm_1:
+            """Handle the cluster command."""
+            if self._timer_handle:
+                self._timer_handle.cancel()
+            self._timer_handle = self._loop.call_later(self.reset_s, self._turn_off)
```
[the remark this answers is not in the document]
Tagnard: I am new to building quirks and I didn't know what is did more than I used it as base for mine. 

After changing the  cluster to the one you showed me the sensor still works. Now I want to override the 30 second timer with a 60 second one, how do you do that using this predefined cluster? 

The motion sensor has a 60 second timeout before being able to trigger again.

Tagnard:
Thanks for the information, I have updated it

javicalle (approved):
LGTM

zhaquirks/tuya/sm0202_motion.py
```diff
@@ -0,0 +1,68 @@
...
+
+    reset_s: int = 60
+
+
+class SM0202Motion(CustomDevice):
+    """Quirk for LH-961ZB motion sensor."""
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        super().__init__(*args, **kwargs)
```
TheJulianJES, line 29: This can be removed now

TheJulianJES (approved):
LGTM

zhaquirks/tuya/sm0202_motion.py
```diff
@@ -0,0 +1,64 @@
...
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.IAS_ZONE,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
+                    Identify.cluster_id,
+                    IasZone.cluster_id,
+                    0xEEFF,  # Unknown
```
TheJulianJES, line 43: Ah, actually. I think we should still add this to the `replacement`?
(But I guess it doesn't really make a difference)

Tagnard: Like this?

```
replacement = {
        ENDPOINTS: {
            1: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.IAS_ZONE,
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    Identify.cluster_id,
                    IasZone.cluster_id,
                    0xEEFF,  # Unknown
                    MotionCluster,
                ],
                OUTPUT_CLUSTERS: [Ota.cluster_id],
            },
        },
    }
```

TheJulianJES: No, I meant that the unknown cluster ID `0xEEFF` could also be added to the lower replacement part (below where you use `MotionCluster`).
But it's probably fine as-is.

Issue #2272, opened 2023-03-13
[BUG] TS0002 by _TZ3000_fbjdkph9, 2 gang light switch, when either switch to turn off/on by ZHA both do the same action


Diramu:
**Describe the bug**
This device has two light entities, when HA turns either on/off, both lights physical switch on/off, then a few seconds later the others entity states updates to match what you did (to the other).  In others words, they are acting like a group.

**To Reproduce**
1. Go to device page of this unit via the ZHA integration, both control switches show off.
2. Click on the toggle switch for the first control switch
3. Yes see both lights connected to the device turn on instantly
4. logbook reports: _Ikuu_2sw Light Left turned on triggered by service light.turn_on_
5. about 5 seconds later, the toggle switch for the second control switch turn on, and the logbook reports: _Ikuu_2sw Light Right turned on_

**Expected behavior**
I would have expected just the light controlled by the first control switch to turn on.

**Screenshots**
![image](https://user-images.githubusercontent.com/973442/224702953-34fb2c29-4df5-4338-8c05-613697eef53d.png)
<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0xe000",
        "0xe001"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0100",
[...]

PR #2296, opened 2023-03-23
Add Tuya `TS0002` 2 gang switch variant

javicalle:
New `Switch_2G_Var03` variation for the `TS0002` Tuya 2 gang devices.

Fix: #2272

### DIFF
--- a/zhaquirks/tuya/ts000x.py
+++ b/zhaquirks/tuya/ts000x.py
@@ -323,7 +323,7 @@ class Switch_2G_Metering(EnchantedDevice, CustomDevice):
 
 
 class Switch_2G_Var03(EnchantedDevice, CustomDevice):
-    """Tuya 2 gang (variation 03))."""
+    """Tuya 2 gang (variation 03)."""
 
     signature = {
         MODEL: "TS0002",
--- a/zhaquirks/tuya/ts000x.py
+++ b/zhaquirks/tuya/ts000x.py
@@ -322,6 +322,98 @@ class Switch_2G_Metering(EnchantedDevice, CustomDevice):
     }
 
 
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts000x.py
```diff
@@ -322,6 +322,98 @@ class Switch_2G_Metering(EnchantedDevice, CustomDevice):
     }
 
 
+class Switch_2G_Var03(EnchantedDevice, CustomDevice):
```
TheJulianJES, line 325: I think just `EnchantedDevice` is enough, as that also extends `CustomDevice`. But I see that it's also done like this in other places, so I guess it's fine too

zhaquirks/tuya/ts000x.py
```diff
@@ -322,6 +322,98 @@ class Switch_2G_Metering(EnchantedDevice, CustomDevice):
...
+    """Tuya 2 gang (variation 03))."""
```
TheJulianJES, line 326: Second `)` can probably be removed

javicalle: Done: https://github.com/zigpy/zha-device-handlers/pull/2296/commits/0dc12e01f3c47aa4ddc24b6c469e77c0a7b3a9d8

zhaquirks/tuya/ts000x.py
javicalle: Yes, you're right. And have warned me already 😞 
I'll update all of them (in this file) in another PR

MattWestb: I think some one was doing most or all of them 😞

TheJulianJES (approved):

javicalle:
Thanks Julian

PR #2299, opened 2023-03-24
Address Tuya `EnchantedDevice` inheritance

javicalle:
Remove the `CustomDevice` inheritance for the quirks that already extend from `EnchantedDevice`. 
Follow up from #2296


TheJulianJES (approved):
LGTM. (Will close + re-open to hopefully get tests to kick-off)

PR #2305, opened 2023-03-28
Don't bind power config for Tuya TS004F remotes, use remote OnOff cluster

TheJulianJES:
This uses `TuyaNoBindPowerConfigurationCluster` for all three remotes/quirks in the `TS004F` file to avoid binding the power configuration cluster.

It also replaces the `TuyaZBOnOffAttributeCluster` with the `TuyaSmartRemoteOnOffCluster` cluster that's used by the other models.
=> This is somewhat to avoid having two enchanted clusters, the other two models already used that cluster, and it also prevents binding the OnOff cluster (due to the different `ep_attribute`, so I guess that's one less cluster bound).
We don't have tests for this, but I don't think it should break anything(?) (Even if, the impact should be somewhat minimal and I can revert/fix it.)
Any ideas if this is okay?

Lastly, it also removes the `TuyaPowerConfigurationClusterEnchantable` cluster class, as it was only used by those TS004F models and since Tuya gateways apparently don't bind that cluster even, we likely won't need it again.

cc @MattWestb (and maybe @javicalle 😄)

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -871,12 +871,6 @@ class TuyaLocalCluster(LocalDataCluster):
         return self._update_attribute(attr.id, value)
 
 
-class TuyaPowerConfigurationClusterEnchantable(
-    TuyaEnchantableCluster, PowerConfiguration
-):
-    """Enchantable PowerConfiguration cluster."""
-
-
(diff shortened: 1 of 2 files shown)


TheJulianJES:
Any idea if the `TuyaZBOnOffAttributeCluster` to `TuyaSmartRemoteOnOffCluster` change for the 3rd device is also fine?

TheJulianJES:
So probably fine to merge as-is?

Issue #2311, opened 2023-03-29
[BUG] _TZE200_t1blo2bj switch doesn't work


kennycoder:
**Describe the bug**
 _TZE200_t1blo2bj switch doesn't work. Whenever the trigger is pressed, alarm doesn't turn on nor off. Looking at the logs I see the following:

`DEBUG (MainThread) [zigpy.zcl] [0x8971:1:0xef00] tuya_mcu_command: cluster_data=TuyaClusterData(endpoint_id=1, cluster_name='on_off', cluster_attr='on_off', attr_value=1, expect_reply=True, manufacturer=-1)`

Volume is set, melody is set, alarm is set to "True", Duration is also set.
Interestingly enough, if I go to the device configuration, TuyaMCUCluster, on_off and send "1", the siren turns on. Reading the value back returns "Bool.true" and once duration is passed, "Bool.false"

In the logs with manual command send I see the following:

`DEBUG (MainThread) [zigpy.zcl] [0x8971:1:0xef00] tuya_mcu_command: cluster_data=TuyaClusterData(endpoint_id=1, cluster_name='on_off', cluster_attr='on_off', attr_value=<Bool.true:1>, expect_reply=True, manufacturer=-1)`

So basically it's sending <Bool.true:1> (works) instead of just 1  (doesn't work)
doesn't work via regular trigger

**To Reproduce**
1. Go to ZHA devices
2. Click on the device you added ( _TZE200_t1blo2bj)
3. Scroll down to Toggle control switch
4. Nothing happens

**Expected behavior**
Siren turns on and off

**Screenshots**
n/a

<details>
{
[...]

PR #2313, opened 2023-03-29
Wrap command value in TuyaMCUSiren

javicalle:
Wrap the value in a `bool()` to fix DataType detection
Fixes: #2311


javicalle:
Duplicate of: #2312

PR #2323, opened 2023-04-08
Use shared CI and add ruff

dmulcahey:


zhaquirks/kof/kof_mr101z.py
```diff
@@ -56,7 +56,7 @@ async def command(self, command, *args, expect_reply=None, **kwargs):
         else:
             cmd_expect_reply = expect_reply
 
-        rsp = await super(NoReplyMixin, self).command(
+        rsp = await super().command(
```
dmulcahey, line 59: @puddly will this break anything??

puddly: Don't think so. This code is explicitly unit tested I think.

dmulcahey:
Checks need to be updated

PR #2324, opened 2023-04-08
Add support for Tuya TS0601 gas sensor GreenPower

akaindin:
Add support for Tuya TS0601 _TZE204_yojqa8xn Gas sensor with GreenPower

### DIFF
--- a/zhaquirks/tuya/ts0601_gas.py
+++ b/zhaquirks/tuya/ts0601_gas.py
@@ -4,7 +4,7 @@ import logging
 import zigpy.profiles.zha
 from zigpy.quirks import CustomCluster, CustomDevice
 import zigpy.types as t
-from zigpy.zcl.clusters.general import Basic, Groups, Ota, Scenes, Time
+from zigpy.zcl.clusters.general import Basic, GreenPowerProxy, Groups, Ota, Scenes, Time
 from zigpy.zcl.clusters.security import IasZone
 
 from zhaquirks import Bus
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts0601_gas.py
```diff
@@ -29,12 +32,9 @@
 class TuyaGasDetectorCluster(TuyaManufClusterAttributes):
     """Manufacturer Specific Cluster of the TS0601 gas detector."""
 
-    attributes = TuyaManufClusterAttributes.attributes.copy()
```
javicalle, line 32: Why to remove the `attributes.copy()`?
Regardless of the fact that it currently does nothing, in the future it may break if attributes are added to the parent classes.

zhaquirks/tuya/ts0601_gas.py
```diff
@@ -72,7 +72,6 @@ def update_zone_status(self, value):
 
 class TuyaGasDetector0601(CustomDevice):
     """TS0601 _TZE200_ggev5fsl quirk."""
-
```
javicalle, line 75: Restore the blank line.

javicalle (changes requested):

luigifcruz:
I can confirm this PR is working with my device. Thanks for this!

Screenshot 2023-06-17 at 12 34 27 PM

Matze89x:
When will this be officially adopted? I'm currently still using the custom version in Home Assistant. Thank you very much.

PR #2329, opened 2023-04-11
Use newer Tuya spell implementation for Lidl plug

TheJulianJES:
This removes a legacy/custom implementation of the Tuya spell for the LIDL plug.
By inheriting `EnchantedDevice`, it makes the LIDL plug use the newer Tuya spell implementation which also casts the spell again when re-configuring a device.

(The removed `TuyaBasicCluster` was never working anyway, as ZHA explicitly doesn't bind the `Basic` cluster)

Follow-up to:
- https://github.com/zigpy/zha-device-handlers/pull/2287

Addresses some parts of this for LIDL devices:
- https://github.com/zigpy/zha-device-handlers/issues/2271

(Not done here, but IMO this LIDL device should probably be moved into the Tuya TS011F file)

### DIFF
--- a/zhaquirks/lidl/ts011f_plug.py
+++ b/zhaquirks/lidl/ts011f_plug.py
@@ -1,13 +1,7 @@
 """LIDL TS011F plug."""
 from __future__ import annotations
 
-import asyncio
-import logging
-
-import zigpy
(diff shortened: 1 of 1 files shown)


javicalle (approved):
LGTM

Issue #2308, opened 2023-03-29
[Device Support Request]  GL-SD-001 by GLEDOPTO


ShogunMan:
The dimmer is of good quality and mostly works as it should. However current state seems not to be reported back to HA correctly.

If changing the dimming level from HA the dimmer dims to the correct level, but the level in HA is not updated to the new level.
It seems to be a timeout issue.

Over at zigbee2mqtt they had the same issue, but solved it: https://github.com/Koenkk/zigbee2mqtt/issues/8349

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4687, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0101",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x1000"
      ],
      "out_clusters": [
        "0x0019"
      ]
    }
  },
  "manufacturer": "GLEDOPTO",
  "model": "GL-SD-001",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant Container",
    "version": "2023.3.6",
    "dev": false,
    "hassio": false,
    "virtualenv": false,
    "python_version": "3.10.10",
    "docker": true,
    "arch": "aarch64",
    "timezone": "Europe/Oslo",
    "os_name": "Linux",
    "os_version": "5.15.84-v8+",
    "run_as_root": true
  },
  "custom_components": {
[...]

PR #2330, opened 2023-04-11
Ignore missing default reply for GLEDOPTO AC dimmer

TheJulianJES:
Fixes https://github.com/zigpy/zha-device-handlers/issues/2308

This GLEDOPTO dimmer does not generate default responses for the `LevelControl` cluster.
The `NoReplyMixin` fakes a reply for ZHA/zigpy.

In the [second commit](https://github.com/zigpy/zha-device-handlers/commit/051a879759139de8310508b613629338d388ec4c), `NoReplyMixin` is refactored out of the KOF quirk into the main quirks `__init__.py` file, as it's now used by different manufacturers. I also expect some other devices to need this.

(`NoReplyMixin` already has units tests with the KOF quirk)

### DIFF
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -390,6 +390,44 @@ class QuickInitDevice(CustomDevice):
         return device
 
 
+class NoReplyMixin:
+    """A simple mixin.
+
+    Allows a cluster to have configurable list of command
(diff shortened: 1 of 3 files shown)


javicalle (approved):
LGTM

PR #2332, opened 2023-04-11
Remove color-support for Tuya dimmer variation `_TZ3210_i680rtja`

ShogunMan:
Remove invalid color-support for `_TZ3210_i680rtja` Tuya dimmer.


TheJulianJES:
You tested if the dimmer works correctly with this?

ShogunMan:
Yes, it only removes the color cluster from the endpoints.  As the _TZ3210_i680rtja is a single channel led controller it is not needed.

PR #2340, opened 2023-04-14
Add Philips RDM002 Tap Dial switch support, refactor Hue remotes

fholzer:
Add suport for [Philps Hue Tap Dial Switch](https://www.philips-hue.com/en-us/p/hue-tap-dial-switch/046677578800) (RDM002).
This PR also refactors all Hue remotes to "generate" the device triggers and removes a lot of duplicated code.
The device triggers changed are backwards-compatible (verified in tests).

### DIFF
--- a/zhaquirks/philips/__init__.py
+++ b/zhaquirks/philips/__init__.py
@@ -197,7 +197,7 @@ class PhilipsRemoteCluster(CustomCluster):
         ] = None,
     ):
         """Handle the cluster command."""
-        self.debug(
+        _LOGGER.debug(
             "%s - handle_cluster_request tsn: [%s] command id: %s - args: [%s]",
             self.__class__.__name__,
             hdr.tsn,
--- a/zhaquirks/philips/__init__.py
+++ b/zhaquirks/philips/__init__.py
@@ -1,6 +1,7 @@
 """Module for Philips quirks implementations."""
 
 import asyncio
+import itertools
 import logging
 import time
 from typing import Any, Optional, Union
@@ -15,7 +16,11 @@ from zhaquirks.const import (
     ARGS,
     BUTTON,
     COMMAND,
+    COMMAND_HOLD,
     COMMAND_ID,
+    COMMAND_M_LONG_RELEASE,
+    COMMAND_M_SHORT_RELEASE,
+    COMMAND_PRESS,
     DIM_DOWN,
     DIM_UP,
     DOUBLE_PRESS,
@@ -35,41 +40,6 @@ from zhaquirks.const import (
 PHILIPS = "Philips"
 SIGNIFY = "Signify Netherlands B.V."
 _LOGGER = logging.getLogger(__name__)
-
-HUE_REMOTE_DEVICE_TRIGGERS = {
-    (SHORT_PRESS, TURN_ON): {COMMAND: "on_press"},
-    (SHORT_PRESS, TURN_OFF): {COMMAND: "off_press"},
-    (SHORT_PRESS, DIM_UP): {COMMAND: "up_press"},
-    (SHORT_PRESS, DIM_DOWN): {COMMAND: "down_press"},
-    (LONG_PRESS, TURN_ON): {COMMAND: "on_hold"},
-    (LONG_PRESS, TURN_OFF): {COMMAND: "off_hold"},
-    (LONG_PRESS, DIM_UP): {COMMAND: "up_hold"},
-    (LONG_PRESS, DIM_DOWN): {COMMAND: "down_hold"},
-    (DOUBLE_PRESS, TURN_ON): {COMMAND: "on_double_press"},
-    (DOUBLE_PRESS, TURN_OFF): {COMMAND: "off_double_press"},
-    (DOUBLE_PRESS, DIM_UP): {COMMAND: "up_double_press"},
-    (DOUBLE_PRESS, DIM_DOWN): {COMMAND: "down_double_press"},
-    (TRIPLE_PRESS, TURN_ON): {COMMAND: "on_triple_press"},
-    (TRIPLE_PRESS, TURN_OFF): {COMMAND: "off_triple_press"},
-    (TRIPLE_PRESS, DIM_UP): {COMMAND: "up_triple_press"},
-    (TRIPLE_PRESS, DIM_DOWN): {COMMAND: "down_triple_press"},
-    (QUADRUPLE_PRESS, TURN_ON): {COMMAND: "on_quadruple_press"},
-    (QUADRUPLE_PRESS, TURN_OFF): {COMMAND: "off_quadruple_press"},
-    (QUADRUPLE_PRESS, DIM_UP): {COMMAND: "up_quadruple_press"},
-    (QUADRUPLE_PRESS, DIM_DOWN): {COMMAND: "down_quadruple_press"},
-    (QUINTUPLE_PRESS, TURN_ON): {COMMAND: "on_quintuple_press"},
-    (QUINTUPLE_PRESS, TURN_OFF): {COMMAND: "off_quintuple_press"},
-    (QUINTUPLE_PRESS, DIM_UP): {COMMAND: "up_quintuple_press"},
-    (QUINTUPLE_PRESS, DIM_DOWN): {COMMAND: "down_quintuple_press"},
-    (SHORT_RELEASE, TURN_ON): {COMMAND: "on_short_release"},
-    (SHORT_RELEASE, TURN_OFF): {COMMAND: "off_short_release"},
-    (SHORT_RELEASE, DIM_UP): {COMMAND: "up_short_release"},
-    (SHORT_RELEASE, DIM_DOWN): {COMMAND: "down_short_release"},
-    (LONG_RELEASE, TURN_ON): {COMMAND: "on_long_release"},
-    (LONG_RELEASE, TURN_OFF): {COMMAND: "off_long_release"},
-    (LONG_RELEASE, DIM_UP): {COMMAND: "up_long_release"},
-    (LONG_RELEASE, DIM_DOWN): {COMMAND: "down_long_release"},
-}
 
 
 class PhilipsOccupancySensing(CustomCluster):
@@ -138,6 +108,39 @@ class ButtonPressQueue:
         self._task = asyncio.ensure_future(self._job())
 
 
+class Button:
+    """Represents a remote button, including string literals used in triggers and actions."""
+
+    def __init__(
+        self, id: str, trigger: str | None = None, action: str | None = None
+    ) -> None:
+        """Init."""
+        self.id = id
+        self.trigger = trigger or id
+        self.action = action or id
+
+    def __repr__(self) -> str:
+        """Repr."""
+        return f"<Button id={self.id}, trigger={self.trigger}, action={self.action}>"
+
+
+class PressType:
+    """Represents a remote button press-type, including string literals used in triggers and actions."""
+
+    def __init__(
+        self, name: str, action: str, arg: str | None = None, trigger: str | None = None
+    ) -> None:
+        """Init."""
+        self.name = name
+        self.action = action
+        self.arg = arg or action
+        self.trigger = trigger or name
+
+    def __repr__(self) -> str:
+        """Repr."""
+        return f"<PressType name={self.name}, action={self.action}, trigger={self.trigger}, arg={self.arg}>"
+
+
 class PhilipsRemoteCluster(CustomCluster):
     """Philips remote cluster."""
 
@@ -152,15 +155,35 @@ class PhilipsRemoteCluster(CustomCluster):
                 "param2": t.uint24_t,
                 "press_type": t.uint8_t,
                 "param4": t.uint8_t,
-                "param5": t.uint8_t,
-                "param6": t.uint8_t,
+                "param5": t.uint16_t,
             },
             False,
             is_manufacturer_specific=True,
         )
     }
-    BUTTONS = {1: "on", 2: "up", 3: "down", 4: "off"}
-    PRESS_TYPES = {0: "press", 1: "hold", 2: "short_release", 3: "long_release"}
+
+    BUTTONS: dict[int, Button] = {}
+
+    PRESS_TYPES: dict[int, PressType] = {
+        # We omit "short_press" and "short_release" on purpose, so it
+        # won't interfere with simulated multi-press events. We emit
+        # them in the multi-press code later on.
+        # 0: SHORT_PRESS,
+        1: PressType(LONG_PRESS, COMMAND_HOLD),
+        # 2: SHORT_RELEASE,
+        3: PressType(LONG_RELEASE, COMMAND_M_LONG_RELEASE),
+    }
+
+    MULTI_PRESS_EVENTS: dict[int, PressType] = {
+        2: PressType(DOUBLE_PRESS, "double_press"),
+        3: PressType(TRIPLE_PRESS, "triple_press"),
+        4: PressType(QUADRUPLE_PRESS, "quadruple_press"),
+        5: PressType(QUINTUPLE_PRESS, "quintuple_press"),
+    }
+    SIMULATE_SHORT_EVENTS = [
+        PressType(SHORT_PRESS, COMMAND_PRESS),
+        PressType(SHORT_RELEASE, COMMAND_M_SHORT_RELEASE),
+    ]
 
     button_press_queue = ButtonPressQueue()
 
@@ -174,48 +197,133 @@ class PhilipsRemoteCluster(CustomCluster):
         ] = None,
     ):
         """Handle the cluster command."""
-        _LOGGER.debug(
-            "PhilipsRemoteCluster - handle_cluster_request tsn: [%s] command id: %s - args: [%s]",
+        self.debug(
+            "%s - handle_cluster_request tsn: [%s] command id: %s - args: [%s]",
+            self.__class__.__name__,
             hdr.tsn,
             hdr.command_id,
             args,
         )
 
-        button = self.BUTTONS.get(args[0], args[0])
-        press_type = self.PRESS_TYPES.get(args[2], args[2])
+        button = self.BUTTONS.get(args[0])
+        _LOGGER.debug(
+            "%s - handle_cluster_request button id: [%s], button name: [%s]",
+            self.__class__.__name__,
+            args[0],
+            button,
+        )
+        # Bail on unknown buttons. (This gets rid of dial button "presses")
+        if button is None:
+            return
+
+        press_type = self.PRESS_TYPES.get(args[2])
+        if (
+            press_type is None
+            and self.SIMULATE_SHORT_EVENTS is not None
+            and args[2] == 2
+        ):
+            press_type = self.SIMULATE_SHORT_EVENTS[1]
+        if press_type is None:
+            return
+
+        duration = args[4]
+        _LOGGER.debug(
+            "%s - handle_cluster_request button press type: [%s], duration: [%s]",
+            self.__class__.__name__,
+            press_type,
+            duration,
+        )
 
         event_args = {
-            BUTTON: button,
-            PRESS_TYPE: press_type,
+            BUTTON: button.id,
+            PRESS_TYPE: press_type.arg,
             COMMAND_ID: hdr.command_id,
+            "duration": duration,
             ARGS: args,
         }
 
         def send_press_event(click_count):
             _LOGGER.debug(
-                "PhilipsRemoteCluster - send_press_event click_count: [%s]", click_count
+                "%s - send_press_event click_count: [%s]",
+                self.__class__.__name__,
+                click_count,
             )
             press_type = None
             if click_count == 1:
-                press_type = "press"
-            elif click_count == 2:
-                press_type = "double_press"
-            elif click_count == 3:
-                press_type = "triple_press"
-            elif click_count == 4:
-                press_type = "quadruple_press"
-            elif click_count > 4:
-                press_type = "quintuple_press"
+                press_type = self.PRESS_TYPES.get(0) or self.SIMULATE_SHORT_EVENTS[0]
+            elif click_count > 1:
+                press_type = self.MULTI_PRESS_EVENTS[min(click_count, 5)]
 
-            if press_type:
+            _LOGGER.debug(
+                "%s - send_press_event evaluated press_type: [%s]",
+                self.__class__.__name__,
+                press_type,
+            )
+            if press_type is not None:
                 # Override PRESS_TYPE
-                event_args[PRESS_TYPE] = press_type
-                action = f"{button}_{press_type}"
+                event_args[PRESS_TYPE] = press_type.arg
+                event_args[ARGS] = list(event_args[ARGS])
+                event_args[ARGS][2] = 0 if click_count < 2 else 2 + min(click_count, 5)
+                action = f"{button.action}_{press_type.action}"
+                _LOGGER.debug(
+                    "%s - send_press_event emitting action: [%s] event_args: %s",
+                    self.__class__.__name__,
+                    action,
+                    event_args,
+                )
                 self.listener_event(ZHA_SEND_EVENT, action, event_args)
 
+            # simulate short release event, if needed for this device type
+            if (
+                press_type.name == SHORT_PRESS
+                and self.SIMULATE_SHORT_EVENTS is not None
+            ):
+                press_type = self.PRESS_TYPES.get(2) or self.SIMULATE_SHORT_EVENTS[1]
+                sim_event_args = event_args.copy()
+                sim_event_args[PRESS_TYPE] = press_type.arg
+                sim_event_args[ARGS] = sim_event_args[ARGS].copy()
+                sim_event_args[ARGS][2] = 2
+                action = f"{button.action}_{press_type.action}"
+                _LOGGER.debug(
+                    "%s - send_press_event emitting simulated action: [%s]",
+                    self.__class__.__name__,
+                    action,
+                )
+                self.listener_event(ZHA_SEND_EVENT, action, sim_event_args)
+
         # Derive Multiple Presses
-        if press_type == "press":
-            self.button_press_queue.press(send_press_event, button)
+        if press_type.name == SHORT_RELEASE:
+            self.button_press_queue.press(send_press_event, button.id)
         else:
-            action = f"{button}_{press_type}"
+            action = f"{button.action}_{press_type.action}"
             self.listener_event(ZHA_SEND_EVENT, action, event_args)
+
+    @classmethod
+    def generate_device_automation_triggers(cls, additional=None):
+        """Generate automation triggers based on device buttons and press-types."""
+        triggers = {}
+        for button in cls.BUTTONS.values():
+            press_types = [cls.PRESS_TYPES.values()]
+            if cls.SIMULATE_SHORT_EVENTS is not None:
+                press_types.append(cls.SIMULATE_SHORT_EVENTS)
+                press_types.append(cls.MULTI_PRESS_EVENTS.values())
+            for press_type in itertools.chain(*press_types):
+                triggers[(press_type.trigger, button.trigger)] = {
+                    COMMAND: f"{button.action}_{press_type.action}"
+                }
+
+        if additional:
+            triggers.update(additional)
+
+        return triggers
+
+
+class PhilipsRwlRemoteCluster(PhilipsRemoteCluster):
+    """Philips remote cluster for RWL devices."""
+
+    BUTTONS = {
+        1: Button("on", TURN_ON),
+        2: Button("up", DIM_UP),
+        3: Button("down", DIM_DOWN),
+        4: Button("off", TURN_OFF),
+    }
--- a/zhaquirks/philips/rdm002.py
+++ b/zhaquirks/philips/rdm002.py
@@ -1,4 +1,4 @@
-"""Philips ROM001 device."""
+"""Signify RDM002 device."""
 
 from zigpy.profiles import zha
 from zigpy.quirks import CustomDevice
@@ -15,17 +15,24 @@ from zigpy.zcl.clusters.general import (
 from zigpy.zcl.clusters.lightlink import LightLink
 
 from zhaquirks.const import (
-    COMMAND_HOLD,
-    COMMAND_ON,
+    BUTTON_1,
+    BUTTON_2,
+    BUTTON_3,
+    BUTTON_4,
+    CLUSTER_ID,
+    COMMAND,
+    COMMAND_STEP_ON_OFF,
     DEVICE_TYPE,
+    DIM_DOWN,
+    DIM_UP,
+    ENDPOINT_ID,
     ENDPOINTS,
     INPUT_CLUSTERS,
-    LONG_PRESS,
-    LONG_RELEASE,
     MODELS_INFO,
     OUTPUT_CLUSTERS,
+    PARAMS,
     PROFILE_ID,
-    TURN_ON,
+    SHORT_PRESS,
 )
 from zhaquirks.philips import (
     PHILIPS,
@@ -33,39 +40,44 @@ from zhaquirks.philips import (
     Button,
     PhilipsBasicCluster,
     PhilipsRemoteCluster,
-    PressType,
 )
 
-DEVICE_SPECIFIC_UNKNOWN = 64512
+DIAL_TRIGGERS = {
+    (SHORT_PRESS, DIM_UP): {
+        COMMAND: COMMAND_STEP_ON_OFF,
+        CLUSTER_ID: 8,
+        ENDPOINT_ID: 1,
+        PARAMS: {"step_mode": 0},
+    },
+    (SHORT_PRESS, DIM_DOWN): {
+        COMMAND: COMMAND_STEP_ON_OFF,
+        CLUSTER_ID: 8,
+        ENDPOINT_ID: 1,
+        PARAMS: {"step_mode": 1},
+    },
+}
 
 
-class PhilipsRom001RemoteCluster(PhilipsRemoteCluster):
-    """Philips remote cluster for ROM001."""
+class PhilipsRdm002RemoteCluster(PhilipsRemoteCluster):
+    """Philips remote cluster for RDM002."""
 
     BUTTONS = {
-        1: Button(COMMAND_ON, TURN_ON, COMMAND_ON),
-    }
-
-    PRESS_TYPES: dict[int, PressType] = {
-        # We omit "short_press" and "short_release" on purpose, so it
-        # won't interfere with simulated multi-press events. We emit
-        # them in the multi-press code later on.
-        # 0: SHORT_PRESS,
-        1: PressType(LONG_PRESS, COMMAND_HOLD),
-        # 2: SHORT_RELEASE,
-        3: PressType(LONG_RELEASE, "long_release", "hold_release"),
+        1: Button(BUTTON_1),
+        2: Button(BUTTON_2),
+        3: Button(BUTTON_3),
+        4: Button(BUTTON_4),
     }
 
 
-class PhilipsROM001(CustomDevice):
-    """Philips ROM001 device."""
+class PhilipsRDM002(CustomDevice):
+    """Philips RDM002 device."""
 
     signature = {
         #  <SimpleDescriptor endpoint=1 profile=260 device_type=2096
         #  device_version=1
         #  input_clusters=[0, 1, 3, 64512, 4096]
         #  output_clusters=[25, 0, 3, 4, 6, 8, 5, 4096]>
-        MODELS_INFO: [(PHILIPS, "ROM001"), (SIGNIFY, "ROM001")],
+        MODELS_INFO: [(PHILIPS, "RDM002"), (SIGNIFY, "RDM002")],
         ENDPOINTS: {
             1: {
                 PROFILE_ID: zha.PROFILE_ID,
@@ -74,7 +86,7 @@ class PhilipsROM001(CustomDevice):
                     Basic.cluster_id,
                     PowerConfiguration.cluster_id,
                     Identify.cluster_id,
-                    DEVICE_SPECIFIC_UNKNOWN,
+                    PhilipsRdm002RemoteCluster.cluster_id,
                     LightLink.cluster_id,
                 ],
                 OUTPUT_CLUSTERS: [
@@ -95,12 +107,12 @@ class PhilipsROM001(CustomDevice):
         ENDPOINTS: {
             1: {
                 PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: zha.DeviceType.NON_COLOR_SCENE_CONTROLLER,
+                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
                 INPUT_CLUSTERS: [
                     PhilipsBasicCluster,
                     PowerConfiguration.cluster_id,
                     Identify.cluster_id,
-                    PhilipsRom001RemoteCluster,
+                    PhilipsRdm002RemoteCluster,
                     LightLink.cluster_id,
                 ],
                 OUTPUT_CLUSTERS: [
@@ -118,5 +130,5 @@ class PhilipsROM001(CustomDevice):
     }
 
     device_automation_triggers = (
-        PhilipsRom001RemoteCluster.generate_device_automation_triggers()
+        PhilipsRdm002RemoteCluster.generate_device_automation_triggers(DIAL_TRIGGERS)
     )
--- a/tests/test_philips.py
+++ b/tests/test_philips.py
@@ -1,0 +1,502 @@
+"""Tests for Philips quirks."""
+
+from unittest import mock
+
+import pytest
+from zigpy.zcl.foundation import ZCLHeader
+
+import zhaquirks
+from zhaquirks.const import (
+    COMMAND,
+    COMMAND_HOLD,
+    COMMAND_M_LONG_RELEASE,
+    DIM_DOWN,
+    DIM_UP,
+    DOUBLE_PRESS,
+    LONG_PRESS,
+    LONG_RELEASE,
+    QUADRUPLE_PRESS,
+    QUINTUPLE_PRESS,
+    RIGHT,
+    SHORT_PRESS,
+    SHORT_RELEASE,
+    TRIPLE_PRESS,
+    TURN_OFF,
+    TURN_ON,
+)
+import zhaquirks.philips
+from zhaquirks.philips import ButtonPressQueue
+from zhaquirks.philips.rdm001 import PhilipsROM001 as PhilipsRDM001
+from zhaquirks.philips.rom001 import PhilipsROM001
+from zhaquirks.philips.rwl022 import PhilipsRWL022
+from zhaquirks.philips.rwlfirstgen import PhilipsRWLFirstGen, PhilipsRWLFirstGen2
+
+zhaquirks.setup()
+
+
+@pytest.mark.parametrize(
+    "classes, triggers",
+    (
+        (
+            [PhilipsRWLFirstGen, PhilipsRWLFirstGen2, PhilipsRWL022],
+            {
+                (SHORT_PRESS, TURN_ON): {COMMAND: "on_press"},
+                (SHORT_PRESS, TURN_OFF): {COMMAND: "off_press"},
+                (SHORT_PRESS, DIM_UP): {COMMAND: "up_press"},
+                (SHORT_PRESS, DIM_DOWN): {COMMAND: "down_press"},
+                (LONG_PRESS, TURN_ON): {COMMAND: "on_hold"},
+                (LONG_PRESS, TURN_OFF): {COMMAND: "off_hold"},
+                (LONG_PRESS, DIM_UP): {COMMAND: "up_hold"},
+                (LONG_PRESS, DIM_DOWN): {COMMAND: "down_hold"},
+                (DOUBLE_PRESS, TURN_ON): {COMMAND: "on_double_press"},
+                (DOUBLE_PRESS, TURN_OFF): {COMMAND: "off_double_press"},
+                (DOUBLE_PRESS, DIM_UP): {COMMAND: "up_double_press"},
+                (DOUBLE_PRESS, DIM_DOWN): {COMMAND: "down_double_press"},
+                (TRIPLE_PRESS, TURN_ON): {COMMAND: "on_triple_press"},
+                (TRIPLE_PRESS, TURN_OFF): {COMMAND: "off_triple_press"},
+                (TRIPLE_PRESS, DIM_UP): {COMMAND: "up_triple_press"},
+                (TRIPLE_PRESS, DIM_DOWN): {COMMAND: "down_triple_press"},
+                (QUADRUPLE_PRESS, TURN_ON): {COMMAND: "on_quadruple_press"},
+                (QUADRUPLE_PRESS, TURN_OFF): {COMMAND: "off_quadruple_press"},
+                (QUADRUPLE_PRESS, DIM_UP): {COMMAND: "up_quadruple_press"},
+                (QUADRUPLE_PRESS, DIM_DOWN): {COMMAND: "down_quadruple_press"},
+                (QUINTUPLE_PRESS, TURN_ON): {COMMAND: "on_quintuple_press"},
+                (QUINTUPLE_PRESS, TURN_OFF): {COMMAND: "off_quintuple_press"},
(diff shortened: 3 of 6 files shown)


zhaquirks/philips/__init__.py
```diff
@@ -145,8 +145,7 @@ class PhilipsRemoteCluster(CustomCluster):
                 "param2": t.uint24_t,
                 "press_type": t.uint8_t,
                 "param4": t.uint8_t,
-                "param5": t.uint8_t,
-                "param6": t.uint8_t,
+                "duration": t.uint16_t,
```
fholzer, line 148: This seems to be supported by many(/all?) of Philips' button devices. It's measured in 10th of seconds.

Confirmed that for ROM001, RWL022, RDM002.

zhaquirks/philips/__init__.py
```diff
@@ -168,25 +167,30 @@ def handle_cluster_request(
     ):
         """Handle the cluster command."""
         _LOGGER.debug(
-            "PhilipsRemoteCluster - handle_cluster_request tsn: [%s] command id: %s - args: [%s]",
+            "%s - handle_cluster_request tsn: [%s] command id: %s - args: [%s]",
+            self.__class__.__name__,
```
fholzer, line 171: Not sure if you are fine with this. Main purpose is to show `PhilipsRdmRemoteCluster` instead of `PhilipsRemoteCluster` in the logs for RDM002 devices. Can remove it if you want. Same a couple lines down.

zhaquirks/philips/__init__.py
fholzer: Also same for RDM001

zhaquirks/philips/rdm002.py
```diff
@@ -0,0 +1,144 @@
...
+        2: BUTTON_2,
+        3: BUTTON_3,
+        4: BUTTON_4,
+    }
+    PRESS_TYPES = {
+        0: COMMAND_PRESS,
+        1: COMMAND_HOLD,
+        2: SHORT_RELEASE,
+        3: LONG_RELEASE,
+    }
```
TheJulianJES, line 60: The press types are exactly the same as in `PhilipsRemoteCluster`, right?
There, the constants aren't used though. Can you change that there (and remove the above code)?:
https://github.com/zigpy/zha-device-handlers/blob/ffdf2fbfe559f74e6fc00d61130b8025e18141ec/zhaquirks/philips/__init__.py#L162

zhaquirks/philips/__init__.py
TheJulianJES: Hmm, I guess it should be fine.

Can you also use `self.debug` instead of `_LOGGER.debug` and remove the logger? 
That should include the NWK address of the actual device in the logs.

zhaquirks/philips/__init__.py
TheJulianJES: Is there a chance that you also have a `RWL020` or `RWL021` dimmer (one of the older ones) to test this with?

If not, I'll search for one of mine to test this with.

MattWestb:
The only test  is not liking is `zhaquirks/philips/__init__.py` and line 3 is not used `import logging` then the logging method is being changed.

MattWestb:
Ops by the way:
Servus !!!

zhaquirks/philips/rdm002.py
fholzer: The `(SHORT|LONG)_RELEASE` constants used in `rdm002.py` are set to `"remote_button_short_release"` and `"remote_button_long_release"`, while the string literals `"short_release"` and `"long_release"` are used in `__init__.py`. `"short_release"` and `"long_release"` correspond to `COMMAND_M_SHORT_RELEASE` and `COMMAND_M_LONG_RELEASE`. Not entirely sure if something would break when changing the `PhilipsRemoteCluster`'s `PRESS_TYPE` values. Which constants would you prefer?

Also, I'm not sure if changing these values would break anything for existing users. Would it *not* break anything if we also change `PhilipsROM001`'s `device_automation_triggers` values?

Actually I was thinking about changing the implementation, introducing new `PhilipsRDMRemoteCluster` and `PhilipsRWLRemoteCluster` classes inheriting from `PhilipsRemoteCluster`, then change `PhilipsROM001` `PhilipsROM002` to inherit from `PhilipsRDMRemoteCluster`, and `PhilipsRWL022` from `PhilipsRWLRemoteCluster`, but as I'm not sure this can be done without breaking anything for existing users when changing things like button name and press type values, I haven't started working on this.

Servus MattWestb :)

fholzer:
lint issue fixed

zhaquirks/philips/__init__.py
MattWestb: RWL020 is the first gen Dimmer switch ?
That i have in the box for bad Zigbee devices.
@fholzer If you is near W1110 you can having my for testing its easy picking up then im at home if you like (U3 Enkplatz).

MW

fholzer: Yeah, that would be appreciated. From the RWLs I only have the RWL022. My email is ferdinand.holzer ( at ) gmail.com. Get in touch and we can arrange a time later today if possible.

MattWestb: Mailed !!

zhaquirks/philips/rdm002.py
TheJulianJES: > The (SHORT|LONG)_RELEASE constants used in rdm002.py are set to "remote_button_short_release" and "remote_button_long_release", while the string literals "short_release" and "long_release" are used

Ah, I missed that.  The [`HUE_REMOTE_DEVICE_TRIGGERS`](https://github.com/zigpy/zha-device-handlers/blob/c01893fc3ed47bdce3f1d447ce13e18bb46ceec8/zhaquirks/philips/__init__.py#L38) that are used for all(?) other Hue remotes at the moment use those suffixes in the triggers.
IMO, the constants used in `rdm002.py` with `remote_button_xxx_release` aren't optimal. Personally, I think the press types that are currently used make more sense:
https://github.com/zigpy/zha-device-handlers/blob/ffdf2fbfe559f74e6fc00d61130b8025e18141ec/zhaquirks/philips/__init__.py#L162

Do you maybe want to remove the overriding `PRESS_TYPES` from your class, so we use all same press types for all devices?

And btw., your device triggers don't actually include short/long release ones for the buttons. Not sure if intended or not?

TheJulianJES: > Actually I was thinking about changing the implementation, introducing new PhilipsRDMRemoteCluster...

Hmm, the current implementation isn't optimal, as `rdm001` also copies the complete custom cluster.
RWL022 also uses the "wrong button types".

If you want to have a go at changing the implementation, feel free to do so. If possible, we shouldn't really break device triggers / event data (used for blueprints).
The RWL022 event data is somewhat wrong though and should be changed maybe(?)

zhaquirks/philips/__init__.py
TheJulianJES: To get back to this again: I've just thought about this the other day and while including the Nwk address of the device is nice, this also changes the logging prefix to look like it's coming from zigpy (and not zhaquirks).

This will (hopefully) be generally addressed in the future, as quirks sometimes use the zigpy mixin and sometimes a new logger at the moment.

For now, please use a custom logger (again). (Sorry for the back and forth with this one)

fholzer: No worries. Done

zhaquirks/philips/__init__.py
fholzer: Got Matt's RWL020, though have to wait for new CR2450s to arrive.

zhaquirks/philips/rdm002.py
fholzer: Off the bat, sorry for the long post. I'd like to understand how I can help.

> Personally, I think the press types that are currently used make more sense: [...]

Is it OK to use constants here, or do you prefer string literals? Could still update `PRESS_TYPES` in `__init__.py` to use constants with the same value as the string literals currently used.

> And btw., your device triggers don't actually include short/long release ones for the buttons. Not sure if intended or not?

You're right. Also realized this and took a note below. Will have a look in the next couple of days.

> RWL022 also uses the "wrong button types".

What do you mean?

#### `HUE_REMOTE_DEVICE_TRIGGERS`
* ...is used by all RWL type devices, which are the 4-button dimmer remote controls.
* ROM001, the single button remote, is using separately defined `device_automation_triggers`, which is fine.
* And [RDM001](https://zigbee.blakadder.com/Philips_RDM001.html), the wall switch, is also using separately defined `device_automation_triggers`, but I think the trigger naming could be improved. I.e. instead of `TURN_ON`/`RIGHT`, to either use `LEFT`/`RIGHT` or `BUTTON_1`/`BUTTON_2`.

#### `PhilipsRemoteCluster`
* ... is defined in `__init__.py` and used directly for ROM and RWL devices.
* RDM002 currently inherits from it and overrides the `BUTTONS` and `PRESS_TYPES` without providing triggers for `LONG_PRESS` or `(SHORT|LONG)_RELEASE` events. (Need to check why i didn't add those - it's been some time since i wrote this).
* And finally we have RDM001, which defined it's own `PhilipsRemoteCluster`. Furthermore RDM001's `device_automation_triggers` lists multi-press triggers (e.g. double-, triple-, etc. -press events ) but I wonder how that works, as it's `PhilipsRemoteCluster` is missing the `send_press_event` function that generates similar events in `__init__.py`'s `PhilipsRemoteCluster`. For easy comparison of the two classes, I made [a diff here](https://gist.github.com/fholzer/f485009437d2cd50ef4265ca72e54490).

#### Questions
I would love to fix & streamline all that, but as i'm new to the code base i have a couple of questions:
* My understanding is that changing keys in the `device_automation_triggers` dict influences only how the triggers are presented in the UI, while changing the value (ie. `COMMAND`) would break things for existing users. Is that correct?
  * Even so, changing keys might break things that directly listen for zha_events and inspect the event arguments. Is that acceptable or not?
* Would you like to have a single `PhilipsRemoteCluster` defined in `__init__.py` and used all applicable devices instead of copying the class to device-specific files and use overrides where needed. Then all devices would support features like synthetic multi-press events.

fholzer: I still need to test all this, but already pushed my changes so you can have a look. The `device_automation_triggers` is unchanged for all existing devices.

fholzer: To get back to your original statement about `PRESS_TYPES` not being constants. Right, so in this new implementation I derive the `device_automation_triggers` from the `PRESS_TYPES`, and those `device_automation_triggers` were using these constants before.

TheJulianJES:
Mostly just a note for me when I hopefully get back to this PR:
See if this fixes https://github.com/zigpy/zha-device-handlers/issues/2571 or if the fix can be easily made with this PR or if it should be addressed in a different PR.

Another note:
Although this likely shouldn't be changed in this PR, as it's a breaking change for device triggers, the "buttons" for the new-style dimmer (RWL022) should probably be updated, as they just copy the first-gen device triggers.

(Also sorry for not being able to have a close look yet. Very limited on time at the moment. Your changes look good from a quick glance and are very much appreciated!)

EDIT: Basically, existing device triggers have to stay for now. The `zha_event` `command`s should also stay the same, so that blueprints (and other "weird" stuff) doesn't break.
Breaking changes have to be in another PR (and I don't know if we can currently make any breaking changes at all due to HA's 6 month guideline).

(I'll put this into draft for now, so it's not accidentally merged yet.)

zhaquirks/philips/rdm002.py
TheJulianJES: > Is it OK to use constants here, or do you prefer string literals? Could still update PRESS_TYPES in __init__.py to use constants with the same value as the string literals currently used.

No real preference for now. Constants are nicer if used in multiple places, but string literals should also be fine if it's used just once.

---

>> RWL022 also uses the "wrong button types".

> What do you mean?

The device has a shared on/off button (that's just an "on" button in the triggers + brightness buttons (fine) + a "hue" button ("off" button in the triggers).
It shouldn't be changed in this PR though, as it would break blueprints (react on `command` in `zha_event`) and device triggers, as those should also be "renamed".


--- 

> And [RDM001](https://zigbee.blakadder.com/Philips_RDM001.html), the wall switch, is also using separately defined device_automation_triggers, but I think the trigger naming could be improved. I.e. instead of TURN_ON/RIGHT, to either use LEFT/RIGHT or BUTTON_1/BUTTON_2.

Yes, it would break device triggers + blueprints though, as far as I can see.
All Hue remote related device triggers should probably be changed later in a separate PR (that's clearly marked as a breaking change). I'm not sure how much we're allowed to "break" now though, as there's some 6 month breaking change guideline in HA now. Not sure if/how that applies to libraries.

---

> And finally we have RDM001, which defined it's own PhilipsRemoteCluster. Furthermore RDM001's device_automation_triggers lists multi-press triggers (e.g. double-, triple-, etc. -press events ) but I wonder how that works, as it's PhilipsRemoteCluster is missing the send_press_event function that generates similar events in __init__.py's PhilipsRemoteCluster. For easy comparison of the two classes, I made [a diff here](https://gist.github.com/fholzer/f485009437d2cd50ef4265ca72e54490).

That's a good question on why those triggers are there if they can never be fired.

---

> My understanding is that changing keys in the device_automation_triggers dict influences only how the triggers are presented in the UI, while changing the value (ie. COMMAND) would break things for existing users. Is that correct?
Even so, changing keys might break things that directly listen for zha_events and inspect the event arguments. Is that acceptable or not?

I'd have to check at this point, but I think you're right. For device triggers to not break in automations, the keys (type + subtype) have to stay the same. Technically, the command can change IIRC.

A lot of blueprints listen to the `command` though and we shouldn't really break/change those for now, so (sadly) we should keep both the `command`and the device automation trigger the same. I'll need to properly look at your changes later, but do your current changes have both stay the same?

("Edit": Hmm, from a quick look, I think the commands are changed now, as both `BUTTONS` and `PRESS_TYPES` are changed.
Although I prefer your changes, commands (and device triggers) should stay the same for now (to be non-breaking).

(Also, for the "multi-press" stuff (ll `177-185` in `__init__.py`), should the appropriate press type come from `PRESS_TYPES` instead of string literals? Would be a bit cleaner maybe?)

---

> Would you like to have a single PhilipsRemoteCluster defined in __init__.py and used all applicable devices instead of copying the class to device-specific files and use overrides where needed. Then all devices would support features like synthetic multi-press events.

(I always found the idea of having synthetic multi-press events in the Philips quirks a bit weird, but I guess that's what we're doing for now.)

Since the class would mostly/somewhat be the same code for the devices, having one class in a central places makes the most sense to me. (Seems like that's what you did for now)

cod3gen:
Dimming wheel acts as a switch with this quirk, would it be possible to use ROTATED LEFT/RIGHT instead? And maybe also include params like step_size?

fholzer:
> Dimming wheel acts as a switch with this quirk, would it be possible to use ROTATED LEFT/RIGHT instead? And maybe also include params like step_size?

Will have a look this week.

Also, the RWL020 remote I got from Matt doesn't support the long press events, so will have to remove those as well. Not sure about RWL021 though.

> Breaking changes have to be in another PR

Will also look into changing my PR so that it keeps all existing PRESS_TYPES, commands and device_automation_triggers unchanged.

MattWestb:
I think the RWL020 is running on the latest firmware (=old then its one early HUE device) https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/OTA-Image-Types---Firmware-versions#philips-hue-signify was the latest upgrade i was doing on it if i remember right.
Perhaps Julian have "found" some newer version in his HUE system ?

fholzer:
@cod3gen for me rotating the dial on my RDM002 using the code in this MR yields both, button press events, and "Steps with On Off" events, the latter includes direction & step size. I just pushed some changes. Can you check these?

@TheJulianJES wrt the dial, can you give guidance? In my last commit just now I suppose these duplicate button press events rotating the dial yields, and added device automation triggers. I see COMMAND_STEP_ON_OFF is used by a couple of quirks([1](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/lutron/lzl4bwhl01remote.py#L111), [2](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/lds/cctswitch.py#L113)), though always in combination with "press" events. That would make the UI suggest the wheel is something that is "pressed":
image

Is there some better alternative? I tried finding the strings available in the UI. Found [this](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/strings.json#L228), but not sure if I am looking in the right place. Also, it doesn't list anything that's better suited.

P.S.: Regarding how to deal with breaking changes... I could make a version of this PR that only adds changes needed for RDM002, and keep all other devices unchanged. If you also want to merge those changes to other devices, these could be dealt with separately, as per breaking changes guideline.

fholzer:
With the requirement  to not break anything, device automation trigger keys and values, params, parameter names, button names, press types, I'm not sure this can be done in one generic class without having a lot of code just for handling these "exceptions". I'm leaning towards discarding all changes except those for RDM002.

zhaquirks/philips/rdm002.py
fholzer: > (Also, for the "multi-press" stuff (ll 177-185 in __init__.py), should the appropriate press type come from PRESS_TYPES instead of string literals? Would be a bit cleaner maybe?)

the PRESS_TYPES have keys that correspond to data we get in the event, so that dict can't be easily changes. The single/double/triple/...  entries are not subsequent entried, so the best we can do there is use the same constants  in the multipress part of the code that we use in the PRESS_TYPES like so: c2fdd4d

cod3gen:
Dim up and dim down is now present, step_size can be taken from params. Nice 👍 

Does the double click, triple click etc. work on your tap dial? Seems to have no function here, seems to be responding only to pressed, continuesly pressed, released and released after long press on my dial.

SkywiperSolutions:
> Dim up and dim down is now present, step_size can be taken from params. Nice 👍
> 
> Does the double click, triple click etc. work on your tap dial? Seems to have no function here, seems to be responding only to pressed, continuesly pressed, released and released after long press on my dial.

My RDM0002's only responds to pressed, continuously pressed, released and released after long press as well.
Any outlook on a release?
Thanks!

loicortola:
Hi everyone, thanks @fholzer and all for working on this device!
Any updates on a release? Anyone working on it still? I see latest commits fail

Just so you know, I will completely make an extensive use of it as soon as available ^^

I haven't yet contributed to the custom quirks, but if you need assistance let me know, I will try to dig in.

smoothquark:
Thank you for this. I am hoping that this will eventually get incorporated into the official ZHA quirks. In the meantime, I noticed that on my RDM002, if a button is long pressed, 'Press', 'Hold' and 'Long Release' events are fired, whereas for a single press, 'Short Release' followed by 'Press' events. Double and triple presses have 'Short Release' followed by 'Double Press' and 'Triple Press' respectively, and no 'Press'. So, 'Press' is fired for a single short press and a long press but not double or triple presses. I am wondering if the 'Press' event should not be fired for a long press, but a 'Long Press' event instead? I am trying to create an automation that responds to short and long presses but the short press automation gets activated too. I hope I am making sense!

casparz:
Came here when looking into the workings of the RDM001, since it is not firing multiple button presses at the moment. I would love to see this refactor get merged, since is resolves this issue.
Alternative is copying more code from the init to the RDM001 implementation. As for no breaking changes, would it be considered a breaking change if the double, triple, etc. events were suddenly triggered? In the current version, these events are present, but will never be fired.

loicortola:
@TheJulianJES @MattWestb , would you be able to take a final look at this PR?
From what I understand, all breaking changes have been reverted and it should be ready for merge?

fholzer:
Sorry for the long absence. Just to clarify, at this point this PR is not backwards compatible, and also there's an issue (at least for me) where when turning the RDM002's dial e.g. up, it start dimming my light up, but in between jumps back to lower brightness before going back up. So that definitely needs to be fixed before considering merging this. Also someone reported issues, need to check what exactly that was.
Wrt which events are fired upon which button press behavior, I need to check what I had in mind originally, and post here to clarify, and collect feedback.
Further I don't think making this PR generic enough to be able to use class inheritance for shared features isn't viable given the requirement of backwards compatibility. Also there was some lack of clarity on requirements for breaking changes, so it would be great to better understand if it's OK to have this PR include breaking changes, and if there's anything we need to consider going foward wrt breaking changes.

SkywiperSolutions:
Ok. Thanks!

gtunes-dev:
Is this still being actively worked on? Would be fantastic to see it completed. Thanks!

TheJulianJES:
> so it would be great to better understand if it's OK to have this PR include breaking changes, and if there's anything we need to consider going foward wrt breaking changes.

I think the existing device triggers (and events used for blueprints) should need to stay the same, as we'd break a lot of setups otherwise.
Once "full backwards compatibility" is confirmed with this PR, it can probably be merged.
And thanks for your work of course!

gtunes-dev:
Hi, @fholzer. Thanks for your work on this. Do you have the time and interest in driving this to completion? If not for the entire effort, perhaps for a more focused set of changes to support the RDM002? Thanks!

fholzer:
TBH rewriting the whole Philips integration doesn't make much sense if we can't break things in order to streamline them across devices. At the moment my time is quite limited. I could probably get around to looking into this again sometime in the next few weeks. If someone else is interested in driving this forward faster, they are welcome to do so, though please say so. (So we don't end up both putting time into this)
I would probably end up stripping this PR of most of the changes and focus only on the RDM002, as you suggested.

gtunes-dev:
Speaking just for myself - my pressing issue is the state of the RDM002 support. It's a fantastic controller. I posted a Z2M blueprint that enables its use for media control with customization options for long/short presses for all buttons (https://community.home-assistant.io/t/z2m-philips-hue-tap-dial-media-controls/656669). I have a ZHA blueprint in the works but it's stalled because of the limited support though I do have a pretty creative solution for dealing with rotation speed.

So I would really appreciate it if you found the time to reduce this PR to RDM002 support and get it finalized. I'm not sure I have either the time or skill to do it though if you end up not being able to finish it, I might try.

In any case, it seems that the best approach here is to accept that the scope needs to be reduced and then figure out how to get it finished by you or someone else. Thanks!

gtunes-dev:
Hi, @fholzer. Checking to see if this is still on your radar and if you have a plan to reduce the scope of the change to get the RDM002 working. Thanks.

Edivion:
Any update here? The RDM002 would be the perfect solution for my current setup but with the very limited features I cannot really work with it. Properly supporting it would be greatly appreciated.

Sno88e:
Is there any way to get this remote working properly?
I am happy to help in any way I can if testing is needed.
These remotes are awesome and there is not really an equivalent alternative out there.
I Know it's possible to switch to zigbee2mqtt but I'm not a fan of it and I want to keep my installation as native as possible...

fholzer:
The regression that broke multi-press events should be fixed now. Also the single press issues. Though for the latter to work properly, I had to remove the "short release" event.

gtunes-dev:
@fholzer Thank you for picking this up again!

loicortola:
A-MA-ZING !
Thanks for getting back to it 😊. Can't wait to use your code in the hopefully next release!

fholzer:
Yeah, sorry for the long hiatus. Still have to change a couple of things to not break backwards compatibility for the other devices for those not using this branch. The downside to this for you will be that if you use this right now, and use other Philips devices like RDM001, ROM or RWL devices, the trigger names will be reverted to what they are on the dev branch. So in that case you'll have to update your automation triggers once that's done. Sorry in advance.

gtunes-dev:
@fholzer - with your current changes, are long presses and/or releases supported on the Hue Tap Dial (rmd002)? 

They don't appear to be, but I might be missing something or there might be an issue with how I'm testing. Thanks.

fholzer:
@gtunes-dev just pushed some more changes. Re-did the whole MR, but not finished yet. This supports single press/release, multi press and long press/release actions. Though it might break automation triggers for non-RDM002 devices. Will look into that next.

FYI, below are the test cases I want all devices I have to pass (RDM001, RDM002, RWL021, RWL022, ROM001). I know that on some devices that's broken (without this MR).

* On short press, should trigger a short press, followed by a short release event.
* On long hold, if supported, no short press/release events should be triggered.
* On long hold, if supported, a long press event should be triggered one or more times while the button is held, followed by a long release event once the button is released.
* On multi-press, if supported, no short press/release events should be triggered when pressing two or more times.

Also, are you testing by putting just the rdm002.py in your `custom_quirks` path? I haven't figured out how that should work if my rdm002.py file relies on changes to classes exported by `philips/__init__.py`. Personally, I test my changes by running homeassistant in a docker container based on `lscr.io/linuxserver/homeassistant:latest`, and I mount  `zha-device-handlers/zhaquirks` from my git repo clone onto the container's `/usr/local/lib/python3.12/site-packages/zhaquirks` path. For me, that looks like so: https://gist.github.com/fholzer/a2e876f9a06f9557cd5815c8272cbe4e

gtunes-dev:
Thanks for the update @fholzer. I'll take a look at your latest changes tomorrow.

I'm testing by putting everything from your branch's version of the Philips folder into the root of my custom quirks folder. That seems to work though I've had to delete the contents of the __pycache__ folder in custom quirks and restart HA to otherwise it picks up old versions.

If there's an issue with this approach, I can do what you're suggesting with Docker but I'm not currently set up to attach a Zigbee antenna to my Docker server and that would take a bit of work.

Thanks again for doing this - I'll let you know what I see with the test cases you listed.

gtunes-dev:
I understand the issue with __init__.py now - please disregard my previous message. I'm not sure there's an easy fix for this. I think I'll try to replicate your Docker-based environment tomorrow.

gtunes-dev:
@fholzer - ok...I have your updates working now. Thanks for sharing your docker compose - the only way I was able to get it to work was to modify this volume mapping:

      - ./zha-device-handlers/zhaquirks:/usr/local/lib/python3.12/site-packages/zhaquirks

to map in just the "philips" directory. I don't think the enlistment I have is the right version but limiting the quirks changes to just philips does work.

I have rdm002 devices. No others. 

I tried all test cases and they all work correctly.

A short press and release of buttons 1, 2, 3, 4 generates a short press followed by a short release and all events identify the correct button. A long press of any button generates a series of long press events and, when the button is released, a long release. All button identifiers are correct.

Double, Triple, Quadruple, Quintuple presses of all four buttons work correctly. The correct multi-press events are generated and no short press/release events.

Dial rotation events appear unchanged, which I assume is intentional.

In addition to the new events, short presses generate the recall events that were present before your changes.

These changes are great. I can now complete a blueprint I've had in the works for a long time that enables the rdm002 to act as a media controller with extensibility and velocity sensitivity. Thank you for doing this and I hope you're able to get it into production builds!

gtunes-dev:
I have a first pass at my media controller blueprint that relies on the changes from @fholzer.

It supports simple configuration of a dial as a media controller with a new approach for velocity-sensitive volume control.

It also supports custom actions for everything else this new update generates including double, triple, quadruple, and quintuple presses.

It also has logbook support so you can see the events the blueprint receives. To install, create a folder called "gTunes" in config\blueprints and drop this file in:

https://gist.github.com/gtunes-dev/b2636c6f1c4932cb15fa9b5a10ae820b

Thank you again, @fholzer - I hope you're able to get this out to production so I can post this publicly. I could also easily create one that is fully customizable without biasing towards media. I'll think about that.

fholzer:
Glad you got this working, @gtunes-dev. Do note that my initial changes on this branch changed the automation triggers of other Philips devices (ROM*, RWL*, RDM01) and I a preparing a commit to fix that. Though that means if you are working on blueprints for any of these devices, the trigger will change back to the original ones before this can be merged. I will also share my updated blueprint for controlling my LG TV's volume using the dial :)

Also I will be adding a lot of unit tests soon, most of which I have implements locally already. This revealed a minor bug that caused args.args[2] and to have the wrong value for multi-press events, and args.press_type and args.args[2]  to have the wrong value for single-press release events, though i guess it's unlikely these values are used by anyone.

gtunes-dev:
@fholzer - I'm working only with rdm002 controllers, but thanks for the warning.

I created an additional blueprint for the rdm002 that supports the dial and custom events for every event that I'm aware of....every press, release, long press, double, triple, quadruple, quintuple press. I suspect this might be helpful for you for debugging.

Couple of key features - it uses an approach to velocity sensing that turns the dial events into "small", "medium", "large" clockwise or counterclockwise actions. The user can configure the thresholds for differentiating small, medium, large. I think it's clever. We'll see if anyone else does.

It also emits actions to logbook and this can be toggled.

In any case, you might want to look at it to see if it helps with development. Again, I very much appreciate you doing this!

https://gist.github.com/gtunes-dev/b2636c6f1c4932cb15fa9b5a10ae820b#file-zha-philips-hue-tap-dial-yaml

fholzer:
Had a brief look at the step_size related part of you blueprint. I see you use a much more comple approach than I do. The blueprint i currently use for my TV volume as well as dimming lights directly use the step_size and a customizable multiplier (i.e. the resulting change is step_size * multiplier). This worked well for me so far. Might have a closer look once I am done with the unit tests related to DIM_UP/DOWN events in the coming days.

gtunes-dev:
> Had a brief look at the step_size related part of you blueprint. I see you use a much more comple approach than I do.

I have a bunch of these controlling various media endpoints driven by Roon - currently with Z2MQTT. What I've found is that, for me, what works best is adjusting specifically by 1, 3, or 5 increments based on the speed of rotation. I didn't want to hardcode that, though, so I made it possible for the user to choose their own "small, medium, large" detection windows and increments. With my media-specific blueprint, the defaults should work well. With the generic template, there are detection defaults but no opinion about 1,3,5...just the ability to put your own script in place.

Once your changes are on a path to release, I'll get my blueprints posted publicly.

gtunes-dev:
@fholzer - just confirming that all new events continue to work with an rdm002 with your latest changes.

What's the next step here? It looks like this is ready to go, right?

fholzer:
Well, would be great if some more people that hat Philips remote could test this, though I am not sure how to reach out to them. Also this needs to be reviewed by maintainers, and if they have feedback, that needs to be incorporated. That is, if they're happy with this refactoring in the first place. Let see. Will mark it as "ready for review". Though note that i will be traveling for the next 2 weeks, so probably won't be able to work on this a lot until I'm back.

fholzer:
I mentioned it already earlier on this thread, but just to recap: When I started working on RDM002 support, I found that the current implementation of many of the Philips remotes have issues. I decided to refactor code for all remotes, while adding RDM002 support. As part of that work I looked into all of those issues, and came up with 4 test cases that all remotes should pass. Here are those test cases, as well as a table showing which device implementations support which features, and pass/fail these test cases:

| Test Case                             | Behavior                                                                                                                                                                |
|---------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| test_PhilipsRemoteCluster_short_press | On short press, should send a short press, followed by a short release event.                                                                                           |
| test_PhilipsRemoteCluster_long_press  | On long hold, if supported, no short press/release events should be sent.                                                                                               |
| test_PhilipsRemoteCluster_long_press  | On long hold, if supported, a long press event should be sent one or more times while the button is held, followed by a long release event once the button is released. |
| test_PhilipsRemoteCluster_multi_press | On multi-press, if supported, no short press/release events should be sent when pressing two or more times.                                                             |

Here what the implementation supports without this MR. t1, t2, .. refere to above mentioned test cases; P=pass, F=fail
|        | press | released | multi | long press | released after long press | duration tracking | t1 | t2 | t3 | t4  |
| ------ | ----- | -------- | ----- | ---------- | ------------------------- | ----------------- | -- | -- | -- | --- |
| RDM001 | Y     | Y        | N     | Y          | broken                    | param5            | P  | F  | P  | N/A |
| ROM001 | Y     | Y        | Y     | Y          | Y                         | param5            | F  | F  | P  | F   |
| RWL022 | Y     | Y        | Y     | Y          | Y                         | param5            | F  | F  | P  | F   |
| RWL021 | Y     | Y        | Y     | Y          | Y                         | param5            | F  | F  | P  | F   |

I added a unit test that checks the device automation triggers before I made any other changes, so i can say with some level of certainty that my changes won't break existing automations. Additionally I added test cases for the desired behaviors outlined above. Before making my changes, I marked some things that the old implementation fails with "TODO"s in the unit tests, which I then removed in the second commit that include refactoring as well as adding support for RDM002.

bernhard-efler:
> Well, would be great if some more people that hat Philips remote could test this, though I am not sure how to reach out to them. Also this needs to be reviewed by maintainers, and if they have feedback, that needs to be incorporated. That is, if they're happy with this refactoring in the first place. Let see. Will mark it as "ready for review". Though note that i will be traveling for the next 2 weeks, so probably won't be able to work on this a lot until I'm back.

I have a device and can test your changes. But how can I get them installed/copied to my home assistant?

fholzer:
Thank you for offering to test. I am not entirely sure what the best way to "install" the files for testing is. I think usually files are self-sufficient, so can be put in the custom_quirks folder, though that won't work for this MR. I would suggest to copy the `zhaquirks/philips/` folder into the directory where the zhaquirks python dependency of your HA lives in your installation. That location varies depending on how you run home assistant. If you run it without docker, you could run `sudo find / -type d -name zhaquirks` to find the folder. If you run it inside a docker container, you'd have to run that (maybe without the `sudo`) inside the container, then mount the `philips` folder into that directory.

Personally I am using docker compose to run home assistant, and my `docker-compose.yml` looks like [this](https://gist.github.com/fholzer/a2e876f9a06f9557cd5815c8272cbe4e). This requires a clone of the zhaquirks repo, checked out to this branch, inside the folder where this `docker-compose.yml` file is located.

gtunes-dev:
@TheJulianJES - I hope it's not an inconvenience for me to mention you. You're the identified assignee on this and it's now ready for review. @fholzer has done the work - I'm just an enthusiastic consumer.

TheJulianJES:
All good. It's a huge PR, but hoping to review and test it soon.

bernhard-efler:
I tested the RDM002 with my own device. Works great and it is a great improvement. 
Only the multi press threshold could be longer (500ms) in my opinion. Sometimes I missed the time frame, especially when the dial was not positioned well for my fingers. But I'm not aware what other devices are using here.

fholzer:
The multi-press threshold is the same it was for other Philips remotes before my changes. Yes it could be changed, but there's a trade-off here. These remotes don't have built-in support for multi-press events. Instead these are simulated by this code. The problem is that all remotes only send short press and short release events (and corresponding long press/release events for long presses). In order for this code to tell whether the user just pressed a single time, or is starting a multi-press sequence, it has to wait for some time, the threshold you mentioned. As a result, single press events are not immediately triggering automations, as we have to wait for this threshold to be able to tell it's not the start of a multi press sequence. Increasing the threshold will probably annoy users that are wondering why there's such a long delay after a single press until their single press automations are triggered.
Sorry for the delayed response. I'm traveling at the moment.

gtunes-dev:
> As a result, single press events are not immediately triggering automations, as we have to wait for this threshold to be able to tell it's not the start of a multi press sequence.

This is a very helpful explanation. Given this, my user feedback is that I would prefer to not see the detection window increased.

I suppose the more complex fix would be to extend the window once you're in accumulation mode. You could, for example, have the detection window increase non-linearly as you wait to see if there are third or fourth presses. That said, since the current behavior pre-exists your changes, I'd like to see them go to production without any changes. I'm sure there will be people that use triple and quadruple presses but I'm pretty happy with single press, long press release, and the rotation events.

gtunes-dev:
I have updated blueprints available for use and testing. These versions include updated documentation, a couple of small bugfixes, and much improved organization including collapsible sections.

Media controls. This blueprint is for the use of the Tap Dial to control a media endpoint. It has built-in support for volume and transport controls while allowing additional customization:

Generic Controls. This is the second file in the gist. It doesn't assume a media controller but does allow the user to fully customize rotational actions as well as all button presses.

Both support generating logbook events.

https://gist.github.com/gtunes-dev/b2636c6f1c4932cb15fa9b5a10ae820b

I hope someone finds these useful. I'll post them to the community when these changes near release.

tverlaan:
@gtunes-dev I tested your general blueprint. I had to rename `button_2_double` to `button_2_double_press` and so forth for the other buttons and amount of presses. Is this a bug or did I load the patch from @fholzer incorrectly?

It works really well with this as custom quirk and the blueprint after that minor fix. I hope this can be merged soon :)

gtunes-dev:
Thanks for looking at the blueprint, @tverlaan!

> @gtunes-dev I tested your general blueprint. I had to rename `button_2_double` to `button_2_double_press` and so forth for the other buttons and amount of presses. Is this a bug or did I load the patch from @fholzer 

I thought I fixed that in the update I posted a couple of weeks ago. Is it possible you're using the older version?

In addition to fixing the issue you found, the newer version has collapsible sections for each button - they start off collapsed and you need expand them to customize. Does the version you're using have these collapsed sections?

The latest version is the second file (you need to scroll down) in this gist:

https://gist.github.com/gtunes-dev/b2636c6f1c4932cb15fa9b5a10ae820b

I also hope this gets reviewed and approved soon!

gtunes-dev:
@TheJulianJES - can you possibly review this? Thanks.

zhaquirks/philips/__init__.py
```diff
@@ -174,48 +197,133 @@ def handle_cluster_request(
         ] = None,
     ):
         """Handle the cluster command."""
-        _LOGGER.debug(
-            "PhilipsRemoteCluster - handle_cluster_request tsn: [%s] command id: %s - args: [%s]",
+        self.debug(
```
TheJulianJES: Please also replace this with `_LOGGER.debug` for now.

TheJulianJES (commented):
I think this looks very good. Definitely planning to have a proper look soon, but it seems fine tbh.

zhaquirks/philips/__init__.py
gtunes-dev: Sorry to be a pest about this, but I'd really like to see it go forward. @fholzer - did you see the request from @TheJulianJES to make the logger change?

fholzer: hey, yes saw it. Was waiting to see if more changes are needed. Was hoping to keep the number of commits for small changes like this low.

gtunes-dev:
Hi, @fholzer. I believe you and I are still running similar environments for testing - docker with your changes mapped into the container.

This has been working fine for me but is now broken with the latest Home Assistant container from linuxserver.io. There have been multiple 2024.10.2 images released. As of this post, the latest has the tag "2024.10.2-ls35". With this one, your changes prevent Home Assistant from starting. If I remove the mapping of your changes, it works fine.

If I revert to the previous 10.2 release, or anything earlier than that, your changes also work fine. So I'm back to the image with the tag "2024.10.2-ls34". 

In my case, this problem exists with the "latest" tag, these are just the version specific tags. For me, at least, something broke not in 10.2 but in something that is different in the .35 release.

If you can't repro this, I can get you logs. Thanks.

fholzer:
Thanks for pointing this out. Will try to look into it this week.

gtunes-dev:
One of those moments where you wish github had direct messages.

Some notes on the issue I'm experience. @TheJulianJES - this issue is very specific to my environment and I hope you don't defer your code review until it's root caused. 

@fholzer - I think the issue is related to specific changes the linuxserver team made to python package management. They went from pip to uv a few months ago but in this latest ls35 release, they made additional changes. My theory at the moment is that there's a permissions issue with the subfolder of quirks that I've mapped in and the consequence is that it can't write to it (might be trying to download a directory or to update __pycache__).

So this is specifically about the use of Docker, the use of this specific Docker image, and the fact that I'm trying to map just your changes, in a Philips directory) directly into the zhaquirks directory with this mapping:

      - /volume1/docker/homeassistant/quirks_philips_240908:/usr/local/lib/python3.12/site-packages/zhaquirks/philips

I believe that what's happening is that when uv tries to do this update, it's doing it as a user within the container called abc:abc, not using the user and group ids that I configured my container with. I don't think that user/group can write to my mapped in folder and I'm not sure how to make it possible. 

Here's a compare of the ls34 to ls35 that will help you see the changes they made:
      
 https://github.com/linuxserver/docker-homeassistant/compare/2024.10.2-ls34...2024.10.2-ls35
 
In any case, If I'm remotely right about this, this is very specific to my (our?) environments and what we're doing to try make zhaquirks work for testing purposes. This problem won't exist when the changes go public.

zhaquirks/philips/__init__.py
```diff
@@ -152,15 +155,35 @@ class PhilipsRemoteCluster(CustomCluster):
                 "param2": t.uint24_t,
                 "press_type": t.uint8_t,
                 "param4": t.uint8_t,
-                "param5": t.uint8_t,
-                "param6": t.uint8_t,
```
TheJulianJES, line 156: What was the reason again that `param6` is removed here?

TheJulianJES (commented):
I've had a look at the whole code now. I think it can stay as-is. Looks really good!
Thanks for doing this whole rework. The tests are especially nice! 😄

gtunes-dev:
Thanks, @TheJulianJES! And another Thank You! to @fholzer for this work!!

zhaquirks/philips/__init__.py
fholzer: I think these params were added at a time when it wasn't yet fully understood what these bytes in the message actually mean. By now we know that param5 and param6 (previously both uint8) are actually one single uint16, representing the time a button is held down during a long press event. This code removes replaces param5 (uint**8**) and param6 (uint**8**) with a single param5 (uint**16**). I was a bit conflicted about this too. If anyone realized that at least one of them represent long press duration, they would probably have seen the value of param6 increase - maybe not pressing long enough for the byte in param5 to change. Given this, there's maybe a case to be made to keep the name param6 have it be a uint16, and remove and skip param5.

TheJulianJES: Ah, thanks for the explanation. I think leaving it like this (`param5`) should be fine.

TheJulianJES:
Apparently, there's a new revision of the Hue wall switch module. Just another model string to add (`RDM004`) and maybe rename the class? Do you want to do it in this PR or do we want to rebase https://github.com/zigpy/zha-device-handlers/pull/3075/ after merging this PR and do it there (or in another PR)?

EDIT: Merged this PR first. Afterwards merged:
- https://github.com/zigpy/zha-device-handlers/pull/3075

(Feel free to PR changes if I missed something in that PR)

zhaquirks/philips/__init__.py
TheJulianJES: I'll add this as a commit to your branch, so that part should be done.
```suggestion
        _LOGGER.debug(
```
(All commits will be squashed when merging the PR anyways)

gtunes-dev:
> Apparently, there's a new revision of the Hue wall switch module. Just another model string to add (`RDM004`) and maybe rename the class?

I'm just a cheerleader and somewhat helpful tester but I've been hoping to see the RDM002 supported for a very, very long time and would prefer to see this PR go forward without converging with the wall switch. I understand if that's not what you prefer.

TheJulianJES (commented):
Unrelated to this PR, but I just noticed we're overriding `bind()` in the `PhilipsBasicCluster` to set some attributes.
I'm pretty sure `bind()` isn't actually called for the `Basic` cluster by default, as ZHA explicitly doesn't bind it (some devices can only bind a few clusters).

Do we know what the "philips" attribute (`0x0031`) does?

TheJulianJES (commented):
The `PhilipsBasicCluster` for `RDM001` also has another "mode" attribute (`0x0034`) that should be set to `0x2`. I don't think that'll happen either.
Also not sure what both attributes do tbh.

It's not something that needs to be addressed in this PR, but we should still take a look at it.

TheJulianJES:
To answer the second question myself, the "mode" options seems to configure how the wall switch modules act.
We should ideally expose that as an enum via quirks v2, as a lot of people use that apparently: https://community.home-assistant.io/t/how-to-configure-the-philips-hue-wall-module-to-use-push-button-momentary-type-wall-switches-zha/451125

TheJulianJES:
To get back to the first question about the "philips" attribute, I only found this: https://github.com/zigpy/zha-device-handlers/pull/388/files#r450561439 ... which still isn't quite clear to me 😄 
We do bind the Philips remote clusters in ZHA since some years, maybe that's the same as setting the "philips" attribute?
Everything appears to work at least..

TheJulianJES (approved):
LGTM. Thanks again! Let's merge this

DerGehrmann:
Hey @fholzer,
I try to figure out what was the idea of the dial functionality of the RDM002.

I'm using the RDM002 with the Firmware 0x02003b19 and the actual quirk 
grafik

but I can not see any attribute via zha_event to differ between a start of a long press of any button
grafik

and the left dial of the ring
grafik

Am I missing something?
I would like to implement a blueprint for the HACS switch manager but if a event is not unique it is not possible.

gtunes-dev:
Hi, @DerGehrmann. 

I believe your issue is that you're looking at zha_event rather than the "commands" generated by the quirk.

I have a blueprint up on Blueprint Exchange that you might want to use as a reference. If you look at the source of the blueprint, you'll see how it allows a user to specifically choose an RDM002 controller and then matches on "commands" coming from that controller. You'll see that the generated commands uniquely identify button presses (short, long, press, release, multiple, etc.) as well as rotation events. This will hopefully get you started.

https://community.home-assistant.io/t/zha-philips-hue-tap-switch-mini-custom-controls-rdm002/789654

If you'd like to see these commands in the Developer tools event listener, you can see them by listening to "call_service" events. You should also see them in the Logbook.

I hope this helps.

TheJulianJES:
The "commands" should be `zha_event`. That's also what the linked blueprint uses for triggers, so it likely also wouldn't work if you don't get proper `zha_event`s when pressing button. Try reconfiguring the remote via the HA UI on the device page. Press a button whilst doing that. Otherwise, try to re-pair the remote.

DerGehrmann:
Thanks for the quick responses.
@gtunes-dev I thought that everything on `zha_event` is the outcome of native zigbee or the corresponding quirk like in this case.
If I look for `call_service` I see the outcome of my switch manager configuration like `light.toogle` after `button_1_short_release`.

@TheJulianJES I tried reconfiguring and re-pairing but I have the same behaviour as before.
I get the following `commands`:
**1x Button:** `recall`, `button_1_press`, `button_1_short_release`
**2x Button:** `recall`, `recall`, `button_1_double_press`
**...**
**Long Press Button:** `step_with_on_off` {step_mode: 1, step_size: 255, transition_time: 4}, `button_1_hold`, `button_1_long_release`

**Dial left, slow (one step):** `step_with_on_off` {step_mode: 1, step_size: 8, transition_time: 4}
**Dial left, fast (multiple steps):** `step_with_on_off` {step_mode: 1, step_size: 255, transition_time: 6}

Like this it is not possible to differ between `Long Press Button` and `Dial left`. 
Dial right works fine because the step_mode = 0 :-)

What is the expected behaviour?
If I get the blueprint of @gtunes-dev right, the `step_mode` should be a text like `StepMode.Down` but it doesn't appear like this in my case.

Thanks in advance

gtunes-dev:
@fholzer - I think this is actually a bug.

My previous comments were misleading and incorrect. Thanks @TheJulianJES for correcting. 

I've now actually tested this. It looks like a bug to me.

As @DerGehrmann notes, a long press of a button generates a spurious step_with_on_off before generating the expected press event. The first post from @DerGehrmann suggested that the long press only generated this one event but their second post clarified that the sequence they see when doing a long press (and release) is actually:

step_with_on_off (down)
button_x_hold
button_x_release

This is what I see as well. I haven't noticed this in the past. I'm not sure how I missed it in my own testing. Now that I see it in the event stream, I can clearly also see it impacting my own blueprints. 

Here's what the sequence looks like in Logbook when doing a long press / release of button 1.

My blueprints end up handling  the counterclockwise rotation event and then the press / release. 

image

This happens on all of my controllers. It's not specific to one.

@DerGehrmann - you asked about my blueprint's use of "StepMode.Down". When looking at zha_events in Developer tools > Events, I see step_mode as a param with the value of 0 or 1 depending on rotation direction. My blueprints, however,  get an event that has a parameter of "step_mode" that has a value of either "StepMode.Down" or "StepMode.Up". You should be able to use the code in my blueprints to model your own.

My blueprints also have the ability to write events to Logbook. I found this helpful during the development process. If you turn Logbook logging on in one of my blueprints, you'll see the params broken out in Logbook entries -  you may find that helpful.

Hopefully @fholzer can take a look at this. Thanks!

fholzer:
Traveling this week. Will have a look on Sat/Sun.

fholzer:
I can confirm I see the same behavior, though I think it's unrelated to this PR. The step_with_on_off events triggered by both, holding any button and rotating the dial to the left is handled via the LevelControl cluster, and this PR isn't changing anything related to that. I created a new issue for this.

PR #2341, opened 2023-04-15
Add another version of Aqara roller shade driver E1

badrpc:
Extends existing quirk. The only difference is DEVICE_TYPE in signature.


javicalle (approved):
LGTM

PR #2350, opened 2023-04-21
Fix battery percentage for Sercomm XHS2-SE contact sensor

sjbarkey:


TheJulianJES:
@sjbarkey This PR is for correctly updating the battery sensor entity of that Sercomm XHS2-SE contact sensor, right?

PR #2358, opened 2023-04-27
Ensure quirk clusters are proper subclasses

puddly:
A few quirk subclasses are "broken" because they subclass an existing cluster but modify it in an incompatible way: the Konke button, for example, uses an `OnOff` cluster subclass that does not have an `on_off` attribute. This breaks ZHA, which expects the `on_off` attribute to exist on an `OnOff` cluster.

(ZHA currently does not match by cluster class, however, and instead matches on cluster ID alone, so this theoretically could be fixed in just ZHA. I think it's better to not have this be a problem in the future).

This code is not runtime tested but unit tests pass with these corresponding ZHA changes: https://github.com/puddly/core/tree/puddly/20230427-fix-bad-quirk-cluster-attrs

### DIFF
--- a/zhaquirks/konke/__init__.py
+++ b/zhaquirks/konke/__init__.py
@@ -18,6 +18,19 @@ from zhaquirks.const import (
 KONKE = "Konke"
 
 
+class KonkeButtonEvent(t.enum8):
+    Single = 0x80
+    Double = 0x81
+    Hold = 0x82
+
+
+PRESS_TYPES = {
+    KonkeButtonEvent.Single: COMMAND_SINGLE,
+    KonkeButtonEvent.Double: COMMAND_DOUBLE,
+    KonkeButtonEvent.Hold: COMMAND_HOLD,
+}
+
+
 class OccupancyCluster(LocalDataCluster, OccupancyOnEvent):
     """Occupancy cluster."""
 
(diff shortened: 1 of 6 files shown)


zhaquirks/konke/__init__.py
```diff
@@ -60,30 +76,13 @@ def handle_cluster_general_request(
...
 
         attr = args[0][0]
-        if attr.attrid != 0x0000:
+
+        if attr.attrid != self.attributes_by_name["konke_button_event"].id:
             return
 
         value = attr.value.value
         event_args = {
-            PRESS_TYPE: self.PRESS_TYPES.get(value, value),
```
puddly, line 68: Are there other press types?

zhaquirks/konke/__init__.py
```diff
@@ -60,30 +76,13 @@ def handle_cluster_general_request(
...
-            COMMAND_ID: value,
+            PRESS_TYPE: PRESS_TYPES[value],
+            COMMAND_ID: value.value,  # to maintain backwards compatibility
         }
         self.listener_event(ZHA_SEND_EVENT, event_args[PRESS_TYPE], event_args)
-
-    def deserialize(self, data):
```
puddly, line 73: This `deserialize` override seems unnecessary. Are the unit tests accurate? They pass without it.

TheJulianJES: Doesn't seem to be required (for my button model at least). Seems to work fine without it.

zhaquirks/philips/__init__.py
```diff
@@ -71,13 +71,19 @@
 }
 
 
-class OccupancyCluster(CustomCluster, OccupancySensing):
+class PhilipsOccupancySensing(CustomCluster):
     """Philips occupancy cluster."""
 
+    cluster_id = OccupancySensing.cluster_id
+    ep_attribute = "philips_occupancy"
```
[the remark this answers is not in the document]
puddly: I have a branch of ZHA linked above in the PR that works with this specific sensor (I'm running it right now). It did require a small tweak.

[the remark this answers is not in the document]
TheJulianJES: Ah, missed that somehow. Small question there: https://github.com/puddly/core/commit/4de50036f364380ec5d9745d5e1c1cf4de026b7d#r110865576

TheJulianJES (approved):
Quirk changes LGTM when paired with the ZHA changes.

(Small question for the HA Core changes: https://github.com/puddly/core/commit/4de50036f364380ec5d9745d5e1c1cf4de026b7d#r110865576)

PR #2363, opened 2023-04-28
Add support for Tuya single button scene switch variant `TS0041A`

ShogunMan:
## Proposed change

  Added support for variant TS0041A

## Additional information

 Sold as Cleverio in Sweden/Norway
https://www.kjell.com/no/produkter/smarte-hjem/smarte-hjem-losninger/cleverio-smarte-hjem/cleverio-multifunksjonsknapp-med-zigbee-3.0-p51830

Blueprint here:
https://community.home-assistant.io/t/zha-cleverio-tuya-1-button-scene-switch-ts0041a-tyzb01-4qw4rl1u/552120

## Checklist


MattWestb:
Thats interesting its one `A` version that is normally only being made for LIDL and other that need have Zigbee / CSA-ITO certificated device and the normal `CE` (**C**hina **E**xport not EU cert) is not getting certified like without letter or `B` in the end of the device type.

ShogunMan:
> Thats interesting its one `A` version that is normally only being made for LIDL and other that need have Zigbee / CSA-ITO certificated device and the normal `CE` (**C**hina **E**xport not EU cert) is not getting certified like without letter or `B` in the end of the device type.

Do you think it would be smart to add  vendor specific data to the class ( _TYZB01_4qw4rl1u) to avoid mispairings?

MattWestb:
I think it shall being in this quirk then its the same functionality as all other TS0041.
But i also think its good putting the **A** in the device class name and in the description so users is not trying adding the class to no **A** versions that is not working then it cant being matched.

We have not seen any TS004X that need model info matching and all is working well if the cluster / end points is matching OK with exception for the TS004F that is one different story and if its needed we is implanting it later for getting all working the same way.

zhaquirks/tuya/ts0041.py
```diff
@@ -184,3 +184,56 @@ class TuyaSmartRemote0041TOPlusA(CustomDevice):
...
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
+                    TuyaSmartRemoteOnOffCluster,
+                ],
+                OUTPUT_CLUSTERS: [
+                    TuyaSmartRemoteOnOffCluster,
```
TheJulianJES, line 227: Are both clusters used by that remote?
Or is replacing one `OnOff` cluster enough?

ShogunMan: There is only one button, but i really dont know why it has one input and one output OnOff cluster.
It does not throw any errors or warnings as it is now, but We could try with replacing only one.

MattWestb: Its one tuya thing and deleting it in replacement is making ZHA not making on switch in the GUI that is not working (at least for TS004F devices).
Testing if it working without it and its still sending commands OK and not getting problems / delays with fast key pressings.

ShogunMan:
From testing it looks like both in and out clusters must be replaced for proper function.

zhaquirks/tuya/ts0041.py
ShogunMan: @TheJulianJES From testing and the device signature, it seems like both clusters has to be replaced. Works fine here.

TheJulianJES: So when only one cluster is replaced, it doesn’t work?

ShogunMan: I'll have to do another test. Should I replace TuyaSmartRemoteOnOffCluster with OnOff.cluster_id in the output cluster?

TheJulianJES: Try leaving it out completely, so you have just one Tuya OnOff cluster.
(If one doesn’t work, try the other one.)

ShogunMan: I thought that the number of output clusters must match the number in the device signature, otherwise it will fail loading the quirk. This is my proposed replacement:

```python
replacement = {
        ENDPOINTS: {
            1: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    TuyaSmartRemoteOnOffCluster,
                ],
                OUTPUT_CLUSTERS: [
                    OnOff.cluster_id,,
                    Time.cluster_id,
                    Ota.cluster_id,
                ],
            },
        },
    }
```

ShogunMan: The above code is the only one that is working reliably. PR has been updated with tuya cluster in only the input cluster.

MattWestb: Ar you getting one switch in ZHA GUI then having it as input cluster after restarting HA ?
All TS004X is having  problems with the dealt  response shall being sent the wrong way or the device is making one pause between commands if making more fast. But it can being different  with the **A** version you is having.

I have making one issue for sending it in the wrong detection but its not solved yet.

ShogunMan: @MattWestb It does expose a switch entity in the GUI. I will check how it works. Might need some more work, but basic functionality of the button is in place.

ShogunMan: This replacement/quirk does not provide the switch entity in the GUI as previous versions did. 
If you have used a previous version, you will have to delete the switch entity manually as it is no longer provided by the integration, it will still be there after a restart if not deleted manually.

ShogunMan:
How about pulling this one? It works as it should.
Or should we just let it die :-)

I am currently using this custom quirk.

TheJulianJES (approved):
LGTM, thanks!

PR #2376, opened 2023-05-03
Add support for Onesti EasyCode Touch lock

1bard:
## Proposed change

This adds support for Onesti EasyCode Touch door lock.
The quirk removes incorrect MainsPowered-flag from NODE_DESCRIPTOR, making battery percentage visible in HA.

## Checklist

### DIFF
--- a/zhaquirks/onesti/easycodetouch.py
+++ b/zhaquirks/onesti/easycodetouch.py
@@ -1,25 +1,24 @@
 from zigpy.profiles import zha
-from zigpy.quirks import CustomDevice, CustomCluster
+from zigpy.quirks import CustomCluster, CustomDevice
+from zigpy.zcl.clusters.closures import DoorLock
 from zigpy.zcl.clusters.general import (
     Basic,
-    PowerConfiguration,
     Groups,
     Identify,
-    Scenes,
     Ota,
-)
-from zigpy.zcl.clusters.closures import (
-    DoorLock,
+    PowerConfiguration,
+    Scenes,
 )
 import zigpy.zdo.types
+
 from zhaquirks.const import (
     DEVICE_TYPE,
     ENDPOINTS,
     INPUT_CLUSTERS,
     MODELS_INFO,
+    NODE_DESCRIPTOR,
     OUTPUT_CLUSTERS,
     PROFILE_ID,
-    NODE_DESCRIPTOR,
 )
 
 MANUFACTURER_SPECIFIC_PROFILE_ID = 0xFEA2
@@ -28,7 +27,8 @@ MANUFACTURER_SPECIFIC_PROFILE_ID = 0xFEA2
 class BatteryBasicCluster(CustomCluster, Basic):
     """Basic cluster implementation.
     This implementation sets power source to battery
-    and removes incorrect MainsPowered-flag."""
+    and removes incorrect MainsPowered-flag.
+    """
 
     cluster_id = Basic.cluster_id
     POWER_SOURCE = 0x0007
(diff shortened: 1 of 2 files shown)


1bard:
I have not checked "Tests have been added to verify that the new code works".
I really don't know what tests to add (or how).
I'm new to Python-development and there's not much code to test.

zhaquirks/onesti/easycodetouch.py
```diff
@@ -0,0 +1,99 @@
...
+MANUFACTURER_SPECIFIC_PROFILE_ID = 0xFEA2
+
+
+class BatteryBasicCluster(CustomCluster, Basic):
+    """Basic cluster implementation.
+    This implementation sets power source to battery
+    and removes incorrect MainsPowered-flag.
+    """
+
+    cluster_id = Basic.cluster_id
```
TheJulianJES, line 33: This line shouldn't be needed, as inheriting `Basic` also sets that `cluster_id`.
(But I know it's also done in some other places still. You can just remove it though)

zhaquirks/onesti/easycodetouch.py
```diff
@@ -0,0 +1,99 @@
...
+    POWER_SOURCE = 0x0007
+
+    _CONSTANT_ATTRIBUTES = {
+        POWER_SOURCE: Basic.PowerSource.Battery,
```
TheJulianJES, line 37: Just curious, is this "power source" attribute actually used anywhere (in HA)?
Showing the battery percentage sensor only relies on the node descriptor IIRC(?)

zhaquirks/onesti/easycodetouch.py
```diff
@@ -0,0 +1,99 @@
...
+            aps_flags=0,
+            frequency_band=8,
+            mac_capability_flags=136,
+            manufacturer_code=4660,
+            maximum_buffer_size=108,
+            maximum_incoming_transfer_size=127,
+            server_mask=11264,
+            maximum_outgoing_transfer_size=127,
+            descriptor_capability_field=0,
+        ),
```
TheJulianJES, line 82: "To compare": can you also post the original/unquirked device signature of the device?

zhaquirks/onesti/easycodetouch.py
```diff
@@ -0,0 +1,99 @@
...
+    cluster_id = Basic.cluster_id
+    POWER_SOURCE = 0x0007
+
+    _CONSTANT_ATTRIBUTES = {
+        POWER_SOURCE: Basic.PowerSource.Battery,
+    }
+
+
+class EasyCodeTouch(CustomDevice):
+    """onesti.easycodetouch"""
```
TheJulianJES, line 42:
```suggestion
    """Onesti EasyCodeTouch quirk."""
```

zhaquirks/onesti/easycodetouch.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class BatteryBasicCluster(CustomCluster, Basic):
+    """Basic cluster implementation.
+    This implementation sets power source to battery
+    and removes incorrect MainsPowered-flag.
+    """
+
+    cluster_id = Basic.cluster_id
+    POWER_SOURCE = 0x0007
```
TheJulianJES, line 34: Getting attribute ids will be easier in the future, but right now, we can already do this:
```suggestion
    Basic.attributes_by_name["power_source"].id
```

TheJulianJES (commented):
Thanks for the PR! Some small comments/questions:

zhaquirks/onesti/easycodetouch.py
1bard: This was my initial attempt to make it act as a battery-powered device.
I've tested and you're right. Power source attribute is not required. I've removed it.

zhaquirks/onesti/easycodetouch.py
1bard: Added

PR #2238, opened 2023-02-28
Improve Aqara E1 thermostat quirk

TheJulianJES:
This PR does the following:
- it fixes system mode (couldn't turn off before)
- it adds other Xiaomi attributes (supersedes https://github.com/zigpy/zha-device-handlers/pull/2237 and part of https://github.com/zigpy/zha-device-handlers/pull/2109 
- it adds the OTA cluster

Tested and working so far.

Possible TODOs:
- add basic tests ✅ 
- do the external sensor temp stuff (-> hopefully in a future PR)
- implement more?
- implement entities in ZHA (separate PR)
- attribute for 0x270 (write 1 to calibrate?)

To test this, you should be able to download the modified `zhaquirks/xiaomi/aqara/thermostat_agl001.py` file and put that in your custom quirks directory.

PR #2380, opened 2023-05-04
Fix reading PM2.5 attribute in IKEA Starkvind quirk

michaelmulley:
## Proposed change
An IKEA Starkvind device was not showing a PM2.5 sensor in Home Assistant. Upon investigation, Python was throwing a `TypeError: 'set' object is not subscriptable` in `read_attributes`. This appears to be because code in `starkvind.py` was calling `read_attributes` with a set for the attributes arg, but the type signature says that argument should be a list. So I made that very simple change.

## Additional information
I'm brand new to this project and unfamiliar with the code. But the way the code is structured here (a call to one attribute in the PM2.5 cluster is redirected to another attribute in a custom manufacturer cluster) feels a little odd -- is there a different best practice to do this? In particular, with the current code, it looks like if `PM25Cluster.read_attributes` is called with multiple attributes at once, one of which is `measured_value`, it will quietly return only one attribute in its response and ignore any others.

## Checklist

### DIFF
--- a/tests/test_ikea.py
+++ b/tests/test_ikea.py
@@ -1,6 +1,14 @@
 """Tests for Ikea Starkvind quirks."""
 
+from unittest import mock
+
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.measurement import PM25
+
+import zhaquirks
 import zhaquirks.ikea.starkvind
+
+zhaquirks.setup()
 
 
 def test_ikea_starkvind(assert_signature_matches_quirk):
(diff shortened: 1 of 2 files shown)


TheJulianJES:
Nice catch!

(Note: Attribute reporting also works without this change, as that doesn't use the `read_attributes` method, but only `update_attribute`.
However, the sensor entity in HA is missing up until restarting (under the requirement that you have received at least one attribute report prior to the restart).)

> it looks like if `PM25Cluster.read_attributes` is called with multiple attributes at once, one of which is `measured_value`, it will quietly return only one attribute in its response and ignore any others.

Yes.

Although the following approach also isn't ideal IMO, see the changes in this PR (they should properly deal with multiple attributes in one read call): https://github.com/zigpy/zha-device-handlers/pull/2238/files

TheJulianJES (approved):

PR #2385, opened 2023-05-08
Support for Aqara H1 single rocker wall switch with neutral (n1aeu1)

szamfirov:
## Proposed change
Add missing support for the Aqara H1 single rocker wall switch (with neutral wire). The model is also known as: `WS-EUK03` or `lumi.switch.n1aeu1`

## Additional information
Everything at home is zigbee based but this is my first ever in-depth look into zigbee as well as contributing in that space so I only have basic understanding about how things should work. The contribution is heavily _inspired from the OppleSwitch_ and _OppleRemote_ implementations with some caveats which are better described in the checklist below.

**Some highlights while working on this**:
* **Operation mode is working fine**. You can set the switch to coupled / decoupled mode but I cannot get the switch_mode / switch_type functionality to work (I'm always receiving None values).
* I don't seem to be getting any ZHA events on any button press - short, double, long etc. Not sure what can be improved but I'm guessing the secret lies in some of the clusters being used (and whether they are set as input vs output clusters).

## Conclusion
Seems to be quite a straightforward addition but I'm currently lacking some additional context / knowledge to make this complete so I'd appreciate some support from the community on this one.

## Checklist
- [x] Operation mode is working as expected (coupled / decoupled)
- [x] Receive ZHA events on button press (short / double / long)
- [x] Automation triggers are properly configured

### DIFF
--- a/zhaquirks/xiaomi/aqara/switch_h1.py
+++ b/zhaquirks/xiaomi/aqara/switch_h1.py
@@ -1,0 +1,176 @@
+import copy
+import logging
+from enum import Enum
+
+from zhaquirks.const import (ALT_DOUBLE_PRESS, ALT_SHORT_PRESS, ARGS, ATTR_ID,
+                             BUTTON, CLUSTER_ID, COMMAND, COMMAND_DOUBLE,
+                             COMMAND_HOLD, COMMAND_OFF, COMMAND_SINGLE,
+                             COMMAND_TOGGLE, COMMAND_TRIPLE, DEVICE_TYPE,
+                             DOUBLE_PRESS, ENDPOINT_ID, ENDPOINTS,
+                             INPUT_CLUSTERS, LONG_PRESS, MODELS_INFO,
+                             OUTPUT_CLUSTERS, PROFILE_ID, SHORT_PRESS,
+                             TRIPLE_PRESS, ZHA_SEND_EVENT)
+from zhaquirks.xiaomi import (LUMI, DeviceTemperatureCluster,  # BasicCluster,
+                              OnOffCluster, XiaomiCustomDevice,
+                              XiaomiMeteringCluster)
+from zhaquirks.xiaomi.aqara.opple_remote import (MultistateInputCluster,
+                                                 OppleCluster)
+from zhaquirks.xiaomi.aqara.opple_switch import (OppleIndicatorLight,
+                                                 OppleOperationMode)
+from zigpy import types as t
+from zigpy.profiles import zha
+from zigpy.zcl.clusters.general import (Alarms, Basic, GreenPowerProxy, Groups,
+                                        Identify, OnOff, Ota, Scenes, Time)
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class AqaraH1SwitchCluster(OppleCluster):
+    """Xiaomi mfg cluster implementation."""
+
+    attributes = copy.deepcopy(OppleCluster.attributes)
+
+    attributes.update(
+        {
+            0x0002: ("power_outage_count", t.uint8_t, True),
+            0x000A: ("switch_type", t.uint8_t, True),  # values can be: 1 or 2
+            0x00F0: ("reverse_indicator_light", OppleIndicatorLight, True),
+            0x0125: ("switch_mode", t.uint8_t, True),  # values can be: 1 or 2
+            0x0200: ("operation_mode", OppleOperationMode, True),
+            0x0201: ("power_outage_memory", t.Bool, True),
+            0x0202: ("auto_off", t.Bool, True),
+            0x0203: ("do_not_disturb", t.Bool, True),
+        }
+    )
+
+    def _update_attribute(self, attrid, value):
+        super()._update_attribute(attrid, value)
+        if attrid == 0x00FC:
+            self._current_state = PRESS_TYPES.get(value)
+            event_args = {
+                PRESS_TYPE: self._current_state,
+                ATTR_ID: attrid,
+                VALUE: value,
+            }
+            self.listener_event(ZHA_SEND_EVENT, self._current_state, event_args)
+            # show something in the sensor in HA
+            super()._update_attribute(0, self._current_state)
+
+
+class AqaraH1SingleRockerSwitch(XiaomiCustomDevice):
+    """Aqara H1 Single Rocker Switch (with neutral)."""
+
+    signature = {
+        MODELS_INFO: [(LUMI, "lumi.switch.n1aeu1")],
+        ENDPOINTS: {
+            #  input_clusters=[0, 2, 3, 4, 5, 6, 18, 64704], output_clusters=[10, 25]
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,  # 0
+                    DeviceTemperatureCluster.cluster_id,  # 2
+                    Identify.cluster_id,  # 3
+                    Groups.cluster_id,  # 4
+                    Scenes.cluster_id,  # 5
+                    OnOff.cluster_id,  # 6
+                    Alarms.cluster_id,  # 9
+                    XiaomiMeteringCluster.cluster_id,  # 0x0702
+                    0x0B04,
+                ],
+                OUTPUT_CLUSTERS: [
+                    Time.cluster_id,  # 0x000a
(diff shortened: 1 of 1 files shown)


szamfirov:
Managed to find the missing link - there was a cluster missing in the replacement definition (see below for more details). It was tied to the events not being emitted when pressing the button face.

https://github.com/zigpy/zha-device-handlers/blob/3cd8b4e1e2a2699ec681fb05660f145daf41fe93/zhaquirks/xiaomi/aqara/switch_h1.py#L115-L122

Once I added this, the switch now reports
**single, double, triple** and **long** presses 🎉 

Device trigger automations work correctly as well 💪 

@TheJulianJES, do you mind having another look? Please include other people in the discussion if needed.

zhaquirks/xiaomi/aqara/switch_h1.py
```diff
@@ -0,0 +1,162 @@
...
+    BasicCluster,
+    DeviceTemperatureCluster,
+    OnOffCluster,
+    XiaomiCustomDevice,
+    XiaomiMeteringCluster,
+)
+from zhaquirks.xiaomi.aqara.opple_remote import MultistateInputCluster
+from zhaquirks.xiaomi.aqara.opple_switch import OppleSwitchCluster
+
+_LOGGER = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
szamfirov: Thanks for the quick reaction. I removed the unused logger 👍

zhaquirks/xiaomi/aqara/switch_h1.py
```diff
@@ -0,0 +1,158 @@
...
+        (TRIPLE_PRESS, BUTTON): {
+            ENDPOINT_ID: 1,
+            CLUSTER_ID: 0,
+            COMMAND: COMMAND_ATTRIBUTE_UPDATED,
+            ARGS: {
+                ATTRIBUTE_ID: 5,
+                ATTRIBUTE_NAME: "model",
+                VALUE: "lumi.switch.n1aeu1",
+            },
+        },
```
TheJulianJES, line 151: This is interesting.
Normally, the Aqara sensors send an attribute report with their model when the "reset button" is pressed once (which doesn't actually reset the device).

But I guess if -- for this device -- it's only sent when tripple-pressing the button, it should be fine.

TheJulianJES: ... although I do wonder if this can cause a "tripple press" event to be fired when:
(1) the button is first paired
(2) the button is re-paired (probably doesn't re-read the attribute again though)
(3) the model attribute is read through the clusters UI. <- Could you check that?

zhaquirks/xiaomi/aqara/switch_h1.py
```diff
@@ -0,0 +1,158 @@
...
+                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
+            },
+        },
+    }
+
+    device_automation_triggers = {
+        (SHORT_PRESS, BUTTON): {
+            ENDPOINT_ID: 41,
+            CLUSTER_ID: 18,
+            COMMAND: f"41_{COMMAND_SINGLE}",
```
TheJulianJES, line 132: Can you put the three commands (single, double, long) into constants in this file (at the top)?
Similar to this:
https://github.com/zigpy/zha-device-handlers/blob/ffdf2fbfe559f74e6fc00d61130b8025e18141ec/zhaquirks/xiaomi/aqara/opple_remote.py#L59-L63

zhaquirks/xiaomi/aqara/switch_h1.py
szamfirov: (2) and (3) do indeed cause "triple press" events to be fired so I removed this trigger altogether since it may cause some unexpected behaviour for people who may want to use it.

I also noticed the following comment in the general cluster implementation which further strengthens my doubts for having "triple press" events for these switches: https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/xiaomi/__init__.py#L236-L239

zhaquirks/xiaomi/aqara/switch_h1.py
szamfirov: Good shout. Cheers

zhaquirks/xiaomi/aqara/switch_h1.py
```diff
@@ -0,0 +1,147 @@
...
+)
+from zhaquirks.xiaomi.aqara.opple_remote import MultistateInputCluster
+from zhaquirks.xiaomi.aqara.opple_switch import OppleSwitchCluster
+
+XIAOMI_COMMAND_SINGLE = "41_single"
+XIAOMI_COMMAND_DOUBLE = "41_double"
+XIAOMI_COMMAND_HOLD = "1_hold"
+
+
+class AqaraH1SingleRockerSwitch(XiaomiCustomDevice):
```
TheJulianJES, line 52: Since this is a device without battery (right?), you can change `XiaomiCustomDevice` to just `CustomDevice`.

szamfirov: That's correct - the device doesn't have a battery, it's fed through mains. I've implemented the change and it all works fine with it 👍

zhaquirks/xiaomi/aqara/switch_h1.py
```diff
@@ -0,0 +1,147 @@
...
+                    GreenPowerProxy.cluster_id,  # 0x0021
+                ],
+            },
+        },
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
```
TheJulianJES, line 92: Ah, one last thing: the `PROFILE_ID` is missing in the replacement. You can just copy it over.

szamfirov: Added

zhaquirks/xiaomi/aqara/switch_h1.py
```diff
@@ -0,0 +1,148 @@
...
+                    0x0B04,
+                ],
+                OUTPUT_CLUSTERS: [
+                    Time.cluster_id,
+                    Ota.cluster_id,
+                ],
+            },
+            41: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
```
TheJulianJES, line 114: Hopefully one last thought:

You're changing the device type here.
Simple switch functionality already works without the quirk, right?
So this could be a breaking change for existing users.

The type can also be changed in the `configuration.yaml` without using a quirk (IEEE and endpoint needed).

At least for Tuya (light) switches, we've tried to avoid changing the device type using quirks.

Any opinions on changing the device type for this device?
Honestly, the change shouldn't be too breaking, but we should still consider if we want to make it.

szamfirov: I understand your concern. However, I'm basing this quirk on the [double rocker version](https://zigbee.blakadder.com/Aqara_WS-EUK04.html) of the same type of switch which is already being actively used and, judging from all the comments from the community, people are keen on having the same functionality for the [single rocker switch](https://zigbee.blakadder.com/Aqara_WS-EUK03.html) as well which I happen to own so I figured why don't I give something back to the community.

My personal preference also matches the existing implementation for the double rocker switch which is to expose them as switches instead of lights. For reference: https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/xiaomi/aqara/opple_switch.py#L124

szamfirov: Screenshot 2023-05-22 at 21 22 01



Time for these bad boys to get some much deserved support 🙂

TheJulianJES: Since it's consistent with the other switches, it's probably fine. 👍

TheJulianJES (approved):
LGTM. Thanks! 🎉

szamfirov:
@TheJulianJES, thank you for merging it! 🙇

Could you please shed some light on what the release cadence for new quirks is? I'm just wondering when folks could officially make use of it.

TheJulianJES:
@szamfirov I'd expect a new quirks release for the first HA Core 2023.6.0 beta or at least the final release.

HA schedule can be found here: https://developers.home-assistant.io

Icelk:
Hi, sorry for necroing the thread. I have a question about the H1 switches: do they send hold and release events? Or is the "long press / hold" event just when a button is held for e.g. 1 second, and no duration information is gatherable?

Thanks for the integration!

PR #2401, opened 2023-05-16
Add Aqara FP1 region support

### LINKED ISSUE #2202: Aqara FP1 - implement region support
Would you be able to implement regions  for Aqara FP1?
It makes the sensor actually usable and reduces the false positives.
Zigbee2mqtt has just implemented it but I would rather use zha if possible.

The following options should be implemented:
- possibility to add/remove regions
- parse region info from sensor and make it available in a event or sensor
- even better, the device could create one sensor for each region and update it's status :D

<details>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered: 140>, manufacturer_code=4660, maximum_buffer_size=108, maximum_incoming_transfer_size=127, server_mask=11264, maximum_outgoing_transfer_size=127, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0xfff0",
      "in_clusters": [
        "0x0000",
        "0x0002",
        "0x0003",
        "0x0406",
        "0xfcc0"
      ],
      "out_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "aqara",
  "model": "lumi.motion.ac01",
  "class": "zhaquirks.xiaomi.aqara.motion_ac01.AqaraLumiMotionAc01"
}
```

</details>

<details>

```yaml
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>

```
Paste any additional debug logs here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

**Additional context**

michalmo:
## Proposed change
Extends the Aqara FP1 quirk with support for regions.

## Additional information
This PR adds support for regions and their configuration, as well as support for position events for setup and debugging. 
Fixes #2202.

This is implemented by adding 10 extra endpoints with OccupancySensing clusters for each of the regions, and custom commands to the main OccupancySensing cluster and the region OccupancySensing clusters for configuration.

Implementation is based on reverse engineering efforts from https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5928#issuecomment-1166545226 and https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5928#issuecomment-1422392105, and my own work to complete the understanding of how position reporting works.

### About regions
The FP1 allows up to 10 detection regions, each composed of areas from a 4x7 grid as shown below:

The device seems to allow overlapping regions.

### Entities and events
Presence in configured regions will be reported in the OccupancySensing clusters of the extra endpoints (which will appear in ZHA as extra occupancy entities). Additionally the quirk will report `Region_Enter` and `Region_Leave` events for these regions.

### Configuration
Exit, interference and edge regions can be written and read, but detection regions can only be written. For a consistent experience I've opted to use cluster commands to implement configuration using a custom schema that allows a fairly intuitive UI (as long as you are familiar with the grid), where each area of the grid is represented by a check box.

image

New commands that have been added:
- Endpoint 1, FP1OccupancySensing cluster, set_exits (0x0001) - sets entrance/exit region
- Endpoint 1, FP1OccupancySensing cluster, set_interference (0x0002) - sets interference region
- Endpoint 1, FP1OccupancySensing cluster, set_edges (0x0003) - sets edge region
- Endpoint 1, FP1OccupancySensing cluster, report_position (0x0004) - enables position reporting mode (see below)
- Endpoint 2-11, FP1RegionXOccupancySensing cluster, set_region (0x0001) - set region configuration
- Endpoint 2-11, FP1RegionXOccupancySensing cluster, clear_region (0x0002) - clear region configuration

It's important to note that the FP1 needs to be awake (listening for zigbee messages, as opposed to just reporting) for these to work, I have had to sometime power-cycle the device before sending configuration.

### Position reporting
The FP1 supports a live position reporting mode, that is used by the Aqara Home app. When enabled, the device reports information about where exactly presence has been detected, before disabling again after a few minutes.

When enabled by issuing the report_position command (or writing 1 to the report_position attribute/0x155), the quirk will send a stream of ZHA events with command="position" as well as writing debug information to the log (which can be viewed by configuring the logger).

```yaml
logger:
  default: warning
  logs:
    zhaquirks.xiaomi.aqara.mootion_ac01: debug
```

This position information includes x and y values to make it easier to map zones.

### Things I'm not quite happy with
The quirk works fine and I'm using it, but there are some things that could be better.

- It adds 10 extra occupancy sensors in HA by default. The user can disable them, but is there a way to make them disabled by default?
- The occupancy sensors are all called occupancy which makes it harder than necessary to identify which is which. Can we influence the sensor naming from the quirk?
- Where would be the place to document this?

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/motion_ac01.py
+++ b/zhaquirks/xiaomi/aqara/motion_ac01.py
@@ -2,15 +2,18 @@
 from __future__ import annotations
 
 import logging
+import math
 from typing import Any
 
 from zigpy.profiles import zha
-from zigpy.quirks import CustomDevice
+from zigpy.quirks import CustomCluster, CustomDevice
 import zigpy.types as types
 from zigpy.zcl.clusters.general import Basic, DeviceTemperature, Identify, Ota
 from zigpy.zcl.clusters.measurement import OccupancySensing
+from zigpy.zcl.foundation import CommandSchema, ZCLCommandDef
 
 from zhaquirks.const import (
+    ARGS,
     COMMAND,
     DEVICE_TYPE,
     ENDPOINTS,
@@ -22,6 +25,7 @@ from zhaquirks.const import (
 )
 from zhaquirks.xiaomi import XiaomiAqaraE1Cluster
 
+MAX_REGIONS = 10
 OCCUPANCY = 0x0000
 PRESENCE = 0x0142
 PRESENCE2 = 101
@@ -30,8 +34,19 @@ PRESENCE_EVENT2 = 102
 MONITORING_MODE = 0x0144
 MOTION_SENSITIVITY = 0x010C
 APPROACH_DISTANCE = 0x0146
+REGION_CONFIG = 0x150
+REGION_PRESENCE_EVENT = 0x151
+EXITS_CONFIG = 0x153
+INTERFERENCE_CONFIG = 0x154
+REPORT_POSITION = 0x155
+EDGES_CONFIG = 0x156
 RESET_NO_PRESENCE_STATUS = 0x0157
+POSITION_EVENT = 0xFFF2
 SENSOR = "sensor"
+REGION = "region"
+POSITION = "position"
+REGION_ANY = "region_any"
+REGION_FORMAT = "region_{}"
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


evnvc:
Is there anyway I can help test this on my end? Have about 4 sensors in the house waiting for this magic to transform them :)

evnvc:
Additional feedback after personally testing the quirk: all functions as described. 
Trouble is it indeed created 10 extra sensors for each FP1, so had quite a lot to delete once disabling. I've used listening to zha_events in the developer tools to set my zones, found it pretty reliable. 

As an observation or further improvement to UX, sensor has 120 degree field of view with a 5m detection distance, so that can perhaps help when mapping out your zones in the blind and then fine tunning them. The event listener also lists the distance in cms and the degree of detection, so that might also be useful data (confirmed with a tape measure).

michalmo:
Thanks for the testing @evnvc.

The device reports the angle and distance, and that is used to compute the zone coordinates in a way familiar from the Aqara app. But since those values can be useful I kept them in the report.

Regarding the extra sensors, that seems to me to be the biggest UX problem. I would like to:
1.  rename them (so that we would have Occupancy, OccupancyZone1, OccupancyZone2, ...).
2. make the zone sensors disabled by default, so that they do not appear in the device page in HA until enabled.
3. (optionally) make the sensor enabled automatically when configured (i.e. when a value is reported for a zone).

However, I wasn't able to find any examples of devices that support something like this, so I have no implementation to work from. Still hoping someone can point me in the right direction.

I also would appreciate any feedback on anything missing in the PR. @TheJulianJES, you labelled this PR, can you weight in or point to someone who can?

evnvc:
All very useful suggestions @michalmo, I too was confused on the naming but managed to figure out which is master and which was zone 1 only by moving around 😆. 

Suggestion two would be great in my case, since I have about 4 sensors in the house so if I could choose the default behaviour it would definitely help (plus in some rooms I only use like two zones anyway).

Thank you again for the work on this, it really elevates the experience. Now with the FP2 taking the news everywhere I was a little bummed out of my investment, but with your quirk I can finally do (almost) the same things, but via Zigbee.

PatrikTheDev:
Hey, any updates on this? I accidentally ordered one (only noticed it’s an FP1 after it got shipped) and this would make it a lot more useful. How quickly does it report the zone occupancy/position changes?

Even if it isn’t ideal, would it be possible to at least get an early version merged in? Thank you for your work

PR #2407, opened 2023-05-22
Add missing attribute to TuyaZBExternalSwitchTypeCluster()

nworbneb:
## Proposed change

Add missing attribute

## Checklist

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -1125,8 +1125,10 @@ class TuyaZBExternalSwitchTypeCluster(CustomCluster):
     name = "Tuya External Switch Type Cluster"
     cluster_id = TUYA_CLUSTER_E001_ID
     ep_attribute = "tuya_external_switch_type"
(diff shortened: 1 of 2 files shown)


zhaquirks/tuya/__init__.py
```diff
@@ -1126,6 +1126,7 @@ class TuyaZBExternalSwitchTypeCluster(CustomCluster):
     cluster_id = TUYA_CLUSTER_E001_ID
     ep_attribute = "tuya_external_switch_type"
     attributes = {0xD030: ("external_switch_type", ExternalSwitchType)}
+    attributes.update({0xD010: ("power_on_state", PowerOnState)})
```
dmulcahey, line 1129: why not just add this to the dict above?

TheJulianJES:
For what exact device do you need this? (Can you post the signature or device diagnostics?)

IIRC there are already multiple Tuya clusters with `power_on_state` (although that may be on the `OnOff`custer).

nworbneb:
This is Moes MS-104BZ identified as _TZ3000_pmz6mjyu TS011F

PR #2408, opened 2023-05-23
Add display config cluster for Aqara TVOC

elupus:
## Proposed change

Add display config cluster for Aqara TVOC

## Additional information

Solution in herdsman: https://github.com/Koenkk/zigbee-herdsman-converters/commit/2d103776fd138ad488ad4ee3025cb9884c1659d5

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/tvoc.py
+++ b/zhaquirks/xiaomi/aqara/tvoc.py
@@ -1,6 +1,5 @@
 """Quirk for lumi.airmonitor.acn01 tvoc air monitor."""
 
-from enum import Enum
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster
 import zigpy.types as t
--- a/zhaquirks/xiaomi/aqara/tvoc.py
+++ b/zhaquirks/xiaomi/aqara/tvoc.py
@@ -1,5 +1,6 @@
 """Quirk for lumi.airmonitor.acn01 tvoc air monitor."""
 
+from enum import Enum
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster
 import zigpy.types as t
@@ -28,6 +29,7 @@ from zhaquirks.xiaomi import (
 )
 
 MEASURED_VALUE = 0x0000
+DISPLAY_UNIT = 0x0114
 
 
 class AnalogInputCluster(CustomCluster, AnalogInput):
@@ -64,6 +66,19 @@ class EmulatedTVOCMeasurement(LocalDataCluster):
             self.MIN_CHANGE,
         )
         return result
+
+
+class TVOCCluster(XiaomiAqaraE1Cluster):
+    """Aqara LUMI Config cluster."""
+
+    ep_attribute = "aqara_cluster"
+    attributes = {
+        # mgm3_celsius    = 0x00
+        # ppb_celsius     = 0x01
+        # mgm3_fahrenheit = 0x10
+        # ppb_fahrenheit  = 0x11
+        DISPLAY_UNIT: ("display_unit", t.uint8_t, True),
+    }
 
 
 class TVOCMonitor(XiaomiCustomDevice):
@@ -114,7 +129,7 @@ class TVOCMonitor(XiaomiCustomDevice):
                     RelativeHumidityCluster,
                     AnalogInputCluster,
                     EmulatedTVOCMeasurement,
-                    XiaomiAqaraE1Cluster,
+                    TVOCCluster,
                 ],
                 OUTPUT_CLUSTERS: [Ota.cluster_id],
             }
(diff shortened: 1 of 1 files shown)


elupus:
I tried using something like `class TVOCCluster(t.uint8_t, Enum)`. but seem the ability to set fields like that from HA is broken. It will try to do `TVOCCluster("TVOCCluster.somename")` or number in string form like: TVOCCluster("0x10")

zhaquirks/xiaomi/aqara/tvoc.py
```diff
@@ -66,6 +67,24 @@ async def bind(self):
         return result
 
 
+class TVOCDisplayUnit(t.enum_factory(t.uint8_t)):
```
TheJulianJES, line 70: Using `t.enum_factory(t.uint8_t)` is fine, but directly using `t.enum8` is preferable.

zhaquirks/xiaomi/aqara/tvoc.py
```diff
@@ -66,6 +67,24 @@ async def bind(self):
...
+    mgm3_celsius = 0x00
+    ppb_celsius = 0x01
+    mgm3_fahrenheit = 0x10
+    ppb_fahrenheit = 0x11
+
+
+class TVOCCluster(XiaomiAqaraE1Cluster):
+    """Aqara LUMI Config cluster."""
+
+    ep_attribute = "aqara_cluster"
```
TheJulianJES, line 82: Generally, `opple_cluster` is used for Aqara devices. This is what ZHA mostly uses to match Aqara devices.
The cluster handler there is also called something with "Opple" at the moment.

However, I think ZHA doesn't yet create a config entity for this. This could be added by doing the following:

- Set `ZCL_INIT_ATTRS` to `display_unit` if the model matches `lumi.airmonitor.acn01` [here](https://github.com/home-assistant/core/blob/9d91cfa27fc10b584114378ee13989e546c03267/homeassistant/components/zha/core/cluster_handlers/manufacturerspecific.py#L111-L116).
- Create a config entity [similar to this here](https://github.com/home-assistant/core/blob/9d91cfa27fc10b584114378ee13989e546c03267/homeassistant/components/zha/select.py#L350-L359).
Note: You'd probably just use `TVOCDisplayUnit`, since `zhaquirks` already defines that (with this PR). Most other quirks do not do that in `zhaquirks`, but directly in ZHA.

If you want (and have some time), feel free to implement that (and PR to HA Core).

TheJulianJES:
> but seem the ability to set fields like that from HA is broken

This might have been improved in the latest HA Core release.

For items that need to be configured, config entities in ZHA should be created. Having users use the cluster UI is not really meant for "doing config stuff", but rather "debugging".

zhaquirks/xiaomi/aqara/tvoc.py
elupus: Does not work. The device reject it. Enum8 maps to some standard type identifier.

zhaquirks/xiaomi/aqara/tvoc.py
elupus: Its somewhere on my todo list. Though if we want to rename anything, we should do that first.

zhaquirks/xiaomi/aqara/tvoc.py
TheJulianJES: It should not. Are you sure you used the zigpy `enum8`?: https://github.com/zigpy/zigpy/blob/a3e449582fe2a1203316dce0955ccfe19d6fdc3b/zigpy/types/basic.py#L461

elupus: Yup: Here it maps the type: https://github.com/zigpy/zigpy/blob/a3e449582fe2a1203316dce0955ccfe19d6fdc3b/zigpy/zcl/foundation.py#L215 

It seem the device only accept the write if its of type 0x20. Which is also how herdsman mapped it: https://github.com/Koenkk/zigbee-herdsman-converters/commit/2d103776fd138ad488ad4ee3025cb9884c1659d5#diff-2d16489d92b9b8bfd6140c0045ff1f2f7e9caf7df7e92445313f1d1a7609da1aR133

elupus:
> > but seem the ability to set fields like that from HA is broken
> 
> This might have been improved in the latest HA Core release.

Yup. I provided a fix for that.

zhaquirks/xiaomi/aqara/tvoc.py
TheJulianJES: Ah, I see. That’s probably why most custom attributes for Xiaomi devices just have their type as `t.uint8_t` in zha-quirks and then have an enum for the config entity in ZHA (as that does not change the "Zigbee type").

PR #2421, opened 2023-06-01
Add Inovelli VZM35-SN support, update Inovelli device triggers

codyhackw:
## Proposed change

This change adds support for Inovelli's VZM35-SN Fan Switch. It also updates the device triggers to allow for config held and released.

## Checklist


zhaquirks/inovelli/VZM35SN.py
```diff
@@ -0,0 +1,124 @@
...
+                    OnOff.cluster_id,
+                    LevelControl.cluster_id,
+                ],
+            },
+            242: {
+                PROFILE_ID: 41440,
+                DEVICE_TYPE: 0x0061,
+                INPUT_CLUSTERS: [],
+                OUTPUT_CLUSTERS: [0x0021],
+            },
```
TheJulianJES:
```suggestion
            242: {
                PROFILE_ID: zgp.PROFILE_ID,
                DEVICE_TYPE: zgp.DeviceType.PROXY_BASIC,
                INPUT_CLUSTERS: [],
                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
            },
```
This should also be changed in the replacement then. (ref: https://github.com/zigpy/zha-device-handlers/pull/2421/commits/e18db60488868533274aa840876cf91a6cf78111)

zhaquirks/inovelli/__init__.py
```diff
@@ -247,12 +247,44 @@ def handle_cluster_request(
...
+    )
+
+
 INOVELLI_AUTOMATION_TRIGGERS = {
     (COMMAND_PRESS, ON): {COMMAND: f"{BUTTON_2}_{COMMAND_PRESS}"},
     (COMMAND_PRESS, OFF): {COMMAND: f"{BUTTON_1}_{COMMAND_PRESS}"},
     (COMMAND_PRESS, CONFIG): {COMMAND: f"{BUTTON_3}_{COMMAND_PRESS}"},
     (COMMAND_HOLD, ON): {COMMAND: f"{BUTTON_2}_{COMMAND_HOLD}"},
     (COMMAND_HOLD, OFF): {COMMAND: f"{BUTTON_1}_{COMMAND_HOLD}"},
+    (COMMAND_HOLD, CONFIG): {COMMAND: f"{BUTTON_3}_{COMMAND_HOLD}"},
```
TheJulianJES, line 287: This device trigger (and the other one below) would also apply to [all dimmers/switches](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/inovelli/VZM31SN.py).
Do those devices also produce these triggers?

TheJulianJES:
Changes seem to be fine IMO. The "copying attributes but excluding some" part isn't that elegant, but I'd be fine with it.
Edit: Although this could be "dangerous" if new attributes are added that we don't want to copy.
~~(Question regarding new device triggers for other devices outstanding)~~ (✅ )

@codyhackw Just to make sure, did you test if your changes work correctly with that device?

Note: the recent commits with the "`ZGP` part" require at least HA Core 2023.7.x (beta right now), but they don't change anything important.

zhaquirks/inovelli/__init__.py
codyhackw: They do, but hadn't been included previously as the original form of factory resetting the device relied on holding the config button (got changed to a combination) and no one had asked about them until recently.

codyhackw:
> @codyhackw Just to make sure, did you test if your changes work correctly with that device?

I can confirm they work correctly with the new device. (Haven't tested the ZGP piece as this is running in prod as a custom quirk for me, but would expect that to work as well once 2023.7.x is released)

TheJulianJES:
@puddly Since you pushed some commits to this, are you fine with this now or is there anything else you wanted to do?

puddly:
I'm fine with the merge, I just fixed the failing unit test and replaced the ZGP constants.

TheJulianJES (approved):
LGTM

We have to be "cautious" when adding new attributes to the `Inovelli_VZM31SN_Cluster` class, as they’d get copied automatically (unless explicitly excluded), but that should be fine probably.

PR #2423, opened 2023-06-02
Sync with upcoming zigpy release

puddly:
## Proposed change

Zigpy's new `attribute_updated` and ZCL schema changes break unit tests. The Sinope leak sensor also had two attributes called `zone_status`.

## Checklist

### DIFF
--- a/zhaquirks/sinope/sensor.py
+++ b/zhaquirks/sinope/sensor.py
@@ -67,7 +67,7 @@ class SinopeTechnologiesIasZoneCluster(CustomCluster, IasZone):
             # FIXME: ZHA currently relies on `zone_status` to exist and hold the correct
             # value. Until this can be properly fixed, we need to manually sync these
             # two attributes.
-            super()._update_attribute(self.AttrDefs.zone_status.id, value)
+            super().update_attribute(self.AttrDefs.zone_status.id, value)
 
 
 class SinopeTechnologiesSensor(CustomDevice):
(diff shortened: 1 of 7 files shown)


zhaquirks/sinope/sensor.py
```diff
@@ -45,7 +45,7 @@ class SinopeManufacturerCluster(CustomCluster):
 class SinopeTechnologiesIasZoneCluster(CustomCluster, IasZone):
     """SinopeTechnologiesIasZoneCluster custom cluster."""
 
-    class ZoneStatus(t.enum8):
+    class LeakStatus(t.enum8):
         """zone_status values."""
```
javicalle, line 49: """leak_status values"""

javicalle (changes requested):
Fix comment

zhaquirks/sinope/sensor.py
```diff
@@ -45,16 +45,16 @@ class SinopeManufacturerCluster(CustomCluster):
...
+        """leak_status values."""
 
         Dry = 0x00
         Leak = 0x01
 
     attributes = IasZone.attributes.copy()
     attributes.update(
         {
-            0x0030: ("zone_status", ZoneStatus, True),
+            0x0030: ("leak_status", LeakStatus, True),
```
TheJulianJES, line 57: Haven't checked ZHA yet, but did this quirk rely on HA creating a binary sensor for an attribute called `zone_status`? (And possibly the higher attribute id overwriting the actual `zone_status`?)

So, is another "binary sensor entity" necessary in ZHA for this `leak_status` now?

puddly: You're right, it does rely on an IASZone `zone_status` attribute existing. I'll add a new matcher should to ZHA in the corresponding PR.

PR #208, opened 2019-12-01
Linear battery voltage

roblandry:
Initial implementation of linear battery voltages.

Fixes: #199
Potential fix: #176

Reported values:
* Existing batteries:
  *  [zhaquirks] CentraLite 3321-S Voltage [RAW]:29 [Max]:2.8 [Min]:1.5, Battery Percent: 100.0
  *  [zhaquirks] CentraLite 3325-S Voltage [RAW]:29 [Max]:2.8 [Min]:1.5, Battery Percent: 100.0
*  New batteries:
   * [zhaquirks] SmartThings tagv4, Voltage [RAW]:30 [Max]:2.8 [Min]:1.5, Battery Percent: 100.0

PR #2430, opened 2023-06-13
Use minimum battery threshold to calc percent

elupus:
## Proposed change

Use minimum battery threshold to calc percent

This fixes the Aqara TVOC sensor battery reporting which seem to have a minimum voltage of 2.7 set for its battery alarm by default.

## Additional information

This could for example change the behaviour of sensors with a nominal value of 3.0, which may not always report 100% even on new batteries. I do think that might be less of a concern. We could of course let max be nominal minus 0.1 volt. 

Reference: https://github.com/zigpy/zha-device-handlers/pull/208/files#r352357340

## Checklist

### DIFF
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -196,9 +196,11 @@ class PowerConfigurationCluster(CustomCluster, PowerConfiguration):
 
     cluster_id = PowerConfiguration.cluster_id
     BATTERY_VOLTAGE_ATTR = 0x0020
+    BATTERY_RATED_VOLTAGE = 0x0034
+    BATTERY_VOLTAGE_MIN_THRESHOLD = 0x0036
     BATTERY_PERCENTAGE_REMAINING = 0x0021
-    MIN_VOLTS = 1.5  # old 2.1
-    MAX_VOLTS = 2.8  # old 3.2
+    MIN_VOLTS = 15  # old 2.1
+    MAX_VOLTS = 28  # old 3.2
 
     def _update_attribute(self, attrid, value):
         super()._update_attribute(attrid, value)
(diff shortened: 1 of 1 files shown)


elupus:
Should i split this feature into some new custom cluster? It seems rather useful. Likely for other devices too.

For example: Av201032PowerConfigurationCluster is likely redundant now, since it will report it's min threshold.

TheJulianJES:
It's not guaranteed that those attribute values are already cached. Currently, you'd probably need to read them manually for this to work.
ZHA could initialize/read those attributes during pairing, but this won't work retroactively for currently paired devices.

It would probably be best to just apply this to the Aqara TVOC sensor with static MIN/MAX values (that you read once and hard-code in the quirk).

Use something like this:
https://github.com/zigpy/zha-device-handlers/blob/7120cbddc6411c8bd5d7a8e8ee8929ee0df1cdf0/zhaquirks/xiaomi/aqara/remote_h1.py#L80-L86
and then use the new cluster in the replacement.

TheJulianJES:
@elupus (ping) in case you haven't seen my comment above and still want to work on this.

elupus:
I saw it.. Not decided how i want to approach it. Quirking it for for aquara seems like cheating and will require more work in the future for other devices so i dont like it.

I want a solution that is beneficial for future devices too.

TheJulianJES:
> Quirking it for for aquara seems like cheating and will require more work in the future

Using a custom cluster like mentioned in my [previous comment](https://github.com/zigpy/zha-device-handlers/pull/2430#issuecomment-1657226556) is not really a workaround.
It's the cleanest way to fix that device without breaking others.

But even Aqara devices themselves behave very differently with their battery reporting. Some use a custom cluster for reporting voltage only, some only report percentage, ...

[Z2M uses `2.850V` to `3V`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/dd71c7b50e7cdba87869d202e398f5d28abc1a25/src/devices/xiaomi.ts#L2880) for the Aqara TVOC device.
If that matches the "min voltage threshold" + "rated voltage" attributes for that device, we should probably use them too.

Hedda:
Any updates on this?

PR #2455, opened 2023-07-02
Add support for Moes Smart Button (Tuya TS004F variant)

### LINKED ISSUE #1697: [Device Support Request] not standard TS004F?
I got a [Moes Smart Button](https://www.moeshouse.com/products/ip55-waterproof-smart-zigbee-scene-switch-wireless-remote-dimmer?_pos=87&_sid=426664fbd&_ss=r). I can register it with ZHA, but I can create event for press, double press or long press.

I don't really understant the device reference.  
It reports it-self like a TS004F _TZ3000_kjfzuycl. TS004F is the reference, and TZ3000_kjfzuycl a kind of manufacturer ID?

There is already a quirk for [TS004F](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts004f.py), but it seems to be a 4 gang switch? with different manufacturer?

Is it really the same product, or do I have to try to make a news TS004F quirk?

A would like to catch event on press, double press and long press, in order to make automations.

<details>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0104",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0004",
        "0x0006",
        "0x1000",
        "0xe001"
      ],
      "out_clusters": [
        "0x0003",
        "0x0004",
        "0x0006",
        "0x0008",
        "0x000a",
        "0x0019",
        "0x1000"
      ]
    }
  },
  "manufacturer": "_TZ3000_kjfzuycl",
  "model": "TS004F",
}
```

</details>

<details>

```yaml
I don't have any diagnostic button.
```
image
</details>

<details>

**with a 10s long press**
```
_TZ3000_kjfzuycl TS004F Stop event was fired with parameters: {'options_mask': None, 'options_override': None}
17:05:42 - Maintenant
_TZ3000_kjfzuycl TS004F Step event was fired with parameters: {'step_mode': <StepMode.Up: 0>, 'step_size': 85, 'transition_time': 10, 'options_mask': None, 'options_override': None}
17:05:41 - Maintenant
```

**with a single press**
```
_TZ3000_kjfzuycl TS004F Move To Level event was fired with parameters: {'level': 255, 'transition_time': 1, 'options_mask': None, 'options_override': None}
17:06:22 - Il y a 2 minutes
```

**with a double press**
```
_TZ3000_kjfzuycl TS004F 92aa42fe on_of  
éteint
Add any other context or screenshots about the feature request here.
17:06:16 - Il y a 2 minutes
```
</details>

evgenln:
## Proposed change
Add support for [Moes Smart Button](https://www.moeshouse.com/products/ip55-waterproof-smart-zigbee-scene-switch-wireless-remote-dimmer?_pos=87&_sid=426664fbd&_ss=r).

## Additional information
At the moment known 2 versions of the Moes Smart Button (_TZ3000_kjfzuycl, _TZ3000_ja5osu5g), both look the same.
It was possible to add it with ZHA as a switch but not possible to create the events for press, double press, or long press.
Fixes [1697](https://github.com/zigpy/zha-device-handlers/issues/1697)

## Checklist


zhaquirks/tuya/ts004f.py
```diff
@@ -412,3 +414,86 @@ class TuyaSmartRemote004F(EnchantedDevice):
...
+        (LONG_RELEASE, BUTTON): {COMMAND: COMMAND_STOP, ENDPOINT_ID: 1, CLUSTER_ID: 8},
+        (ALT_SHORT_PRESS, BUTTON): {
+            ENDPOINT_ID: 1,
+            CLUSTER_ID: 8,
+            PARAMS: {
+                "transition_time": 1,
+                "options_mask": None,
+                "options_override": None,
+            },
+        },
```
TheJulianJES: When does this trigger (vs normal `SHORT_PRESS`)? Is there no event "command" sent?

(If it's the `step` command for example, matching by `step` isn't enough? The params are needed?)

zhaquirks/tuya/ts004f.py
```diff
@@ -412,3 +414,86 @@ class TuyaSmartRemote004F(EnchantedDevice):
             ENDPOINT_ID: 1,
         },
     }
+
+
+class TuyaSmartRemote004FSK(EnchantedDevice):
```
TheJulianJES: Can you move this whole class above `TuyaSmartRemote004F`? (because that class only matches by model)

evgenln: Done

zhaquirks/tuya/ts004f.py
evgenln: The button works in two modes Dimmer switch and Scene.
In dimer mode, when lights are off, the device generates an ON command (normal SHORT_PRESS), but when it is already ON, it generates move_to_color_temp, move_to_level, or move_to_hue_and_saturation commands. The alternate mode is for a single press when the lights are on. Listed PARAMS are the only way to catch SHORT_PRESS 

https://github.com/zigpy/zha-device-handlers/issues/1697#issuecomment-1221291611

TheJulianJES (approved):
LGTM. Thanks!

Maybe @javicalle can have a quick look too? 😄

javicalle:
Hi Julian,
 any concerns here? Nothing chimes to me.
Maybe the `TuyaZBExternalSwitchTypeCluster` in the replacement side is not needed but don't harm neither so... is fine to me.
The `zha.DeviceType.NON_COLOR_CONTROLLER` vs the new `Color` cluster in the output is something that would me make check the code and implications but I have no time to check myself (just to my knowledge and better understand of code). 

So LGTM too.

javicalle (approved):

TheJulianJES:
Good catch with the `Color` cluster. Probably not needed either, as I don’t see any device triggers (commands) from that cluster, but the other quirks already have that so probably fine 😄

Thanks for checking as always!

javicalle:
> Good catch with the `Color` cluster. Probably not needed either

Ummmm, not sure because the device seems to trigger the command (or at least report from `0x0300` cluster):
https://github.com/zigpy/zha-device-handlers/issues/1697#issuecomment-1221291611

So the cluster should be needed and is consistent with all the others quirks.

MattWestb:
**If** changing from the original `zha.DeviceType.DIMMER_SWITCH` is shall being one color controller and not `DeviceType.NON_COLOR_CONTROLLER` then one` COLOR_CONTROLLER = 0x0800` but i dont knowing if its one problem for ZHA and if it is it OK having more cluster then the device profile is saying by the standard but all moderate must being implanted.

PS: I have using scene controller or dimmer for the other TS004F = using the profile that is matching best.

And great work done getting the PR made after very long time and remote testing with outer for getting it working !!

MattWestb:
The color cluster is sending commands but its not used in the DA then using the level command instead that is being sent with the same function as normal tuya standard Zigbee devices is doing.

Issue #2441, opened 2023-06-25
[Device Support Request] Tuya HC-T020-ZIGBEE Series Thermostat


adamantivm:
### Problem description

Device is recognized quickly as TS0601 by _TZE200_ztvwu4nk but doesn't expose any sensors or controls

### Solution description

At the bare minimum, it would be really nice to be able to read current temperature and set desired temperature, that would already allow quite a bit of automation through Home Assistant.

In an ideal scenario, exposing all the functionality available in the device (usable through the Tuya app) would be great.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

Here is a great writeup by @notenoughtech that describes the device in some detail: https://notenoughtech.com/home-automation/a-budget-zigbee-thermostat/

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.FullFunctionDevice|MainsPowered|RxOnWhenIdle|AllocateAddress: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_ztvwu4nk",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2023.6.3",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
[...]

PR #2466, opened 2023-07-08
Adds support for HC-T020 thermostat

adamantivm:
## Proposed change

Adds basic support for Tuya HC-T020 Zigbee thermostat, addressing #2441

Aliexpress link: https://www.aliexpress.us/item/3256803286637740.html?gatewayAdapt=glo2usa4itemAdapt

With this, it's possible to read current temperature, set temperature and know if the device is currently heating or idle. 
It's the minum necessary to be able to use it in Home Assistant for basic automations.

## Additional information

It's heavily based on the [MoesBHT quirk](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts0601_electric_heating.py)

## Checklist


adamantivm:
@TheJulianJES any suggestions on what else to do to get this reviewed and incorporated?

I have this successfully working with one device. I had to fix a regression today which appeared after HA core update to 2023.8.0 which included zha_quirks 0.0.102, but that's solved now.

If you can give me any guidance / pointers on what else should be done, I'd appreciate it.

TheJulianJES:
First of all, thanks for your PR!

The already existing code for the Tuya thermostats is somewhat of a mess and I'm not very familiar with it.
For example, everything related to `Bus` could be cleaned up, as that doesn't need to be used anymore.
Updating an attribute on the thermostat cluster can be done like this now:
```python
self.endpoint.thermostat.update_attribute(
    Thermostat.AttributeDefs.local_temperature.id,
    local_temp,
)
```
A method defined in a(n underlying) `CustomCluster` can also be called like that (just instead of `update_attribute`, use the method name).

Like you mentioned, the code is very similar to the Moes quirk (which could/should also be cleaned up).

Lastly, this PR doesn't have any tests for the newly added code ([`test_tuya.py`](https://github.com/zigpy/zha-device-handlers/blob/dev/tests/test_tuya.py) as somewhat of an example).
This is also not done for some other Tuya thermostat, so I'll have to ask what we should do about this.
(**If** you could write code that tests the quirk/"emulates the thermostat", that would be beneficial of course)

MattWestb:
The [ts0601_electric_heating.py](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts0601_electric_heating.py) is for thermostats.
TRVs is in [ts0601_trv.py](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts0601_trv.py) and some have getting own that have different implementations but its always possible changing names on the files for getting it more logic like [ts0601_haozee.py](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts0601_haozee.py) so its looks like this one [ts0601_trv_sas.py](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts0601_trv_sas.py).

adamantivm:
Thanks @TheJulianJES and @MattWestb 
I have to admit I'm completely new to ZHA and the quirks, but I can give a shot at adding tests and perhaps using a more modern approach by following your pointers.

@MattWestb I didn't understand the suggestion implied by your comment. Are you suggesting any changes in the file naming perhaps?

MattWestb:
@adamantivm It was more for Julian and the other maintainers for getting information and searing the quirk naming.
I have done some renaming of tuya quirks but the responsibility is by the maintainers and i can only suggesting changes and must living with how its being done.

Edit: Your naming is very good but 2 of my examples shall being renamed i think.
perhaps putting the Thermostat in front so its being sorted OK
Like `ts0601_thermostat_gas_heating.py`

MattWestb:
@TheJulianJES I think `ts0601_electric_heating.py` shall being `ts0601_thermostat_electric_heating.py` or `ts0601_thermostat_electric.py`  and `ts0601_haozee.py` `ts0601_trv_haozee.py`.

I dont do renaming PR then i cant fixing if i braking the test linked to them.

PR #2491, opened 2023-07-26
Migrate to `pyproject.toml`

puddly:
## Proposed change

Migrate the zha-quirks repo to `pyproject.toml`. This should fix CI.

~~To align with the other zigpy repositories, version info is no longer stored in the repo itself (e.g. `setup.py:VERSION`) and instead will be pulled from the Git tag present on release.~~

## Checklist


puddly:
The tag-based versioning will have to go in a second PR, after consolidating `dev` and `master`.

PR #2501, opened 2023-08-01
Use shared PyPi release workflow

TheJulianJES:
## Proposed change

This should fix the PyPi GitHub release action. It uses the shared zigpy PyPi release workflow.
Failing log: <https://github.com/zigpy/zha-device-handlers/actions/runs/5729959163/job/15527723986>

As far as I can see, the issue was that `python3 setup.py sdist bdist_wheel` was still used for building instead of `python3 -m build`.

This change was somewhat tested in a private repo and seems to build correctly (with metadata now).

## Checklist

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -6,7 +6,7 @@ on:
       - released
 
 jobs:
-  shared-ci:
+  shared-build-and-publish:
     uses: zigpy/workflows/.github/workflows/publish-to-pypi.yml@main
     with:
       PYTHON_VERSION_DEFAULT: 3.9.15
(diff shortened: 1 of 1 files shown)


TheJulianJES:
I also noticed that only the zigpy-znp repo was using "on: published" instead of "on: released" for the trigger.

It seems like the only difference would be whether or not we want to publish pre-releases on PyPi. As we're not using them anyway, I don't think this matters.
But this should probably match for all repos in the future.

TheJulianJES:
@puddly If you're around, would you mind having a quick look at this?

Publishing for `0.0.102` failed because the PyPi workflow was still outdated and I missed that before publishing the release.
This PR now uses the shared workflow and it seems to work as expected (tested a bit in a private repo).

I'm guessing the next release tag should be `0.0.103` then (and 0.0.102 will be left as a tag, but not published to PyPi)(?)

Issue #2493, opened 2023-07-28
[Device Support Request] Fingerbot (_TZ3210_dse8ogfy, TS0001)


brettjurgens:
### Problem description

I purchased this [Zigbee Fingerbot from aliexpress](https://www.aliexpress.us/item/3256805414231144.html) and it has limited support in home assistant.

I've attempted to write a custom quirk based on the code from [zigbee-herdsman-converters](https://github.com/Koenkk/zigbee-herdsman-converters/blob/8181aefa9344bac21a95d2033bee3c8833ec61e1/src/devices/tuya.ts#L5175-L5216). I can read most of the values, but writing does not work.

### Solution description

I'd like to be able to write the `mode`, `down_movement`, and `up_movement` data points.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

![Screenshot 2023-07-27 at 19 32 41](https://github.com/zigpy/zha-device-handlers/assets/1226910/3661de92-fa58-446a-9253-4c02632c45a0)

![Screenshot 2023-07-27 at 19 33 05](https://github.com/zigpy/zha-device-handlers/assets/1226910/a959dcce-99d3-4abb-b0ab-cf5007bc7a13)

![Screenshot 2023-07-27 at 19 33 42](https://github.com/zigpy/zha-device-handlers/assets/1226910/50aad00c-3876-4308-ad68-70c4cbd545da)

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0103",
      "input_clusters": [
        "0x0000",
[...]

PR #2506, opened 2023-08-02
Add Tuya fingerbot `_TZ3210_dse8ogfy` support

brettjurgens:
## Proposed change

Add support for Tuya Zigbee Fingerbot

## Additional information

Device support request: #2493

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0001.py
+++ b/zhaquirks/tuya/ts0001.py
@@ -1,0 +1,148 @@
+from typing import Any, Dict, Optional, Union
+
+from zigpy.profiles import zha
+import zigpy.types as t
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.general import Basic, Groups, Ota, Scenes, Time
+
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+from zhaquirks.tuya import TUYA_SEND_DATA
+from zhaquirks.tuya.mcu import (
+    DPToAttributeMapping,
+    EnchantedDevice,
+    TuyaMCUCluster,
+    TuyaOnOff,
+    TuyaEnchantableCluster,
+)
+
+
+class FingerBotMode(t.enum8):
+    CLICK = 0x00
+    SWITCH = 0x01
+    PROGRAM = 0x02
+
+
+class FingerBotReverse(t.enum8):
+    UP_ON = 0x00
+    UP_OFF = 0x01
+
+
+class TuyaFingerbotCluster(TuyaEnchantableCluster, TuyaMCUCluster):
+    attributes = TuyaMCUCluster.attributes.copy()
+    attributes.update({101: ("mode", FingerBotMode)})
+    attributes.update({102: ("down_movement", t.uint16_t)})
+    attributes.update({103: ("sustain_time", t.uint16_t)})
+    attributes.update({104: ("reverse", FingerBotReverse)})
+    attributes.update({105: ("battery", t.uint16_t)})
+    attributes.update({106: ("up_movement", t.uint16_t)})
+    attributes.update({107: ("touch_control", t.Bool)})
+
+    async def command(
+        self,
+        command_id: Union[foundation.GeneralCommand, int, t.uint8_t],
+        *args,
+        manufacturer: Optional[Union[int, t.uint16_t]] = None,
+        expect_reply: bool = True,
+        tsn: Optional[Union[int, t.uint8_t]] = None,
+        **kwargs: Any,
+    ):
+        """Override the default Cluster command."""
+
+        return await super().command(
+            TUYA_SEND_DATA,
+            *args,
+            manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID,
+            expect_reply=expect_reply,
+            tsn=tsn,
+            **kwargs,
+        )
+
+    dp_to_attribute: Dict[int, DPToAttributeMapping] = {
+        1: DPToAttributeMapping(TuyaOnOff.ep_attribute, "on_off"),
+        # Mode
+        101: DPToAttributeMapping(
+            TuyaMCUCluster.ep_attribute,
+            "mode",
+            converter=lambda x: FingerBotMode(x),
+        ),
+        # Down Movement
+        102: DPToAttributeMapping(
+            TuyaMCUCluster.ep_attribute,
+            "down_movement",
+        ),
+        # Sustain Time
+        103: DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "sustain_time"),
+        # Reverse
+        104: DPToAttributeMapping(
+            TuyaMCUCluster.ep_attribute,
+            "reverse",
+            converter=lambda x: FingerBotReverse(x),
+        ),
+        # Battery
+        105: DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "battery"),
+        # Up Movement
+        106: DPToAttributeMapping(
+            TuyaMCUCluster.ep_attribute,
+            "up_movement",
+        ),
+        107: DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "touch_control"),
+    }
+
+    data_point_handlers = {
+        1: "_dp_2_attr_update",
+        101: "_dp_2_attr_update",
+        102: "_dp_2_attr_update",
+        103: "_dp_2_attr_update",
+        104: "_dp_2_attr_update",
+        105: "_dp_2_attr_update",
+        106: "_dp_2_attr_update",
+        107: "_dp_2_attr_update",
+    }
+
+
+class TuyaFingerbot(EnchantedDevice):
+    signature = {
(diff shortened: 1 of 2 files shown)


zhaquirks/tuya/ts0001.py
```diff
@@ -0,0 +1,148 @@
...
+    SWITCH = 0x01
+    PROGRAM = 0x02
+
+
+class FingerBotReverse(t.enum8):
+    UP_ON = 0x00
+    UP_OFF = 0x01
+
+
+class TuyaFingerbotCluster(TuyaEnchantableCluster, TuyaMCUCluster):
```
javicalle: Maybe I'm wrong, but I would say that you don't need extend the `TuyaEnchantableCluster` here.
You have an `EnchantedDevice` and the `TuyaOnOff` is already and enchantable cluster.

Are you sure that this device needs the Tuya spell?

zhaquirks/tuya/ts0001.py
```diff
@@ -0,0 +1,148 @@
...
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
```
javicalle: Not in the signature part.
Are these clusters needed?

zhaquirks/tuya/ts0001.py
```diff
@@ -0,0 +1,148 @@
...
+        expect_reply: bool = True,
+        tsn: Optional[Union[int, t.uint8_t]] = None,
+        **kwargs: Any,
+    ):
+        """Override the default Cluster command."""
+
+        return await super().command(
+            TUYA_SEND_DATA,
+            *args,
+            manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID,
```
javicalle: Can you confirm if the `NO_MANUFACTURER_ID` is needed too?

javicalle (changes requested):
Not happy with the filename because `ts0001` devices are defined in `ts000x.py` file. But this is a very especific ts0001 device.
Maybe `ts0001_fingerbot.py`?

zhaquirks/tuya/ts0001.py
```diff
@@ -0,0 +1,148 @@
...
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    TuyaOnOff.cluster_id,
```
TheJulianJES: The `OnOff` cluster isn't actually replaced by the custom `TuyaOnOff` cluster.
If you want to do that, you have to leave out the `.cluster_id`.

The "Tuya spell" will also work then, since that cluster is bound by ZHA, it extends `TuyaEnchantableCluster` and you're using `EnchantedDevice`.
Z2M seems to at least "cast the Tuya spell" for this device.

zhaquirks/tuya/ts0001.py
TheJulianJES: Yep, and AFAIK ZHA doesn't bind this cluster by default and the "Tuya spell is cast" during binding.
So I'd remove `TuyaEnchantableCluster` from this cluster.

zhaquirks/tuya/ts0001.py
TheJulianJES: ~~Yup, and if it's really needed, I think you should also be able to have `TuyaFingerbotCluster` extend `NoManufacturerCluster`.~~
EDIT: Nvm, I missed your comment about replacing `TUYA_SET_DATA` with `TUYA_SEND_DATA`.

zhaquirks/tuya/ts0001.py
MattWestb: Z2M have adding the spell to all tuya device but "normal Zigbee" and MCU devices what  i have seen and looks working OK.

TheJulianJES:
Thanks your for PR!
We've added some comments and it would be amazing if you could look at the requested changes.

zhaquirks/tuya/ts0001.py
javicalle: It can be done, but it needs also overwritte the `command_id` to `TUYA_SEND_DATA` to make it work:
* https://github.com/zigpy/zha-device-handlers/issues/2493#issuecomment-1661338801

zhaquirks/tuya/ts0001_fingerbot.py
```diff
@@ -0,0 +1,145 @@
...
+
+class TuyaFingerbotCluster(TuyaMCUCluster):
+    attributes = TuyaMCUCluster.attributes.copy()
+    attributes.update({101: ("mode", FingerBotMode)})
+    attributes.update({102: ("down_movement", t.uint16_t)})
+    attributes.update({103: ("sustain_time", t.uint16_t)})
+    attributes.update({104: ("reverse", FingerBotReverse)})
+    attributes.update({105: ("battery", t.uint16_t)})
+    attributes.update({106: ("up_movement", t.uint16_t)})
+    attributes.update({107: ("touch_control", t.Bool)})
```
TheJulianJES: I think you should be able to use one `update` call with all attributes:

```suggestion
    attributes = TuyaMCUCluster.attributes.copy()
    attributes.update(
        {
            101: ("mode", FingerBotMode),
            102: ("down_movement", t.uint16_t),
            103: ("sustain_time", t.uint16_t),
            104: ("reverse", FingerBotReverse),
            105: ("battery", t.uint16_t),
            106: ("up_movement", t.uint16_t),
            107: ("touch_control", t.Bool),
        }
    )
```
(I know some Tuya code does it differently, but this seems a bit nicer.)

zhaquirks/tuya/ts0001_fingerbot.py
```diff
@@ -0,0 +1,145 @@
...
+        # Sustain Time
+        103: DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "sustain_time"),
+        # Reverse
+        104: DPToAttributeMapping(
+            TuyaMCUCluster.ep_attribute,
+            "reverse",
+            converter=lambda x: FingerBotReverse(x),
+        ),
+        # Battery
+        105: DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "battery"),
```
TheJulianJES: This alone doesn't create a battery sensor in HA, right?

Maybe you could try something like this:
```python
        105: DPToAttributeMapping(
            TuyaPowerConfigurationCluster.ep_attribute,
            "battery_percentage_remaining",
        ),
```
and then include `TuyaPowerConfigurationCluster` in the `replacement`.

(And for whatever reason, we have `TuyaPowerConfigurationCluster` twice. Use the one from the `mcu` module/folder.)

EDIT: Pushed a commit with this change now. Please let me know if the battery entity shows up (and works).
You might need to restart HA twice or remove + re-pair the device.

TheJulianJES:
@brettjurgens Just in case you didn't see the PR feedback (and since we're releasing a new zha-quirks version very soon), here's a ping.

Since I don't have the device to test with, it would be really nice if you could ~~implement~~ **test** the changes (especially the ones from [this comment regarding the battery](https://github.com/zigpy/zha-device-handlers/pull/2506#discussion_r1287325404)) to see if they work.

EDIT: See comment below.

TheJulianJES:
I've pushed a commit with the changes to your branch/PR now.
Please let me know if they work and if the battery entity shows up. You might need to restart HA twice and/or remove and re-pair the device (if the battery entity doesn't show up in the beginning).

brettjurgens:
Sorry, been really busy 😞

I'll give it a try tomorrow!

TheJulianJES:
Changes look good to me. To get this into HA 2023.9.0 betas, I'll merge it for now.
IMO, the battery entity *should* work now. If it doesn't, it can be fixed later, as it's not needed for the main functionality.
(Please still report back if it works, as soon as you have some time)

The (`movement`) attributes could be exposed as ZHA entities in the future (code needs to go into ZHA component @ HA core repo).
I'll leave the original issue open for now, so we/I don't forget about it (and we should probably include tests for this device as well).

brettjurgens:
Battery entity doesn't seem to be working. 

Unfortunately, moving to TuyaOnOff also broke the switch. I opened #2554 to move to the regular OnOff which does work

ifuller1:
~~I found this post via a home assistant page. Would you mind giving me a pointer on how I can leverage this change? I'm running home assistant on my own box via docker.~~

I found it here, sorry for the ping: https://community.home-assistant.io/t/tuya-fingerbot-plus-in-zha-not-responding-to-switch-state-in-home-assistant/613481

TheJulianJES:
Just upgrade to Home Assistant Core 2023.10.0 (or later) and do not use any outdated custom quirks. They'll break at some point.

ifuller1:
Thanks!

ifuller1:
~For the uninitiated. Where can I find documentation related to how to
leverage this? I can see more options but nothing seems to result in the
switch staying depressed. Do I need to write some config for that somehow?~

Worked it out. Changed mode to "SWITCH" and fiddled with up/down until they didn't move too much. Thanks all!

Issue #2344, opened 2023-04-16
[Device Support Request] Tuya 3 gang switch `_TZ3000_qewo8dlz`


Horley900:
**Is your feature request related to a problem? Please describe.**I
I have an issue with tuya 2 gang and 3 gang light switch's, i was running a zha quirk up untill yesterday, this morning it doesnt seem to be having any effect.
the quirk resolved the issue of when pressing one button on the switch  within HA it would turn all lights on and not treat them as seperat entiteis.
This is on a 3 gang switch , TS0013 by _TZ3000_qewo8dlz and a 2 gang switch, TS0012by _TZ3000_18ejxno0.

**Describe the solution you'd like**
each button to trigger its corresponding light seperatly.

<details>

3 gang signature:

{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006"
      ],
      "out_clusters": [
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0004",
        "0x0005",
        "0x0006"
      ],
      "out_clusters": []
    },
    "3": {
      "profile_id": 260,
      "device_type": "0x0100",
      "in_clusters": [
        "0x0004",
        "0x0005",
        "0x0006"
      ],
      "out_clusters": []
    }
  },
  "manufacturer": "_TZ3000_qewo8dlz",
  "model": "TS0013",
[...]

PR #845, opened 2021-04-13
Add quirk ts001x.py

sylvaing:
Add quirk for Tuya no neutral device TS001x (zigpy#826)

PR #2524, opened 2023-08-12
Fix `SKIP_CONFIGURATION` being set for enchanted Tuya devices

TheJulianJES:
## Proposed change

The "Tuya spell" is required for many Tuya devices and is currently called when binding certain clusters.
These Tuya devices had `SKIP_CONFIGURATION` set which prevents the spell from being called at all, as ZHA does not call `bind()` for any cluster.

This PR adds a test to detect such invalid quirks and also removes `SKIP_CONFIGURATION` for the affected devices.

### TODO:
~~potentially figure out why `SKIP_CONFIGURATION` was added in the first place~~

@MattWestb @javicalle Sorry for the ping, but any ideas why this was used?
I guess it's just unnecessary to bind the `OnOff` and `Scenes` cluster (for this device)?
It shouldn't cause any harm if we try bind it though, right? I think that's what we're doing with other (similar?) Tuya devices.

Confirmed working in: https://github.com/home-assistant/core/issues/96452#issuecomment-1679185854

### Note:
The quirk only seems to:
1. remove the `Time` cluster from the replacement, as that was [apparently causing problems with the device turning itself off randomly ](https://github.com/zigpy/zha-device-handlers/pull/845#issuecomment-829218739)
2. add some custom attributes
3. cast the Tuya spell

## Additional information

Should address the issues in:
- https://github.com/home-assistant/core/issues/96452
- https://github.com/zigpy/zha-device-handlers/issues/2344
- https://github.com/zigpy/zha-device-handlers/issues/1981

## Checklist

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -32,6 +32,7 @@ from zhaquirks.const import (
     ON,
     OUTPUT_CLUSTERS,
     PROFILE_ID,
+    SKIP_CONFIGURATION,
     ZONE_STATUS_CHANGE_COMMAND,
 )
 from zhaquirks.tuya import (
@@ -1638,6 +1639,13 @@ async def test_tuya_spell(zigpy_device_from_quirk):
 
         for quirk in ENCHANTED_QUIRKS:
             device = zigpy_device_from_quirk(quirk)
+
+            # fail if SKIP_CONFIGURATION is set, as that will cause ZHA to not call bind()
+            if getattr(device, SKIP_CONFIGURATION, False):
+                pytest.fail(
+                    f"Enchanted quirk {quirk} has SKIP_CONFIGURATION set. "
+                    f"This is not allowed for enchanted devices."
+                )
 
             for cluster in itertools.chain(
                 device.endpoints[1].in_clusters.values(),
--- a/zhaquirks/tuya/ts001x.py
+++ b/zhaquirks/tuya/ts001x.py
@@ -10,7 +10,6 @@ from zhaquirks.const import (
     MODEL,
     OUTPUT_CLUSTERS,
     PROFILE_ID,
-    SKIP_CONFIGURATION,
 )
 from zhaquirks.tuya import (
     TuyaSwitch,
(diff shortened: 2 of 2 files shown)


MattWestb:
I was suggesting testing it with some device then xiaomi sensor devices that is not needing binding then they is using the coordinator for all uni cast.
In sniff tuya is not setting up binding and reporting and some tricky devices is getting problems if doing it. (IKEA is also very much self configuration and TF is doing nothing then getting group and address thru touch link paring).

Some new TS004X devices that is daring batteries if binding the power cluster but the last new versions of them is also need the tuya magic spell for working OK.

I think its great getting the spell also working for this devices that need both trix for working OK.

I think our "Hola" JC is having better memory as i is having.

javicalle:
Hi Julian,
 no problem at all for the ping but sorry for my late reply.

Some time ago I tried a couple of times to understand the functionof the `SKIP_CONFIGURATION` configuration but I really couldn't understand it, so if I have used it somewhere it has been only for a code homogeneity or by mistake, so I can't explain the reason why it is included in those quirks.

It seems to first come in the https://github.com/zigpy/zha-device-handlers/commit/674238873c4a5ec4708467d771877fa9cc2dc102, but there is no clue about where that quirk comes from. I could suspect that comes from other quirk, but it doesn't seem to be the `ts000x` (doesn't include any `SKIP_CONFIGURATION` configuration).

Also I'm not fluent in the Zigbee commands/behavior, so I'm not much help in changes impact analysis either 😓

TheJulianJES:
Confirmed working in: https://github.com/home-assistant/core/issues/96452#issuecomment-1679185854

dr3amr:
Works fine for me with both devices with 2gang TS0012 and 3gang TS0013. I had 1x2gang and 1x3gang (newer revisions) that when switched from ZHA it would switch on all lights of the same switch as mentioned here (working fine when triggered by touch). The older ones TS0011/TS0012/TS0013 all worked fine. Tried with the modified quirk as bellow, and now older ones and also new ones work as expected 👍 
https://raw.githubusercontent.com/zigpy/zha-device-handlers/6a2693f9fd179124cf09d2f204c9b714ecc077f4/zhaquirks/tuya/ts001x.py

PR #2528, opened 2023-08-17
Clean up Xiaomi quirks by removing `Bus` usage

TheJulianJES:
## Proposed change

This PR refactors the Xiaomi quirks to do the following:
- remove specific Xiaomi usage of `Bus` (custom clusters in the main/quirks `__init__.py` and more still use `Bus`)

Other smaller changes:
- uses zigpy `AttributeDefs` in more places for getting attribute IDs (/removes "magic numbers")
- removes redundant `cluster_id` for custom clusters extending ZCL clusters
- improves one test by no longer calling the Bus directly, but instead creating a Xiaomi attribute report (new method `create_aqara_attr_report` will very likely be used in future PRs)
- adds test for Aqara P1 sensor
- adds test for old Aqara motion sensor
- adds test for Aqara weather sensor
- general cleanup 

## Additional information

While unit tests pass, some stuff isn't unit-tested, ~~so this PR is basically untested and needs to be tested manually with all the sensors.~~
I've tested the PR in my production setup with different Aqara devices (which caused me to add more unit tests for multiple motion sensor models, weather/temp sensors, and so on) and it seems to work just fine.

### TODO:
- see if we even want to do this? (missing a `Bus()` initializer when creating a quirk is a somewhat common problem though and `Bus` is technically no longer needed)
- ✅ test changes (-> didn't notice any issues in production) 
- ✅ maybe also replace `attributes_by_name` with `AttributeDefs` where possible? (-> done in tests for now)
- ✅ think about [this comment](https://github.com/zigpy/zha-device-handlers/pull/2528#issuecomment-1685396364) (-> [addressed in comment below](https://github.com/zigpy/zha-device-handlers/pull/2528#issuecomment-1685488920))

## Checklist

- [x] The changes are tested and work correctly (-> mostly, tested with many Aqara devices with no issues)
- [x] Tests have been added to verify that the new code works (-> mostly yes)


TheJulianJES:
So the battery bus stuff is slightly more tricky.
Apparently most Xiaomi quirks implement the `XiaomiCluster`, so they could receive the "mV" battery reading via that.
Previously, the `XiaomiCustomDevice` implemented a `battery_bus` for all devices (regardless of whether another cluster actually used it).
I'm guessing this was the reason, because they can always receive a mV battery attribute report via a `XiaomiCluster`.
It was then up to the devices to see if they showed that report via the `XiaomiPowerConfiguration` cluster.

All of the following devices implement a `XiaomiCluster`, but not the `XiaomiPowerConfiguration` cluster:
(Some even implement other power clusters, even though there are mV reports sent.)

```python
'zhaquirks.xiaomi.aqara.ctrl_neutral.CtrlNeutral'
'zhaquirks.xiaomi.aqara.ctrl_neutral.CtrlNeutral_2G'
'zhaquirks.xiaomi.aqara.illumination.Illumination'
'zhaquirks.xiaomi.aqara.illumination.IlluminationT1'
'zhaquirks.xiaomi.aqara.light_aqcn2.LightAqcn02'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB286OPCN01V4'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB286OPCN01V3'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB286OPCN01Alt'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB286OPCN01V2'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB286OPCN01'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB486OPCN01V4'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB486OPCN01V3'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB486OPCN01V2'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB486OPCN01'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB686OPCN01V5'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB686OPCN01V4'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB686OPCN01V3'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB686OPCN01V2'
'zhaquirks.xiaomi.aqara.opple_remote.RemoteB686OPCN01'
'zhaquirks.xiaomi.aqara.plug.Plug2'
'zhaquirks.xiaomi.aqara.plug.Plug'
'zhaquirks.xiaomi.aqara.plug_maus01.Plug'
'zhaquirks.xiaomi.aqara.relay_c2acn01.Relay'
'zhaquirks.xiaomi.aqara.remote_h1.RemoteH1SingleRocker'
'zhaquirks.xiaomi.aqara.remote_h1.RemoteH1DoubleRocker2'
'zhaquirks.xiaomi.aqara.remote_h1.RemoteH1DoubleRocker1'
'zhaquirks.xiaomi.aqara.remote_e1.RemoteE1SingleRocker1'
'zhaquirks.xiaomi.aqara.remote_e1.RemoteE1DoubleRocker1'
'zhaquirks.xiaomi.aqara.roller_curtain_e1.RollerE1AQ_3'
'zhaquirks.xiaomi.aqara.roller_curtain_e1.RollerE1AQ_2'
'zhaquirks.xiaomi.aqara.roller_curtain_e1.RollerE1AQ'
'zhaquirks.xiaomi.aqara.tvoc.TVOCMonitor2'
'zhaquirks.xiaomi.aqara.tvoc.TVOCMonitor'
```

I didn't check, but I hope at least the "Plug" devices do not send a mV readout in their attribute report.

But with the changes, the `XiaomiCluster` (most of the time, a custom `Basic` cluster) now tries to call `battery_reported` with the mV read-out on the `PowerConfiguration` cluster. If it's not a `XiaomiPowerConfiguration` cluster, that will fail.

For now, the missing method is ignored and the behavior is basically as without these changes.
IMO, the quirks above should be looked at though (if they should use `XiaomiPowerConfiguration`) (and if the mV read-out is even provided for them).

Otherwise, the calculation logic could theoretically be moved into the `XiaomiCluster` and then just call `update_attribute` (which would also work for non- `LocalDataCluster`s, but that's more of a hack then).

EDIT: See next comment for answer to this.
TL;DR is that the same behavior (ignoring those reports) still happens now (when `XiaomiPowerConfiguration` isn't used), but we a log a message now.

TheJulianJES:
Regarding my last comment:
Now, the battery voltage readouts are ignored IF the `XiaomiPowerConfiguration` is not used (same as previous behavior).

However, a debug message is now logged:
``Xiaomi battery voltage attribute received but XiaomiPowerConfiguration not used``

dmulcahey (approved):

PR #2535, opened 2023-08-18
Add Tuya cover TS0601 `_TZE200_dng9fn0k` variant

mandishshah:
## Proposed change

- One of the blinds that I own wasn't working with ZHA. Upon looking further, this quirk seems to make it work locally. This quirk seems to be the same that some of my other blinds are using.

## Checklist


TheJulianJES (approved):
LGTM. Thanks!

PR #2540, opened 2023-08-23
Add ThirdReality multifunction night light

llhappier:
## Checklist

### DIFF
--- a/zhaquirks/thirdreality/night_light.py
+++ b/zhaquirks/thirdreality/night_light.py
@@ -28,6 +28,7 @@ from zhaquirks.const import (
 from zhaquirks.thirdreality import THIRD_REALITY
 
 ThirdReality_Specific_CLUSTER_ID = 0xFC00
+THIRD_REALITY_MOTION_EVENT_ATTR_ID = 0x0002
 
 
 class ThirdRealitySpecificCluster(CustomCluster):
@@ -37,9 +38,10 @@ class ThirdRealitySpecificCluster(CustomCluster):
 
     def _update_attribute(self, attrid, value):
         super()._update_attribute(attrid, value)
-        self.endpoint.ias_zone.update_attribute(
-            IasZone.AttributeDefs.zone_status.id, value
-        )
+        if attrid == THIRD_REALITY_MOTION_EVENT_ATTR_ID:
+            self.endpoint.ias_zone.update_attribute(
+                IasZone.AttributeDefs.zone_status.id, value
+            )
 
 
 class LocalIasZone(LocalDataCluster, IasZone):
--- a/zhaquirks/thirdreality/night_light.py
+++ b/zhaquirks/thirdreality/night_light.py
@@ -27,14 +27,14 @@ from zhaquirks.const import (
 )
 from zhaquirks.thirdreality import THIRD_REALITY
 
-ThirdReality_Specific_CLUSTER_ID = 0xFC00
+THIRD_REALITY_CLUSTER_ID = 0xFC00
 THIRD_REALITY_MOTION_EVENT_ATTR_ID = 0x0002
 
 
 class ThirdRealitySpecificCluster(CustomCluster):
     """Manufacturer specific cluster to relay motion event to IAS Zone cluster."""
 
-    cluster_id = ThirdReality_Specific_CLUSTER_ID
+    cluster_id = THIRD_REALITY_CLUSTER_ID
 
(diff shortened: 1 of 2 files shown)


llhappier:
Hi, 
Because I am not very familiar with submission rules, there were too many commits for pull requests earlier. 
So I think we should re pull request.
Thanks.

TheJulianJES:
The amount of commits in a Pull Request do not matter at all. They will get squashed (so automatically merged into one) when we're merging a PR for this project.

Closing and opening new PRs is messy though.

For reference, old PR with feedback:
- https://github.com/zigpy/zha-device-handlers/pull/2509

Part of the suggested changes in that PR (when looking at the full dif) still isn't implemented.
I'll add it again here soon.

zhaquirks/thirdreality/night_light.py
```diff
@@ -0,0 +1,104 @@
...
+        self.endpoint.ias_zone.update_attribute(
+            IasZone.AttributeDefs.zone_status.id, value
+        )
+
+
+class LocalIasZone(LocalDataCluster, IasZone):
+    """Local IAS Zone cluster."""
+
+    _CONSTANT_ATTRIBUTES = {
+        IasZone.attributes_by_name["zone_type"].id: IasZone.ZoneType.Motion_Sensor
```
TheJulianJES: You can now also use `AttributeDefs` here (that's the newer way, instead of using `attributes_by_name`).
```suggestion
        IasZone.AttributeDefs.zone_type.id: IasZone.ZoneType.Motion_Sensor
```
EDIT: Merged.

zhaquirks/thirdreality/night_light.py
```diff
@@ -0,0 +1,104 @@
...
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+from zhaquirks.thirdreality import THIRD_REALITY
+
+ThirdReality_Specific_CLUSTER_ID = 0xFC00
+
+
+class ThirdRealitySpecificCluster(CustomCluster):
+    """privatecluster, only used to relay ias zone information to ias zone Cluster."""
```
TheJulianJES: Would be nice to improve the comment a bit here.
Something like this maybe:
```suggestion
    """Manufacturer specific cluster to relay motion event to IAS Zone cluster."""
```
EDIT: Merged.

zhaquirks/thirdreality/night_light.py
```diff
@@ -0,0 +1,104 @@
...
+
+    cluster_id = ThirdReality_Specific_CLUSTER_ID
+
+    def _update_attribute(self, attrid, value):
+        super()._update_attribute(attrid, value)
+        self.endpoint.ias_zone.update_attribute(
+            IasZone.AttributeDefs.zone_status.id, value
+        )
```
TheJulianJES: Please check that `attrid` matches `0x0002` before forwarding to the `ias_zone` cluster.
Otherwise, this *will* break when another attribute is used in the future. Not all HA users might update their HA version, so if they have updated devices/firmware, but outdated HA, this will break.

So please add a check that's something like this:
```suggestion
        if attrid == 0x0002:
            self.endpoint.ias_zone.update_attribute(
                IasZone.AttributeDefs.zone_status.id, value
            )
```

EDIT: Merged.

Relevant comments:
- https://github.com/zigpy/zha-device-handlers/pull/2509#discussion_r1287281874
- https://github.com/zigpy/zha-device-handlers/pull/2509#issuecomment-1676580102

TheJulianJES (dismissed):
Copied my feedback from the other PR again with the small things that need to be addressed before merging.

And please, leave this PR open. Do not open another PR.
Just add new commits to your  forked `dev` branch (don't squash, don't force-push, ...).

TheJulianJES:
I've made the requested changes now.

---

The main change is the `attrid` check. You mentioned (in the previous PR):
> Re: Yes, only the attribute id 0x0002 is used on the private/manufacturer specific cluster rightnow.

So that [added check](https://github.com/zigpy/zha-device-handlers/pull/2540/commits/67a6839fd31897f38fae7714dcc0c37378403c25) should work. This check is to guarantee that older HA versions won't break if newer firmware/device versions use more attribute on that cluster.

---

I've also added a basic test to see if the forwarding works as expected.

I'll merge this for the next HA 2023.9.0 release/beta (relasing today/tomorrow). Should there be any issues, please open a new issue/PR and tag me, so I see it as soon as possible.
Thanks for the PR!

PR #2548, opened 2023-08-28
Apply humidity correction to Sonoff SNZB-02 sensor

wtom:
## Proposed change
The Sonoff SNZB-02 sensor reports too high humidity values. I compared to the newer SNZB-02D and Aqara Temperature/Humidity sensor. Those two are very close to each other in humidity values. The SNZB-02 is about 6% off in my messured environments 

Here is a screenshot before and after applying the quirk:
Bildschirmfoto 2023-08-28 um 14 41 09

## Checklist

### DIFF
--- a/zhaquirks/sonoff/snzb-02.py
+++ b/zhaquirks/sonoff/snzb-02.py
@@ -1,0 +1,66 @@
+"""Sonoff SNZB-02 temperature/humidity sensor."""
+from zigpy.profiles import zha
+from zigpy.profiles.zha import DeviceType
+from zigpy.quirks import CustomCluster, CustomDevice
+from zigpy.zcl.clusters.general import Basic, Identify, PowerConfiguration
+from zigpy.zcl.clusters.measurement import RelativeHumidity, TemperatureMeasurement
+
+from zhaquirks.const import (
+    DEVICE_TYPE,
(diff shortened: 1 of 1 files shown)


MattWestb:
I like the approach that all sensors shall have possibility to calibrating in ZHA before the data is saved in HA so its correct done and not being junk in the database.

I have some Aqara `lumi.weather` and its 2 different batches using the quirk `zhaquirks.xiaomi.aqara.weather.Weather` or `zhaquirks.xiaomi.aqara.weather.Weather2` and in the same batch is all around the same but between the batches is looks very bad also with the pressure readings.

I have done one future request for implanting the sensor function / device but its closed of the slate boot.

TheJulianJES:
Hmm, I don't know if we should take this approach and do it as a quirk, as we don't know that all Sonoff sensors are offset the same way.

In Home Assistant 2023.9.0, you'll be able to create template sensors using the UI, so that should at least improve the situation a bit. (I do hope we'll eventually get a simple "offset" for changing existing sensor entities.)

darkxst:
Just to add a datapoint, I've not noticed this 6% offest here. I have beside me an SNZB-02, SNZB-02D and SNZB-02P which are reading 63.6, 66.3 and 62.3 respectively.

PR #2554, opened 2023-08-30
Fix Tuya fingerbot TS0001 `OnOff` cluster not working

brettjurgens:
Follow-up to #2506 which fixes the On/Off switch not working


zhaquirks/tuya/ts0001_fingerbot.py
```diff
@@ -34,7 +34,7 @@ class FingerBotReverse(t.enum8):
     UP_OFF = 0x01
 
 
-class TuyaFingerbotCluster(TuyaMCUCluster):
+class TuyaFingerbotCluster(TuyaEnchantableCluster, TuyaMCUCluster):
```
brettjurgens: I tried without this and switching the device to use CustomDevice instead of Enchanted, but that did not work. This was the only way to get it to work

brettjurgens:
@TheJulianJES I'm not sure why these tests are failing, but when I tried without using the Enchantable stuff nothing worked

TheJulianJES:
Some Tuya devices need to get attribute read requests for certain attributes before they'll completely unlock. Sending those attribute reads is often referred to as "casting the Tuya spell".
An `EnchantedDevice` is a device that needs to be unlocked with that spell.

Due to the current architecture, there's no real way for a quirk to hook into the pairing process of a device.
However, ZHA binds many default clusters (like `OnOff`) (and some manufacturer ones). We hook into the `bind()` method and "cast the spell" there.

The tests emulate ZHA calling `bind()` for the standard clusters that are bound. They don't do this for manufacturer specific clusters. This is probably why the tests fail, but it still works for you. (As you hooked into the `bind()` method of a Tuya manufacturer specific cluster by making it an "enchantable" cluster.)
(An "enchantable cluster" will cast the Tuya spell in its `bind()` method if that's called and the device is an `EnchantedDevice`. There's more comments/docs in the [src](https://github.com/zigpy/zha-device-handlers/blob/efa4828a018f68dabd3348969728b289831672d8/zhaquirks/tuya/__init__.py#L530-L548)/tests regarding this behavior.)

It would be best if we could make the `OnOff` cluster "enchantable" instead of the Tuya one to fix this for now.

We could probably just create a new class and use that instead of `OnOff`:
```python
class OnOffEnchantable(TuyaEnchantableCluster, OnOff):
    """Enchantable OnOff cluster."""
```

--- 

However, if the custom `TuyaOnOff` cluster from `tuya.mcu` doesn't work, that makes me wonder if it's even a special Tuya datapoint?

So can you try to **remove** the following:
- `1: DPToAttributeMapping(OnOff.ep_attribute, "on_off"),` (from `dp_to_attribute`)
- `1: "_dp_2_attr_update",` (from `data_point_handlers`)

Then see if on/off works (with the normal `OnOff` cluster (and test failures) or my `OnOffEnchantable` example above).
If that works, then `OnOff` is indeed not a Tuya datapoint, but just uses the normal Zigbee-spec `OnOff` cluster.

Also, we have another `TuyaOnOff` cluster that's not from `tuya.mcu`, but just from `tuya`. It overrides the `command()` method, but you could try to use that. (This really needs to be cleaned up a bit)
If that doesn't work, use the `OnOffEnchantable` example from above.

brettjurgens:
@TheJulianJES Manufacturer clusters seem to actually be valid for the test, but the cluster range is incorrect see [here](https://github.com/zigpy/zigpy/blob/dev/zigpy/zcl/clusters/manufacturer_specific.py#L9) - the cluster is actually `0xEF00` (not in the defined range of `0xFC00`-`0xFFFF`)

This definition is used [here](https://github.com/zigpy/zigpy/blob/dev/zigpy/zcl/clusters/__init__.py#L30) and in turn in the test [here](https://github.com/zigpy/zha-device-handlers/blob/dev/tests/test_tuya.py#L1658-L1659)

I can change the OnOff cluster to be Enchantable, but I think the cluster definition should also be fixed.

I can also update the test to include this cluster (i.e. clone CLUSTERS_BY_ID and add the Tuya Manufacturer to the clone)

What do you suggest I do?

TheJulianJES:
Yeah, that range is given by the "Zigbee Cluster Library Specification" (ZCL). Unfortunately, Tuya does not follow them (hence why they're out of the range). We can't/shouldn't update that range though.

I'll need to have a look at the test later though. It's (originally) supposed to ignore any manufacturer specific clusters (whether inside the "manufacturer specific range" or not), as it's somewhat complicated on which clusters ZHA calls `bind()` on. (They need to be defined inside ZHA)

Since no other quirks hook into the `bind()` method of a Tuya manufacturer specific cluster (and `bind()` for `OnOff` is called earlier), it's probably best to make the `OnOff` cluster "enchantable".

The `TuyaOnOff` (from `tuya` and `tuya.mcu`) are already enchantable clusters. You can try the non-mcu one from `tuya` to see if that works. (Currently, the `mcu` one is used and that doesn't work as you mentioned.)
Otherwise, the `OnOffEnchantable ` example from above should do the trick.

(again, the `TuyaOnOff` clusters should really be renamed -- hopefully I can clean that up at some point)

TheJulianJES:
Oh, and from my [first comment](https://github.com/zigpy/zha-device-handlers/pull/2554#issuecomment-1698420377), see if you can remove the datapoint (handler) for `1`/`OnOff` if you haven't done so already:

> So can you try to **remove** the following:
> - `1: DPToAttributeMapping(OnOff.ep_attribute, "on_off"),` (from `dp_to_attribute`)
> - `1: "_dp_2_attr_update",` (from `data_point_handlers`)

brettjurgens:
Thanks, will push an update in the morning (pacific time) after I can test it out

MattWestb:
I was testing how the TS004F commands is being debug logged and with `zigpy.zcl` looks being OK logged (its tuya Zigbee commands on  the OnOff cluster):
```
2023-08-30 07:57:38.900 DEBUG (MainThread) [zigpy.zcl] [0x7497:1:0x0006] Received ZCL frame: b'\x01F\x00'
2023-08-30 07:57:38.902 DEBUG (MainThread) [zigpy.zcl] [0x7497:1:0x0006] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl(frame_type=<FrameType.CLUSTER_COMMAND: 1>, is_manufacturer_specific=0, direction=<Direction.Server_to_Client: 0>, disable_default_response=0, reserved=0, *is_cluster=True, *is_general=False), tsn=70, command_id=0, *direction=<Direction.Server_to_Client: 0>)
2023-08-30 07:57:38.906 DEBUG (MainThread) [zigpy.zcl] [0x7497:1:0x0006] Decoded ZCL frame: TuyaSmartRemoteOnOffCluster:off()
2023-08-30 07:57:38.908 DEBUG (MainThread) [zigpy.zcl] [0x7497:1:0x0006] Received command 0x00 (TSN 70): off()
2023-08-30 07:57:38.909 DEBUG (MainThread) [zigpy.zcl] [0x7497:1:0x0006] TS004X: send default response
2023-08-30 07:57:38.915 DEBUG (MainThread) [zigpy.zcl] [0x7497:1:0x0006] Sending reply header: ZCLHeader(frame_control=FrameControl(frame_type=<FrameType.GLOBAL_COMMAND: 0>, is_manufacturer_specific=False, direction=<Direction.Client_to_Server: 1>, disable_default_response=1, reserved=0, *is_cluster=False, *is_general=True), tsn=70, command_id=<GeneralCommand.Default_Response: 11>, *direction=<Direction.Client_to_Server: 1>)
2023-08-30 07:57:38.917 DEBUG (MainThread) [zigpy.zcl] [0x7497:1:0x0006] Sending reply: Default_Response(command_id=0, status=<Status.SUCCESS: 0>)
```
So can easy see if its using this commands or normal ZCL OnOff commands.
And my tuya TRV is sending MCU commands and looks like this:
```
2023-08-30 08:03:03.735 DEBUG (MainThread) [zigpy.application] Received a packet: ZigbeePacket(src=AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x8C10), src_ep=1, dst=AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x0000), dst_ep=1, source_route=None, extended_timeout=False, tsn=158, profile_id=260, cluster_id=61184, data=Serialized[b'\t\x05\x01\x0f\xfb\x02\x02\x00\x04\x00\x00\x002'], tx_options=<TransmitOptions.NONE: 0>, radius=0, non_member_radius=0, lqi=198, rssi=-51)
2023-08-30 08:03:03.736 DEBUG (MainThread) [zigpy.zcl] [0x8C10:1:0xef00] Received ZCL frame: b'\t\x05\x01\x0f\xfb\x02\x02\x00\x04\x00\x00\x002'
2023-08-30 08:03:03.738 DEBUG (MainThread) [zigpy.zcl] [0x8C10:1:0xef00] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl(frame_type=<FrameType.CLUSTER_COMMAND: 1>, is_manufacturer_specific=0, direction=<Direction.Client_to_Server: 1>, disable_default_response=0, reserved=0, *is_cluster=True, *is_general=False), tsn=5, command_id=1, *direction=<Direction.Client_to_Server: 1>)
2023-08-30 08:03:03.740 DEBUG (MainThread) [zigpy.zcl] [0x8C10:1:0xef00] Decoded ZCL frame: SiterwellManufCluster:get_data(param=Command(status=15, tsn=251, command_id=514, function=0, data=[4, 0, 0, 0, 50]))
2023-08-30 08:03:03.745 DEBUG (MainThread) [zigpy.zcl] [0x8C10:1:0xef00] Received command 0x01 (TSN 5): get_data(param=Command(status=15, tsn=251, command_id=514, function=0, data=[4, 0, 0, 0, 50]))
2023-08-30 08:03:03.746 DEBUG (MainThread) [zhaquirks.tuya] [0x8c10:1:0xef00] Received value [0, 0, 0, 50] for attribute 0x0202 (command 0x0001)
```
So i think its good adding `zhaquirks.tuya` to debug logging then you see both commands in the log.

And IKEA Rodret is sending normal ZCL OnOff looks like this:
```
2023-08-30 08:08:06.889 DEBUG (MainThread) [zigpy.application] Received a packet: ZigbeePacket(src=AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x86AC), src_ep=1, dst=AddrModeAddress(addr_mode=<AddrMode.Group: 1>, address=0x0000), dst_ep=255, source_route=None, extended_timeout=False, tsn=170, profile_id=260, cluster_id=6, data=Serialized[b'\x11\x19\x00'], tx_options=<TransmitOptions.NONE: 0>, radius=0, non_member_radius=0, lqi=208, rssi=-52)
2023-08-30 08:08:06.890 DEBUG (MainThread) [zigpy.zcl] [0x86AC:1:0x0006] Received ZCL frame: b'\x11\x19\x00'
2023-08-30 08:08:06.892 DEBUG (MainThread) [zigpy.zcl] [0x86AC:1:0x0006] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl(frame_type=<FrameType.CLUSTER_COMMAND: 1>, is_manufacturer_specific=0, direction=<Direction.Server_to_Client: 0>, disable_default_response=1, reserved=0, *is_cluster=True, *is_general=False), tsn=25, command_id=0, *direction=<Direction.Server_to_Client: 0>)
2023-08-30 08:08:06.894 DEBUG (MainThread) [zigpy.zcl] [0x86AC:1:0x0006] Decoded ZCL frame: OnOff:off()
2023-08-30 08:08:06.895 DEBUG (MainThread) [zigpy.zcl] [0x86AC:1:0x0006] Received command 0x00 (TSN 25): off()
2023-08-30 08:08:06.896 DEBUG (MainThread) [zigpy.zcl] [0x86AC:1:0x0006] No explicit handler for cluster command 0x00: off()
```
So with `zigpy.zcl` and `zhaquirks.tuya` you shall see what method tuya is using for fornding the OnOff commands without problems.

For the tuya mess i think the best is renaming all "normal tuya Zigbee" commands / classes and so on to somthing `tuyaZB` and the MCU to `tuyaMCU` and do it all thru all tuya quirks but its not one small work if not braking somthings.

brettjurgens:
`TuyaOnOff` (non-mcu) didn't work (`AttributeError: 'TuyaFingerbot' object has no attribute 'switch_bus'`), but using the new `OnOffEnchantable` did!

Removing the datapoints also seemed to work :)

TheJulianJES (approved):
Changes LGTM. Thanks for testing!

TheJulianJES:
Can't guarantee this PR will make it into Home Assistant Core 2023.9.x, as it's a "minor issue".
`zha-quirks` 0.0.103 release with the initial PR will be in 2023.9.0b1 (missed b0).

If there are some more PRs for a new version, a new release might be made for a beta or later hotfix version.
This fix should definitely be in 2023.10.0 though.

TheJulianJES:
@brettjurgens Also, although unrelated to this PR, I just noticed that battery might not be working because `TuyaPowerConfigurationCluster` is missing in the `INPUT_CLUSTERS` for the `replacement`.
(It's a fake/local data cluster that should be populated with `battery_percentage_remaining` from the Tuya datapoint.)

If you have some time, feel free to modify like mentioned above and PR that fix too if it works :)

EDIT: Created:
- https://github.com/zigpy/zha-device-handlers/pull/2578

PR #2567, opened 2023-09-04
Add new model for TS0001 Fingerbot: `_TZ3210_j4pdtz9v`

abmantis:
## Proposed change

I have a model that is different that the one that was already supported.

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

PR #2578, opened 2023-09-08
Add Tuya battery cluster for TS0001 fingerbot

TheJulianJES:
## Proposed change

This adds the missing `LocalDataCluster` that accepts the battery reports from the Tuya datapoints.
`TuyaPowerConfigurationCluster` also doubles the battery reports, as ZCL/ZHA expect 200% to actually be 100%.

@abmantis You seem to have this device, is it possible for you to test if this change works? (and the battery entity shows up in HA + gets populated after some time?)

## Additional information

Related Tuya fingerbot PRs:
- https://github.com/zigpy/zha-device-handlers/pull/2506
- https://github.com/zigpy/zha-device-handlers/pull/2554
- (https://github.com/zigpy/zha-device-handlers/pull/2567)

(Python 3.12.x test failure is expected, as zigpy broke with Python 3.12 rc2)

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0001_fingerbot.py
+++ b/zhaquirks/tuya/ts0001_fingerbot.py
@@ -143,6 +143,7 @@ class TuyaFingerbot(EnchantedDevice):
                 DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
+                    TuyaPowerConfigurationCluster,
                     OnOffEnchantable,
                     TuyaFingerbotCluster,
                 ],


abmantis:
image
Yep! I don't know if the value is correct, since it is a new device so it may be at 100% still.
The value was available just a few seconds after pairing.

TheJulianJES:
Awesome, thanks for testing!
The value is from Tuya datapoint `105` (same as Z2M uses), so I assume it's at least what the device reports. 

---

Merging as a follow-up small bugfix to the previous PRs.

Python 3.12 test failure is expected, as zigpy isn't updated for Python 3.12 rc2 yet. CI on `dev` is already failing for 3.12 action (before this PR). Should be possible to re-run failing test/step on `dev` later when zigpy gets updated.

All other tests are passing/test coverage basically unchanged.

PR #2609, opened 2023-09-28
Add IKEA Rodret dimmer support

inventorofthingies:
## Proposed change
Added signature to match Ikea's Rodret dimmer button.

## Additional information
Rodret is basically same as Tradfri without alarm input cluster and without window covering output cluster. Tested as a standalone quirk

## Checklist

### DIFF
--- a/zhaquirks/ikea/twobtnremote.py
+++ b/zhaquirks/ikea/twobtnremote.py
@@ -202,8 +202,8 @@ class IkeaRodretRemote2Btn(CustomDevice):
     signature = {
         # <SimpleDescriptor endpoint=1 profile=260 device_type=2080
         # device_version=1
-        # input_clusters=[0, 1, 3, 9, 32, 4096, 64636]
-        # output_clusters=[3, 4, 6, 8, 25, 258, 4096]>
+        # input_clusters=[0, 1, 3, 32, 4096, 64636]
+        # output_clusters=[3, 4, 6, 8, 258, 4096]>
         MODELS_INFO: [(IKEA, "RODRET Dimmer")],
         ENDPOINTS: {
             1: {
--- a/zhaquirks/ikea/twobtnremote.py
+++ b/zhaquirks/ikea/twobtnremote.py
@@ -85,7 +85,7 @@ class IkeaTradfriRemote2Btn(CustomDevice):
                 DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
-                    DoublingPowerConfig1CRCluster,
+                    PowerConfiguration.cluster_id,
                     Identify.cluster_id,
                     Alarms.cluster_id,
                     PollControl.cluster_id,
--- a/zhaquirks/ikea/twobtnremote.py
+++ b/zhaquirks/ikea/twobtnremote.py
@@ -196,6 +196,7 @@ class IkeaTradfriRemote2BtnZLL(CustomDevice):
 
     device_automation_triggers = IkeaTradfriRemote2Btn.device_automation_triggers.copy()
 
+
 class IkeaRodretRemote2Btn(CustomDevice):
     """Custom device representing IKEA of Sweden RODRET remote control."""
 
@@ -236,7 +237,7 @@ class IkeaRodretRemote2Btn(CustomDevice):
                 DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
-                    DoublingPowerConfig1CRCluster,
+                    PowerConfiguration.cluster_id,
                     Identify.cluster_id,
                     PollControl.cluster_id,
                     LightLink.cluster_id,
--- a/zhaquirks/ikea/twobtnremote.py
+++ b/zhaquirks/ikea/twobtnremote.py
@@ -195,3 +195,63 @@ class IkeaTradfriRemote2BtnZLL(CustomDevice):
     }
 
     device_automation_triggers = IkeaTradfriRemote2Btn.device_automation_triggers.copy()
+
+class IkeaRodretRemote2Btn(CustomDevice):
+    """Custom device representing IKEA of Sweden RODRET remote control."""
+
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=2080
+        # device_version=1
+        # input_clusters=[0, 1, 3, 9, 32, 4096, 64636]
+        # output_clusters=[3, 4, 6, 8, 25, 258, 4096]>
+        MODELS_INFO: [(IKEA, "RODRET Dimmer")],
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
+                    Identify.cluster_id,
+                    PollControl.cluster_id,
+                    LightLink.cluster_id,
+                    IKEA_CLUSTER_ID,
+                ],
+                OUTPUT_CLUSTERS: [
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    OnOff.cluster_id,
+                    LevelControl.cluster_id,
+                    Ota.cluster_id,
+                    LightLink.cluster_id,
+                ],
+            }
+        },
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    DoublingPowerConfig1CRCluster,
+                    Identify.cluster_id,
+                    PollControl.cluster_id,
+                    LightLink.cluster_id,
+                    IKEA_CLUSTER_ID,
+                ],
+                OUTPUT_CLUSTERS: [
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    OnOff.cluster_id,
+                    LevelControl.cluster_id,
+                    Ota.cluster_id,
+                    LightLink.cluster_id,
+                ],
+            }
+        }
+    }
+
+    device_automation_triggers = IkeaTradfriRemote2Btn.device_automation_triggers.copy()
--- a/zhaquirks/ikea/twobtnremote.py
+++ b/zhaquirks/ikea/twobtnremote.py
@@ -40,7 +40,12 @@ from zhaquirks.const import (
     TURN_OFF,
     TURN_ON,
 )
-from zhaquirks.ikea import IKEA, IKEA_CLUSTER_ID, DoublingPowerConfig1CRCluster
+from zhaquirks.ikea import (
+    IKEA,
+    IKEA_CLUSTER_ID,
+    DoublingPowerConfig1CRCluster,
+    PowerConfig1AAACluster,
+)
 
 
 class IkeaTradfriRemote2Btn(CustomDevice):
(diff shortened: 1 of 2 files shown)


zhaquirks/ikea/twobtnremote.py
```diff
@@ -195,3 +195,63 @@ class IkeaTradfriRemote2BtnZLL(CustomDevice):
...
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    DoublingPowerConfig1CRCluster,
```
MattWestb: Gen 3 devices or earlier with gen 3 firmware (2X.X) is not needing dabbling the battery % then IKEA have fixing the "defect" in there implantation (Styrbar was the first but its not reporting OK its reporting 255 = not valid reading).
All my Rodret is reporting battery OK like this on the device card:
```
Diagnostic
Battery 100%
```

zhaquirks/ikea/twobtnremote.py
```diff
@@ -195,3 +195,63 @@ class IkeaTradfriRemote2BtnZLL(CustomDevice):
     }
 
     device_automation_triggers = IkeaTradfriRemote2Btn.device_automation_triggers.copy()
+
+class IkeaRodretRemote2Btn(CustomDevice):
+    """Custom device representing IKEA of Sweden RODRET remote control."""
```
MattWestb, line 206: Better description of the device then its 2 more upcoming Shortcut and OpenClose and the first is CSA certed for some weeks ago.

MattWestb (changes requested):

MattWestb:
You need putting one more empty line between the old last class and the new one (shall being 2 lines is only one) (Line 198) then the test shall working OK.

zhaquirks/ikea/twobtnremote.py
```diff
@@ -195,3 +195,63 @@ class IkeaTradfriRemote2BtnZLL(CustomDevice):
     }
 
     device_automation_triggers = IkeaTradfriRemote2Btn.device_automation_triggers.copy()
+
```
MattWestb, line 203: One more empty line for the test here so 2 in all between classes.

zhaquirks/ikea/twobtnremote.py
MattWestb: Pleas marking it fixed after you have testing its working OK in your system (my 2 in production is reporting OK without quirk).

inventorofthingies: Yes, I've just been checking all mine -- no need for doubling.

zhaquirks/ikea/twobtnremote.py
MattWestb: SOMRIG shortcut button Aka. E2213

https://csa-iot.org/csa_product/somrig-shortcut-button/

OpenCloce is coming then IKEA have using all produces hardware they have in storage at the factory's (and they need do new packages for all blinds).

inventorofthingies: I've not seen a SOMRIG yet -- happy to have a go at adding when I get one. Let me know how to handle different firmware variants and I'll add that later too.

MattWestb: I have not seen it in one store but its very like weeks and not months before its start pooing up little here and there.
Its very likely its using Matter commands on one private cluster then its not fixed in Zigbee (next ZCL is command with Matter) like they have on Symfonisk Gen 2 scene scene buttons but we will see then we is getting them.

### By the way good work done !!!

zhaquirks/ikea/twobtnremote.py
```diff
@@ -85,7 +85,7 @@ class IkeaTradfriRemote2Btn(CustomDevice):
                 DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
-                    DoublingPowerConfig1CRCluster,
+                    PowerConfiguration.cluster_id,
```
javicalle: Don't seems related to your quirk. It is needed?

javicalle (changes requested):

zhaquirks/ikea/twobtnremote.py
TheJulianJES: Yeah, let's not change the old two button dimmer in this PR.

The problem is that newer IKEA firmware versions didn't change the device signature for the (old) two button dimmer.
We shouldn't double battery percentage on those newer firmware versions, but still on old firmware. But we can't match by firmware version at the moment.

The signature also isn't "refreshed" when just updating the firmware through ZHA. I think it needs to be removed from ZHA and completely re-added. (I don't think re-pairing is enough, but I'm not sure right now)

---

Without matching per firmware version, maybe it's possible to find out what voltage 0% and 100% respond to on IKEA remotes. @MattWestb Any idea? 😄 
At least for the two button dimmer, we could use a linear function to calculate the battery percentage based on the battery voltage (like we do for some other devices). That way it would work regardless of the firmware version -- assuming IKEA sends a proper battery voltage.

Or maybe we just do it the Z2M way and remove doubling everywhere -- regardless of firmware version.

zhaquirks/ikea/twobtnremote.py
```diff
@@ -195,3 +195,64 @@ class IkeaTradfriRemote2BtnZLL(CustomDevice):
...
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    PowerConfiguration.cluster_id,
```
TheJulianJES: This isn't important, but using a cluster like this (important: non-doubling), it would add information about battery type/quantity/voltage:
https://github.com/zigpy/zha-device-handlers/blob/7385465bd6f04f295893fb20d053433011183935/zhaquirks/ikea/__init__.py#L146-L154
At least the battery type should be displayed as an additional attribute when clicking on the battery entity in HA then.

(The cluster above would need to be "duplicated" and quantity changed to 1, as the device only seems to take one AAA battery.)

zhaquirks/ikea/twobtnremote.py
TheJulianJES: > Better description of the device then its 2 more upcoming Shortcut and OpenClose and the first is CSA certed for some weeks ago.

Is this a request about something that should be changed? IMO, class name and comment are fine.

zhaquirks/ikea/twobtnremote.py
MattWestb: My "beloved" d(f)conz have just removing it for all IKEA controllers then they is expecting all user updating the firmware and its one way to going.
But i think its better doing one separate PR for that and doing all but not the old rotating dimmer then its still on 2.3.X firmware and the classes for older ZLL and also ZB3 versions. 

If i remember right is most reporting OK battery voltage but first gen Symfonisk is no (i think 25V is full) and perhaps some more but its not easy testing that all my controllers is on latest firmware without j-tag flashing all types them with one older firmware and testing.

Z2M is doing matching on SW_Build for gen 3 controller and i think on the old ones 2.

ZHA is having the SW_build and version saved in the Zigbe.DB so one way is implanting quirk to using it and its all good but is it needed for most users ?

TheJulianJES: Reverted that unrelated change in https://github.com/zigpy/zha-device-handlers/pull/2609/commits/cbfb4d50a78da0fe04f5c99580be6c2e7eed0f22

We still need to address this, but preferably without breaking outdated or updated dimmers.
Since the battery voltage should be reported the same (regardless of if it's updated or not), we should be able to map voltage to percentage manually and just ignore the percentage the device sends.
We really need to find out how the device maps battery voltage to percentage though.
Would be amazing if somebody can give us multiple data points on which voltage corresponds to which percentage.

zhaquirks/ikea/twobtnremote.py
TheJulianJES: Addressed in: https://github.com/zigpy/zha-device-handlers/pull/2609/commits/b736f27ad727311ce62aef4d920e1121d18faf79

TheJulianJES (approved):
I've pushed two commits to resolve the remaining issues with this and it should be good now. Thanks for the PR!

(This should make it into HA Core 2023.12.x.)

zhaquirks/ikea/twobtnremote.py
MattWestb: Then you need verifying that all firmware versions is sending correct voltage then i have see dome IKEA controllers sending no or wrong voltage (Symfonisk first gen was sending 25V and not 2.5V in some version).
I think better taking the double for the quirk class that also being loaded by 24.x firmware if have the same as the classes and letting all old have the doubling and user shall updating there devices for fixing battery darning and network loss then running battery low.

zhaquirks/ikea/twobtnremote.py
joshuapfl: > I've not seen a SOMRIG yet -- happy to have a go at adding when I get one. Let me know how to handle different firmware variants and I'll add that later too.

Hey. I have a somrig Remote and its not recognizing any button press in the log, will your quirk help? What do i need to change in order to make it work for somrig? 
I am a very beginner, so keeop that in mind. Thank you soo much..

MattWestb: I s´was getting little parts today https://github.com/zigpy/zigpy/discussions/660#discussioncomment-7775801 and all is working without quirks but Sommerig (Suller like in Eng) is using matter ZCL  commands on one manufacture cluster like Symfonisk 2 is doing so need little work foe getting it working but im sure its coming very son.

joshuapfl: Thank MattWestb
I dindt quite understand everything you said, since i am sort of a newbie :) But you made me hopefull again haha.
So the SOMRig is not using the same hardware protocls that Rodret i assume. Even togh they seem identical..

MattWestb: Hardware is identical only the printing on it is different and the firmware is different and sending commands on one different cluster that is not standard and need one quirk for fixing it,

joshuapfl: Ok. Thank you a lot. looking forward to your quirk :)

joshuapfl: Any news? I would also donate you something. 
Cheers

MattWestb: Im on one 2  weeks trip and have only laptop with and no connection to my test networks so i cant do any testings but if some one like coding and testing i can helping with little knowledge that i was taking with ;-)

leehadassin:
Hi, I noticed the turn on, turn off and dim up automation triggers work but the dim down device automation trigger doesn't work for my Rodret - looking at the zha_events the release of the down dim send the same _command:stop_with_on_off_ as the release of the up i.e.

## up button 
### press and hold
command: move_with_on_off
### release
command: stop_with_on_off

## down button 
### press and hold
command: move
### release
**_command: stop_with_on_off_** (NOT command: stop)

MattWestb:
The device is sending light commands and by Zigbee standard its only have one stop command for move light level,
The same is with the open / close button but some have implemented to remember if the last was up or down and is using it for tagging the release.
https://github.com/zigpy/zha-device-handlers/blob/d9481d36a8483810696fe1c0e796895b92d6479b/zhaquirks/ikea/opencloseremote.py#L66-L76
If you like to having it look on the quirk how its made and making on PR for getting it in ZHA of by one Sommrig short cut button then you is getting more HA functions you can using for your automatons bu no light commands for steering lights.

MattWestb:
By the way with the late firmware (`rodret_dimmer_soc-0x11cd-1.0.57-prod.ota.ota.signed`) the quirk is broke then IKEA have putting back the group cluster (in => its showing up in adding lights to gorps alla tuya).

Ping @TheJulianJES Shall we patching one more device class or upgrading the quirk to V2 for adding the latest version of the device ?

The signature for the last production firmware:
```
{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4476,
    "maximum_buffer_size": 82,
    "server_mask": 11264,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0820",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0004",
        "0x0020",
        "0x1000",
        "0xfc7c"
      ],
      "output_clusters": [
        "0x0003",
        "0x0004",
        "0x0006",
        "0x0008",
        "0x0019",
        "0x1000"
      ]
    }
  },
  "manufacturer": "IKEA of Sweden",
  "model": "RODRET Dimmer",
}
```

PR #2674, opened 2023-10-21
Add quirk IDs to Tuya plugs

TheJulianJES:
## Proposed change
This adds quirk IDs which are later used in another Home Assistant PR to let ZHA match against these `quirk_ids`.

## Additional information
- Reference for base ZHA quirks ID PR: https://github.com/home-assistant/core/pull/102482

This PR can already be merged independently.

- These quirk IDs will be used by: https://github.com/home-assistant/core/pull/102489

## Checklist
- [x] The changes are tested and work correctly (with linked PRs above)

### DIFF
--- a/zhaquirks/quirk_ids.py
+++ b/zhaquirks/quirk_ids.py
@@ -1,4 +1,4 @@
-"""Quirk IDs used for matching quirked devinces in ZHA."""
+"""Quirk IDs used for matching quirked devices in ZHA."""
 
 # Tuya
 TUYA_PLUG_ONOFF = "tuya.plug_on_off_attributes"
--- a/zhaquirks/quirk_ids.py
+++ b/zhaquirks/quirk_ids.py
@@ -1,5 +1,5 @@
 """Quirk IDs used for matching quirked devices in ZHA."""
 
 # Tuya
-TUYA_PLUG_ONOFF = "tuya.plug_on_off_attributes"
-TUYA_PLUG_MANUFACTURER = "tuya.plug_manufacturer_attributes"
+TUYA_PLUG_ONOFF = "tuya.plug_on_off_attributes"  # plugs with configurable attributes on the OnOff cluster
+TUYA_PLUG_MANUFACTURER = "tuya.plug_manufacturer_attributes"  # plugs with configurable attributes on a custom cluster
--- a/zhaquirks/quirk_ids.py
+++ b/zhaquirks/quirk_ids.py
@@ -1,0 +1,5 @@
+"""Quirk IDs used for matching quirked devinces in ZHA."""
+
+# Tuya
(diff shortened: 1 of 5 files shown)


zhaquirks/quirk_ids.py
```diff
@@ -0,0 +1,5 @@
+"""Quirk IDs used for matching quirked devinces in ZHA."""
+
+# Tuya
+TUYA_PLUG_ONOFF = "tuya.plug_on_off_attributes"
```
elupus: Seems there really should be a description about what these quirk is all about. Impossible for a reviewer or contributer to know when this should be added.

TheJulianJES: Added a small comment about what the quirks that are assigned the specific quirk IDs do: https://github.com/zigpy/zha-device-handlers/pull/2674/commits/22e7036ff954aa7f8c9ce754373995bc60ce8b1d
Or did you mean something else?

(The "quirks ID" file just consists of constants (the quirk IDs) which are referenced by both the quirk and ZHA, so the entities match as expected. The quirk ID can technically be chosen completely "random".
I used a format which is something along the lines of: `manufacturer.roughly_what_the_quirk_does/provides`)

elupus: Those comments looks alright. I would prefer really python doc strings, but this is okey.

There is no situation you need two quirks active at the same time? You could have made this a flags enum for example.

TheJulianJES: > I would prefer really python doc strings, but this is okey.

So our existing constants file doesn't have any documentation: [zhaquirks/const.py](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/const.py), same with [homeassistant/components/zha/core/const.py](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/core/const.py).
They're just "constant strings" after all. This is the same with quirk IDs. The quirk IDs themselves "don't do anything".

I do understand why you'd maybe want to have a bit of documentation, but the actual implementation for what entities are created is in ZHA (where the quirk ID is used) and the quirk implementation is in zha-quirks where the quirk ID is assigned to a quirk (using these constants).

So I fear that if we add too many comments to the constant quirks ID file, (1) the file will become cluttered quickly and (2) -- more importantly -- might contain outdated information if the quirks or the ZHA code is changed, but the documentation for the quirks ID is not.
That could likely happen, because the quirk IDs likely don't need to be touched in those cases.

---

> There is no situation you need two quirks active at the same time? You could have made this a flags enum for example.

The quirk ID approach mostly replaces the existing "quirk class path" matching approach (which was suboptimal).
At the moment, I don't think that we need to assign multiple quirk IDs to a quirk. If we need to do that in the future, the code in quirks can be updated relatively easily. It'll also require an update in ZHA too though (more important part).

PR #2678, opened 2023-10-23
Add "power outage memory" support for Aqara LED Bulb T1 (ZNLDP13LM)

corvis:
## Proposed change
This quirk adds support for power on behavior configuration to [Aqara LED Bulb T1 aka Xiaomi ZNLDP13LM](https://zigbee.blakadder.com/Aqara_ZNLDP13LM.html).

## Additional information
None

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/light_acn.py
+++ b/zhaquirks/xiaomi/aqara/light_acn.py
@@ -1,15 +1,23 @@
 from zigpy import types as t
-from zigpy.profiles import zha
+from zigpy.profiles import zgp, zha
 from zigpy.zcl.clusters.general import (
+    Alarms,
+    AnalogInput,
+    BinaryInput,
+    DeviceTemperature,
+    GreenPowerProxy,
     Groups,
     Identify,
     LevelControl,
+    MultistateInput,
     OnOff,
     Ota,
     Scenes,
     Time,
 )
+from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
 from zigpy.zcl.clusters.lighting import Color
+from zigpy.zcl.clusters.smartenergy import Metering
 
 from zhaquirks.const import (
     DEVICE_TYPE,
@@ -19,7 +27,12 @@ from zhaquirks.const import (
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-from zhaquirks.xiaomi import BasicCluster, XiaomiAqaraE1Cluster, XiaomiCustomDevice
+from zhaquirks.xiaomi import (
+    LUMI,
+    BasicCluster,
+    XiaomiAqaraE1Cluster,
+    XiaomiCustomDevice,
+)
 
 
 class OppleClusterLight(XiaomiAqaraE1Cluster):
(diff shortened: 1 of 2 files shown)


TheJulianJES:
Interesting that the bulb has `ElectricalMeasurement` clusters and so on. Can you send a screenshot of what entities ZHA generates?
Otherwise, PR LGTM.

corvis:
Yep, here it is:
Entities:
image

Device Page:
image

TheJulianJES:
Are the "Active power", "Binary input", "Summation delivered" sensors working? Like, do they increase when the bulb is turned on?

corvis:
Hm it looks like no. Summation is still 0 despite I use the bulb for a week already, also checked active power history at I see no values. I didn't noticed this initially because I disabled all entities I don't need right after adding the device.

So shell we remove this clusters from replacements section?

TheJulianJES:
> So shell we remove this clusters from replacements section?

Yep. So probably `ElectricalMeasurement`, `Metering`, and possibly also `MultistateInput`, `BinaryInput`, `AnalogInput` too? I guess just remove the clusters that create non-working entities.

TheJulianJES:
Alright, I've added a commit that removes the three clusters that create useless entities.
The other clusters probably aren't needed too, but I've left them in for now, since they create no new entities.

zhaquirks/xiaomi/aqara/light_acn.py
```diff
@@ -84,3 +97,104 @@ class LumiLightAcn003(XiaomiCustomDevice):
...
+                OUTPUT_CLUSTERS: [],
+            },
+            31: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.COLOR_TEMPERATURE_LIGHT,
+                INPUT_CLUSTERS: [
+                    AnalogInput.cluster_id,  # 0x000C
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
```
TheJulianJES, line 156: As a note, these clusters are removed, since they're not in the replacement.
IIRC, Xiaomi normally uses those those endpoint numbers for power measurement plugs. Since this bulb isn't one though, these can likely be removed as well (which is already done).

zhaquirks/xiaomi/aqara/light_acn.py
```diff
@@ -84,3 +97,104 @@ class LumiLightAcn003(XiaomiCustomDevice):
...
+                    BasicCluster,  # 0x0000
+                    DeviceTemperature.cluster_id,  # 0x0002
+                    Identify.cluster_id,  # 0x0003
+                    Groups.cluster_id,  # 0x0004
+                    Scenes.cluster_id,  # 0x0005
+                    OnOff.cluster_id,  # 0x0006
+                    LevelControl.cluster_id,  # 0x0008
+                    Alarms.cluster_id,  # 0x0009
+                    AnalogInput.cluster_id,  # 0x000C
+                    MultistateInput.cluster_id,  # 0x0012
```
TheJulianJES, line 182: These clusters do not create any unnecessary entities in ZHA, but I still wonder if the device actually has these clusters (does it support reading any attribute on those clusters?) or if it just reports it has them and nothing actually reads/writes.

TheJulianJES (approved):
Changes LGTM. Added some comments (that don't block a merge).

PR #1460, opened 2022-04-03
Add quirks for _tz3210 TS011F socket

Brazen00:
This PR adds quirks for the Mercator IKUU TS011F AC 1- and 2-gang sockets.

Fixes https://github.com/zigpy/zha-device-handlers/issues/1403

PR #2679, opened 2023-10-24
Add custom Tuya `OnOff` cluster to two TS011F models

TheJulianJES:
## Proposed change

This also adds the custom Tuya `TuyaZBOnOffAttributeCluster` cluster to two more TS011F Tuya plug variations.
It seems like it was forgotten in the initial PR. All other plugs support this.

I noticed this during https://github.com/zigpy/zha-device-handlers/pull/2674, but wanted to have another commit for this, as  this change is more than just adding quirk IDs.

Having this custom cluster also allows us to add the `TUYA_PLUG_ONOFF` quirk ID, so ZHA will create config entities for this plug (in case the attributes are supported).

## Additional information

Original PR (that likely missed this): https://github.com/zigpy/zha-device-handlers/pull/1460

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -600,7 +600,7 @@ class Plug_4AC_2USB(CustomDevice):
 class Plug_TZ3210_2AC(CustomDevice):
     """TS0011F 2 outlet plug."""
 
-    # quirk_id = TUYA_PLUG_ONOFF  # TODO: Why does this quirk not have OnOffAttributeCluster??
+    quirk_id = TUYA_PLUG_ONOFF
 
     signature = {
         MODEL: "TS011F",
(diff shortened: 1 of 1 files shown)


TheJulianJES:
cc @Brazen00 since you made the original PR. From what I can see in Z2M, these two plug models should also support the custom attributes on the `OnOff` cluster.

So, if everything goes right, you'll see some config entities in HA Core 2023.11.x for these plugs too.
EDIT: [PR likely won't be merged in time](https://github.com/home-assistant/core/pull/102489), so you'll see them in HA Core 2023.12.x (hopefully).

Brazen00:
Thanks for this. How do you tell if a certain model supports this cluster? I have TS0501 fan controllers all over my house and the quirk just uses the original OnOff cluster.

TheJulianJES:
You can use a custom quirk that just uses the custom cluster in the `replacement`. Then, check if the attributes can be read using the clusters UI. If they all return `None`, it's not supported.

PR #2688, opened 2023-10-29
Fix attribute id for Inovelli fan switch VZM35-SN

codyhackw:
Smart Fan LED Display Levels is P263 instead of P262,

## Proposed change

Smart Fan LED Display Levels is P263 instead of P262, updating the quirk to reflect that correction.

## Checklist


TheJulianJES (approved):
LGTM ([same as Z2M uses](https://github.com/Koenkk/zigbee-herdsman-converters/blob/73c0b11bd6ba9d5b3ac6be57c0c15a6d56201ac5/src/devices/inovelli.ts#L852-L859))

PR #1039, opened 2021-09-21
Add iCasa wall mounted remote support: ICZB-KPD12, ICZB-KPD14S, ICZB-KPD18S 

Emrvb:
This pull requests replaces https://github.com/zigpy/zha-device-handlers/pull/855

I have made an initial implementation for support for the icasa wall mounted zigbee remotes ICZB-KPD12, ICZB-KPD14S and ICZB-KPD18S. Remotes are mentioned here #503

ICZB-KPD12: This remote has ON and OFF buttons (hold for dimming up/down) 
ICZB-KPD14S: This remote has ON and OFF buttons (hold for dimming up/down) and 2 SCENE buttons (hold for storing, push to recall).
ICZB-KPD18S: This remote has ON and OFF buttons (hold for dimming up/down) and 6 SCENE buttons (hold for storing, push to recall).

I myself am not very familiar with zigpy in general and the device handlers in particular, so I'm left with a few questions I would like answered in order to complete this device handler. I'll raise issues for the shortly.

Issue #1041, opened 2021-09-21
[Device Support Request] Stop command level control does not specify the button


Emrvb:
I'm working on the device handler for several iCasa remotes and I'm running into an issue with the level control. https://github.com/zigpy/zha-device-handlers/pull/1039

With these remotes you can dim up or down with on and off buttons. I want to add LONG_RELEASE for the on and off buttons, however the stop command does not include the button that was released. It seems zigpy does not keep track of it either.

How should I proceed?

1. Is there a workaround for this situation?
2. Is there a generic trigger is should use?
3. Should I just ignore this trigger?

MattWestb:
I was doing the tuya TS0404F that is one 4 button dimmer switch but i was not able doing the nice stop command with from up or down dimming.
It shall being possible doing if looking on IKEA Open/Close bottom have implanting logic for it:
https://github.com/zigpy/zha-device-handlers/blob/1c2b7ece00f494d16e38bb41093a525b1ed64f42/zhaquirks/ikea/opencloseremote.py#L48
If you is implanting it for the dimmer i can using it for the tuya dimmer switch :-))

Emrvb:
Thank you for pointing me in that direction. After a quick glance this indeed seems a solution to a similar problem. It should be no problem for me to implement that. 

However, if multiple remotes need this particular workaround, perhaps a more generalized (custom) cluster should be implemented, instead of copy and pasting the same fix around.

MattWestb:
I like to do but i and my laptop is on Holiday in Spain but coming "home" on Monday and rejoining with my Zigbee networks, then i can trying if i can getting it working with my new IKEA Styrbar i have baying in France (the stainless steel version for 10€) that have the same commands but little more for making some magic with scenes in IKEA lights.

O I see you is storing the options from last move and using it for the stop commands is being receiver and can getting the from up or down :-)).

The devs have more knowledge and is knowing if its one better way to do it or if its can making  problems in the higher level (ZHA).

Great work done Eric !!!

PING @dmulcahey
Sorry for addressing you directly Dave but is this one OK way for getting the "long press release" as 2 different automation triggers (stop from moving up and stop from moving down) from dimmer switches that many user like to having as device trigger from there devices or is it making problems in ZHA ?
Or is it one better way implanting it in the quirk root so all quirks can using it ?

I like implanting it for the IKEA dimmer swishes and also tuya TS004F if is working OK and not making problems.

MattWestb:
[...]

PR #2693, opened 2023-10-31
Add device automation triggers for Icasa wall-mount remotes

Emrvb:
This pull request supersedes #1039 

Quirk for icasa wall mounted zigbee remotes ICZB-KPD12, ICZB-KPD14S and ICZB-KPD18S. Remotes are mentioned here https://github.com/zigpy/zha-device-handlers/issues/503

ICZB-KPD12: This remote has ON and OFF buttons (hold for dimming up/down)
ICZB-KPD14S: This remote has ON and OFF buttons (hold for dimming up/down) and 2 SCENE buttons (hold for storing, push to recall).
ICZB-KPD18S: This remote has ON and OFF buttons (hold for dimming up/down) and 6 SCENE buttons (hold for storing, push to recall).

## Checklist


Emrvb:
If somebody cares to test this, that would be great. I've cleaned up this quirk a little while back, but couldn't be bothered to test it. Just did some updates, had to restart HA and the custom quirk came alive. _Appears_ to work as intended.

Emrvb:
And if I understand the CI logs correctly a little refactoring is required, will try to do that soon.

TheJulianJES:
Thanks for the PR!
Since there's no "real code" (just defining device triggers), we don't really need new tests for this device.
However, CI complains about some issues where triggers aren't unique and that `ARGS` need to be migrated to `PARAMS`.

Emrvb:
@TheJulianJES Yeah, that was fixed already but my commits are a bit messy on this one, so I accidentally reverted that fix when reverting something else.

My apologies to anyone who was waiting for this quirk, but better late than never.

zhaquirks/icasa/iczb_kpd12.py
```diff
@@ -0,0 +1,129 @@
...
+    OUTPUT_CLUSTERS,
+    PARAMS,
+    PROFILE_ID,
+    SHORT_PRESS,
+    TURN_OFF,
+    TURN_ON,
+)
+
+COMMAND_STORE = "store"
+COMMAND_RECALL = "recall"
```
TheJulianJES: Do you want to move this to https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/const.py?

TheJulianJES (commented):
Thanks, looks good! One minor note about adding the constants used everywhere to the global constants file.

Emrvb:
@TheJulianJES Sure, np.

TheJulianJES (approved):
LGTM, thanks!
Just to confirm, triggers all work as expected? 😄

javicalle:
I'm not really sure but if there's no changes in the `replacement` part I believe that you don't need to include it.
Can you test your quirk only with the signature and the device_triggers part?

Emrvb:
@TheJulianJES To be brutally honest, I haven't tested the last 2 commits and I only own the ICZB-KPD18S.

@javicalle I suspect you're right, the replacement part is there because of an experiment with a custom cluster, but I never got any feedback on whether I should do that (#1041), so I ripped that out for this pull request.

Do you happen to know if I should remove the replacement variable entirely or must it be defined with some empty value?

javicalle:
> Do you happen to know if I should remove the replacement variable entirely or must it be defined with some empty value?

Try to remove completely all the `replacement` definition.

Emrvb:
@javicalle

I get failed tests when I remove the replacement definition. It is required when inheriting from CustomDevice. I couldn't find a relevant example to see if I should inherit from something else.

The cleanest quirk I came across is https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/sonoff/button.py which has a replacement that's identical to the signature.

Feel free to merge.

PR #2710, opened 2023-11-07
Fix Tuya Haozee TRV deprecated `attridx` and usage

TheJulianJES:
## Proposed change
This replaces the deprecated `attridx` property with `attributes_by_name`.
The code was also somewhat broken, as it tried to call `update_attribute` with a `ZCLAttributeDef` (and not an `int`).

So, that should be fixed too.

## Additional information
Untested, but similar to other Tuya TRVs now. The code was just broken before.
In the future, this should probably all be changed to use new-style zigpy `AttributeDefs` classes. But this isn't for that yet.

Note: **Codecov tests will fail when merging, as nothing of this was/is covered. All other tests pass.**

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_haozee.py
+++ b/zhaquirks/tuya/ts0601_haozee.py
@@ -228,8 +228,10 @@ class HY08WEThermostat(TuyaThermostatCluster):
         elif value == 2:
             prog_mode = self.ProgrammingOperationMode.Simple
             occupancy = self.Occupancy.Unoccupied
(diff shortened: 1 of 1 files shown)


TheJulianJES:
Changed code is expected to not be covered, as the code also wasn't covered before.
This should at least fix an issue (and the deprecated usage) though.

I can't add tests for this at the moment, as I'm generally not too familiar with the Tuya TRV code and some of it will be (hopefully) rewritten in the future: "fake config clusters", ...

---

**Codecov check will fail when merging. All other tests pass.** This fix is needed.

PR #2715, opened 2023-11-07
Fix missing asserts in Tuya cluster tests

TheJulianJES:
## Proposed change
Fix missing asserts in Tuya cluster tests.
It looks like these were supposed to be checking `r.payload`.

However, the values they were supposedly trying to check for differ from the actual values.
The new values seem like they could be correct, but I'd like some confirmation on if this is correct, so if the tested code and tests are good.
@Shulyaka You seem to also have worked on this in the past. Can you say if the changes are correct?

## Checklist


Shulyaka (approved):
I think you are correct.

Issue #2286, opened 2023-03-20
[Device Support Request] Niko Double connectable switch,10A (552-721X2)


arvidc:
**Is your feature request related to a problem? Please describe.**
The device can be readily connected and the basic functionality (i.e. switching either relay; e.g. to control lights) seems to work as is. However, there are additional functionalities in a manufacturer specific cluster that are not currently supported by ZHA. More specifically, the following functions (per relay) could be added: Operation_mode (control_relay or decoupled), Led_enable (ON or OFF) and Led_state. In particular the first would be highly welcome, as it allows the switch to operate better with smart lights (by not actually switching the relay, but instead sending out an action event that could in turn be used to control the smart light).

**Describe the solution you'd like**
Adding support for missing functions in ZHA.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4703, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x010a",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0b05",
        "0xfc00",
        "0xfc01"
      ],
      "out_clusters": [
        "0x0019"
      ]
    },
    "2": {
      "profile_id": 260,
      "device_type": "0x010a",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0b05",
        "0xfc00",
        "0xfc01"
      ],
      "out_clusters": []
    },
    "242": {
      "profile_id": 41440,
[...]

PR #2723, opened 2023-11-07
Add Niko Connectable switch support

svenjochems:
## Proposed change
Add support for Niko Connectable switch (Both Double and Single rocker switch)

## Additional information
Based on input from https://github.com/zigpy/zha-device-handlers/issues/2286

## Checklist


zhaquirks/niko/switch.py
```diff
@@ -0,0 +1,300 @@
...
+                INPUT_CLUSTERS: [],
+                OUTPUT_CLUSTERS: [
+                    GreenPowerProxy.cluster_id,
+                ],
+            },
+        },
+    }
+
+    device_automation_triggers = {
+        (COMMAND_CLICK, BUTTON_1): {ARGS: {"value": 64}},
```
svenjochems, line 323: I have chosen following buttons for now:
- BUTTON_1: left
- BUTTON_2: right
- BUTTON_3: left_external
- BUTTON_4: right_external

For the single rocker switch, this is:
- BUTTON_1: left
- BUTTON_2: left_external

Is this acceptable or should I create extra entries in `zhaquirks/const.py` to add more descriptive names?

zhaquirks/niko/switch.py
```diff
@@ -0,0 +1,300 @@
...
+    cluster_id = 0xFC00
+    ep_attribute = "niko_config_cluster"
+    attributes = {
+        # If this would ever be expanded to outlets:
+        # WARNING: 0x0000 has different datatypes!
+        # enum8 (switch) vs. bitmap8 (outlet)
+        # unknown usage/function on outlet
+        0x0000: ("button_operation_mode", ButtonOperationMode, True),
+        0x0100: ("led_state", LedState, True),
+        0x0104: ("led_operation_mode", LedOperationMode, True),
```
svenjochems: Will it be better to change `LedState` and `LedOperationMode` to enums?

tests/test_niko.py
```diff
@@ -0,0 +1,48 @@
+"""Tests for Niko quirks."""
+
+import zhaquirks
+
+zhaquirks.setup()
+
+
+def test_motion_signature(assert_signature_matches_quirk):
+    """Test Niko Double Connectable switch signature is matched to its quirk."""
```
svenjochems, line 9: Only added a test for the double rocker switch as I don't own a single rocker switch

zhaquirks/niko/switch.py
```diff
@@ -0,0 +1,300 @@
...
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+from zhaquirks.niko import NIKO
+
+
+class NikoConfigCluster(CustomCluster, ManufacturerSpecificCluster):
```
TheJulianJES: Remove `ManufacturerSpecificCluster` here. AFAIK it's only really used internally by zigpy.

zhaquirks/niko/switch.py
```diff
@@ -0,0 +1,300 @@
...
+        # WARNING: 0x0000 has different datatypes!
+        # enum8 (switch) vs. bitmap8 (outlet)
+        # unknown usage/function on outlet
+        0x0000: ("button_operation_mode", ButtonOperationMode, True),
+        0x0100: ("led_state", LedState, True),
+        0x0104: ("led_operation_mode", LedOperationMode, True),
+    }
+
+
+class NikoActionCluster(EventableCluster, ManufacturerSpecificCluster):
```
TheJulianJES: Remove `ManufacturerSpecificCluster` here too.

svenjochems:
@TheJulianJES thanks for the review, I updated the PR based on your remarks. Do you happen to have some answers to my own questions in my review?

TheJulianJES:
> Do you happen to have some answers to my own questions in my review?

Hmm, maybe I'm missing something. But I do not see any comment/questions from you?
The changes seem to look good (from a quick review at least).

svenjochems:
> > Do you happen to have some answers to my own questions in my review?
> 
> Hmm, maybe I'm missing something. But I do not see any comment/questions from you? The changes seem to look good (from a quick review at least).

Apparently my review comments where still pending. They should be visible now.

zhaquirks/niko/switch.py
TheJulianJES: Can you still successfully read and write the attributes using the clusters UI then?

svenjochems: @TheJulianJES 
When I read `ButtonOperationMode` (enum) I get following output in the ui:
image

When I read `LedOperationMode` (uint24) I get following output in the ui:
image

When I do a quick search in the repository, I see that mostly enum type is used for these sort of types, I will adapt the code

svenjochems:
@TheJulianJES can you give this another review? I would like to have this PR merged as I have to install more of these devices in the near future :)

zhaquirks/niko/switch.py
TheJulianJES: I think this is fine. The names should be translated in ZHA (HA Core), so let's go with these.

zhaquirks/niko/switch.py
```diff
@@ -0,0 +1,313 @@
...
+        # If LedOperationMode == Decoupled and/or ButtonOperationMode == Decoupled, then LED can nevertheless by controlled by writing LedState ('always on' or 'always off')
+
+    cluster_id = 0xFC00
+    ep_attribute = "niko_config_cluster"
+    attributes = {
+        # If this would ever be expanded to outlets:
+        # WARNING: 0x0000 has different datatypes!
+        # enum8 (switch) vs. bitmap8 (outlet)
+        # unknown usage/function on outlet
+        0x0000: ("button_operation_mode", ButtonOperationMode, True),
```
TheJulianJES: Ideally, we should switch all attributes to use the new-style zigpy `AttributeDefs`, similar to this:

zhaquirks/niko/switch.py
```diff
@@ -0,0 +1,313 @@
...
+
+    class ButtonOperationMode(t.enum8):
+        # WARN: while one can technically write 0x00 to the operationMode attribute
+        #       this seems to brick the device and it will need to be rejoined
+        Decoupled = 0x01
+        ControlRelay = 0x02
+
+    class LedState(t.uint8_t):
+        Off = 0x00
+        On = 0xFF
```
TheJulianJES: Hmm, I'm not sure we can/should extend `t.uint8_t` like this (and "abuse" it as an enum). For example, it might break if it's used as a select config entity in HA(?), as it's not really an enum.
(But for that, it would be preferred to have an on/off switch anyway.)

~~For now, wrap it in an `t.enum_factory` like in the first example below. It makes it a proper enum whilst keeping the "unsigned integer" "Zigbee type". Still not optimal though – it should be handled better in the future.~~

Actually, leave the enum for now, but have it extend `t.enum8` (to be used for a future config entity).
In the actual attribute definition, use just the raw `t.unit8_t` type. Do not use `LedState` there.

I do see it's already done somewhat similar in other places though:
- https://github.com/zigpy/zha-device-handlers/blob/0df6f2b6a7c3c895a73bedcc5846108c3050897b/zhaquirks/xiaomi/aqara/tvoc.py#L70-L76 (^ this is an interesting way to do it, as it avoids zigpy setting the type to "enum" instead of "int" [here](https://github.com/zigpy/zigpy/blob/6aeb8a6816585a4bda67e9ea1b125568c18fa702/zigpy/zcl/foundation.py#L215). Seems to be somewhat of a hack though.)
- https://github.com/zigpy/zha-device-handlers/blob/0df6f2b6a7c3c895a73bedcc5846108c3050897b/zhaquirks/xiaomi/aqara/opple_switch.py#L63-L67 (^ also interesting, but IMO less good)
- https://github.com/zigpy/zha-device-handlers/blob/0df6f2b6a7c3c895a73bedcc5846108c3050897b/zhaquirks/sinope/thermostat.py#L65-L69 (^ not an Enum at all, should probably be changed)

The way we can implement this might slightly change with [quirks v2](https://github.com/zigpy/zigpy/discussions/1312#discussioncomment-8191961).
With quirks v2, we should be able to define a custom select entity for this. I'm pretty sure that would also need this class to be a "proper enum" (e.g. with the `t.enum_factory`)
(IMO, an on/off switch would be better, but I think the on/off values are hard-coded to 0x0 / 0x1 right now?)

zhaquirks/niko/switch.py
svenjochems: @TheJulianJES 
Do you have an other example of this?
I tried changing the attributes to
```
    button_operation_mode = ZCLAttributeDef(
        id=0x0000,
        type=t.enum8,
    )
    led_state = ZCLAttributeDef(
        id=0x0100,
        type=t.uint8_t,
    )
    led_operation_mode = ZCLAttributeDef(
        id=0x0104,
        type=t.enum8,
    )
```
But then I get `enum8.undefined_0x01` when reading the attribute in homeassistant ui.
image

Changing type to `ButtonOperationMode` does not seem to work either.

TheJulianJES: `type = ButtonOperationMode` (and so on) would be correct and should work.
There are more examples in the zigpy project in the `zcl` folder where all ZCL clusters are defined.

svenjochems: @TheJulianJES When defining the `ButtonOperationMode` class outside the `NikoConfigCluster` class and adding `ButtonOperationMode: Final = ButtonOperationMode` in the `NikoConfigCluster`, it works.

zhaquirks/niko/switch.py
```diff
@@ -0,0 +1,335 @@
...
+    Decoupled = 0x01
+    ControlRelay = 0x02
+
+
+class LedState(t.enum_factory(t.uint8_t)):
+    Off = 0x00
+    On = 0xFF
+
+
+class LedOperationMode(t.enum8):
```
svenjochems, line 52: Today I did some more testing. ButtonOperation works successfully but LedOperationMode does not seem to make any changes on the device.
@arvidc were you able to test this in your tests from https://github.com/zigpy/zha-device-handlers/issues/2286?
Any pointer how I can debug this?

TheJulianJES: So the attribute isn't written correctly? Writing `0` or `1` doesn't work?
If so, did you try setting the `type` for the `ZCLAttributeDef` to `t.uint8_t` and see if it works then?

arvidc: Hey @svenjochems , I've been using it through a custom quirk on my end and it seems to work (apart from the potentially unintuitive behaviour described in the comments; which makes me think `LedOperationMode` is perhaps better called `LedEnabled` with states `Disabled` and `Enabled`, but that's beyond this).
More specifically, I wrote `LedOperationMode` to `0` for most/all(?) of my switches, and actively control `LedState` on one to reflect our terrace light.

I do have them as:
```
   class LedState(t.uint24_t):
        Off = 0
        On = 255

    class LedOperationMode(t.uint8_t):
        Decoupled = 0
        ControlledByRelay = 1
```
        
and I can't succesfully write (and it doesn't read like that either) e.g. `LedOperationMode.Decoupled` (in analogy to `ButtonOperationMode`), but only `0` or `1`. Maybe because this isn't a proper enum, as I believe @TheJulianJES was pointing out earlier?

(Sorry, I barely know what an enum is, let alone a `t.uint24_t`. So pretty much all I do is copy and adapt pieces from somewhere and trial-and-error until I think it works well enough for me.)

TheJulianJES: > Maybe because this isn't a proper enum

Yes, the "Zigbee type" is just an integer. Not an enum. If the zigpy enum is used, zigpy also sends it as a Zigbee enum which we do not want in this case. This might be changed in the future.

For now, just set the `type` for the `ZCLAttributeDef` to `t.uint8_t` (to force zigpy to send it as a Zigbee int).
We can keep the enum for future-use (with quirks v2 for exposing the entity or when this is fixed. But that's what required for now.
Just inheriting `t.uint8_t` in a class "works", but doesn't really do anything different from just using `t.uint8_t` when other class attributes are added.

PR #2758, opened 2023-11-18
Add Sinope G2 devices support

claudegel:
## Proposed change

The new G2 light and dimmer add new functionality like doubleup full which open at 100% intensity and keep current dimmer value in memory. It add also support for the DM2550ZB phase control

## Additional information

This PR also fix device temperature that was incorrectly reported as celsius instead of celsius*100

## Checklist


claudegel:
I don't understand what's wrong with codecov. How can I correct this. I've the same in another PT #2759 and there is no error in codecov. Please tell me what's wrong or direct me to doc so I can do RTFM to correct this

TheJulianJES:
There's a percentage threshold for how much newly added code can be untested.
Ideally, we should have a basic unit test for the added code (for both quirks). I've left [a comment in the other PR](https://github.com/zigpy/zha-device-handlers/pull/2759#discussion_r1398065213) with some example.

I can also try to write this test in the next few days or so.

claudegel:
I've no idea on what to do for these test. I'll appreciate if you can do it but if you can direct me to doc I can read it should help for futur PR.
I've tested those attrid multiplication or division and they fix wrong value for device temperaure

TheJulianJES:
Linking my comment in the other PR for reference: https://github.com/zigpy/zha-device-handlers/pull/2759#issuecomment-1817381385

I've now added tests in that PR. It will need to be merged before this, so we can reference the (then existing) `CustomDeviceTemperatureCluster` in this PR.
This PR will require a rebase, but I can do that after the other one is merged.

claudegel:
Hi, as this SINOPE_MANUFACTURER_CLUSTER_ID = 0xFF01 is present in each 4 files is it better to put it in __init__.py and import into each files instead ?
Thank you for your help

TheJulianJES:
Yeah, sounds like a good idea.

claudegel:
Ok so I'll change it in this PR

TheJulianJES (approved):
LGTM, thanks!

PR #2759, opened 2023-11-18
Add Sinope valve flow meter support

claudegel:
## Proposed change

Added support for the new Sedna valve 2n gen with flow meter.
Fix device temperature that was wrong 
Fix flow meter level that was wrong

## Additional information

Added many cluster attributes for switches

## Checklist

### DIFF
--- a/zhaquirks/sinope/switch.py
+++ b/zhaquirks/sinope/switch.py
@@ -103,7 +103,7 @@ class SinopeManufacturerCluster(CustomCluster):
         Gal_80 = 0x04
 
     class FlowDuration(t.uint32_t):
-        """tank_size values."""
+        """Abnormal flow duration."""
 
         M_15 = 0x0384
         M_30 = 0x0708
--- a/zhaquirks/sinope/switch.py
+++ b/zhaquirks/sinope/switch.py
@@ -37,7 +37,7 @@ from zhaquirks.const import (
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-from zhaquirks.sinope import SINOPE, FLOWMETER, CURTEMP
+from zhaquirks.sinope import CURTEMP, FLOWMETER, SINOPE
 
 SINOPE_MANUFACTURER_CLUSTER_ID = 0xFF01
 
--- a/zhaquirks/sinope/switch.py
+++ b/zhaquirks/sinope/switch.py
@@ -23,7 +23,11 @@ from zigpy.zcl.clusters.general import (
 )
 from zigpy.zcl.clusters.homeautomation import Diagnostic, ElectricalMeasurement
 from zigpy.zcl.clusters.lightlink import LightLink
-from zigpy.zcl.clusters.measurement import RelativeHumidity, TemperatureMeasurement, FlowMeasurement
+from zigpy.zcl.clusters.measurement import (
+    RelativeHumidity,
+    TemperatureMeasurement,
+    FlowMeasurement,
+)
 from zigpy.zcl.clusters.security import IasZone
 from zigpy.zcl.clusters.smartenergy import Metering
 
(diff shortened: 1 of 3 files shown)


zhaquirks/sinope/switch.py
```diff
@@ -56,36 +103,114 @@ class TankSize(t.enum8):
...
+            0x0300: ("unit_of_measure", UnitOfMeasure, True),
+        }
+    )
+
+
+class CustomDeviceTemperatureCluster(CustomCluster, DeviceTemperature):
+    """Custom DeviceTemperature Cluster."""
+
+    def _update_attribute(self, attrid, value):
+        if attrid == 0x0000:
```
TheJulianJES: `attrid == self.AttributeDefs.current_temperature.id`

zhaquirks/sinope/switch.py
```diff
@@ -56,36 +103,114 @@ class TankSize(t.enum8):
...
+            super()._update_attribute(attrid, value * 100)
+
+
+class CustomFlowMeasurementCluster(CustomCluster, FlowMeasurement):
+    """Custom FlowMeasurement Cluster."""
+
+    def _update_attribute(self, attrid, value):
+        if attrid == 0x0000:
```
TheJulianJES: `attrid == self.AttributeDefs.measured_value.id`

zhaquirks/sinope/switch.py
```diff
@@ -56,36 +103,114 @@ class TankSize(t.enum8):
...
+        }
+    )
+
+
+class CustomDeviceTemperatureCluster(CustomCluster, DeviceTemperature):
+    """Custom DeviceTemperature Cluster."""
+
+    def _update_attribute(self, attrid, value):
+        if attrid == 0x0000:
+            super()._update_attribute(attrid, value * 100)
```
TheJulianJES: Ideally, we'd also have a very small test for this.

It could just call `update_attribute` (forwards to `_update_attribute` in zigpy) on this cluster and then check if it's multiplied correctly.

As somewhat of an example, see this:

important part:
https://github.com/zigpy/zha-device-handlers/blob/fb794323fb11014c41843bebc5825bc9004cd23a/tests/test_xiaomi.py#L582-L589

first part:
https://github.com/zigpy/zha-device-handlers/blob/fb794323fb11014c41843bebc5825bc9004cd23a/tests/test_xiaomi.py#L528-L539

Otherwise, I can also see if I get around to creating this short test in (hopefully) the next few days.

zhaquirks/sinope/switch.py
```diff
@@ -56,36 +103,114 @@ class TankSize(t.enum8):
...
+
+
+class CustomFlowMeasurementCluster(CustomCluster, FlowMeasurement):
+    """Custom FlowMeasurement Cluster."""
+
+    def _update_attribute(self, attrid, value):
+        if attrid == 0x0000:
+            super()._update_attribute(attrid, value / 10)
```
TheJulianJES, line 204: Same as other comment about adding a small test case for this (can both go in the same test).

TheJulianJES (commented):
Mostly looks good. Added some small comments.

If you can't add the test (to check for correct multiplication/division), I should (hopefully) be able to do it in the next few days.

zhaquirks/sinope/switch.py
TheJulianJES: Resolved with https://github.com/zigpy/zha-device-handlers/pull/2759/commits/3ec99ba6ab734db14c8c19a1d052a8e9a5cced42

zhaquirks/sinope/switch.py
TheJulianJES: Resolved with added test in https://github.com/zigpy/zha-device-handlers/pull/2759/commits/3ec99ba6ab734db14c8c19a1d052a8e9a5cced42

TheJulianJES:
I've added some unit tests for the attribute multiplication and division for this PR now. Please take a look.

I also moved the `CustomDeviceTemperatureCluster` to ``sinope/__init__.py``, since your [G2 device support PR](https://github.com/zigpy/zha-device-handlers/pull/2758) also seems to use that class.
That other PR will need to be rebased after this one is merged, so that `CustomDeviceTemperatureCluster` can be used. (I can do that after this is merged if you want.)
That PR will not require any added tests, as the `CustomDeviceTemperatureCluster` code is already tested using this PR.

claudegel:
Thanks that look a lot better like this  I'll do the same for futur PR that are comming.
I'll wait for the test merge and rebase the other PR

Issue #2203, opened 2023-02-18
[Device Support Request] IKEA controller with 24.X firmware battery readings.


MattWestb:
**Is your feature request related to a problem? Please describe.**

All 24.X controllers have fixed battery reporting and dont need the doubling of the % renaming.
Only Styrbar looks reporting 255% and its something wrong in the firmware (it was the first out).

Styrbar and 5 button and old motion sensor update is in the normal IKEA feed.
Dirigera is updating OnOff, OpenClose new motion sensor and blinds and all need being updated.

**Describe the solution you'd like**

The problem is that most of the device is using the same EP / Cluster setting like the privies version so we cant fixing it in the quirk in the normal way.

Is it possible using software version from basic cluster and if its larger then 2.0 we using no battery doubling and if its lower we is doing it (Z2M is start using that) ??
If its possible i suggesting implanting it in IKEA  INIT for not spamming all quirks with the same code.

<details>

<summary>Device signature</summary>

```yaml
Paste the device signature here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
Paste the diagnostic information here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

<details>
<summary>Additional logs</summary>

```
Paste any additional debug logs here.
Don't remove the extra line breaks outside the ``` marks.
```

</details>

**Additional context**

MattWestb:
Matrix before and after updating to 24.X firmware
Device  | Type | TRADFRI | DIRIGERA | Battery | PM
-- | -- | -- | --| -- | --
E1525 | Old Motion | 24.4.5 | 24.4.5 | OK | No Quirk
E1745 | New Motion | 2.0.022 | 24.4.5 | OK | No Quirk
E1743 | OnOff / OpenClose | 2.3.079 | 24.4.5 | 200% | With Quirk
E1524 | Remote 5Bt | 24.4.5 | 24.4.5 | OK
E2001/2| Styrbar | 24.4.5 | 24.4.5 | 255%
E1744 | Symfonisk | 2.3.080 | 24.4.5 | OK | No Quirk
E1812 | Shortcut | 2.3.080 | ? |  ?

All is not made and coming later.

TheJulianJES:
Can you also post all the new signatures? (after re-pairing the devices)

MattWestb:
[...]

PR #2773, opened 2023-11-26
Remove doubling of battery for symfonisk sound

elupus:
## Proposed change

Sound controllers does not need doubling of battery percentage.

## Additional information

Debug output of one of my sound controller, verified with another too.
[zha-810b434c4e1088344d3010f1e76eeb64-IKEA of Sweden SYMFONISK Sound Controller-e9fda80013051afd7cf7fd9a9982c3d7.json.txt](https://github.com/zigpy/zha-device-handlers/files/13467958/zha-810b434c4e1088344d3010f1e76eeb64-IKEA.of.Sweden.SYMFONISK.Sound.Controller-e9fda80013051afd7cf7fd9a9982c3d7.json.txt)

## Checklist

### DIFF
--- a/zhaquirks/ikea/symfonisk.py
+++ b/zhaquirks/ikea/symfonisk.py
@@ -80,7 +80,7 @@ class IkeaSYMFONISK1(CustomDevice):
                 DEVICE_TYPE: zha.DeviceType.REMOTE_CONTROL,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
-                    DoublingPowerConfig1CRCluster,
+                    PowerConfiguration.cluster_id,
                     Identify.cluster_id,
                     PollControl.cluster_id,
                     LightLink.cluster_id,


TheJulianJES:
Ah yeah, this is an issue for most IKEA remotes at this point ([more information here](https://github.com/zigpy/zha-device-handlers/issues/2203)).
Your remote is running firmware `24.4.5`, but I think there are older firmware versions for the Symfonisk that still need doubling(?) (cc @MattWestb)

We don't have the ability to match quirks per firmware version though. So this is somewhat of an issue.
And since ZHA doesn't have firmware updates enabled by default, we don't know for how many users are running old vs new firmware.

Another idea could be to completely ignore the battery percentage sent by the device and just calculate it ourselves depending on the battery voltage. We'd need to match the scale basically 1:1 though. This *should* work for new and old firmware.

Lastly, we could just remove battery doubling for all/most IKEA remotes at this point. I'm not sure how much of a breaking change this would be though.

elupus:
Given that we are reporting firmware version in GUI it must be available somewhere. Can't we use this cached value to choose? (not using different quirks, just adapting inside existing quirk)

elupus:
Ps. Also sort of make more sense to adapt to latest available firmware even if it's hard to update.

TheJulianJES:
> Given that we are reporting firmware version in GUI it must be available somewhere. Can't we use this cached value to choose? (not using different quirks, just adapting inside existing quirk)

I think that value is only stored in the HA registry which we can't access from zha-quirks: https://github.com/home-assistant/core/blob/670e5a2eae0a49ffd0552b7171bac9f2b363b20f/homeassistant/components/zha/core/device.py#L421-L430
That method is also only fired when the `query_next_image` command is sent by the device. At least some devices don't send this automatically, so the `image_notify` command has to be sent from ZHA, even after the firmware update to refresh the value (displayed in the UI).

We could access the ZCL `sw_build_id` attribute if cached, but it's never cached automatically AFAIK (only cached when read manually). And we can't trigger a read, as the devices don't respond when sleeping (which is basically always for remotes).
zigpy has the `sw_build_id` attribute cached for your remote, so I'm guessing you read it manually at some point(?)

If the attribute is cached, we could execute the proper behavior. If it's not cached (default), we'd need to default to doubling or not-doubling. This could cause different behavior for the same exact device/firmware depending on the cache, so maybe it's not the best idea.

> Ps. Also sort of make more sense to adapt to latest available firmware even if it's hard to update.

Yeah, agreed. I'll have a look (soon-ish) to see if the "percentage depending on voltage" idea could work. Otherwise, we could just remove doubling for all IEKA remotes that can be updated to 24.x firmware. Worst case is that older remotes shows half the percentage which might cause some confusion, but shouldn't stop anything from working.

MattWestb:
First is the wrong device class then the latest firmware using `Quirk: zhaquirks.ikea.symfonisk.IkeaSYMFONISK2` so line 179 is the right place doing the change then the line 83 is being used of older firmware.

If users **not** updating the firmware then they is not knowing it or like keeping the group binding its no problem then they is shipped with 2.X firmware and using older quirk classes.
But all users that have activating IKEA OTA is facing the same problem so i facing the same problem with nearly all IKEA controllers (have installing 6 motion 2 last week and its not funny updating them but needed for getting battery daring being fixed).

Fix the quirk class and it shall being OK and shall being merged.

PS: New OnOff Dimmer switch bayed last week in the new combo pack with one gen 3 WW is produced 21XX and running 2,2X firmware so the user must updating it for getting the problem.

elupus:
I don't follow. My device seem to be using zhaquirks.ikea.symfonisk.IkeaSYMFONISK1, see supplied dump.

TheJulianJES:
I'm guessing the device was paired with an older firmware and only later updated. ZHA doesn't re-interview when the firmware is updated (which could cause the signature to change).

If it's (removed and) re-paired, it might have a slightly different signature causing zigpy to match another quirk.

elupus:
It was re-configured. That must be enough? If not that is very confusing for a user.

TheJulianJES:
Reconfiguration only binds all relevant clusters and sets up attribute reporting again.
It doesn't re-interview the device and I agree we should add an option for this.

Right now, I'm not even sure if just allowing joins and resetting the device (which causes it to pair again) is enough for the new signature to be "recognized". It's also possible that the device has to be removed from ZHA first before re-pairing.

elupus:
I tried removing and re-adding. It did something: 

[zha-810b434c4e1088344d3010f1e76eeb64-IKEA of Sweden SYMFONISK Sound Controller-2d6ff0e4db97b763cce17e24a386e9bb.json (1).txt](https://github.com/zigpy/zha-device-handlers/files/13468180/zha-810b434c4e1088344d3010f1e76eeb64-IKEA.of.Sweden.SYMFONISK.Sound.Controller-2d6ff0e4db97b763cce17e24a386e9bb.json.1.txt)

Strangely quirk class is now: "quirk_class": "zigpy.device.Device",

elupus:
So apparently none of the quirks are applied now. Ie the device triggers are lost. Events for move still seem to work thou.

Also strangely. it started with reporting 125% battery, (ie value 255). Then it went back to normal 100%. Should 255 be considered an invalid value and reported as unavailable in HA perhaps?

-edit-
Yes according to NXP docs at least "The special value 0xFF indicates an invalid or unknown measurement. "

Fix for battey unavailable for HA here: https://github.com/home-assistant/core/pull/104566
 -edit-

TheJulianJES:
> I tried removing and re-adding. It did something: [...] Strangely quirk class is now: "quirk_class": "zigpy.device.Device",
> So apparently none of the quirks are applied now. Ie the device triggers are lost.

Looks like the signature for was really missing. The new firmware has an added an (unused) "Works With All Hubs" cluster.
Can you try [this file](https://raw.githubusercontent.com/TheJulianJES/zha-device-handlers/tjj/ikea_symfonisk_new_sig/zhaquirks/ikea/symfonisk.py) as a custom quirk to see if it matches and you get your device triggers back?

> it started with reporting 125% battery, (ie value 255). Then it went back to normal 100%. Should 255 be considered an invalid value and reported as unavailable in HA perhaps?

This is somewhat weird behavior of IEKA remotes. As you've already seen, [ZCL](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf) considers 0xFF to be invalid. I'm guessing we don't have it in HA, as no other device ever used `255`. Might be a good idea to add it though.

MattWestb:
Thanks i was also **not** doing one delete and adding it new so the new signature is more then likely the right and my wrong.

About reading attribute from / sending commands to the device i see 2 possible way and one is like IKEA is doing by configuring reporting if the` sw_build_id` attribute so if doing the same we is getting it in the zigbee.db like charm.
The next is all IKEA controllers is Zigbee PRO R22 then not running ZLL firmware and is doing checkins every 55 minutes. If implanting sending queered commends then the device is doing checkins its possible reading attributes and sending commands to it without problems (Z2M have implanting some of this functionality).

MattWestb:
> > it started with reporting 125% battery, (ie value 255). Then it went back to normal 100%. Should 255 be considered an invalid value and reported as unavailable in HA perhaps?
> 
> This is somewhat weird behavior of IEKA remotes. As you've already seen, [ZCL](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf) considers 0xFF to be invalid. I'm guessing we don't have it in HA, as no other device ever used `255`. Might be a good idea to add it though.

I was changing battery in one OnOff dimmer switch before going to bed and i was seen the same.
But i think its logic then the device is not reading the batter so oft (i think only some times a day) so after putting in one new battery its sending the last red reading that is out of range / not valid and after some time its sending the new updated one (my device was doing it one hour after getting it online  with new battery).

PS: Only Styrbar is looking sending 255 oft then putting in one new batteries and i think its one firmware defect then its was the first "fixes" gen 3 device but its reporting better then the battery is being partly used but not good.

elupus:
Ive provided a fix for 255 value in HA. But feels like this should be fixed in zigpy to return a none value in case of unavailable values.

MattWestb:
`0xff` shall not being filtered out but shall being flagged and not valid reading from the device in ZHA then its one OK value in ZCL but is having other meaning then `0x00-0xfe` that is the battery % and the ZCL is having many this kind of specials in the standard and the same is ruling Matter that is using the same commands.

elupus:
i'm not saying 255 should always be removed. I'm saying in the case of a battery percentage in the power cluster, it should be replaced by a None value. It will still be reported, but as a indication of unavailable data.

elupus:
@TheJulianJES Tested the quirk. Loaded up fine here. So that should be alright.

TheJulianJES:
Superseded by:
- https://github.com/zigpy/zha-device-handlers/pull/2948

That fix works for devices that support both old firmware and firmware by using the cached `sw_build_id` attribute (and reading it on an incoming attribute report if not cached).

If you still see new "double battery percentage" reports, try to press a button on the remote or manually read the `sw_build_id` attriubte. Re-pairing/reconfiguring should also work.

PR #2774, opened 2023-11-26
Add alternative IKEA Symfonisk Gen1 signature

TheJulianJES:
## Proposed change
Adds an alternative signature for the IKEA Symfonisk Gen1 device.
This is a signature from 24.x firmware when newly paired.

It also cleans up the comments a bit and changes from the copying the device triggers to having all alternative classes inherit from `IkeaSYMFONISK1` (`device_automation_triggers` is still the same for all then).

Since 24.x firmware no longer requires power doubling, this quirk does not double the power. It uses the [`PowerConfig1CRCluster`](https://github.com/zigpy/zha-device-handlers/blob/3eb660a907829df0a7e52e5c8c45a11e88eaf02c/zhaquirks/ikea/__init__.py#L173-L180) cluster class to provide some information about battery quantity, size, and voltage.

## Additional information
Device diagnostics with signature originally provided in: https://github.com/zigpy/zha-device-handlers/pull/2773#issuecomment-1826910690

Note: We still need a proper long-term solution for quirks with the older signatures (that could still match this device if updated and not re-paired) regarding battery percentage doubling. Related:
- https://github.com/zigpy/zha-device-handlers/issues/2203
- https://github.com/zigpy/zha-device-handlers/pull/2773 (discussion)

## Checklist

### DIFF
--- a/zhaquirks/ikea/symfonisk.py
+++ b/zhaquirks/ikea/symfonisk.py
@@ -38,7 +38,13 @@ from zhaquirks.const import (
     TRIPLE_PRESS,
     TURN_ON,
 )
-from zhaquirks.ikea import IKEA, IKEA_CLUSTER_ID, DoublingPowerConfig1CRCluster
+from zhaquirks.ikea import (
+    IKEA,
+    IKEA_CLUSTER_ID,
+    WWAH_CLUSTER_ID,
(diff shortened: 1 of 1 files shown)


MattWestb:
Looks great and i trying getting time in the afternoon testing it !!!

PR #2779, opened 2023-11-27
Fix Aqara motion sensor erroneous motion events

TheJulianJES:
## Proposed change
The Aqara T1 motion sensor sends false attribute reports about every hour or so on the `Occupancy` cluster. Just using a `LocalDataCluster` doesn't ignore them, so we override `handle_cluster_general_request`.
The same fix is also applied for the Aqara "high-precision" agl04 motion sensor.

This was reported via Discord.

## Additional information
We should also consider using a `LocalMotionCluster` for this sensor in the future (like the P1 quirk already does).
A 60 second timeout also works for this sensor.

Ideally, we should also figure out if this is needed for the "high-precision" agl04 motion sensor.
-> yes, see https://github.com/home-assistant/core/issues/114129

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/__init__.py
+++ b/zhaquirks/xiaomi/__init__.py
@@ -11,6 +11,7 @@ from zigpy import types as t
 import zigpy.device
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
+from zigpy.typing import AddressingMode
 from zigpy.zcl import foundation
 from zigpy.zcl.clusters.general import (
     AnalogInput,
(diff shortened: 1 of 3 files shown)


TheJulianJES:
Unfortunately, the person who had these issues and was testing this is no longer responding.

For anyone else with the Aqara T1 motion sensor who experiences "ghost detections", please test [this](https://raw.githubusercontent.com/zigpy/zha-device-handlers/2cf6736987fea7de780c1cd08880e5aed0d8df98/zhaquirks/xiaomi/aqara/motion_agl02.py) as a custom quirk and report back if that corrects it.

HA-smart:
image
I don't really understand what I need to change

TheJulianJES:
> I don't really understand what I need to change

There are fixes for the Aqara T1 motion sensor and Aqara high-precision motion sensor in Home Assistant Core 2024.4.0 (in beta currently). There's nothing for you to change.

stast1:
lumi.motion.agl04 does not work after update. Does not detect motion. Never.

TheJulianJES:
@stast1 Please create a new issue. Also, try re-pairing the sensor.

stast1:
https://github.com/home-assistant/core/issues/115027
Re-pairing doesn't help

PR #2787, opened 2023-11-29
Fix tests with zigpy 0.60

mweinelt:
## Proposed change

In zigpy 0.60 the permit_with_link_key function became abstract, which means we now need to provide it on the concrete class.

This fixes error of the following kind:

```
    @pytest.fixture(name="MockAppController")
    def app_controller_mock():
        """App controller mock."""
        config = {"device": {"path": "/dev/ttyUSB0"}, "database": None}
        config = MockApp.SCHEMA(config)
>       app = MockApp(config)
E       TypeError: Can't instantiate abstract class MockApp with abstract method permit_with_link_key

tests/conftest.py:88: TypeError
```

## Additional information

Breaking change here:

https://github.com/zigpy/zigpy/commit/b0ffd5b624c3d1529f1aebd5a0548891572a2193#diff-066be728ff64be6f84d4f0ea83512608a64bead3804e27ef5bee4c73344f5287R1172

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -70,6 +70,9 @@ class MockApp(zigpy.application.ControllerApplication):
     async def start_network(self, *args, **kwargs):
         """Mock start_network."""
 
+    async def permit_with_link_key(self, *args, **kwargs):
+        """Mock permit_with_link_key"""
+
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
Thanks!

Issue #1630, opened 2022-06-23
[Device Support Request] TS004F _TZ3000_rco1yzb1


TomBursch:
**Is your feature request related to a problem? Please describe.**
I connected a smart button, which I bought from Lidl in Germany. The button is not firing any events and shows up as a switch inside of home assistant.

**Describe the solution you'd like**
Although mentioned [here](https://github.com/zigpy/zha-device-handlers/pull/1437), that this device should work, I couldn't get it to recognize a tap.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0104",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0004",
        "0x0006",
        "0x1000",
        "0xe001"
      ],
      "out_clusters": [
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x000a",
        "0x0019",
        "0x1000"
      ]
    }
  },
  "manufacturer": "_TZ3000_rco1yzb1",
  "model": "TS004F",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant Container",
    "version": "2022.6.7",
    "dev": false,
    "hassio": false,
    "virtualenv": false,
    "python_version": "3.9.12",
    "docker": true,
    "arch": "x86_64",
    "timezone": "Europe/Berlin",
    "os_name": "Linux",
[...]

PR #2790, opened 2023-11-30
Add Tuya Moes Smart Knob ERS-10TZBVK-AA `_TZ3000_qja6nq5z`

avzhuiko:
## Proposed change
Developed full support of Tuya Moes Smart Knob ERS-10TZBVK-AA.
Fixed bug, that rotating is triggered by random, because there is used parameter "step_size": 13 and "step_size": 37, but there are much more options (25, 49, 61, 73 and so on) depending on rotation speed.
Also now it supports "press and rotate".

## Additional information
It has full backward compatibility. Tested on real device.
Maybe it will help with #1630

## Checklist


zhaquirks/tuya/ts004f.py
```diff
@@ -132,55 +143,152 @@ class TuyaSmartRemote004FROK(EnchantedDevice):
...
+            ENDPOINT_ID: 1,
+            CLUSTER_ID: 6,
+        },
+        (COMMAND_STEP, RIGHT): {
             COMMAND: RIGHT,
             ENDPOINT_ID: 1,
             CLUSTER_ID: 6,
         },
-        (ROTATED, LEFT): {
+        (COMMAND_STEP, LEFT): {
```
MattWestb: Way renaming it to somthing  the user dont understanding ?

zhaquirks/tuya/ts004f.py
```diff
@@ -132,55 +143,152 @@ class TuyaSmartRemote004FROK(EnchantedDevice):
...
+            ENDPOINT_ID: 1,
+            CLUSTER_ID: 6,
+        },
+        # only on releasing after 3 seconds
+        (LONG_PRESS, BUTTON_1): {
+            COMMAND: LONG_PRESS,
+            ENDPOINT_ID: 1,
+            CLUSTER_ID: 6,
+        },
+        (COMMAND_STEP, RIGHT): {
```
MattWestb: Way renaming it to somthing  the user dont understanding ?

zhaquirks/tuya/ts004f.py
```diff
@@ -132,55 +143,152 @@ class TuyaSmartRemote004FROK(EnchantedDevice):
...
+            COMMAND: COMMAND_STEP_COLOR_TEMP,
             ENDPOINT_ID: 1,
-            PARAMS: {"step_mode": 1, "step_size": 37},
+            CLUSTER_ID: 768,
+            PARAMS: {"step_mode": 1},
         },
-        (SHORT_PRESS, BUTTON_1): {ENDPOINT_ID: 1, COMMAND: SHORT_PRESS},
-        (LONG_PRESS, BUTTON_1): {ENDPOINT_ID: 1, COMMAND: LONG_PRESS},
-        (DOUBLE_PRESS, BUTTON_1): {ENDPOINT_ID: 1, COMMAND: DOUBLE_PRESS},
-        (ROTATED, RIGHT): {
```
MattWestb, line 174: Way renaming it to somthing  the user dont understanding ?

zhaquirks/tuya/ts004f.py
```diff
@@ -69,19 +76,23 @@ class TuyaSmartRemote004FROK(EnchantedDevice):
...
+            # Moes Smart Knob ERS-10TZBVK-AA
             ("_TZ3000_qja6nq5z", "TS004F"),
             ("_TZ3000_csflgqj2", "TS004F"),
             ("_TZ3000_abrsvsou", "TS004F"),
         ],
         ENDPOINTS: {
             1: {
                 PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: zha.DeviceType.DIMMER_SWITCH,
+                DEVICE_TYPE: zha.DeviceType.NON_COLOR_CONTROLLER,
```
MattWestb: If changing it do it as the class its it representing (you is using color commands original was not)

MattWestb (changes requested):
We was not implanting the fates rotate and press rotate then it was not working well and was useless the test users was saying (perhaps step_size was the wrong way implanting it but i was not having the device). I think its more then welcome if have test all functions is working OK also on older versions but changing the name of the DA i dont like at all then we have using the same for the DMS and other devices with different tuya working modes and is one **braking change** for all users that have automatons done with them so only making problems.

So big thanks for digging more in this very _interesting_ tuya device and using more of its functionality !!

javicalle:
I'm not familiar with the rotating knob or anything similar, but these 'magic numbers' must be something else:
| Decimal | Binary | 
| --------- | ------- |
| 13 | 0001101|
| 25 | 0011001|
| 37 | 0100101|
| 49 | 0110001|
| 61 | 0111101|
| 73 | 1001001|

If you trim the 2 right bits (always `01`) you get:
| Binary | Decimal  |
| --------- | ------- |
| 00011 | 3 |
| 00110 | 6 |
| 01001 | 9 |
| 01100 | 12 |
| 01111 | 15 |
| 10010 | 18 |

I believe that the calculation would be more simple than:
https://github.com/zigpy/zha-device-handlers/blob/b6098daa461473efe649b7ba31c5e613750c6601/zhaquirks/tuya/ts004f.py#L162

And maybe the 2 right bits are `10` if pressing?

I'm speaking just from the mathematical point of view. I have no idea what means these `step_size` or anything 🤷🏻‍♂️

MattWestb:
Hola @javicalle 
The problem is that the device is sending its one series so the user must knowing how many steps  to triggering but if implanting one "true counting" its great but if like using 4 steps you is getting first `step 1`, `step 2`, `step 3` and then `step 4` and perhaps more so its nearly useless doing good automatons of it then other automation using lesser steps is being triggered (Its still light commands we is using not real steps rotated) (but looks very nice !!!)

avzhuiko:
Hi @MattWestb
I decided to make an another class with my improvements and now I'm covering it with tests. 
But I do not have any other TS004F devices, so I can't cover tle old code. Could you help me to find some info about "old" devices?

About renaming DA: I reverted it, but how users will figure out that a _"right" step_ and _"right" device_rotated_slow/medium/fast_ are the same, and that device_rotated is just a more precise version of the _"right" step_?

MattWestb:
I have one of the original and can testing it until Thursday (then im not at home for 10 days) then i think its better having only one device class for the same type of devices.

For the DA only Rotate and click / press = Scene mode. With speed Light controller mode.
And yes its using the same commands only different "speed" for the light command.
Here is the problem doing it so users is understanding it but them making more DA is being little messy. So in the end it can being better the user is doing automatons of events but if like ding more DA is very welcome.

And i dont knowing if all different devices is running in the same firmware or is tuya have doing new versions. The SW build is not implanted (as expected from tuya) and only useful information i can getting from the device is:
```
Attributes of the selected cluster
app_version (id: 0x0001)
 
Value
67
```
What app version is your device have ?

avzhuiko:
And some additional explanation about connection between _transition_time_ and _step_size_.
Without pressing:
- transition_time = 1 when step_size is one of [13, 25]
- transition_time = 2 when step_size is one of [37, 49, 61]
- transition_time = 3 when step_size more than 61  

With pressing (pretty same):
- transition_time = 1 when step_size is one of [18, 35]
- transition_time = 2 when step_size is one of [52, 69, 86]
- transition_time = 3 when step_size more than 86

So yes, fast rotating is different because now is more than 61, but I think nobody used it because it didn't work at all with other step sizes.

avzhuiko:
> What app version is your device have ?

The same - 67

MattWestb:
Then i think all is using the same firmware (version) and only changing the manufacture name.
Do the change and i cant testing it if i have time before i going on my trip.

zhaquirks/tuya/ts004f.py
avzhuiko: It was mistake =(

zhaquirks/tuya/ts004f.py
```diff
@@ -132,45 +143,138 @@ class TuyaSmartRemote004FROK(EnchantedDevice):
...
+            COMMAND: COMMAND_STOP_MOVE_STEP,
+            ENDPOINT_ID: 1,
+            CLUSTER_ID: 768,
+        },
+        # rotation without pressing contains step_size where you can get ticks count by formula ((step_size - 1) / 12)
+        (COMMAND_STEP, RIGHT): {
+            COMMAND: COMMAND_STEP,
+            ENDPOINT_ID: 1,
+            CLUSTER_ID: 8,
+            PARAMS: {"step_mode": 0},
```
MattWestb, line 168: Its the same function as the slow, medium and fast so i think the "ground name" shall being the same.
Perhaps `ROTATED_STEP`, `ROTATED_LAZILY`.
Or making the fast as "Worp" or "Fastest" ? (and moving all naming one step up)
More suggestions is very welcome for getting it better.

MattWestb (commented):
Ned feedback from @javicalle and @TheJulianJES  and others !!

zhaquirks/tuya/ts004f.py
avzhuiko: maybe we can use `SPIN`?

TheJulianJES (commented):
I think the changes look good from a quick review.

> It has full backward compatibility. Tested on real device.

Nice! And also a big thanks for adding tests!

TheJulianJES:
@javicalle If you have a minute, are you also good with this now?

PR #2803, opened 2023-12-06
Inovelli custom attribute updates

codyhackw:
## Proposed change

Fixes/Adds Parameters from 2.15 for the Inovelli 2-1 switch and updates for the Fan switch.

## Additional information

Will require an update in HA Core for the leading/trailing edge parameter to be exposed in the device page UI.

## Checklist


zhaquirks/inovelli/__init__.py
```diff
@@ -248,6 +250,7 @@ def handle_cluster_request(
     0x0013,
     0x0014,
     0x0019,
+    0x001A,
```
TheJulianJES, line 253: Not for this PR, but we should probably have like three lists for the attributes:
- base (used for all devices)
- light switch (copies base attributes, adds new light switch attributes)
- fan switch (copies base attributes, adds new fan switch attributes)

ZHA has a completely separated list: https://github.com/home-assistant/core/pull/102924/files

codyhackw: Yeah, that'd be a nice way of going about it. The list will only continue to get more involved with the other stuff they've got listed in flight (canopy module, mmwave switch, scene controller).

TheJulianJES (approved):
LGTM. Seems to match [Z2M code](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/inovelli.ts), although it seemingly doesn't have the `leading_or_trailing_edge` attribute yet.

PR #2806, opened 2023-12-06
Fix Xiaomi H1 alternative signature

dumpfheimer:
## Proposed change

It seems like this commit kicked a device signature for Xiaomi H1 switches: https://github.com/zigpy/zha-device-handlers/commit/4f5abdd2b7e5d636d5bce1a7173f1544a506493c
This PR adds it back.

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/switch_h1_double.py
+++ b/zhaquirks/xiaomi/aqara/switch_h1_double.py
@@ -78,8 +78,8 @@ class AqaraH1DoubleRockerSwitchWithNeutralAlt(XiaomiOpple2ButtonSwitchBase):
     """Aqara H1 Double Rocker Switch (with neutral) alternative signature."""
 
     signature = {
+        MODELS_INFO: [(LUMI, "lumi.switch.n2aeu1")],
         ENDPOINTS: {
-            MODELS_INFO: [(LUMI, "lumi.switch.n2aeu1")],
             # input_clusters=[0, 2, 3, 4, 5, 6, 18, 64704], output_clusters=[10, 25]
             1: {
                 PROFILE_ID: zha.PROFILE_ID,
--- a/zhaquirks/xiaomi/aqara/switch_h1_double.py
+++ b/zhaquirks/xiaomi/aqara/switch_h1_double.py
@@ -2,6 +2,7 @@
 from zigpy.profiles import zgp, zha
 from zigpy.zcl.clusters.general import (
     Alarms,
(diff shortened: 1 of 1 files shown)


TheJulianJES:
Hmm, unless I'm missing something, I don't see where that quirk was removed. Maybe it was using a quirk without `MODELS_INFO` before and the `MODELS_INFO` is missing `lumi.switch.n2aeu1` now.
Do you still know which (exact) quirk it was using before? (EDIT: Likely `XiaomiOpple2ButtonSwitchFace1`)

I'll have a look.

Please also attach device diagnostics and/or the device signature. It can be downloaded from the device page.
(Without the quirk applied!)

TheJulianJES:
Looks like it matches the `XiaomiOpple2ButtonSwitchFace1` signature too:
https://github.com/zigpy/zha-device-handlers/blob/aeca76da0ea81ecba5b17e0f7314620cd686faae/zhaquirks/xiaomi/aqara/opple_switch.py#L277-L281

So, `(LUMI, "lumi.switch.n2aeu1")` just needs to be added to the `MODELS_INFO` there.
Maybe we should move it to the "H1 double" file too then? Although the `lumi.switch.b2naus01` isn't a H1 switch from what I can see.

TheJulianJES:
Okay, it seems like (at least the US models of) Aqara switches sometimes report a different signature when paired. This might explain some things.
In quirks, we need to match all signatures then.

I'll fix https://github.com/zigpy/zha-device-handlers/issues/2824 first and then see how many signatures also match the H1 models.

TheJulianJES:
Superseded by https://github.com/zigpy/zha-device-handlers/pull/2826/
I'll see if we can get that into a patch release of HA.

Thanks for making me aware btw!

dumpfheimer:
Great, thank you!

In addition to this, I made some configuration entities for the H1 and was wondering if there was a better way to match all of the devices using a similar configuration, to prevent a HA PR every time a device is added:
https://github.com/home-assistant/core/commit/a698090e09ddff34abcc008a67bf6baf7aab7d4d

right now the matching works by using a quirk ID, eg
```
@CONFIG_DIAGNOSTIC_MATCH(
    cluster_handler_names="opple_cluster", quirk_ids="zhaquirks.xiaomi.aqara.opple_switch.XiaomiOpple2ButtonSwitchFace1"
)
```

Could you point me in the right direction for how I could match these configuration entities to catch all devices?

PR #2823, opened 2023-12-09
Add `mode` attribute to Aqara T1 relay

dmulcahey:
## Proposed change

Add the mode attribute to the aqara T1 relay

TODO:
- [ ] change mode to an enum?
- [ ] add config entities to ZHA

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/switch_t1.py
+++ b/zhaquirks/xiaomi/aqara/switch_t1.py
@@ -43,6 +43,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
     """Xiaomi Aqara T1 relay cluster."""
 
     attributes = {
+        0x0009: ("mode", t.uint8_t, True),
         0x000A: ("switch_type", t.uint8_t, True),
         0x0201: ("power_outage_memory", t.Bool, True),
         0x0203: ("do_not_disturb", t.Bool, True),


TheJulianJES:
> change mode to an enum?

This `mode`/`0x0009` attribute is only `0` ("ZCL mode") or `1` ("Aqara/Xiaomi hub mode"), so not like the `mode` attribute from your T2 relay PR (which should probably be renamed to `switch_mode` there).
Since this attribute isn't really meant to be user-accessible (ideally, it should auto-set to `1`), I don't think we need to change it to an enum. If we do, we should probably change it in all quirks though.

It might make sense to change `switch_type` to an enum though (maybe that's also what you meant?). Just like done in `opple_switch.py` (scroll in code field below):
https://github.com/zigpy/zha-device-handlers/blob/529112195dcdc3c8108df05e3a838478ee75ac6e/zhaquirks/xiaomi/aqara/opple_switch.py#L63-L106
Maybe that implementation/enum can be used/shared if it's the same for this T1 relay.

TheJulianJES:
Also, regarding "auto-setting" `mode`:

Although `mode` is likely only needed for old firmware versions of this relay, we could probably try to set it to `1` every time (when binding the cluster), since Z2M does it too. Like this:
https://github.com/zigpy/zha-device-handlers/blob/aeca76da0ea81ecba5b17e0f7314620cd686faae/zhaquirks/xiaomi/aqara/opple_remote.py#L142
https://github.com/zigpy/zha-device-handlers/blob/aeca76da0ea81ecba5b17e0f7314620cd686faae/zhaquirks/xiaomi/aqara/opple_remote.py#L149-L153

If that request errors out (when changing from `0` to `1`), this might be needed instead [zhaquirks/xiaomi/aqara/plug_eu.py#L65-L76](https://github.com/zigpy/zha-device-handlers/blob/aeca76da0ea81ecba5b17e0f7314620cd686faae/zhaquirks/xiaomi/aqara/plug_eu.py#L65-L76), as it seems like that particular device briefly reboots (or something) and doesn't send back a response.

dmulcahey:
> Also, regarding "auto-setting" `mode`:
> 
> Although `mode` is likely only needed for old firmware versions of this relay, we could probably try to set it to `1` every time (when binding the cluster), since Z2M does it too. Like this:
> 
> https://github.com/zigpy/zha-device-handlers/blob/aeca76da0ea81ecba5b17e0f7314620cd686faae/zhaquirks/xiaomi/aqara/opple_remote.py#L142
> 
> 
> https://github.com/zigpy/zha-device-handlers/blob/aeca76da0ea81ecba5b17e0f7314620cd686faae/zhaquirks/xiaomi/aqara/opple_remote.py#L149-L153
> 
> If that request errors out (when changing from `0` to `1`), this might be needed instead [zhaquirks/xiaomi/aqara/plug_eu.py#L65-L76](https://github.com/zigpy/zha-device-handlers/blob/aeca76da0ea81ecba5b17e0f7314620cd686faae/zhaquirks/xiaomi/aqara/plug_eu.py#L65-L76), as it seems like that particular device briefly reboots (or something) and doesn't send back a response.

I debated doing this initially... the quirks work for multiple models and multiple fw differences... I didn't want to add this without more information. The goal was just to get the attribute there so we can set it from the UI and collect more info from users. thoughts?

TheJulianJES:
> I debated doing this initially... the quirks work for multiple models and multiple fw differences... I didn't want to add this without more information. The goal was just to get the attribute there so we can set it from the UI and collect more info from users. thoughts?

Yeah, let's not set it automatically for now then. I can't recall seeing any issues regarding this, so maybe everyone just uses a new firmware version on these.

TheJulianJES (approved):

PR #2829, opened 2023-12-10
Improve support for sinope Calypso and Eco Mode

claudegel:
## Proposed change

- Code cleanup for calypso controller.
- Add support for Eco Mode on thermostats, Eco Sinopé
- Add  many attributes for sensor and light in manufacturer cluster

## Checklist

### DIFF
--- a/zhaquirks/sinope/switch.py
+++ b/zhaquirks/sinope/switch.py
@@ -97,14 +97,6 @@ class SinopeManufacturerCluster(CustomCluster):
         Active = 0x00
         Off = 0x01
 
-    class TankSize(t.enum8):
-        """tank_size values."""
-
-        Gal_40 = 0x01
-        Gal_50 = 0x02
-        Gal_60 = 0x03
-        Gal_80 = 0x04
-
     class FlowDuration(t.uint32_t):
         """Abnormal flow duration."""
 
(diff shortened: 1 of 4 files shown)


zhaquirks/sinope/switch.py
```diff
@@ -127,12 +119,14 @@ class FlowDuration(t.uint32_t):
         0x0003: ("firmware_number", t.uint16_t, True),
         0x0004: ("firmware_version", t.CharacterString, True),
         0x0010: ("outdoor_temp", t.int16s, True),
-        0x0013: ("tank_size", TankSize, True),
+        0x0013: ("unknown_attr_1", t.enum8, True),
```
TheJulianJES, line 122: Just to confirm, this attribute wasn't actually the `TankSize`, right?

claudegel:
It was an error as the TankSize finally is not in the attribute but only in Neviweb cloud. It is not used by the Calypso. I've left the attribute there as other could be able to find out the usage.
I have many other attribute that are still unknown. is it better to leave them out until I find the usage or put them in the quirks so other have access to play with them.

claudegel:
Another thing, I'm thinking of splitting the 4 files into specific file for each device model. Could it be better ? It's a big job to separate everything but if it is better I can try to split them.

TheJulianJES:
> I'm thinking of splitting the 4 files into specific file for each device model.

Generally yeah, it's nicer to have different device types separated. For example, `switch.py` has smart plugs, valve controllers, ... from what I can see. IMO, it's still *fine* for these, as there aren't too many models, but we're open for a "cleanup PR" if you want.

TheJulianJES (approved):
LGTM

claudegel:
I have many attributes found in manufacturer cluster but still need to find out the use. Generally I don't add them to the quirks but could it be better to add them as unknown_attr_xx so other can try to understand the usage for them so we can put a name on them ?

TheJulianJES:
> many attributes found in manufacturer cluster but still need to find out the use. Generally I don't add them to the quirks but could it be better to add them as unknown_attr_xx

I don't have a real preference. If they're likely not needed, I wouldn't add them, but it should also be fine if you do add them.

PR #2885, opened 2024-01-05
Update tests for zigpy TSN changes

TheJulianJES:
## Proposed change
This updates zha-quirks tests to account for the [per-device TSNs](https://github.com/zigpy/zigpy/pull/1309) in zigpy [0.60.3](https://github.com/zigpy/zigpy/releases/tag/0.60.3) and later.
The tests do not pass without this change when using zigpy 0.60.3 or later.

The minimum required zigpy version is also bumped to 0.60.3, although only the tests "need" this.

## Additional information
cc @puddly 

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ readme = "README.md"
 license = {text = "Apache License Version 2.0"}
 requires-python = ">=3.8"
 dependencies = [
-    "zigpy>=0.56.0",
+    "zigpy>=0.60.3",
 ]
 
 [tool.setuptools.packages.find]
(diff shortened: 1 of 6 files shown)


TheJulianJES:
cc @dmulcahey
The zigpy per-device [TSN PR](https://github.com/zigpy/zigpy/pull/1309) from some weeks ago broke quirks tests when using a newer zigpy version (>=0.60.3).

I think the GitHub CI currently only passes because it uses an older version (cached environment).
I've also bumped the minimum required zigpy version in this PR. It's only really required for tests (and forcing GH Actions to not use the old cache I guess), but bumping that should be fine, right?

puddly (approved):

PR #2888, opened 2024-01-05
Add Tuya `QUERY_DATA` command

tschiex:
## Proposed change

Some tuya devices require a query_data to be responsive and provide data.
As suggested by tjj, this just provides this ability to quirks.

## Additional information

Contributes to fixing https://github.com/zigpy/zha-device-handlers/issues/2565

EDIT:
Follow-up PR that adds the "Tuya query data spell":
- https://github.com/zigpy/zha-device-handlers/pull/2940

## Checklist

Couldn't find a significant test.

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -37,6 +37,7 @@ TUYA_CLUSTER_1888_ID = 0x1888
 TUYA_SET_DATA = 0x00
 TUYA_GET_DATA = 0x01
 TUYA_SET_DATA_RESPONSE = 0x02
+TUYA_QUERY_DATA = 0x03
 TUYA_SEND_DATA = 0x04
 TUYA_ACTIVE_STATUS_RPT = 0x06
 TUYA_SET_TIME = 0x24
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/__init__.py
```diff
@@ -1488,6 +1489,9 @@ class TuyaNewManufCluster(CustomCluster):
     ep_attribute: str = "tuya_manufacturer"
 
     server_commands = {
+        TUYA_QUERY_DATA: foundation.ZCLCommandDef(
+            "query_data", {}, False, is_manufacturer_specific=True
```
javicalle, line 1493: There's no `data` structure for this command?
Can you point us to any command message or use case?

tschiex: There are Tuya devices that will not start nor send information w/o DATA_QUERY messages. For these, a DATA_QUERY package has to be sent on startup and also later for polling: no information is sent except if a DATA_QUERY packet is sent. 

See the forthcoming [zha-quirk](https://gist.github.com/tschiex/28237907a4f0c569015b1869c8ffe66a) for an example of usage in ZHA. I followed the suggestion of TheJulianJES here. 

See also the corresponding code in Z2MQTT [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/a60d4f6e3b709160a045e836200910a1ab8660dc/src/lib/tuya.ts#L77)

TheJulianJES (approved):
LGTM, thanks!

PR #2869, opened 2023-12-28
Add Sonoff SNZB-06P presence sensor support

theorlangur:
## Proposed change

I was working on a quirk for SNZB-06P, but got stuck.
Decided to make a PR, although right now it's not at the state I'd like it to be. Maybe I'll get some pointers here

## Additional information

When finished, fixes #2702

What I would like is to do is:
* remove motion entity (it's non-functional) (done)
* add access to occupancy timeout (done, kinda, via virtual AnalogOutput cluster on the virtual endpoint, but it works)
* add access (read-only) to the latest captured illuminance state (dark/bright)
* add access to a sensitivity setting (in case of this device controlled by the "ultrasonic_u_to_o_threshold" standard Zigbee attribute and allows values 1,2,3 - Low,Medium, High)

Access to occupancy timeout seems to work but looks kinda hacky. Is there a better way? We're talking about a standard attribute "ultrasonic_o_to_u_delay" for OccupancySensing cluster with a manufacturer-specific limitations (minimum value that may be set is 15)

A read-only access to the last captured illuminance state was made by creating a BinaryInput-derived cluster. It created a 'binary input' entity in HA and it shows the right thing, but I wanted to provide an adequate name for it and didn't find a proper way.

How to describe a sensitivity in the quirk so that ZHA would "automagically" create an enum-based selector entity for it?

Current state looks like on the screenshot:
Screenshot_20231228-155404

'Binary input' is supposed to be 'Last illuminance state', but I haven't found a way to specify that in the quirk so that it would be taken over to HA's entity.

I'm opened for suggestions of better ways to do any of it.

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [x] Tests have been added to verify that the new code works

Issue #2702, opened 2023-11-02
[Device Support Request] Sonoff SNZB-06P fixes and improvements


blakadder:
### Problem description

Sonoff SNZB-06P creates two entities out of which only the Occupancy sensor updates correctly. Motion sensor updates rarely and doesn't reflect the actual state.

Environment light state is reported only on detected presence (similar to old Xiaomi sensor) on cluster FC11/2001 where 0 is darker and 1 brighter.

### Solution description

Remove motion sensor in the quirk and add a light state sensor from FC11/2001 cluster

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.FullFunctionDevice|MainsPowered|RxOnWhenIdle|AllocateAddress: 142>, manufacturer_code=4742, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0107",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0406",
        "0x0500",
        "0xfc11",
        "0xfc57"
      ],
      "output_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "SONOFF",
  "model": "SNZB-06P",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
{
  "home_assistant": {
    "installation_type": "Home Assistant Container",
    "version": "2023.11.0",
    "dev": false,
    "hassio": false,
    "virtualenv": false,
    "python_version": "3.11.6",
    "docker": true,
    "arch": "x86_64",
[...]

PR #2895, opened 2024-01-09
Add Sonoff SNZB-06P presence sensor support

peterkh:
## Proposed change

Improving the support for the Sonoff SNZB-06P

Main changes by this quirk:
* remove redundant IaS Zone cluster as it doesn't provide anything useful that is not already provided from the Occupancy cluster and removes a second entity being created.
* Adds support for Sonoff specific cluster which has the illumination state (Light or Dark) when presence was last detected. 

## Additional information

Thanks to @theorlangur  and their work on https://github.com/zigpy/zha-device-handlers/pull/2869.

When combined with https://github.com/home-assistant/core/pull/107379, solves https://github.com/zigpy/zha-device-handlers/issues/2702

## Checklist

### DIFF
--- a/zhaquirks/sonoff/snzb06p.py
+++ b/zhaquirks/sonoff/snzb06p.py
@@ -29,9 +29,8 @@ class IlluminationStatus(t.enum8):
 class SonoffFC11Cluster(CustomCluster):
     """Sonoff manufacture specific cluster that provides illuminance."""
 
-    name = "Sonoff Manufacture Specific Cluster at 0xFC11"
     cluster_id = SONOFF_CLUSTER_FC11_ID
-    ep_attribute = "sonoff_manufacturer_specific_FC11"
+    ep_attribute = "sonoff_manufacturer"
     attributes = {
         ATTR_SONOFF_ILLUMINATION_STATUS: ("last_illumination_state", IlluminationStatus)
     }
--- a/zhaquirks/sonoff/snzb06p.py
+++ b/zhaquirks/sonoff/snzb06p.py
@@ -1,0 +1,87 @@
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
+from zigpy.zcl.clusters.general import Basic, Identify, Ota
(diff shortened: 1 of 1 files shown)


peterkh:
@TheJulianJES just wanted to check if there is anything more I need to do with this quirk before it can be reviewed? If its just in the queue, all good and apologies for the ping :)

TheJulianJES:
It's still on my list to review. The changes look good from a quick glance though.

zhaquirks/sonoff/snzb06p.py
```diff
@@ -0,0 +1,88 @@
...
+    Dark = 0x00
+    Light = 0x01
+
+
+class SonoffFC11Cluster(CustomCluster):
+    """Sonoff manufacture specific cluster that provides illuminance."""
+
+    name = "Sonoff Manufacture Specific Cluster at 0xFC11"
+    cluster_id = SONOFF_CLUSTER_FC11_ID
+    ep_attribute = "sonoff_manufacturer_specific_FC11"
```
TheJulianJES: I'll need to think about the `name` and `ep_attribute` later.
We technically don't need the `name` for this. They're both a bit long maybe.

I'd rename the `ep_attribute` to something like `sonoff_manufacturer` or `sonoff_cluster`.

zhaquirks/sonoff/snzb06p.py
```diff
@@ -0,0 +1,88 @@
...
+    attributes = {
+        ATTR_SONOFF_ILLUMINATION_STATUS: ("last_illumination_state", IlluminationStatus)
```
TheJulianJES, line 35: We could also add an `Illuminance` cluster extending `LocalDataCluster` and then update that cluster/attribute cache from this cluster using something like this: [zhaquirks/xiaomi/aqara/driver_curtain_e1.py#L56-L64](https://github.com/zigpy/zha-device-handlers/blob/a014ef5969045bf3e3be8664d2778c7a5ed8c2a5/zhaquirks/xiaomi/aqara/driver_curtain_e1.py#L56-L64)

Tbh, I'm not sure if it's better to have a custom entity in this case or fake the cluster and use the "default" one.

peterkh: Maybe I miss understand what you mean, but wouldn't that be a worse UX as the last illumination state from the device is only ever "light" and "dark" and we'd need to choose arbitrary values in lx for what those two states are?

TheJulianJES: Yep, probably. Just thought about it. I'll have a closer look at your ZHA PR soon to see how you implemented it right now.

TheJulianJES (approved):
LGTM, thanks!

PR #2896, opened 2024-01-09
Update Tuya RCBO quirk power factor calculation for home-assistant/core#107641

jeverley:
## Proposed change

Amend power_factor calculation to respect Zigbee cluster specification - required for release of zha fix:

https://github.com/home-assistant/core/pull/107641

## Additional information

This updates the power factor calculation to multiply by 100 rather than 1000, this is because the ElectricalMeasurement attribute is supposed to expose values between -100 and 100 (see the pre-requisite PR 107641 for details).

The current code exposes values between 0 and 1000, which is subsequently divided by 10 due to the behaviour of the zha cluster sensor corrected in https://github.com/home-assistant/core/pull/107641

If released without https://github.com/home-assistant/core/pull/107641 the value reported will be an order of magnitude too low.

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_rcbo.py
+++ b/zhaquirks/tuya/ts0601_rcbo.py
@@ -251,7 +251,7 @@ class TuyaRCBOElectricalMeasurement(ElectricalMeasurement, TuyaAttributesCluster
                 power_factor = value / apparent_power * 100
                 if power_factor > 100:
                     power_factor = 100
-                super().update_attribute("power_factor", int(power_factor))
+                super().update_attribute("power_factor", int(round(power_factor)))
 
 
 class TuyaRCBODeviceTemperature(DeviceTemperature, TuyaAttributesCluster):
--- a/zhaquirks/tuya/ts0601_rcbo.py
+++ b/zhaquirks/tuya/ts0601_rcbo.py
@@ -248,9 +248,9 @@ class TuyaRCBOElectricalMeasurement(ElectricalMeasurement, TuyaAttributesCluster
         if attr_name == "active_power":
             apparent_power = self.get("apparent_power")
             if apparent_power:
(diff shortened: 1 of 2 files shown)


TheJulianJES:
The Tuya quirks tests probably need to be corrected to expect the new value then. See:
- https://github.com/zigpy/zha-device-handlers/actions/runs/7553664099/job/20564936540?pr=2896
- https://github.com/zigpy/zha-device-handlers/pull/2896/files (test file should be annotated on web)

(and ideally, the HA PR needs to have tests for the added code)

dmulcahey:
@TheJulianJES once you are good with this you can merge this and the corresponding HA PR: https://github.com/home-assistant/core/pull/107641

zhaquirks/tuya/ts0601_rcbo.py
```diff
@@ -248,10 +248,10 @@ def update_attribute(self, attr_name: str, value: Any) -> None:
...
             apparent_power = self.get("apparent_power")
             if apparent_power:
-                power_factor = value / apparent_power * 1000
-                if power_factor > 1000:
-                    power_factor = 1000
-                super().update_attribute("power_factor", int(power_factor))
+                power_factor = value / apparent_power * 100
+                if power_factor > 100:
+                    power_factor = 100
+                super().update_attribute("power_factor", int(round(power_factor)))
```
TheJulianJES: Minor, but I think just `round(power_factor)` should be enough, no `int(x)` needed in that case.

TheJulianJES (approved):
LGTM, thanks!

zhaquirks/tuya/ts0601_rcbo.py
jeverley: Good point, thanks for highlighting that!

TheJulianJES (approved):

PR #2913, opened 2024-01-16
Add Zbeacon DS01 quirk removing `PollControl` cluster

Caius-Bonus:
## Proposed change

The Zbeacon DS01 (no idea who made this, but it's a cheap chinese door sensor that runs on 2xAAA and is long rectangular in shape).

Add quirk for Zbeacon DS01, which removes the PollControl cluster, which is incorrectly implemented by the device I think.

## Additional information

The device will send "PollControl:checkin()" on PollControl cluster, but doesn't respond when checkin_response is sent after that from the coordinator. 
This results in error messages every 30-60 minutes with the message that the device didn't respond.

This is easily avoided by just disabling the PollControl Cluster.

## Checklist


MattWestb:
With witch coordinator do you have testing and also is you having one sniffing of the pairing that we can taking one look on.
Interesting to see if all flags is set OK also if the end device time out is being OK replayed from the devices parent then joining
TI coordinators was having all pull control disabled but was fixed for one half year ago and some is not doing it OK.

Edit: I have not seen any tuya end device using  pull control and checkins only good working ones real Zigbee 3 that is doing it well is IKEA with current production firmware.

zhaquirks/tuya/doorsensor.py
```diff
@@ -0,0 +1,63 @@
...
+                ],
+                OUTPUT_CLUSTERS: [Ota.cluster_id],
+            }
+        },
+    }
+
+    replacement = {
+        ENDPOINTS: {
+            1: {
+                INPUT_CLUSTERS: [Basic, PowerConfiguration, Identify, IasZone],
```
TheJulianJES: I'd use this formatting and also leave `.cluster_id` at the end:
```suggestion
                INPUT_CLUSTERS: [
                    Basic.cluster_id,
                    PowerConfiguration.cluster_id,
                    Identify.cluster_id,
                    IasZone.cluster_id,
                ],
```

zhaquirks/tuya/doorsensor.py
```diff
@@ -0,0 +1,63 @@
...
+                OUTPUT_CLUSTERS: [
+                    Ota,
```
TheJulianJES:
```suggestion
                    Ota.cluster_id,
```
(and maybe also change that formatting to be the same as the one from `signature`)

zhaquirks/tuya/doorsensor.py
```diff
@@ -0,0 +1,63 @@
+"""Doorsensors"""
+
+import zigpy.profiles.zha as zha_p
```
TheJulianJES: Import this as just `zha` to be more consistent with other quirks.

TheJulianJES (commented):
Left some small comments.

(Note for the next time, we also have an experimental "[quirk-generator](https://github.com/zigpy/quirk-generator)" to generate a quirk that just matches a device based on the provided signature/diagnostics file.)

zhaquirks/tuya/doorsensor.py
Caius-Bonus: Done, Thanks

Caius-Bonus:
@TheJulianJES Great that there is a tool to generate quirks now, nice job.

@MattWestb My coordinator is a Conbee II with the latest firmware (15-May-2022). The testing is done with 2 devices and I have consistent errors with both every ~45 minutes without this quirk (even after re-pairing). I could send a log, but it'll take a little while for me to get around to it.

About the classification: I wasn't sure where to put it, because it is unclear to me who made it. The sellers advertise it as being TuYa. I could move it to Elivco or IHseno IH-MC01 (both seem identical in the pictures, but mine are from Elivco).

MattWestb:
Then its not one TI-Coordinator problem then one tuya is trying implanting pull control and more then likely we is getting more device from them with the same problem very sooooooon.
If being one own brand its not so importing for my then its not looking like normal tuya IDs like LIDL and other is using.

Caius-Bonus:
I moved it to its own manufacturer for now; I just put it under Zbeacon. It is probably using a TuYa ZTU module, but that doesn't mean it's programmed by TuYa. I cannot find anything about Zbeacon, Elivco or IHsense. Although Elivco seems to yield some results with smart plugs.

TheJulianJES (approved):
LGTM.
I think it's fine to have it separated from Tuya, as it doesn't seem to be from Tuya at first glance at least. We can always move it later of course.

PR #2916, opened 2024-01-17
Add manufacturer cluster for Sonoff SNZB-02D

elupus:
## Proposed change

Add description of sonoff custom cluster controlling the display information on the temperature meter.

## Additional information

The devices requires manufacture code to not be set when reading/writing attributes. Unsetting this is currently not easily done without overriding core functions in zigpy. Here it's done by overriding the creation of request.

## Checklist

### DIFF
--- a/zhaquirks/sonoff/snzb02d.py
+++ b/zhaquirks/sonoff/snzb02d.py
@@ -1,12 +1,8 @@
 """Module for Philips quirks implementations."""
 from __future__ import annotations
 
-from typing import Any, Final
-
 from zigpy.profiles import zha
-from zigpy.quirks import CustomCluster, CustomDevice
-import zigpy.types as t
-from zigpy.types.basic import enum_factory
+from zigpy.quirks import CustomDevice
 from zigpy.zcl.clusters.general import (
     Basic,
     Identify,
@@ -15,13 +11,6 @@ from zigpy.zcl.clusters.general import (
     PowerConfiguration,
 )
 from zigpy.zcl.clusters.measurement import RelativeHumidity, TemperatureMeasurement
-from zigpy.zcl.foundation import (
-    BaseAttributeDefs,
-    Direction,
-    GeneralCommand,
-    ZCLAttributeDef,
-    ZCLHeader,
-)
 
 from zhaquirks.const import (
     DEVICE_TYPE,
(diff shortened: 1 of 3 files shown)


zhaquirks/sonoff/snzb02d.py
```diff
@@ -0,0 +1,147 @@
...
+        schema: dict | t.Struct,
+        manufacturer: int | None = None,
+        tsn: int | None = None,
+        disable_default_response: bool,
+        direction: Direction,
+        # Schema args and kwargs
+        args: tuple[Any, ...],
+        kwargs: Any,
+    ) -> tuple[ZCLHeader, bytes]:
+        """Override all request to disable manufacturer."""
```
elupus: I tried higher level overrides, like skipping the CustomCluster. But seems HA overrides the possibility of using None as signifier of letting the customcluster decide on use here: https://github.com/home-assistant/core/blob/10014838ef32ee1607e3b959c43821a887c5e3ea/homeassistant/components/zha/websocket_api.py#L823

dmulcahey: The linked code only supports the manage device dialog IIRC and it’s a sensible default. 99% of the time you need it for mfg specific things and common users aren’t going to be aware of that. 

I think what you are doing here is the only way to achieve this currently.

dmulcahey:
Are reading and writing attrs the only commands that fail if the mfg id is set? If so, should we limit this to those 2 command ids?

zhaquirks/sonoff/snzb02d.py
elupus: The CustomCluster has this default too. Would make more sense for backend to handle it. Zigpy or Quirks.

Also as it stands now its impossible to call with no manufacturer code using that dialog.

elupus:
> Are reading and writing attrs the only commands that fail if the mfg id is set? If so, should we limit this to those 2 command ids?

Those are the only commands i know how to test. There are no other things on this cluster currently.

elupus:
> > Are reading and writing attrs the only commands that fail if the mfg id is set? If so, should we limit this to those 2 command ids?
> 
> Those are the only commands i know how to test. There are no other things on this cluster currently.

Tested binding to the cluster with a forced zero too:

```
service: zha_toolkit.conf_report
data:
  ieee: sensor.temperatur_kontor
  endpoint: 1
  cluster: 0xFC11
  attribute: 7
  min_interval: 60
  max_interval: 300
  reportable_change: 1
  manf: 0
```

```
zha_toolkit_version: v1.1.8
zigpy_version: 0.60.4
zigpy_rf_version: 0.22.4
ieee_org: sensor.temperatur_kontor
ieee: 34:10:f4:ff:fe:6c:90:03
command: conf_report
command_data: null
start_time: "2024-01-17T21:08:49.852729+00:00"
errors: []
params:
  endpoint_id: 1
  cluster_id: 64529
  attr_id: 7
  min_interval: 60
  max_interval: 300
  reportable_change: 1
  dir: 0
  manf: 18242
  tries: 1
  expect_reply: true
  args: []
  read_before_write: true
  read_after_write: true
success: true
result_conf:
  - - status: 0
      direction: null
      attrid: null
```

zhaquirks/sonoff/snzb02d.py
```diff
@@ -0,0 +1,77 @@
+"""Module for Philips quirks implementations."""
```
SuperSandro2000, line 1:
```suggestion
"""Module for Sonoff quirks implementations."""
```

elupus:
Superseeded by https://github.com/zigpy/zha-device-handlers/pull/3912

PR #1705, opened 2022-08-23
Add Schneider Electric devices

axellebot:
Todo :
- Switch
  - [ ] CH2AX/SWITCH/1 #1714
  - [ ] CH10AX/SWITCH/1 #1696
  - [ ] FLS/AIRLINK/4 #1623 
- Shutter
  - [ ] NHPB/SHUTTER/1 #1685

PR #2928, opened 2024-01-21
Add quirk for Schneider Electric 1GANG/SHUTTER/1 (MEG5113-0300)

cspurk:
## Proposed change

Add a quirk for Schneider Electric 1GANG/SHUTTER/1 (MEG5113-0300) to

* correct the lift percentage (which is inverted).
* add some manufacturer-specific features.

## Additional information

This PR is based on @axellebot’s work in #1705. As the latter is incomplete, I have removed the code of which I don’t know whether it works.

I have successfully tested the quirk with a real device connected to Home Assistant via ZHA.

## Checklist


TheJulianJES (commented):
The changes mostly seem fine. However, I don't know if we want to ship a 6000 line readme file in the source.

Also, the cover implementation in ZHA is currently being reworked a bit:
- https://github.com/home-assistant/core/pull/108238

That might land in HA Core 2024.2.0 (betas). Since these quirks also invert the covers somewhat, it would be nice if they could be tested (with and without(?) quirk) whenever that PR is merged and released to HA Core.

cspurk:
Thanks for the positive review, @TheJulianJES!

Fair enough, I’ve removed the README. Is there maybe a better place to put it? I suppose @axellebot’s idea behind the file was to have something similar to Zigbee2MQTT’s devices DB, just for zigpy. I thought about a wiki, but zha-device-handlers doesn’t have one.

Regarding home-assistant/core#108238: I can certainly test my device with/without the quirk once this has landed in a HA release. I doubt, though, that the new inversion logic will help here, as it mostly seems to be based on reversing the motor direction. I had tested that manually before implementing the quirk, and it just resulted in a swapping of the up/down push buttons – interestingly both on the wall and in the HA GUI. The opened/closed states in HA were still wrong.

MattWestb:
I think device / vendor things is the best place in the Wiki https://github.com/zigpy/zigpy/wiki.

TheJulianJES:
> I thought about a wiki, but zha-device-handlers doesn’t have one.

I think putting quirks-related stuff in a quirks wiki would be good.

zhaquirks/schneiderelectric/devices/shutters.py
```diff
@@ -0,0 +1,112 @@
...
+                DEVICE_TYPE: zha.DeviceType.WINDOW_COVERING_DEVICE,  # 0x0202
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,  # 0x0000
+                    Identify.cluster_id,  # 0x0003
+                    Groups.cluster_id,  # 0x0004
+                    Scenes.cluster_id,  # 0x0005
+                    WindowCovering.cluster_id,  # 0x0102
+                    Diagnostic.cluster_id,  # 0x0B05
+                ],
+                OUTPUT_CLUSTERS: [Ota.cluster_id],  # 0x0019
```
TheJulianJES: To stay consistent, can you remove the comments with the cluster IDs "everywhere"?
(not the `SimpleDescriptor` comments above though)

TheJulianJES:
Also, to structure it like the rest of our repo, I'd get rid of the `devices` and `clusters` folders/package.
Just have `zhaquirks/schneiderelectric` and put the shared code/clusters/constants in `__init__.py`.
The `shutters.py` can go into that folder/package too.

cspurk:
@MattWestb, @TheJulianJES, thanks for your feedback regarding the wiki. Should I maybe just create an initial structure for a zigpy/zha-device-handlers wiki?

@TheJulianJES, thanks again for the review. I’ve made the changes you requested.

TheJulianJES:
> Should I maybe just create an initial structure for a zigpy/zha-device-handlers wiki?

Yup, I think that's fine.

cspurk:
> > Should I maybe just create an initial structure for a zigpy/zha-device-handlers wiki?
> 
> Yup, I think that's fine.

Ok, I’ve created a [wiki home page](https://github.com/zigpy/zha-device-handlers/wiki) and added a Schneider-specific page with the deleted “README” (mostly as in a4d596c8e70fb951478c935ca93d3b48789d16da).

zhaquirks/schneiderelectric/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+
+from zigpy import types as t
+from zigpy.quirks import CustomCluster
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.closures import WindowCovering
+from zigpy.zcl.clusters.general import Basic
+from zigpy.zcl.foundation import ZCLAttributeDef
+
+SE_MANUF_NAME = "Schneider Electric"
+SE_MANUF_ID = 0x4190
```
dflemstr: I think this is the wrong manufacturer ID.  It should be `0x105e` which is `4190` in decimal (note the lack of `0x` prefix)

cspurk: Ha, good catch! I’ve corrected that.

cspurk:
@TheJulianJES, as promised, I’ve tested HA 2024.2.0 (and hence home-assistant/core#108238) with and without the quirk from this PR:

- Without the quirk, the lift percentage is still inverted (just like before).
- The quirk continues to fix this (just like before).
- I can see the new invert switch in the GUI (with and without the quirk), however, it behaves like I described in a previous comment (and which is not useful for this Schneider Electric device).

zhaquirks/schneiderelectric/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+
+    attributes: Dict[int, ZCLAttributeDef] = Basic.attributes.copy()
+
+    attributes.update(
+        {
+            0xE001: (
+                "se_sw_build_id",
+                t.CharacterString,
+                True,
+            ),  # value: "002.004.016 R
```
TheJulianJES: Ideally, we already want to use the new zigpy-style `AttributeDefs` definitions, like this:

cspurk: Ok, I’ve updated the code.

TheJulianJES (approved):
LGTM, thanks!

PR #2939, opened 2024-01-26
Fix switch events for Aqara T2 relay

trapacska:
## Proposed change

1. Fixed event sending, it has 4 args passed in but could only take 3 so the event was not fired.
2. There is an event for each endpoint(1 and 2) when the S1-COM and S2-COM pins are shorted but the original implementation did not separate the events so both S1 and S2 was emitting the same event.

Fixes my issue (partially): #2934

image
image

## Checklist

### DIFF
--- a/tests/test_xiaomi.py
+++ b/tests/test_xiaomi.py
@@ -1639,9 +1639,7 @@ async def test_xiaomi_e1_roller_commands_2(zigpy_device_from_quirk, command, val
     )
 
 
-@pytest.mark.parametrize(
-    "endpoint", [(1), (2)]
-)
+@pytest.mark.parametrize("endpoint", [(1), (2)])
 async def test_aqara_t2_relay(zigpy_device_from_quirk, endpoint):
     """Test Aqara T2 relay."""
 
--- a/tests/test_xiaomi.py
+++ b/tests/test_xiaomi.py
@@ -1640,45 +1640,19 @@ async def test_xiaomi_e1_roller_commands_2(zigpy_device_from_quirk, command, val
 
 
 @pytest.mark.parametrize(
-    "quirk",
-    (zhaquirks.xiaomi.aqara.switch_acn047.AqaraT2Relay,),
+    "endpoint", [(1), (2)]
 )
-async def test_aqara_t2_relay_endpoint_1(zigpy_device_from_quirk, quirk):
-    """Test Aqara T2 relay endpoint 1."""
+async def test_aqara_t2_relay(zigpy_device_from_quirk, endpoint):
+    """Test Aqara T2 relay."""
 
-    device = zigpy_device_from_quirk(quirk)
-    mi_cluster = device.endpoints[1].multistate_input
+    device = zigpy_device_from_quirk(zhaquirks.xiaomi.aqara.switch_acn047.AqaraT2Relay)
+    mi_cluster = device.endpoints[endpoint].multistate_input
     mi_listener = ClusterListener(mi_cluster)
 
     mi_cluster.update_attribute(MultistateInput.AttributeDefs.present_value.id, 1)
     assert len(mi_listener.attribute_updates) == 1
     assert mi_listener.attribute_updates[0][0] == 0
-    assert mi_listener.attribute_updates[0][1] == "switch_1"
-
-    mi_cluster.update_attribute(MultistateInput.AttributeDefs.state_text.id, "foo")
-    assert len(mi_listener.attribute_updates) == 2
-    assert (
-        mi_listener.attribute_updates[1][0]
-        == MultistateInput.AttributeDefs.state_text.id
-    )
-    assert mi_listener.attribute_updates[1][1] == "foo"
-
-
-@pytest.mark.parametrize(
-    "quirk",
-    (zhaquirks.xiaomi.aqara.switch_acn047.AqaraT2Relay,),
-)
-async def test_aqara_t2_relay_endpoint_2(zigpy_device_from_quirk, quirk):
-    """Test Aqara T2 relay endpoint 2."""
-
-    device = zigpy_device_from_quirk(quirk)
-    mi_cluster = device.endpoints[2].multistate_input
-    mi_listener = ClusterListener(mi_cluster)
-
-    mi_cluster.update_attribute(MultistateInput.AttributeDefs.present_value.id, 1)
-    assert len(mi_listener.attribute_updates) == 1
-    assert mi_listener.attribute_updates[0][0] == 0
-    assert mi_listener.attribute_updates[0][1] == "switch_2"
+    assert mi_listener.attribute_updates[0][1] == "switch_" + endpoint
 
     mi_cluster.update_attribute(MultistateInput.AttributeDefs.state_text.id, "foo")
     assert len(mi_listener.attribute_updates) == 2
--- a/tests/test_xiaomi.py
+++ b/tests/test_xiaomi.py
@@ -1610,31 +1610,6 @@ async def test_xiaomi_e1_roller_commands_1(zigpy_device_from_quirk, command, val
 
 
 @pytest.mark.parametrize(
-    "quirk",
-    (zhaquirks.xiaomi.aqara.switch_acn047.AqaraT2Relay,),
-)
-async def test_aqara_t2_relay(zigpy_device_from_quirk, quirk):
-    """Test Aqara T2 relay."""
-
-    device = zigpy_device_from_quirk(quirk)
-    mi_cluster = device.endpoints[1].multistate_input
-    mi_listener = ClusterListener(mi_cluster)
-
-    mi_cluster.update_attribute(MultistateInput.AttributeDefs.present_value.id, 1)
-    assert len(mi_listener.attribute_updates) == 1
-    assert mi_listener.attribute_updates[0][0] == 0
-    assert mi_listener.attribute_updates[0][1] == "single"
-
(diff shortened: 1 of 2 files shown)


tests/test_xiaomi.py
```diff
@@ -1664,6 +1639,56 @@ async def test_xiaomi_e1_roller_commands_2(zigpy_device_from_quirk, command, val
     )
 
 
+@pytest.mark.parametrize(
+    "quirk",
+    (zhaquirks.xiaomi.aqara.switch_acn047.AqaraT2Relay,),
+)
+async def test_aqara_t2_relay_endpoint_1(zigpy_device_from_quirk, quirk):
```
TheJulianJES: Instead of giving the quirk via parametrize, we should be able to give `1` and `2` for the endpoint and avoid the duplicated test.
If you need help with this, I can push the changes later or so.

zhaquirks/xiaomi/aqara/switch_acn047.py
```diff
@@ -54,14 +54,15 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def _update_attribute(self, attrid, value):
-        if attrid == STATUS_TYPE_ATTR:
-            self._current_state = PRESS_TYPES.get(value)
+        if attrid == STATUS_TYPE_ATTR and value == 1:
```
TheJulianJES, line 59: Just to make sure I understand, what is `value` for this device when the presses are triggered? And why do you check for `1`?

zhaquirks/xiaomi/aqara/switch_acn047.py
```diff
@@ -206,5 +207,12 @@ class AqaraT2Relay(XiaomiCustomDevice):
     }
 
     device_automation_triggers = {
-        (SHORT_PRESS, SHORT_PRESS): {COMMAND: SINGLE},
+        (SHORT_PRESS, PRESS_TYPES.get(1)): {
```
TheJulianJES: Use `PRESS_TYPES[1]` instead of `.get(1)` here, as we know the values exist (or should exist at least).

TheJulianJES (commented):
Nice! Added some small comments

zhaquirks/xiaomi/aqara/switch_acn047.py
trapacska: Anyhow I shorted the pins (short, long, ultralong, double, triple etc) I always received 1. So I did not want to check value for 1 but the original implementation did check in a way only index 1 was available in the PRESS_TYPES hash/object. So to be safe 1 checked.

What I could do is to create a nested hash, so the button event names are accessible this way: PRESS_TYPES[endpoint].get(value) so the check for fix 1 could be removed. Wdyt?

TheJulianJES: Hmm, okay. I'd leave it like this for now then. Should be fine, as we don't seem to ignore any events.
(But it would be good to find out if it can ever be something other than 1)

tests/test_xiaomi.py
```diff
@@ -1640,45 +1640,19 @@ async def test_xiaomi_e1_roller_commands_2(zigpy_device_from_quirk, command, val
...
-    mi_cluster = device.endpoints[2].multistate_input
+    device = zigpy_device_from_quirk(zhaquirks.xiaomi.aqara.switch_acn047.AqaraT2Relay)
+    mi_cluster = device.endpoints[endpoint].multistate_input
     mi_listener = ClusterListener(mi_cluster)
 
     mi_cluster.update_attribute(MultistateInput.AttributeDefs.present_value.id, 1)
     assert len(mi_listener.attribute_updates) == 1
     assert mi_listener.attribute_updates[0][0] == 0
-    assert mi_listener.attribute_updates[0][1] == "switch_2"
+    assert mi_listener.attribute_updates[0][1] == "switch_" + endpoint
```
TheJulianJES: Use an f-string: `f"switch_{endpoint}"`
(It'll also resolve "TypeError: can only concatenate str (not "int") to str")

trapacska: Oh didn't notice your comment, fixing.

tests/test_xiaomi.py
```diff
@@ -1664,6 +1639,28 @@ async def test_xiaomi_e1_roller_commands_2(zigpy_device_from_quirk, command, val
     )
 
 
+@pytest.mark.parametrize("endpoint", [(1), (2)])
+async def test_aqara_t2_relay(zigpy_device_from_quirk, endpoint):
+    """Test Aqara T2 relay."""
```
TheJulianJES, line 1646: Note for other reviewers: This test was in a weird place before (between two other related tests), hence it's good to move this.

zhaquirks/xiaomi/aqara/switch_acn047.py
trapacska: I am not sure what are the best practices in this repo but what we could do is to surface "unknown" for the users so they can notice and report it. I guess they can report also if something is not working correctly. I am not a big fan of hardcoding stuff, we could also expose the raw events so it is arbitrary, for example: `event_1_1`, `event_1_2` in the current scenario. Or we can do it when `switch_1` and `switch_2` is not recognised. 🙃 

If we are here, could you please pass me a documentation/reference PR so I can learn how to extend HA to include `interlock` and `pulse_length` inputs on the device UI? 😊

zhaquirks/xiaomi/aqara/switch_acn047.py
```diff
@@ -206,5 +207,12 @@ class AqaraT2Relay(XiaomiCustomDevice):
     }
 
     device_automation_triggers = {
-        (SHORT_PRESS, SHORT_PRESS): {COMMAND: SINGLE},
+        (SHORT_PRESS, PRESS_TYPES[1]): {
```
TheJulianJES, line 212: Would `button_1` and `button_2` also fit? (from [zhaquirks/const.py#L20-L21](https://github.com/zigpy/zha-device-handlers/blob/6ab4376f4947c456a1f1ba05e9218519bc970100/zhaquirks/const.py#L20-L21)) 
They should have a translation in the UI: [homeassistant/components/zha/strings.json#L228-L252](https://github.com/home-assistant/core/blob/5dac5d5c7e9c5a1bf3ec4edc5198a7a0d92d27ee/homeassistant/components/zha/strings.json#L228-L252)

trapacska: I am not sure. Maybe.
Screenshot 2024-01-27 at 8 32 17
It is SW1-COM and SW2-COM shorted. I want to use them with reed so I can report if garage door is fully closed or fully open.

trapacska: The best would be to not to send events IMHO but to have binary sensors if these are shorted or not. This device is a perfect fit for garage doors but this way it requires extra automation and sensor implementation in HA.

Another strange thing:
- When it is in toggle mode then it sends two events when I close circuit and when I open.
- When it is in momentary then it sends one event when I open the circuit.

So it is not very optimal. Is it something can be handled from the quirk or it is the device's firmware?

zhaquirks/xiaomi/aqara/switch_acn047.py
TheJulianJES: > we could do is to surface "unknown" [...] also expose the raw events

I think we generally don't do this. I'd still leave it like this for now (or remove the `value == 1` part?).
Is there a difference when it's in ~~dry~~ decoupled mode or not? 

Maybe the answer is somewhere in the zigbee-herdsman-converters library, but I wasn't able to find it quickly.

---

For HA entities:

So this is already implemented for the T2 relay: [homeassistant/components/zha/select.py#L426-L435](https://github.com/home-assistant/core/blob/5dac5d5c7e9c5a1bf3ec4edc5198a7a0d92d27ee/homeassistant/components/zha/select.py#L426-L435)

For interlock, you just want on/off, so a switch in HA. Here and something like this: [homeassistant/components/zha/switch.py#L293-L302](https://github.com/home-assistant/core/blob/5dac5d5c7e9c5a1bf3ec4edc5198a7a0d92d27ee/homeassistant/components/zha/switch.py#L293-L302), but with the model and attribute changed, maybe also icon.

You also need to add the translation key to the JSON strings file and generate the translations locally: https://developers.home-assistant.io/docs/internationalization/core/#test-translations

For pulse length, look at `number.py` (custom entities are all towards the bottom of each file.

This will be easier to expose _eventually_ (just with quirks).

trapacska: > Is there a difference when it's in dry decoupled mode or not?

Yes, this events will fire only if it is in decoupled mode.

zhaquirks/xiaomi/aqara/switch_acn047.py
TheJulianJES: The device sends events instead of exposing an `OnOff` cluster for example.
Since push buttons (momentary mode) only send one event, a binary sensor wouldn't make sense, as we'd have to artificially turn off the signal after a delay.
Like, you can do many things with a quirk, but we shouldn't do this probably.

I'd just set up an automation in HA with events from the relay (with different trigger ids) that toggle `input_boolean` switches. Something like that.

Also, let's use the `button_1` and `button_2` constants here, so it's properly translated in the HA UI. And it's also better to not confuse it with the switch / relay part of the device.

TheJulianJES (approved):
LGTM, thanks!

PR #2948, opened 2024-01-29
Fix IKEA double battery percentage for old/new firmware

TheJulianJES:
## Proposed change
This fixes the issue of some battery-powered IKEA devices having double the percentage reported.
The "double battery pct" cluster is still used on some devices which can have new or old firmware (where doubling may or may not be needed depending on the installed firmware version).

The fix would be simple if ZHA/zigpy cached the `sw_build_id` attribute when pairing/reconfiguring a device.
But that doesn't happen. So, we need to read the attribute once to cache it while the device is awake.
(That would also be required if we want to use `sw_build_id` with quirks v2, because we can't magically read an attribute of battery-powered devices that are already paired and asleep.)

The diff might look huge at first (but ignore the tests) and the fix is mostly somewhat simple actually.
-> Start by looking at `_update_attribute()`

## The "fix"

The "fix" works as follows:
- a battery percentage report comes in from the device
   - check `sw_build_id` attribute value from cache like:
      - if the value is known (e.g. `24.0.0`), we split on `.` and check the first part, here: 24
      - when the value is 24 or greater, we do not double the percentage.
      - when the value is below 24, we double the percentage.
      - if the value is not known, double the percentage AND start a task to try and read `sw_build_id` in the background
- if a task was started because of `sw_build_id` being unknown AND the read is successful:
   - and if the task returned that new firmware is installed (>= 24), we update the attribute cache again to correct the wrong doubled value and replace it with the correct one
   - and if the task returned that old firmware is installed, we do not do anything, as the value is already correct because we doubled it by default

Note: If the attribute read for `sw_build_id` is successful (or it's already known), no (further) read calls will ever be made.

EDIT: For (new) pairings or reconfigurations, `sw_build_id` is now also read on pairing/`bind()`.

### Note part 1 / TODO:

~~I did not yet test if an IKEA device responds to attribute reads when it sends an attribute report with the remaining battery percentage. I assume it would, as it should be "awake" at that point, but I still have to test this.
(So far, I only wrote the unit tests to test this behavior.)~~

~~If, for some reason, it does not wake up to receive message in that case, I'd move the logic to read the `sw_build_id` attribute once (until it's in cache) to happen when a button is pressed (for a remote).
Because then, a remote is definitely ready to receive commands for a short period.~~

Seems to work as expected. A button push during longer inactivity of the remote also triggers an attribute report, thus causing us to read `sw_build_id` while the device is awake.

### Note part 2:

Or, if we do not want this "read schedule logic" at all (for now), we can also just check the `sw_build_id` from attribute cache and if it's unknown, not schedule a read call.

This might cause weird behavior though if a user who previously manually read `sw_build_id` and has correct percentages then removes and re-pairs the device, only now to have doubled percentages (because `sw_build_id` isn't in the attribute cache).
(That's fixed by scheduling the read at that point.)

### Note part 3 (EDIT):

We're now also reading `sw_build_id` during pairing/reconfiguration, so during `bind()`.
This will not automatically fix the issue for existing users who currently have doubled percentages, but it will fix it if the device is paired again or reconfigured.
I'm hoping that the "task to schedule a read" will still work to fix existing installations automatically.
If it doesn't work, the "making users reconfigure their device to read `sw_build_id` and fix the issue"-approach is also fine I guess.

~~(I'm waiting for an attribute report from the IKEA remote for now to see if it also receives commands for a short period then.)~~

### Initial approach

My initial approach did not re-update the attribute cache after a successful attribute read if the initial doubling was wrong.
The successful attribute read only updated the attribute cache, so only next time the device sends something, the percentage would be right. You can still see this approach in my initial commit.

---

cc @dmulcahey just to know if this approach/PR would be somewhat okay. I know it might not be ideal to have to read attributes in the background, but it's only done once (if successful) and we need the `sw_build_id` sooner or later anyway.

A similar-ish approach with the background tasks is used for some Tuya devices and EU Xiaomi plugs (although not for firmware checking).

## Checklist

### DIFF
--- a/tests/test_ikea.py
+++ b/tests/test_ikea.py
@@ -196,3 +196,74 @@ async def test_double_power_config_firmware(zigpy_device_from_quirk, quirk):
         # check no reads were made when sw_build_id is known
         assert mock_task.call_count == 0
         assert request_mock.call_count == 0
+
+
+@pytest.mark.parametrize(
+    "firmware, pct_device, pct_correct, expected_pct_updates",
+    (
+        ("2.3.075", 50, 100, 1),
+        ("24.4.5", 50, 50, 2),
+    ),
+)
+async def test_double_power_config_firmware_2(
+    zigpy_device_from_quirk, firmware, pct_device, pct_correct, expected_pct_updates
+):
+    """Test battery percentage remaining is doubled for old firmware."""
+
+    device = zigpy_device_from_quirk(zhaquirks.ikea.fivebtnremote.IkeaTradfriRemote1)
+
+    basic_cluster = device.endpoints[1].basic
+    ClusterListener(basic_cluster)
+    sw_build_id = Basic.AttributeDefs.sw_build_id.id
+
+    power_cluster = device.endpoints[1].power
+    power_listener = ClusterListener(power_cluster)
+    battery_pct_id = PowerConfiguration.AttributeDefs.battery_percentage_remaining.id
+
+    # fake read response for attributes: return plug_read argument for all attributes
+    def mock_read(attributes, manufacturer=None):
+        records = [
+            foundation.ReadAttributeRecord(
+                attr, foundation.Status.SUCCESS, foundation.TypeValue(None, firmware)
+            )
+            for attr in attributes
+        ]
+        return (records,)
+
+    p1 = mock.patch.object(power_cluster, "create_catching_task")
+    p2 = mock.patch.object(
+        basic_cluster, "_read_attributes", mock.AsyncMock(side_effect=mock_read)
+    )
+
+    with p1 as mock_task, p2 as request_mock:
+        # update battery percentage with no firmware in attr cache, check pct doubled for now
+        power_cluster.update_attribute(battery_pct_id, pct_device)
+        assert len(power_listener.attribute_updates) == 1
+        assert power_listener.attribute_updates[0] == (battery_pct_id, pct_device * 2)
+
+        # but also check that sw_build_id read is requested in the background for next update
+        assert mock_task.call_count == 1
+        await mock_task.call_args[0][0]  # await coroutine to read attribute
+        assert request_mock.call_count == 1  # verify request to read sw_build_id
+        assert request_mock.mock_calls[0][1][0][0] == sw_build_id
+
+        # battery pct might be updated again when the attribute read returned new firmware, check pct not doubled then
+        # if firmware turned out to be old or still unknown, do not update battery pct again, as we doubled it already
+        assert len(power_listener.attribute_updates) == expected_pct_updates
+        if expected_pct_updates > 2:
+            assert power_listener.attribute_updates[1] == (battery_pct_id, pct_correct)
+
+        # reset mocks for testing when sw_build_id is known next
+        mock_task.reset_mock()
+        request_mock.reset_mock()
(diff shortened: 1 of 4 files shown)


TheJulianJES:
"New": `sw_build_id` is now also read during `bind()`, no matter if it's already cached or not.
(we do it from the `PowerConfig` cluster, but that's fine. The `Basic` cluster isn't bound by ZHA and we also kind of only want to read this for device where we need it which exactly match where this power cluster class is used.)

The device is very likely to be awake when `bind()` is called and it also reads the attribute on (first) pairing.
For existing pairing, the "task approach" would still be needed. But if that somehow doesn't work out, we can at least keep the `bind()` part to read the attribute once and tell users to repair/reconfigure affected IKEA devices.

MattWestb:
If the controller is on current production firmware its Zigbee3 and have / using pull control.
Way not using the checkin the controller is doing to the coordinator and hocking it for reading the attribute. ZHA is getting it and putting the device in fast pull until sending fast pull pull end or timing out (from the time set by the system in the device). IKEA controllers is doing checkin every 50 - 55 minutes and is logged in HA so its working well.
Z2M have implanted some of this logic but i dont knowing if its working 100 % fir all things.
Its one of the best Zigbee 3 futures for end devices and shall working OK if implanting it well.
Also possible sending fast pull to the device for extending the time then doing "other things" (IKEA Dirigera is doing it then pairin devices).

TheJulianJES:
This seems to be working nicely. 
So, this should automatically fix any issues for double percentages reported on newer firmware. No need to re-pair/reconfigure the device.

TheJulianJES:
I've added another [small commit](https://github.com/zigpy/zha-device-handlers/pull/2948/commits/b9bacbe71270f02997dc8877ac85e2d057229b3d) to guard against possible future version formatting where `sw_build_id` might contain letters (we only expect numbers).
The firmware is handled as "new" in that case (so no double percentage), and a warning also gets printed.

Again, this should be good to go and if nobody has anything against this, I'll merge it tomorrow and release/bump quirks for the next HA beta.

MattWestb:
I can normally testing 90% of IKEA controllers with  current production firmware (and reversing to old if needed with SWD flashing if very needed) but im 810 Km from my systems so cant doing testing and im going back in one week if all is going well.

## Great work done !!

ProfDrYoMan:
Just a side note: This version is running on my side since the release and still I see double battery.
I also tried to query that `sw_build_id` manually and also tried to rebind the device(s) (not deleting them before).

Still same issue. Will wait for some more time and if it does not change will report in a new ticket with logs.
What type of loggings should be done and included?

TheJulianJES:
Were you able to read `sw_build_id` successfully? (Remotes need to be waken up by pressing some buttons whilst trying to read the attribute).

Please open a new issue and provide the diagnostic information for the affected devices.
Tag me in that issue (@TheJulianJES).

ProfDrYoMan:
Yes. I could read with success with the help of a press on the button(s).

And rebinding is anyhow completely waking the device.

Since you are so quick answering I will try to get you the diagnostics this evening. :)

TheJulianJES:
Nice. New issue would indeed be good for that, but do you roughly remember the version that it showed off the top of your head? (e.g. 2.3.075 or 24.4.5)

And which device is affected?

PR #2957, opened 2024-01-31
Update ts0601_din_power.py - Added support for newer Zemismart

simowilso:
Added MODELS_INFO for Zemismart Smart Energy Sensor

## Checklist


simowilso:
I think there's more to it than this. After making this change I can see four sensors:

image
 
But, if I run reconfigure - I can see a whole lot more:

image

PR #2960, opened 2024-01-31
Add Inovelli VZM36 support and rework manufacturer clusters

codyhackw:
## Proposed change

This change adds support for the Inovelli VZM36 canopy module and re-organizes the \_\_init__.py file to simplify the parameters across devices.

## Additional information

I believe the ep_attribute has to stay pointed to the inovelli_vzm31sn_cluster as that's what's [referenced within ZHA](https://github.com/home-assistant/core/blob/2aead3eefc9e6945b6cfb663b356eb93ebbda624/homeassistant/components/zha/core/const.py#L101C1-L101C54) so changing that would result in a breaking change. In the interest of avoiding that, I've kept it in place.

## Checklist


zhaquirks/inovelli/__init__.py
```diff
@@ -66,97 +69,42 @@
...
 NOTIFICATION_TYPE = "notification_type"
 
 
-class Inovelli_VZM31SN_Cluster(CustomCluster):
-    """Inovelli VZM31-SN custom cluster."""
-
-    cluster_id = 0xFC31
-    name = "InovelliVZM31SNCluster"
-    ep_attribute = "inovelli_vzm31sn_cluster"
+class Inovelli_Cluster(CustomCluster):
```
TheJulianJES: It's preferred to use "CamelCase" / "CapWords" for class names.

TheJulianJES: From what I can see, `cluster_id` and `ep_attribute` is the same for all Inovelli cluster classes.
Can we just put it in this base class then?

zhaquirks/inovelli/__init__.py
```diff
@@ -66,97 +69,42 @@
...
+    """Inovelli base cluster."""
 
     attributes = {
         0x0001: ("dimming_speed_up_remote", t.uint8_t, True),
```
TheJulianJES, line 94: Not for this PR, but we need to switch to the new zigpy `AttributeDefs` style at some point.

zhaquirks/inovelli/__init__.py
codyhackw: Re the class name, are you saying the preference would be to do "InovelliCluster" as opposed to "Inovelli_Cluster"? This was staying consistent to the existing naming, but can definitely change it if needed. Just have to change it in the device-specific imports as well.

Putting cluster_id and ep_attribute in the base class is giving async_initialize errors...it all appears to work, but putting them across each cluster resolved that and is why I ended up splitting them out.

TheJulianJES: > saying the preference would be to do "InovelliCluster" as opposed to "Inovelli_Cluster"? 

Yes. That's [preferred](https://peps.python.org/pep-0008/#class-names). 

> Putting cluster_id and ep_attribute in the base class is giving async_initialize errors...it all appears to work, but putting them across each cluster resolved that and is why I ended up splitting them out.

Weird. There shouldn't be any difference. Are you sure it wasn't just some connectivity issue?

zhaquirks/inovelli/__init__.py
TheJulianJES: Again, no need to do it in this PR already, but saving a code sample here for reference how it would look:

zhaquirks/inovelli/__init__.py
codyhackw: Updating all the classes, thanks for the clarification!

And yeah agreed it's weird, just re-tested. In 2024.2.0 it no longer shows up as a warning, but it's still in the logs - 

```
2024-02-09 08:05:45.026 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0xfc31]: failed to get attributes '['power_type', 'switch_type', 'button_delay', 'smart_bulb_mode', 'local_protection']' on 'inovelli_vzm31sn_cluster' cluster: 
2024-02-09 08:05:45.027 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0019]: initializing cluster handler: from_cache: False
2024-02-09 08:05:45.027 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0019]: initializing cached cluster handler attributes: ['current_file_version']
2024-02-09 08:05:45.027 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0019]: Reading attributes in chunks: ['current_file_version']
2024-02-09 08:05:45.027 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0019]: finished cluster handler initialization
2024-02-09 08:05:45.082 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0006]: 'async_initialize' stage succeeded
2024-02-09 08:05:45.082 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0008]: 'async_initialize' stage succeeded
2024-02-09 08:05:45.082 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0000]: 'async_initialize' stage succeeded
2024-02-09 08:05:45.082 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0b04]: 'async_initialize' stage succeeded
2024-02-09 08:05:45.082 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0003]: 'async_initialize' stage succeeded
2024-02-09 08:05:45.083 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0702]: 'async_initialize' stage succeeded
2024-02-09 08:05:45.083 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0x0019]: 'async_initialize' stage succeeded
2024-02-09 08:05:45.083 DEBUG (MainThread) [homeassistant.components.zha.core.cluster_handlers] [0x90CF:1:0xfc31]: 'async_initialize' stage failed: 
```

TheJulianJES: The log level was changed indeed. Are there more (debug) log messages when `zigpy` is also set to debug level?
There's no reason I can see why this could happen. Setting `cluster_id`, `name`, and `ep_attribute` in the "base class" vs the one inheriting that doesn't make a difference.
Can you double-confirm that the error happens every time? And not every time when set in the base classes?

Also, did you check that all attributes set here are correct?
https://github.com/home-assistant/core/blob/567a179084e0226a45d1db01ffb5d4e2e37f3d8c/homeassistant/components/zha/core/cluster_handlers/manufacturerspecific.py#L249-L253
(There might be more info in zigpy debug logs)

codyhackw: I've had both zigpy and zhaquirks set to debug while trying to figure out what's going on and haven't seen anything relevant from zigpy unfortunately. Confirmed I end up with at least one Inovelli device giving that error every time. It's not all of them, and the devices do still seem to work, but I do get at least one log message in 100% of my tests but I don't when I have them set as in the current PR state. I'm not certain why that is because I'd expect it to be the same.

I did also double check the attributes, those all check out, I do find it interesting that the attributes that error are split between the base and one inheriting, but not sure if that's relevant?

zhaquirks/inovelli/__init__.py
```diff
@@ -245,66 +193,240 @@ def handle_cluster_request(
...
+class InovelliVZM31SNCluster(InovelliCluster):
+    """Inovelli VZM31-SN custom cluster."""
+
+    cluster_id = 0xFC31
+    name = "InovelliVZM31SNCluster"
+    ep_attribute = "inovelli_vzm31sn_cluster"
+
+    attributes = {
+        key: InovelliCluster.attributes[key] for key in InovelliCluster.attributes
+    }
```
TheJulianJES: Can you use:
```python
attributes = InovelliCluster.attributes.copy()
```
Please also test if you still have the initialization error then.
... although there shouldn't be a difference I think. But still, use the `.copy()` one now that attributes no longer need to removed, as it's what we're using "everywhere" already.

zhaquirks/inovelli/__init__.py
TheJulianJES: Possibly related comment: https://github.com/zigpy/zha-device-handlers/pull/2960#discussion_r1485413406

codyhackw: Updated to that and was still getting them, however when flipping back to how the PR is currently, I also replicated it once...so I'm just going to chalk it up to my network? Not entirely sure what's causing it, but if it happens in both scenarios it's likely unrelated. I'll move these to the base class and thanks for the ideas in trying to narrow it down.

zhaquirks/inovelli/__init__.py
```diff
@@ -245,66 +196,224 @@ def handle_cluster_request(
...
-    0x007D,
-    0x0105,
-]
+class InovelliVZM31SNCluster(InovelliCluster):
+    """Inovelli VZM31-SN custom cluster."""
+
+    name = "InovelliVZM31SNCluster"
+
+    attributes = InovelliCluster.attributes.copy()
+
```
TheJulianJES: Can you also remove this empty line between `attributes = ...` and `attributes.update(...` everywhere?
(To follow what other quirks do)

zhaquirks/inovelli/__init__.py
```diff
@@ -66,97 +69,45 @@
...
+# 6 - aux config button
 
-BUTTONS = {1: BUTTON_1, 2: BUTTON_2, 3: BUTTON_3}
+BUTTONS = {1: BUTTON_1, 2: BUTTON_2, 3: BUTTON_3, 4: BUTTON_4, 5: BUTTON_5, 6: BUTTON_6}
 ON = "Up"
 OFF = "Down"
 CONFIG = "Config"
+AUX_ON = "Aux Up"
+AUX_OFF = "Aux Down"
+AUX_CONFIG = "Aux Config"
```
TheJulianJES: I think these should follow HA standards and be "Aux up", "Aux down", "Aux config"?

TheJulianJES (approved):
LGTM, thanks!

Issue #2418, opened 2023-05-31
[BUG] Ikea tradfri FYRTUR blinds reporting wrong battery % since latest firmware


alvinchen1:
### Bug description

FYRTUR block-out roller blind with firmware 0x24040011, latest firmware deployed May 29, 2023 (https://ww8.ikea.com/ikeahomesmart/releasenotes/releasenotes.html) are reporting odd Battery %. Release notes include:

FYRTUR and KADRILJ roller blinds (24.4.11)
Product ID: E1752, E1757
◆ Optimised battery performance.
◆ Stability improvements.

![image](https://github.com/zigpy/zha-device-handlers/assets/26409535/088e6bb7-0b1d-4384-aa8a-0687bbb767a8)

Blinds which haven't gotten the new firmware (0x23088631) appear fine, but new firmware seems to now report ~2x the actual value.

![image](https://github.com/zigpy/zha-device-handlers/assets/26409535/8742ab9c-cb30-4f53-ac64-aa8589ccd5b4)

Note in screenshot, Back Door Shade received firmware update 0x24040011 on 5/31/2023 at 8:55 AM and battery % doubled.

![image](https://github.com/zigpy/zha-device-handlers/assets/26409535/783d116f-c922-4fbb-b577-2773c7e94ff0)

Another Example

<img width="514" alt="image" src="https://github.com/zigpy/zha-device-handlers/assets/26409535/6761107d-96e6-4991-9cc3-467cd15985b7">

Received firmware today, battery went from 44% to 100% ~2x.

<img width="425" alt="image" src="https://github.com/zigpy/zha-device-handlers/assets/26409535/422b4ea2-f14a-4b8c-bb14-eee8e88638c1">

### Steps to reproduce

1. Device receives firmware update 0x24040011
2. Device's battery doubles.

### Expected behavior

1. Device receives firmware update 0.24040011
2. Device's battery does not double?  *shrugs*

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
[...]

PR #2986, opened 2024-02-17
Add version 3 of Ikea Frytur Blinds firmware 0x24040011

alvinchen1:
## Proposed change

Since with firmware 0x24040011, IKEA introduced a new output cluster 3. (see https://github.com/zigpy/zha-device-handlers/issues/2418). This quirk adds the additional output clusters. I have tested and thanks to https://github.com/zigpy/zha-device-handlers/pull/2948, battery issues appear to also be resolved. Revised from https://github.com/zigpy/zha-device-handlers/pull/2984

## Additional information

I haven't tested a new factory blind getting gets this or a later firmware update, getting this quirk and reports 2x the battery, but see https://github.com/zigpy/zha-device-handlers/issues/2418 for possible resolutions. My testing on devices with the firmware show correct battery and restore a minor annoyance where the shade being fully open will "drift" every so often leaving the shade slightly open.

## Checklist

### DIFF
--- a/zhaquirks/ikea/blinds.py
+++ b/zhaquirks/ikea/blinds.py
@@ -145,7 +145,7 @@ class IkeaTradfriRollerBlinds3(CustomDevice):
                 OUTPUT_CLUSTERS: [
                     Identify.cluster_id,
                     Ota.cluster_id,
-                    LightLink.cluster_id
+                    LightLink.cluster_id,
                 ],
             }
         },
@@ -170,7 +170,7 @@ class IkeaTradfriRollerBlinds3(CustomDevice):
                 OUTPUT_CLUSTERS: [
                     Identify.cluster_id,
                     Ota.cluster_id,
-                    LightLink.cluster_id
+                    LightLink.cluster_id,
                 ],
             }
         }
--- a/zhaquirks/ikea/blinds.py
+++ b/zhaquirks/ikea/blinds.py
@@ -112,3 +112,59 @@ class IkeaTradfriRollerBlinds2(CustomDevice):
     }
 
     replacement = IkeaTradfriRollerBlinds.replacement
+
+class IkeaTradfriRollerBlinds3(CustomDevice):
+    """Custom device representing IKEA of Sweden TRADFRI Fyrtur blinds."""
+
+    signature = {
+        # <SimpleDescriptor endpoint=1 profile=260 device_type=514
+        # device_version=1
+        # input_clusters=[0, 1, 3, 4, 5, 32, 258, 4096, 64636]
+        # output_clusters=[3, 25, 4096]>
+        MODELS_INFO: [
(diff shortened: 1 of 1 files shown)


TheJulianJES:
So, this firmware doesn't need battery percentage doubling, right? Since if this signature appears, it's always guaranteed to be a new firmware version.
The battery cluster will detect that too and just "do nothing", but if we can avoid using that cluster at all, we should.

If that cluster is removed, the quirk just doesn't do anything from what I can see. Like, no other clusters are patched.
So, what's the reason this is needed?

> My testing on devices with the firmware show correct battery and restore a minor annoyance where the shade being fully open will "drift" every so often leaving the shade slightly open.

But the quirk does nothing to change this?

alvinchen1:
So I wondered about that as well and please forgive me if I am new and trying my first time at contributing.  Let me see if I can explain better.  What I found is that AFTER the firmware 0x24040011, none of the two original class quirks would apply.   

class IkeaTradfriRollerBlinds (original)

        # input_clusters=[0, 1, 3, 4, 5, 32, 258, 4096, 64636]
        # output_clusters=[25, 4096]>

class IkeaTradfriRollerBlinds2 (original)

        # input_clusters=[0, 1, 3, 4, 5, 32, 258, 4096]
        # output_clusters=[25, 4096]>

class IkeaTradfriRollerBlinds3 (my contribution)

        # input_clusters=[0, 1, 3, 4, 5, 32, 258, 4096, 64636]
        # output_clusters=[3, 25, 4096]>

Having no quirk applied resulted in two problems (see https://github.com/zigpy/zha-device-handlers/issues/2418).  

1) Battery would double which was immediately apparent.
2) Shade 100% open position would "drift" and not open completely.  Then when shade would close it would be off by the same drift difference.  This wouldn't happen immediately but over time.  I'd say the course of one or three weeks, there would be anywhere from 1" to 2.5" difference when the shade was supposed to be 100% open. I have 20 of these shades so I definitely have a lot of data points.  I would have to go around, pull the battery, wait, put it back in and the shade would reset to 100% closed, but over time, it would begin to drift again.

What I found is that when I wrote my custom quirk and tested with it, no longer get either the battery or the drift issue. When it set to open 100%, it does so perfectly and has for several months (testing since Nov 2023 see https://github.com/zigpy/zha-device-handlers/discussions/2754).  But you are correct, I don't see anything in the quirk itself which remedies this, I assumed (perhaps wrongly) that simply having a quirk was somehow fixing this since it always worked prior to firmware 0x24040011 when a quirk was applied and started working again once I wrote my custom quirk. Since both of my two problems appear fixed, I wanted to contribute it.    Hope that makes more sense and answers your questions.

I believe the firmware still needs the same battery doubling, or at least in my testing, since battery doubling went in with PR 2948, it appears to still be working correctly.  I know this because when I put in a fresh battery, it would show 200% if still a problem and it hasn't been doing so.

PR #2991, opened 2024-02-19
Add Tuya smoke detector TS0205 PST-YG500A

cdalexndr:
## Proposed change

tuya-smoke-detector-smart-zigbee

image

## Additional information

Removed useless Opening sensor.
Changed motion sensor to smoke sensor.

## Checklist


zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,157 @@
...
+
+    attributes = {
+        TUYA_SMOKE_DETECTED_ATTR: ("smoke_detected", t.uint8_t, True),
+    }
+
+    def _update_attribute(self, attrid, value):
+        super()._update_attribute(attrid, value)
+        if attrid == TUYA_SMOKE_DETECTED_ATTR:
+            if value == 0:
+                self.endpoint.device.ias_bus.listener_event(
```
TheJulianJES: Docs still need to be updated, but we can call `update_attribute` on another cluster like this:
`self.endpoint.ias_zone.update_attribute(IasZone.AttributeDefs.zone_status.id, value)`

Please use that and remove all `Bus` usage

zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,157 @@
...
+    _CONSTANT_ATTRIBUTES = {ZONE_TYPE: IasZone.ZoneType.Fire_Sensor}
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        super().__init__(*args, **kwargs)
+        self.endpoint.device.ias_bus.add_listener(self)
+
+    def update_zone_status(self, value):
+        """Update IAS status."""
+        super()._update_attribute(ZONE_STATUS, value)
```
TheJulianJES: This can be removed completely then if the suggestions from the above comment are implemented.

zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,157 @@
...
+                self.endpoint.endpoint_id,
+                self.cluster_id,
+                attrid,
+            )
+
+
+class TuyaIasZone(CustomCluster, IasZone):
+    """IAS Zone."""
+
+    _CONSTANT_ATTRIBUTES = {ZONE_TYPE: IasZone.ZoneType.Fire_Sensor}
```
TheJulianJES: Instead of the old `ZONE_TYPE` constant, let's use the new zigpy AttributeDefs: `IasZone.AttributeDefs.zone_type.id`

zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,157 @@
...
+            _LOGGER.warning(
+                "[0x%04x:%s:0x%04x] unhandled attribute: 0x%04x",
+                self.endpoint.device.nwk,
+                self.endpoint.endpoint_id,
+                self.cluster_id,
+                attrid,
+            )
+
+
+class TuyaIasZone(CustomCluster, IasZone):
```
TheJulianJES: I think we should use `LocalDataCluster` instead of `CustomCluster` here, since it's a "fake cluster".

zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,157 @@
...
+    #   },
+    #   "manufacturer": "_TZ3210_up3pngle",
+    #   "model": "TS0205",
+    #   "class": "zigpy.device.Device"
+    # }
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        self.ias_bus = Bus()
+        super().__init__(*args, **kwargs)
```
TheJulianJES: (This can also be removed then)

zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,157 @@
...
+    #         "0x000a",
+    #         "0x0019",
+    #         "0x1000"
+    #       ]
+    #     }
+    #   },
+    #   "manufacturer": "_TZ3210_up3pngle",
+    #   "model": "TS0205",
+    #   "class": "zigpy.device.Device"
+    # }
```
TheJulianJES: Let's remove this. It's in the test and IMO, it's enough if it's like also in the PR/issue description for this device.

zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,157 @@
...
+        super().__init__(*args, **kwargs)
+
+    signature = {
+        MODELS_INFO: [("_TZ3210_up3pngle", "TS0205")],
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.IAS_ZONE,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,  # 0x0000
```
TheJulianJES: Can you also remove those cluster IDs comment at the end?
We're (mostly) not using them in quirks, so we should try to stay consistent.

TheJulianJES (changes requested):
Thanks for the PR! I've added some suggestions.
Ideally, we should also add a test for the actual code (that translates the Tuya attribute to the `IasZone` attribute).

This test can mostly be copied IMO: https://github.com/zigpy/zha-device-handlers/blob/e4d0663aa4fb2f4bd0b41825f1942e3d29893375/tests/test_xiaomi.py#L781-L809

The opple cluster stuff just needs to be changed to the Tuya stuff (and the quirk path needs to be changed).
If you need help with adding this test, let me know and I'll add it, as soon as I have some time.

mr-bembel:
Hello,

Are there any news to support the product in ZHA soon? I bought 4 of them...

Fr33L4nc3r:
@cdalexndr Seems the pipeline failed at pre-commit step. Since you fixed it in the past (https://github.com/zigpy/zha-device-handlers/pull/2991/commits/fa9848025ab0306627224b8b1adeb5617d9b7a37), any chance you could have a look again ?
It seems like there ain't much to have this PR approved (despite the change request still appearing in red, while it seemed addressed with https://github.com/zigpy/zha-device-handlers/pull/2991/commits/dcd03eb1035c7cd4ab2e283ba8c88b5b5464ff6b).

Looking forward to see this device recognized correctly and automatically in zha thanks to amazing community work like yours.

zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,118 @@
...
+
+TUYA_SMOKE_DETECTED_ATTR = 0x0401  # [0]/[1] [Detected]/[Clear]!
+
+
+class TuyaSmokeDetectorCluster(TuyaManufClusterAttributes):
+    """Manufacturer Specific Cluster of the TS0205 smoke detector."""
+
+    attributes = {
+        TUYA_SMOKE_DETECTED_ATTR: ("smoke_detected", t.uint8_t, True),
+    }
```
TheJulianJES: Can we use the new `AttributeDefs` style here? As an example, see:
https://github.com/zigpy/zha-device-handlers/blob/915ff9b41e7393e1f2eaa449bd6a09267fa84277/zhaquirks/legrand/cable_outlet.py#L51-L68

You can also remove `TUYA_SMOKE_DETECTED_ATTR` then and if it needs to be access anywhere else (except the attribute definition, where it should just be "hard-coded"), `TuyaSmokeDetectorCluster.AttributeDefs.smoke_detected.id` can be used.

zhaquirks/tuya/ts0205.py
```diff
@@ -0,0 +1,118 @@
...
+                    IasZone.AttributeDefs.zone_status.id, 0
+                )
+        else:
+            _LOGGER.warning(
+                "[0x%04x:%s:0x%04x] unhandled attribute: 0x%04x",
+                self.endpoint.device.nwk,
+                self.endpoint.endpoint_id,
+                self.cluster_id,
+                attrid,
+            )
```
TheJulianJES: Not sure if we really need this?

TheJulianJES: Removed it for now.

zhaquirks/tuya/ts0205.py
TheJulianJES: Changed.

TheJulianJES (approved):
Thanks, I think this should be fine.

abmantis:
@cdalexndr why is the `TuyaSmokeDetectorCluster` needed? The sensor worked fine previously, it would just show motion instead of smoke as the sensor type, but that is be fixed with just the addition of the `TuyaIasZone` cluster on line 57. So, why the extra `TuyaSmokeDetectorCluster` cluster?

Fr33L4nc3r:
Just received a Product Safety Recall mail from Aliexpress about this smoke detector, just sharing in case you didn't know, I know what I need to replace ASAP :/

PR #2996, opened 2024-02-22
Change Tuya switch TS000x device type

Migsi:
## Proposed change
Change device type of Switch_4G_GPP to be an actual switch, as the device consists of potential-free relais. Also the type switch is substitutable with other types using HA helpers, but not (or not easily) vice versa.

It would probably make sense to change this for other similar switches as well, but as I don't own any others, I don't want to change stuff blindy, potentially introducing unnecessary errors/misbehaviour.

I found no related tests to this device, so I didn't do anything in this regard as well, as I've got not enough time to read through your testframework (yet) :/

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts000x.py
+++ b/zhaquirks/tuya/ts000x.py
@@ -855,7 +855,7 @@ class Switch_4G_GPP(EnchantedDevice):
         ENDPOINTS: {
             1: {
                 PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
                     Identify.cluster_id,
@@ -869,7 +869,7 @@ class Switch_4G_GPP(EnchantedDevice):
             },
             2: {
                 PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_SWITCH,
                 INPUT_CLUSTERS: [
                     Groups.cluster_id,
                     Scenes.cluster_id,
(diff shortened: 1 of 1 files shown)


TheJulianJES (changes requested):
The issue with this is that it'll break existing installations. Everyone who currently set it up will have the old entity become stale and get a new one.
We've avoided changing the device type for new quirks because of this.

Please see this section on how to change the device type locally:
https://www.home-assistant.io/integrations/zha/#modifying-the-device-type

Migsi:
> The issue with this is that it'll break existing installations. Everyone who currently set it up will have the old entity become stale and get a new one. We've avoided changing the device type for new quirks because of this.
> 
> Please see this section on how to change the device type locally: https://www.home-assistant.io/integrations/zha/#modifying-the-device-type

I see. So this PR becomes obsolete?

While it is possible to work around this, I still see it at least somewhat inconvenient for new users setting up this (and potentialy any other wrongly specified device). Wouldn't it be possible to ensure somehow the original device type continues working by automatically generating a proper helper, just as "change device type" in the UI does? The latter only seems to work when the "base" entity was a generic switch, but not when it is something specific like e.g. a lamp in this case.

Migsi:
@TheJulianJES sorry for the ping, I'd just close the PR if it really became irrelevant.

TheJulianJES:
> Wouldn't it be possible to ensure somehow the original device type continues working by automatically generating a proper helper, just as "change device type" in the UI does? 

I know it's not ideal, but we can't really do this from quirks.

> The latter only seems to work when the "base" entity was a generic switch, but not when it is something specific like e.g. a lamp in this case.

That's a limitation on HA Core side, as the integration responsible for that is called "Switch as X" and only allows you to "convert" switches.

> So this PR becomes obsolete?

For now, probably yes. I'm aware the current situation isn't ideal (especially because you can't go from light to switch using HA UI and have to modify (ZHA) YAML configuration to fix this, but there's no "good" solution right now.)

Migsi:
> > Wouldn't it be possible to ensure somehow the original device type continues working by automatically generating a proper helper, just as "change device type" in the UI does?
> 
> I know it's not ideal, but we can't really do this from quirks.
>
> ...
> 
> > So this PR becomes obsolete?
> 
> For now, probably yes. I'm aware the current situation isn't ideal (especially because you can't go from light to switch using HA UI and have to modify (ZHA) YAML configuration to fix this, but there's no "good" solution right now.)

Bummer :( Hope this gets sorted at some point.

> > The latter only seems to work when the "base" entity was a generic switch, but not when it is something specific like e.g. a lamp in this case.
> 
> That's a limitation on HA Core side, as the integration responsible for that is called "Switch as X" and only allows you to "convert" switches.

I might have a look at that at some point. Trying to lift this restriction could mitigate this inconvenience at least a bit.

PR #2998, opened 2024-02-23
Improve IKEA battery percentage firmware detection

### LINKED ISSUE #2977: ZHA battery doubling on some IKEA devices still present
> Nice. New issue would indeed be good for that, but do you roughly remember the version that it showed off the top of your head? (e.g. 2.3.075 or 24.4.5)
> 
> And which device is affected?

_Originally posted by @TheJulianJES in https://github.com/zigpy/zha-device-handlers/issues/2948#issuecomment-1938164896_

image

I have multiple of them (old and recently bought) . For some the battery seems correct (but not sure, because the battery is old in there).

No idea what `sw_build_id` is for these devices. Diagnosis data will follow this eventing.

TheJulianJES:
## Proposed change
This improves the detection mechanism for IKEA firmware.

Previously, it detected versions that started with `24` or higher as new. Everything else was treated as old and doubled.
This was fine mostly. However, one IKEA remote already uses `2.4.x` firmware which no longer requires doubling.
So this PR now checks the first two parts of the `sw_build_id` (split by `.`) for the firmware version.
Possible upcoming formattings of that string are treated as new too (so no doubling).

Basically, the doubling is now not applied if the firmware is 2.4.x or newer.

## Additional information
Fixes https://github.com/zigpy/zha-device-handlers/issues/2977

## Checklist

### DIFF
--- a/zhaquirks/ikea/__init__.py
+++ b/zhaquirks/ikea/__init__.py
@@ -228,7 +228,7 @@ class DoublingPowerConfigClusterIKEA(CustomCluster, PowerConfiguration):
 
                 # new firmware is either 24.4.5 or above, or 2.4.5 or above
                 # old firmware is 2.3.x or below
-                return first_part >= 24 or (first_part >= 2 and second_part >= 4)
+                return first_part >= 3 or (first_part >= 2 and second_part >= 4)
             except ValueError:
                 _LOGGER.warning(
                     "sw_build_id is not a number: %s for device %s",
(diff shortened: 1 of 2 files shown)


zhaquirks/ikea/__init__.py
```diff
@@ -209,26 +209,36 @@ async def bind(self):
         await self.endpoint.basic.read_attributes([Basic.AttributeDefs.sw_build_id.id])
         return result
 
-    def _is_firmware_old(self):
-        """Checks if firmware is old or unknown."""
+    def _is_firmware_new(self):
+        """Checks if new firmware is installed that does not require battery doubling."""
         # get sw_build_id from attribute cache if available
         sw_build_id = self.endpoint.basic.get(Basic.AttributeDefs.sw_build_id.id, None)
```
TheJulianJES, line 215:
```suggestion
        sw_build_id = self.endpoint.basic.get(Basic.AttributeDefs.sw_build_id.id)
```
For a future PR: `None` is already the default for zigpy clusters, so we can remove that.

PR #3004, opened 2024-02-26
Fix Tuya 3 gang switch TS0601 `_TZE200_vhy3iakz`

rasmus-hagberg:
## Proposed change
Mark the TS0601 _TZE200_vhy3iakz device as a 3 gang switch instead of a 1 gang switch.

## Additional information
I purchased a 3 gang "Tuya Smart ZigBee Push Switch" from [here](https://www.expert4house.com/en/smart-home/wi-fi-switches/tuya-smart-zigbee-push-switch#/40-size-3_gang). In HA the manufacturer is "_TZE200_vhy3iakz" and the model is "TS0601". It uses the ts0601_switch.TuyaSingleSwitchTO quirk, and so I was only able to control the first switch.

Multi gang quirks for tuya switches were added in [this PR](https://github.com/zigpy/zha-device-handlers/pull/1166), which notes that "For those devices whose number of gangs is not clear, they have been defined as 1-gang devices until someone report correct info" and "we have doubts about [the _TZE200_vhy3iakz] device".

I tried the changes by loading them as a custom quirk, and it works fine for me. I am now able to control all three switches remotely, and the status is updated properly in HA.

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_switch.py
+++ b/zhaquirks/tuya/ts0601_switch.py
@@ -84,7 +84,6 @@ class TuyaSingleSwitchTO(TuyaSwitch):
         MODELS_INFO: [
             ("_TZE200_amp6tsvy", "TS0601"),
             ("_TZE200_oisqyl4o", "TS0601"),
-            ("_TZE200_vhy3iakz", "TS0601"),  # ¿1 or 4 gangs?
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
Thanks!

PR #3032, opened 2024-03-11
Add a helper to create devices from v2 quirks

dmulcahey:
## Proposed change

This PR adds a helper for creating devices from v2 quirks. It will not create a complete copy of the device as we no longer have a signature (we don't have the info at all any more). I think what it creates gives us what we need to test what the quirk is actually meant to do. Definitely open to suggestions...

## Checklist


dmulcahey:
I am going to merge this for now to unblock another PR. We can address whatever concerns we have with this approach in a followup

PR #3050, opened 2024-03-18
Convert Sonoff buttons to v2 quirk

TheJulianJES:
## Proposed change
This converts both existing Sonoff button quirks to a combined v2 quirk. This way we can avoid adding other signature variations.
The "signature match test" is also removed, as it doesn't make much sense with a v2 quirk(?)

## Additional information
Also addresses https://github.com/home-assistant/core/issues/104548#issuecomment-2004994990

## Checklist


dmulcahey:
Why not convert these to v2 and forget the sig stuff?

TheJulianJES:
Right, perfect candidate for that. I'll convert this tomorrow.

zhaquirks/sonoff/button.py
```diff
@@ -1,70 +1,47 @@
...
-                    Identify.cluster_id,
-                    OnOff.cluster_id,
-                ],
+(
+    add_to_registry_v2("eWeLink", "SNZB-01P").device_automation_triggers(
+        {
+            (SHORT_PRESS, BUTTON): {
+                COMMAND: COMMAND_TOGGLE,
+                CLUSTER_ID: 6,
+                ENDPOINT_ID: 1,
```
TheJulianJES: Diff looks bad for all changes, but we can probably combine the two different Sonoff button v2 quirks, as they essentially use the same device triggers. Currently, one quirk just narrows it down to endpoint and cluster which we probably don't need.

TheJulianJES: Done in https://github.com/zigpy/zha-device-handlers/pull/3050/commits/55aa0f270c3796d38f5992113502e7dbf8b13a80

TheJulianJES:
Codecov/project coverage decrease is expected, as a lot of code is removed.
This should be good as-is now. Merging to get it in the next quirks release for HA beta.

PR #3088, opened 2024-04-04
Convert IKEA Starkvind fan speed attribute

freundTech:
## Proposed change

IKEA Starkvind has two similar attributes: `fan_mode` and `fan_speed`.

`fan_mode` can take a value from 0-50 where 0 is off, 1 is auto and 10-50 are manual speed modes.
`fan_speed` can take a number from 10-50 or 0, where 0 is off and 10-50 is the fan speed. `fan_speed` is read-only.

If `fan_mode` is set to a manual mode that is not a multiple of 5 it is rounded to a multiple of 5, which can be seen by reading the `fan_speed` attribute. For this reason The current quirk scales that 10-50 range to a 2-10 range, eliminating the undefined and duplicate (rounded) values.

This patch applies the same downscaling to the `fan_speed` attribute, restoring the correspondence between `fan_mode` and `fan_speed` values.

Native values: Set `fan_mode` to 30 => `fan_speed` reads 30
Current state: Set `fan_mode` to 6 => `fan_speed` reads 30
With patch: Set `fan_mode` to 6 => `fan_speed` reads 6

## Additional information

This will be followed up by a PR in home-assistant to eventually fix https://github.com/home-assistant/core/issues/97440.
Because home-assistant currently only uses the `fan_mode` and not the `fan_speed` attribute this does not depend on the corresponding home-assistant PR. Only the other way around.

I also changed the `<` and `>` to `<=` and `>=` to improve understandability of the code by mentioning the limit values and replaced `/` with `//` to prevent introducing floating point values.

## Checklist

### DIFF
--- a/tests/test_ikea.py
+++ b/tests/test_ikea.py
@@ -10,6 +10,7 @@ from zigpy.zcl.clusters.measurement import PM25
 from tests.common import ClusterListener
 import zhaquirks
 import zhaquirks.ikea.starkvind
+from zhaquirks.ikea.starkvind import IkeaAirpurifier
 
 zhaquirks.setup()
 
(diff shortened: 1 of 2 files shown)


freundTech:
Added a test to make codecov happy.

freundTech:
Alternatively it would also be possible to convert neither `fan_mode`, nor `fan_speed` and let the api consumer (Home Assistant in this case) deal with mapping that to meaningful settings.

freundTech:
Any update on the review?
@TheJulianJES

TheJulianJES (approved):
Sorry for the long wait. The quirk part should be good-to-go IMO.
Also thanks for adding tests!

(In the future, the quirk should be updated to use new-style zigpy `AttributeDefs` though)

PR #3111, opened 2024-04-15
Update Schneider quirks using spec docs

dflemstr:
## Proposed change

This is a start of using the leaked Schneider Electric specification docs posted by @uvNikita in #1705, linked here again for convenience: https://drive.google.com/file/d/1rM9iVhAB-AlWwWhvpEtcReSpEATi7aVH/view

So far I have looked at the light related specs and updated the attributes with the information for the specs docs.  More improvements could probably be made by incorporating info from the other device types too, but I only have light-related devices in my home (dimmers, switches, etc) so I can't test any other changes right now.

The comments have a number of grammar issues etc, but I copied them verbatim from the spec, so if there are any ambiguous statements, your guess is as good as mine!

## Additional information

This PR renames some python variables/constant names to match the spec.  I doubt that anything is using these in code already; if so, this could be a breaking change.  I suppose @cspurk is the only person I know of that might be using this?

## Checklist

### DIFF
--- a/zhaquirks/schneiderelectric/__init__.py
+++ b/zhaquirks/schneiderelectric/__init__.py
@@ -7,7 +7,8 @@ from zigpy import types as t
 from zigpy.quirks import CustomCluster
 from zigpy.zcl import foundation
 from zigpy.zcl.clusters.closures import WindowCovering
-from zigpy.zcl.clusters.general import Basic
+from zigpy.zcl.clusters.general import Basic, OnOff
+from zigpy.zcl.clusters.lighting import Ballast
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 SE_MANUF_NAME = "Schneider Electric"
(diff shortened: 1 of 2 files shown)


TheJulianJES (approved):
I think this looks pretty good, thanks!

uvNikita:
@dflemstr thank you for you work on this.
I have one question, my dimmer presents itself as "NHROTARY/UNIDIM/1" and it doesn't seem like these new changes apply to it. Could it be because it needs to be added to this file? https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/schneiderelectric/dimmers.py

If I understand it correctly, the signature/name of my dimmer is listed here: 
image

dflemstr:
@uvNikita You are right that extra changes are needed for your device to be supported. Another `class` should be created for your device, since it has a slightly different device signature.  I didn't add all the devices from the spec, I only adjusted the ones that already were in `zha-device-handlers` to the spec.

dflemstr:
I added support for your device like this in zigbee2mqtt to give you an idea: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7454

I'm not sure if I will have time to make the equivalent change here though since adding devices in `zha-device-handlers` is a lot more involved compared to zigbee2mqtt

prilly-dev:
Just wanted to add that ELKO is selling this devices under its name after it was acquired by schneider. I hope we can have this UNIDIM device included also

PR #3130, opened 2024-05-01
Adds BSEED/Tuya dimmer switch variant `_TZE204_vevc4c6g`

Elyasin:
## Proposed change
Simple addition to the model info. I have this switch at home as well as the `_TZE200_3p5ydos3` they are the same practically.

## Additional information
Used at home and tested it.

## Checklist


TheJulianJES (approved):
Thanks!

PR #3145, opened 2024-05-10
Quirk for Ikea TRETAKT Smart Plug

jasimancas:
## Proposed change

The new Ikea plug exposes entities that are not correct and do not work, this quirk solves them by exposing only the parameters of a plug.

image

## Checklist

### DIFF
--- a/zhaquirks/ikea/tretaktplug.py
+++ b/zhaquirks/ikea/tretaktplug.py
@@ -1,4 +1,5 @@
 """Tretakt Plug Quirk."""
+
 from zigpy.profiles import zgp, zha, zll
 from zigpy.quirks import CustomDevice
 from zigpy.zcl.clusters.general import (
@@ -24,6 +25,7 @@ from zhaquirks.const import (
 )
 from zhaquirks.ikea import IKEA, IKEA_CLUSTER_ID, WWAH_CLUSTER_ID
 
+
 class TretaktPlug(CustomDevice):
     """Tretakt Smart plug."""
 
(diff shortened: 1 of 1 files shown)


MattWestb:
IKEA is using light level with On/Off then doing direct pairing of there controllers and they looks using the same cluster setting doing coordinator paring. 
Very likely its the same with ÅSKVÄDER Zigbee module but i have not looking on it then its no problem for me.

### Good work done !!!

MattWestb:
@TheJulianJES Shall we not having the manufacture cluster in the replacement so not loosing future possible functions ?

jasimancas:
BTW, new cluster in this device: 0xFC85

MattWestb:
Yes i was seen it and its not used in any github repros for Zigbee.
IKEA is some time making manufacture cluster then its not in the current Zigbee ZCL but is coming in the next like Symfonisk gen 2 have one matter switch cluster  (on 2 endpoints) that is not in Zigbee ZCL but they only need changing the number and they can piping it thru the matter bridge and all is 95% matter compatible or they is fixing it the the controller firmware with OTA.
https://github.com/zigpy/zha-device-handlers/blob/bfbfb4876d1811440b4a05cac9e3b83311d194a8/zhaquirks/ikea/__init__.py#L18-L19

jasimancas:
Understood, maybe I should add it to the init?

jasimancas:
@MattWestb Is something else expected from me or am I just waiting for the merge?

MattWestb:
Its depend of the maintainers if they like have some changes made or if no needed the shall merging it sooner or later.

TheJulianJES:
Thanks for the PR!

I think we can simplify this with quirks v2 now, so we don't need to match the whole signature.
If you can, please test the quirk file from this PR:
- https://github.com/zigpy/zha-device-handlers/pull/3174

You can just install it as a custom quirk. See if the same behavior (of removing the `LevelControl` cluster) is achieved (for both plugs).

@MattWestb You're also welcome to test that quirk of course!

jasimancas:
Continue in: https://github.com/zigpy/zha-device-handlers/pull/3174

PR #3144, opened 2024-05-10
IKEA STYRBAR V3

MattWestb:
IKEA STYRBAR V3 for FW 2.4.11 firmware

## Proposed change
  Adding V3 for firmware 2.4.11 that adding real group binding.

DA for long press up and down release is broken for most version then the remote is sending the same stop command (atleast from the last production) = release up also for down.
If like getting it working the same logic as being used for open/close button must being implanted.

## Additional information
Fixes https://github.com/zigpy/zha-device-handlers/issues/3135

Tested in my production system plus 2 test  systems and also one user in the linked issue.

If binding the scene cluster is the "IKEA scene" working on most IKEA (C)WS lights out of the box with right and left buttons !!!

## Checklist

- [X] The changes are tested and work correctly
- [X] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

PR #3175, opened 2024-05-28
Convert IKEA Styrbar quirks to v2 quirk

TheJulianJES:
## Proposed change
Converts existing IKEA Styrbar quirks to a v2 quirk.
This also adds support for new signatures of the Styrbar, as quirks v2 don't need to match by endpoint/clusters (signature) anymore.

## Additional information
The only downside I see is that we use `DoublingPowerConfig2AAACluster` for all versions of the remote, even if it's a "newer signature". I think we should at least change the default to not double for that first, before we merge this.
Related PR for that:
- https://github.com/zigpy/zha-device-handlers/pull/3176

Would supersede #3144 if we want to go this route.
The test `test_quirk_device_automation_triggers_unique` also still needs updating that should be done in a separate PR.
It doesn't deal properly with v2 quirks.

## Checklist

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -641,12 +641,6 @@ KNOWN_DUPLICATE_TRIGGERS = {
             (zhaquirks.aurora.aurora_dimmer.COLOR_DOWN, const.LEFT),
         ],
     ],
-    zhaquirks.ikea.fourbtnremote.IkeaTradfriRemoteV1: [
-        [
-            (const.LONG_RELEASE, const.DIM_UP),
-            (const.LONG_RELEASE, const.DIM_DOWN),
-        ]
-    ],
     zhaquirks.paulmann.fourbtnremote.PaulmannRemote4Btn: [
         [
             (const.LONG_RELEASE, const.BUTTON_1),
@@ -666,6 +660,7 @@ KNOWN_DUPLICATE_TRIGGERS = {
 }
 
 
+# XXX: Test does not handle v2 quirks
 @pytest.mark.parametrize(
     "quirk",
     [q for q in ALL_QUIRK_CLASSES if getattr(q, "device_automation_triggers", None)],
--- a/zhaquirks/ikea/fourbtnremote.py
+++ b/zhaquirks/ikea/fourbtnremote.py
@@ -1,16 +1,5 @@
 """Device handler for IKEA of Sweden TRADFRI remote control."""
-from zigpy.profiles import zha
-from zigpy.quirks import CustomDevice
-from zigpy.zcl.clusters.general import (
-    Basic,
-    Identify,
-    LevelControl,
-    OnOff,
-    Ota,
-    PollControl,
-    PowerConfiguration,
-)
-from zigpy.zcl.clusters.lightlink import LightLink
+from zigpy.quirks.v2 import add_to_registry_v2
 
 from zhaquirks.const import (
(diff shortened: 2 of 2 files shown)


MattWestb:
OnOff and Level looks working OK then doing fast testing.
Scene is not working then you was deleting the "IKEA Scene" cluster replacement so need adding it back for getting some functions on left and right buttons.
Have not testing battery or deeper.

I like V2 quirks its much cleaner !!!

TheJulianJES:
Ah, yeah.
I've re-added the `ScenesCluster` now. Can you try the updated variant?

MattWestb:
Was updating it from your git and now i getting "scene commands" OK.
But i think long press up and down is not working OK then we is not using all parameter (not in the device log or zha_events but have not testing the DA):
```
event_type: zha_event
data:
  device_ieee: 8c:f6:81:ff:fe:1e:f3:70
  unique_id: 8c:f6:81:ff:fe:1e:f3:70:1:0x0008
  device_id: 64a05efdf2bd3e0e6e80f89e7c7bf5b4
  endpoint_id: 1
  cluster_id: 8
  command: move
  args:
    - 1
    - 83
    - 0
    - 0
  params:
    move_mode: 1
    rate: 83
    options_mask: 0
    options_override: 0
origin: LOCAL
time_fired: "2024-05-29T15:30:34.903406+00:00"
context:
  id: 01HZ2FJY8QG68NGJZAKA3ZSSED
  parent_id: null
  user_id: null

 event_type: zha_event
data:
  device_ieee: 8c:f6:81:ff:fe:1e:f3:70
  unique_id: 8c:f6:81:ff:fe:1e:f3:70:1:0x0008
  device_id: 64a05efdf2bd3e0e6e80f89e7c7bf5b4
  endpoint_id: 1
  cluster_id: 8
  command: move_with_on_off
  args:
    - 0
    - 83
  params:
    move_mode: 0
    rate: 83
origin: LOCAL
time_fired: "2024-05-29T15:30:39.254780+00:00"
context:
  id: 01HZ2FK2GPKNGR175CY9APVBQ9
  parent_id: null
  user_id: null
```
Also some users dont like that we is not remember the long press for the release like Open/Close is doing here:

https://github.com/TheJulianJES/zha-device-handlers/blob/9ed616a33e6f973f57f1aec37d67cabcff754aaa/zhaquirks/ikea/opencloseremote.py#L47-L76

TheJulianJES:
Ok, thanks. But those aren't regressions, right? So no new bugs introduced with this?

TheJulianJES:
Codecov project coverage decrease is expected, as the old quirk had quite a few lines and is now fully removed with this.

MattWestb:
> Ok, thanks. But those aren't regressions, right? So no new bugs introduced with this?

I think it was not working OK then we was chaining par/args and it was very much testing for all IKEA controllers then doing that and my memory . . .

I dont have any problems then i using them as Light controller with device and group bindings so its working then ZHA is not OK.

MattWestb:
Thanks for fixing and closing my V3 that is out of date now !!!

MattWestb:
Only for information:

Dirigeras feed have now updated to 0x02040016 and the quirk is working great and then the new firmware have getting one new private cluster but sadly no group binding is being added like all newer IKEA gen 3 controller is having.

Edit: 2 of 4 inn production was loosing there binding to deices and must being redone  for working after the upgrade.

PR #3195, opened 2024-06-09
Add quirks for new Sinopé Load Controller

claudegel:
## Proposed change

- Add quirks for the new load controller RM3250ZB
- Code format

## Additional information

New load controller have cluster 0x0002 that is missing in old load controller

## Checklist


zhaquirks/sinope/switch.py
```diff
@@ -326,6 +327,67 @@ class SinopeTechnologiesLoadController(CustomDevice):
     }
 
 
+class SinopeTechnologiesNewLoadController(CustomDevice):
```
javicalle: IMHO bad naming.
I would prefer `SinopeTechnologiesLoadController_var02` or similar.
Otherwise you can get sometime a `SinopeTechnologiesNewNewLoadController` 😉

claudegel: better like this SinopeTechnologiesLoadController_V2

TheJulianJES (approved):
Thanks!

PR #3211, opened 2024-06-17
Purge quirks when loading custom quirks

puddly:
## Proposed change

Currently, v1 and v2 quirks are loaded multiple times due to us using `importlib` when the integration is reloaded. For v2 quirks, this will cause a startup error. This PR purges the device registries when quirks are setup.

## Checklist


puddly:
This needs more work to work with custom v2 quirks. I think it may need to go into zigpy instead.

PR #3232, opened 2024-06-27
Replace Inovelli quirks with v2 quirks

codyhackw:
## Proposed change

Updating Inovelli Quirks to the v2 formatting, as well as an update to add the leading/training edge parameter for the VZM36, but removing it from the VZM31SN per Inovelli's request (it's still configurable via button press combo at the switch, which is the recommended method).

## Checklist


TheJulianJES:
Oooh, awesome that you're doing this!
There are some very small changes to the quirks v2 API, example here: https://github.com/zigpy/zha-device-handlers/pull/3280
It'll land with the next HA beta, so should also be in the final 2024.8.0 release.

Also, there were some repo changes, so you probably need to rebase this (or merge `dev` into your branch).

TheJulianJES:
This also seems to supersede this PR then:
- https://github.com/zigpy/zha-device-handlers/pull/3020

codyhackw:
> Oooh, awesome that you're doing this! There are some very small changes to the quirks v2 API, example here: #3280 It'll land with the next HA beta, so should also be in the final 2024.8.0 release.
> 
> Also, there were some repo changes, so you probably need to rebase this (or merge `dev` into your branch).

Looks like that was just the change from add_to_registry_v2 to QuirkBuilder being used and add_to_registry() called at the end now? If so, should be fixed now. I believe I chose the right option there on merging dev in, but if not, just let me know and I can get that sorted too.

TheJulianJES:
Yep, that's fine. Pre-commit just complains about some missing docstrings now.
Tests are a bit different with quirks v2, since we no longer have specific classes for the quirks. I'll have to look at the specific test(s), but you should be able to use `device = zigpy_device_from_v2_quirk("manuf", "model")` to get a device with the custom clusters in a test.

dmulcahey:
IMG_8210

Just comments missing

TheJulianJES:
And the custom cluster on endpoint 3 isn't replaced yet (when it exists). But we can't add `.replaces()` for that endpoint yet, since `RemovesMetadata` assumes the endpoint actually exists (or zigpy crashes otherwise). That seems like an issue.
I'll post more details later and/or fix the zigpy issue if it's one.

dmulcahey:
> And the custom cluster on endpoint 3 isn't replaced yet (when it exists). But we can't add `.replaces()` for that endpoint yet, since `RemovesMetadata` assumes the endpoint actually exists (or zigpy crashes otherwise). That seems like an issue. I'll post more details later and/or fix the zigpy issue if it's one.

Ooo… I wonder if we should have another implementation `replace_if_present` or if we should update the existing one 🤔

TheJulianJES:
> Ooo… I wonder if we should have another implementation `replace_if_present` or if we should update the existing one 🤔

Up until seconds ago, I thought we can just update the existing one. It adds `RemovesMetadata` and `AddsMetadata`.
If we call `removes` on a non-existent cluster (on a valid endpoint), nothing will happen already. We ignore that case.
But we error out if we try to remove a cluster on a non-existent endpoint.

To ignore this, we can just check if the endpoint exists:
https://github.com/zigpy/zigpy/compare/dev...TheJulianJES:tjj/quirks_v2_removes_non_exist_endpoint (ignore test line change here)

But the above change is not enough. When `AddsMetadata` is processed, it'll also fail on the non-existent endpoint.
If we also check there that the endpoint exists, a change like this would work: https://github.com/zigpy/zigpy/compare/dev...TheJulianJES:tjj/quirks_v2_removes_non_exist_endpoint_and_add

Currently thinking about if we want to error out when `removes`, `replaces`, or `adds` is called on a non-existent endpoint. Maybe? Like, `removes` already completely ignores if a cluster that is to be replaced is missing. It just errors out when the endpoint is missing too.
But with `adds` (also used by `replaces), I'm not sure. We may want to error out here, as the cluster will be missing if the endpoint is missing (obviously). Silently failing here might not be a good option.
But it's also unlikely to ever do this unintentionally? Like, at least when testing, you'll notice that the cluster you're trying to replace/add is missing.

Maybe we want a separate method after all? Not sure yet...

EDIT: We can use the following for this PR:
- https://github.com/zigpy/zigpy/pull/1441

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,14 @@
...
-                ],
-            },
-        },
-    }
-
-    device_automation_triggers = INOVELLI_AUTOMATION_TRIGGERS
+(
+    QuirkBuilder("Inovelli", "VZM31-SN")
+    .replaces(InovelliVZM31SNCluster)
+    .replaces(InovelliVZM31SNCluster, endpoint_id=2, cluster_type=ClusterType.Client)
```
TheJulianJES: I think we should be able to use `replace_cluster_occurrences` from https://github.com/zigpy/zigpy/pull/1441 for the custom output clusters on all endpoints.

Haven't tested this, but this might work (if zigpy is new enough on the quirks repo, might need to be bumped?):
```suggestion
    .replace_cluster_occurrences(InovelliVZM31SNCluster)
```
(Replaces both client and server clusters by default. Set `replace_server_instances=False` to only replace client clusters.)

codyhackw:
just a bump to avoid closing...will get updates together this weekend!

zhaquirks/inovelli/VZM36.py
```diff
@@ -1,140 +1,18 @@
...
-                OUTPUT_CLUSTERS: [GreenPowerProxy.cluster_id],
-            },
-        },
-    }
-
-    device_automation_triggers = INOVELLI_AUTOMATION_TRIGGERS
+(
+    QuirkBuilder("Inovelli", "VZM36")
+    .replaces(InovelliVZM36LightCluster)
+    .replaces(InovelliVZM36FanCluster, endpoint_id=2, cluster_type=ClusterType.Client)
```
TheJulianJES: `InovelliVZM36FanCluster` was in `INPUT_CLUSTERS` before, so the cluster type should be the default of `Server` (for input), right? I.e. this:

```suggestion
    .replaces(InovelliVZM36FanCluster, endpoint_id=2)
```

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
-                    WWAH_CLUSTER_ID,
-                ],
-                OUTPUT_CLUSTERS: [
-                    Ota.cluster_id,
-                ],
-            },
-            2: {
-                PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: DeviceType.DIMMER_SWITCH,
-                INPUT_CLUSTERS: [Basic.cluster_id, Identify.cluster_id],
```
TheJulianJES, line 123: `Groups` and `Scenes` input/server clusters were removed for the VZM31-SN (in `InovelliVZM31SNv13`) on ep 2.
Should we also do that in the v2 quirk?

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
-                    Identify.cluster_id,
-                    OnOff.cluster_id,
-                    LevelControl.cluster_id,
-                    InovelliVZM31SNCluster,
-                ],
-            },
-            3: {
-                PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: DeviceType.DIMMER_SWITCH,
-                INPUT_CLUSTERS: [Basic.cluster_id, Identify.cluster_id],
```
TheJulianJES, line 134: `Groups` and `Scenes` input/server clusters were removed for the VZM31-SN (in `InovelliVZM31SNv13`) on ep 3.
Should we also do that in the v2 quirk?

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
-                    WWAH_CLUSTER_ID,
-                ],
-                OUTPUT_CLUSTERS: [
-                    Ota.cluster_id,
-                ],
-            },
-            2: {
-                PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: DeviceType.DIMMER_SWITCH,
-                INPUT_CLUSTERS: [Basic.cluster_id, Identify.cluster_id],
```
TheJulianJES, line 228: `Groups` and `Scenes` input/server clusters were removed for the VZM31-SN (in `InovelliVZM31SNv12`) on ep 2.
Should we also do that in the v2 quirk?

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
-                    InovelliVZM31SNCluster,
-                    WWAH_CLUSTER_ID,
-                ],
-                OUTPUT_CLUSTERS: [
-                    Identify.cluster_id,
-                    OnOff.cluster_id,
-                    LevelControl.cluster_id,
-                    Ota.cluster_id,
-                    InovelliVZM31SNCluster,
-                ],
```
TheJulianJES, line 316: `Identify`, `OnOff`, `LevelControl`, and `InovelliVZM31SNCluster` were added for the VZM31-SN (in `InovelliVZM31SNv11`) on ep 1.
Should we also do that in the v2 quirk? Maybe not..?

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
```
TheJulianJES, line 403: `OnOff`, `LevelControl`, and `InovelliVZM31SNCluster` were added for the VZM31-SN (in `InovelliVZM31SNv10`) on ep 1.
Should we also do that in the v2 quirk? Maybe not..?

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
```
TheJulianJES, line 484: `OnOff`, `LevelControl`, and `InovelliVZM31SNCluster` were added for the VZM31-SN (in `InovelliVZM31SNv9`) on ep 1.
Should we also do that in the v2 quirk? Maybe not..?

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
-                    Diagnostic.cluster_id,
-                    InovelliVZM31SNCluster,
-                ],
-                OUTPUT_CLUSTERS: [
-                    Identify.cluster_id,
-                    OnOff.cluster_id,
-                    LevelControl.cluster_id,
-                    Ota.cluster_id,
-                    InovelliVZM31SNCluster,
-                ],
```
TheJulianJES, line 562: `OnOff`, `LevelControl`, and `InovelliVZM31SNCluster` were added for the VZM31-SN (in `InovelliVZM31SN`) on ep 1.
Should we also do that in the v2 quirk? Maybe not..?

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
-                INPUT_CLUSTERS: [
-                    Basic.cluster_id,
-                    Identify.cluster_id,
-                    Groups.cluster_id,
-                    Scenes.cluster_id,
-                    OnOff.cluster_id,
-                    LevelControl.cluster_id,
-                    Metering.cluster_id,
-                    ElectricalMeasurement.cluster_id,
-                    Diagnostic.cluster_id,
```
TheJulianJES, line 553: `Diagnostic` was added for the VZM31-SN (in `InovelliVZM31SN`) on ep 1.
Should we also do that in the v2 quirk? Maybe not..?

zhaquirks/inovelli/VZM31SN.py
```diff
@@ -1,578 +1,12 @@
...
-                ],
-            },
-        },
-    }
-
-    replacement = {
-        ENDPOINTS: {
-            1: {
-                PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: DeviceType.DIMMABLE_LIGHT,
```
TheJulianJES, line 464: Some of these devices also provide `DIMMER_SWITCH` as the device type. All VZM31-SN v1 quirks change the device type to `DIMMABLE_LIGHT`.
We should likely do that as well in the v2 quirk, although I don't think it should affect anything in ZHA.

zhaquirks/inovelli/VZM35SN.py
```diff
@@ -1,387 +1,12 @@
...
-                    Basic.cluster_id,
-                    Identify.cluster_id,
-                    Groups.cluster_id,
-                    Scenes.cluster_id,
-                ],
-                OUTPUT_CLUSTERS: [
-                    Identify.cluster_id,
-                    OnOff.cluster_id,
-                    LevelControl.cluster_id,
-                    InovelliVZM35SNCluster,
```
TheJulianJES, line 115: `InovelliVZM35SNCluster` was added for the VZM35-SN (in `InovelliVZM35SN`) on ep 2.
Should we also do that in the v2 quirk?

zhaquirks/inovelli/VZM35SN.py
```diff
@@ -1,387 +1,12 @@
...
```
TheJulianJES, line 228: `InovelliVZM35SNCluster` was added for the VZM35-SN (in `InovelliVZM35SNv5`) on ep 2.
Should we also do that in the v2 quirk?

zhaquirks/inovelli/VZM35SN.py
```diff
@@ -1,387 +1,12 @@
...
```
TheJulianJES, line 244: `InovelliVZM35SNCluster` was added for the VZM35-SN (in `InovelliVZM35SNv5`) on ep 3.
Should we also do that in the v2 quirk?

zhaquirks/inovelli/VZM35SN.py
```diff
@@ -1,387 +1,12 @@
...
```
TheJulianJES, line 359: `InovelliVZM35SNCluster` was added for the VZM35-SN (in `InovelliVZM35SNv7`) on ep 2.
Should we also do that in the v2 quirk?

zhaquirks/inovelli/VZM35SN.py
```diff
@@ -1,387 +1,12 @@
...
```
TheJulianJES, line 375: `InovelliVZM35SNCluster` was added for the VZM35-SN (in `InovelliVZM35SNv7`) on ep 3.
Should we also do that in the v2 quirk?

TheJulianJES (commented):
More comments on the differences between the v1/v2 quirks.

Do note that some clusters were in `INPUT_CLUSTERS` (`.replaces(cluster)` is enough) and others in `OUTPUT_CLUSTERS` (requires `.replaces(cluster, cluster_type=ClusterType.Client)` for v2 quirks).

zhaquirks/inovelli/VZM31SN.py
codyhackw: Would there be a good reason to not expose those if they're in the signature already? I recognize that it's a change here from v1 to v2, but this ends up closer to the switch as-is?

zhaquirks/inovelli/VZM31SN.py
codyhackw: Would there be a good reason to not expose those if they're in the signature already? I recognize that it's a change here from v1 to v2, but this ends up closer to the switch as-is?

zhaquirks/inovelli/VZM31SN.py
codyhackw: Would there be a good reason to not expose those if they're in the signature already? I recognize that it's a change here from v1 to v2, but this ends up closer to the switch as-is?

zhaquirks/inovelli/VZM31SN.py
codyhackw: I think at this point, these can be ignored. That v11 signature would've only been for firmware pre v2.08, and from looking at their site, 2.08 came out way back in November of 2022.

zhaquirks/inovelli/VZM31SN.py
codyhackw: I think at this point, these can be ignored. The v11 signature would've only been for firmware pre v2.08, and from looking at their site, 2.08 came out way back in November of 2022. (this one being v10 would be before that)

zhaquirks/inovelli/VZM31SN.py
codyhackw: I think at this point, these can be ignored. The v11 signature would've only been for firmware pre v2.08, and from looking at their site, 2.08 came out way back in November of 2022. (this one being v9 would be before that)

zhaquirks/inovelli/VZM31SN.py
codyhackw: I think at this point, these can be ignored. The v11 signature would've only been for firmware pre v2.08, and from looking at their site, 2.08 came out way back in November of 2022. (this one being original would be before that)

zhaquirks/inovelli/VZM31SN.py
codyhackw: I think at this point, these can be ignored. The v11 signature would've only been for firmware pre v2.08, and from looking at their site, 2.08 came out way back in November of 2022. (this one being original would be before that)

zhaquirks/inovelli/VZM31SN.py
codyhackw: I think they fixed this pretty early on in firmware, looks like the v10 quirk was where I see the signature changed. Do we think this would still be necessary vs just updating the switch if you had one of the very early ones and for whatever reason never updated? The main reason was to get the device to show up as a light and be able to change level in HA by default instead of people just getting the on/off toggle if I remember correctly.

zhaquirks/inovelli/VZM35SN.py
codyhackw: This is a good catch, these are used for button press events to come through. It actually looks like the firmware takes care of this for the other switches, but it looks like it's on the INPUT clusters instead of OUTPUT for the VZM35-SN. Fixing now.

zhaquirks/inovelli/VZM35SN.py
codyhackw: This is a good catch, these are used for button press events to come through. It actually looks like the firmware takes care of this for the other switches, but it looks like it's on the INPUT clusters instead of OUTPUT for the VZM35-SN. Fixing now.

zhaquirks/inovelli/VZM35SN.py
codyhackw: This is a good catch, these are used for button press events to come through. It actually looks like the firmware takes care of this for the other switches, but it looks like it's on the INPUT clusters instead of OUTPUT for the VZM35-SN. Fixing now.

zhaquirks/inovelli/VZM35SN.py
codyhackw: This is a good catch, these are used for button press events to come through. It actually looks like the firmware takes care of this for the other switches, but it looks like it's on the INPUT clusters instead of OUTPUT for the VZM35-SN. Fixing now.

zhaquirks/inovelli/VZM35SN.py
codyhackw: This is a good catch, these are used for button press events to come through. It actually looks like the firmware takes care of this for the other switches, but it looks like it's on the INPUT clusters instead of OUTPUT for the VZM35-SN. Fixing now.

zhaquirks/inovelli/VZM36.py
codyhackw: Committed as suggested, yeah this would be an input cluster, I had them backwards I think.

zhaquirks/inovelli/VZM31SN.py
TheJulianJES: > Would there be a good reason to not expose those if they're in the signature already? 

Yeah, not really.

zhaquirks/inovelli/VZM31SN.py
TheJulianJES: Ah, switch vs light entities is a big issue. Since we worked around in the past, we should still do that.

TheJulianJES: Ugh, I thought just doing `replaces_endpoint(1, device_type=zha.DeviceType.DIMMABLE_LIGHT)` would be enough, but that removes all device reported clusters from that endpoint.

That should probably be addressed in zigpy. I doubt we ever want that.

TheJulianJES: This should work with the next zigpy release that includes https://github.com/zigpy/zigpy/pull/1602. Calling the code in the above comment will only replace the endpoint type and profile (default is ZHA profile) then, not touch the existing device reported clusters anymore.

We'll need to address quirks compatibility with zigpy 0.80.0 first though: https://github.com/zigpy/zha-device-handlers/pull/4088
(and then bump the quirks repo again to a newer zigpy version including the linked PR).

TheJulianJES:
Since the above comment threads are becoming a bit hard to go through, I'll go over the two main points left for this PR:

1. The device type was changed on older firmware versions to correctly create (dimmable) light entities in HA, instead of only switch entity. Relevant comment thread with solution (when zigpy is released + bumped in this repo): https://github.com/zigpy/zha-device-handlers/pull/3232#discussion_r2106079434
Since newer versions all seem to use the `DIMMABLE_LIGHT` type, we can just always change that endpoint type to it.

3. Some older quirks adding `Identify`, `OnOff`, `LevelControl`, and `InovelliVZM31SNCluster`, ... on endpoint 1. These seem crucial to the operation, so I wonder how the device would have worked without presenting them, but since we fixed this with v1 quirks, we should also do this with v2 quirks. And it should be easy:

   We can likely just `adds(OnOff.cluster_id)` and so on (possibly specifying `endpoint_id=x` if it's not the default of `1`).
If the clusters are already present, we don't do anything, otherwise, we add them.
Since newer versions all seem to contain those clusters, there's no reason to not always add them.

   EDIT: Actually, the added clusters were all just output clusters, so we can likely just ignore them..?

zhaquirks/inovelli/__init__.py
```diff
@@ -201,255 +276,1379 @@ class InovelliVZM30SNCluster(InovelliCluster):
...
+        increased_non_neutral_output = ZCLAttributeDef(
+            id=0x0019,
+            type=t.Bool,
+            is_manufacturer_specific=True,
+        )
+        leading_or_trailing_edge = ZCLAttributeDef(
+            id=0x001A,
+            type=t.uint8_t,
+            is_manufacturer_specific=True,
+        )
```
TheJulianJES, line 346: `leading_or_trailing_edge` was `t.Bool` before. Now, it's `t.uint8_t`.
Is this change expected? If not, this might need to be changed in multiple places now.

TheJulianJES:
Generated new ZHA diagnostics from the legacy diagnostics provided by @codyhackw (some already with v2 quirks) with the old v1 quirks. Then, re-generated them with the current state of the PR including the `.replaces_endpoint(1, device_type=zha.DeviceType.DIMMABLE_LIGHT)` patch for `VZM31-SN` and the newer zigpy version.

The diff between diagnostics from the old v1 quirks and the new v2 quirks is this: https://github.com/TheJulianJES/zha/commit/89b74095541f21f045bf380e2ad31e70d73b605c (make sure to load the full diff for all files, though `VZM32-SN` apparently didn't have a quirk applied before)

That's where I also noticed the different for `leading_or_trailing_edge`, since it's used on that device.

We should look over the diagnostics to make sure all changes are expected. Some are just caused by reordering of attribute definitions and are fine. Others are caused by v2 quirks no longer removing some Scenes/Group clusters.
There are also some cases where additional attributes have been added.

zhaquirks/inovelli/__init__.py
codyhackw: hey, yep, this is expected and was updated across the rest of the devices, they added a couple other options depending on the device settings.

codyhackw:
> Since the above comment threads are becoming a bit hard to go through, I'll go over the two main points left for this PR:
> 
>     1. The device type was changed on older firmware versions to correctly create (dimmable) light entities in HA, instead of only switch entity. Relevant comment thread with solution (when zigpy is released + bumped in this repo): [Replace Inovelli quirks with v2 quirks #3232 (comment)](https://github.com/zigpy/zha-device-handlers/pull/3232#discussion_r2106079434)
>        Since newer versions all seem to use the `DIMMABLE_LIGHT` type, we can just always change that endpoint type to it.
> 

This has been updated in the most recent merge. Agreed re point 2 with the added clusters being output and think we're fine to leave them as-is.

TheJulianJES (approved):
This all looks good to me. Thanks again for the effort!

TheJulianJES:
To also mention this here:

One thing we can consider doing in a future PR is to move the all the device specific clusters (e.g. `InovelliVZM31SNCluster`, ...) into the device files with the quirk. Since they're all specific to the devices now (and nothing shared between them, other than the base `InovelliCluster`), it might be a bit cleaner. Especially if v2 quirks define those custom entities in the same file.

PR #3246, opened 2024-07-08
Add Tuya vibration door sensor `_TZE200_kzm5w4iz`

adam-zlatniczki:
## Proposed change
  Added quirk for the [Tuya Vibration Door Sensor](https://www.aliexpress.com/item/1005004443361928.html?spm=a2g0o.order_list.order_list_main.53.35a81802wiZASs), which didn't have any matching quirks before.

## Additional information
  --

## Checklist


zhaquirks/tuya/ts601_door.py
```diff
@@ -0,0 +1,251 @@
...
+    The latter two are both IasZone cluster extensions, that's why they had to
+    be put into different endpoints (otherwise we would have two of the same
+    cluster for an endpoint, which is prohibited by the Zigbee standard).
+
+    Extra details about the physical device: https://www.aliexpress.com/item/1005004443361928.html?spm=a2g0o.order_list.order_list_main.53.35a81802wiZASs
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Init constructor."""
+        super().__init__(*args, **kwargs)
```
TheJulianJES: This can be removed now.

zhaquirks/tuya/ts601_door.py
```diff
@@ -0,0 +1,251 @@
...
+
+        self.update_attribute(
+            attr_name=IasZone.AttributeDefs.zone_type.name,
+            value=ZoneType.Contact_Switch,
+        )
+
+        self.update_attribute(
+            attr_name=IasZone.AttributeDefs.zone_status.name,
+            value=0x0000,
+        )
```
TheJulianJES: Can we use `_CONSTANT_ATTRIBUTES` for these?

zhaquirks/tuya/ts601_door.py
```diff
@@ -0,0 +1,251 @@
...
+
+        self.update_attribute(
+            attr_name=IasZone.AttributeDefs.zone_type.name,
+            value=ZoneType.Vibration_Movement_Sensor,
+        )
+
+        self.update_attribute(
+            attr_name=IasZone.AttributeDefs.zone_status.name,
+            value=0x0000,
+        )
```
TheJulianJES: Same here. `_CONSTANT_ATTRIBUTES`

zhaquirks/tuya/ts601_door.py
```diff
@@ -0,0 +1,251 @@
...
+    """Custom Basic cluster to report power source."""
+
+    def __init__(self, endpoint: Endpoint, is_server: bool = True):
+        """Init constructor that sets the power source to battery."""
+        super().__init__(endpoint=endpoint, is_server=is_server)
+
+        self.update_attribute(
+            attrid=Basic.AttributeDefs.power_source.id,
+            value=Basic.PowerSource.Battery,
+        )
```
TheJulianJES: Same here: `_CONSTANT_ATTRIBUTES`

TheJulianJES:
Thanks for adding tests btw!

adam-zlatniczki:
_CONSTANT_ATTRIBUTES is a good proposal, I've moved the constant attributes.

zhaquirks/tuya/ts601_door.py
```diff
@@ -0,0 +1,230 @@
...
+    def __init__(self, endpoint: Endpoint, is_server: bool = True):
+        """Init constructor that initializes battery attributes."""
+        super().__init__(endpoint=endpoint, is_server=is_server)
+
+        # Initialize remaining battery to 100% until a proper measurement arrives.
+        # Device measures battery in 1% steps, while the ZCL specifies it in 0.5% steps.
+        self.update_attribute(
+            attr_name=PowerConfiguration.AttributeDefs.battery_percentage_remaining.name,
+            value=100 * 2,
+        )
```
TheJulianJES: Since this is done in `__init__`, this will update the attribute to 100% if HA is restarted.
Something like `if attr_id not in self._attr_cache` could be added before, but I think I'd just remove this for now.
HA should still initialize the sensor. It should just show something like "unknown". How long does it take for the device to report a value?

zhaquirks/tuya/ts601_door.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def __init__(self, endpoint: Endpoint, is_server: bool = True):
+        """Init constructor that initializes the major IasZone attributes."""
+        super().__init__(endpoint=endpoint, is_server=is_server)
+
+        # Initialize zone status until a proper update comes.
+        self.update_attribute(
+            attr_name=IasZone.AttributeDefs.zone_status.name,
+            value=0x0000,
+        )
```
TheJulianJES: Same thing here. This will update, as soon as the sensor is opened/closed once, right?
Do we really need to put "fake" initial data here?

zhaquirks/tuya/ts601_door.py
```diff
@@ -0,0 +1,230 @@
...
```
TheJulianJES: Same thing here.

TheJulianJES (commented):
I think the quirk looks really good. Added a few small comments.

zhaquirks/tuya/ts601_door.py
TheJulianJES: Let's remove them for now, so we can make tomorrow's beta release. If they're really needed, they can be re-added with adjustments I mentioned [here](https://github.com/zigpy/zha-device-handlers/pull/3246/files#r1731716318). For now, the entities should stay unknown until the first update report.
```suggestion
```

zhaquirks/tuya/ts601_door.py
TheJulianJES: Let's remove them for now, so we can make tomorrow's beta release. If they're really needed, they can be re-added with adjustments I mentioned [here](https://github.com/zigpy/zha-device-handlers/pull/3246/files#r1731716318). For now, the entities should stay unknown until the first update report.
```suggestion
```

zhaquirks/tuya/ts601_door.py
TheJulianJES: Let's remove them for now, so we can make tomorrow's beta release. If they're really needed, they can be re-added with adjustments I mentioned [here](https://github.com/zigpy/zha-device-handlers/pull/3246/files#r1731716318). For now, the entities should stay unknown until the first update report.
```suggestion
```

TheJulianJES: Removed for now.

zhaquirks/tuya/ts601_door.py
TheJulianJES: Removed for now.

zhaquirks/tuya/ts601_door.py
TheJulianJES: Removed for now.

TheJulianJES (approved):
Thanks!

I've removed initializing the attributes for now, as they'd always initialize to the defaults specified when rebooting HA.
If the defaults are really needed, they can be re-added with adjustments I mentioned [here](https://github.com/zigpy/zha-device-handlers/pull/3246/files#r1731716318). For now, the entities should stay unknown until the first update report.

zhaquirks/tuya/ts601_door.py
```diff
@@ -0,0 +1,198 @@
...
+DOOR_HANDLE_EP_ID = 2
+VIBRATION_EP_ID = 3
+
+
+class CustomBasicCluster(CustomCluster, Basic):
+    """Custom Basic cluster to report power source."""
+
+    _CONSTANT_ATTRIBUTES = {
+        Basic.AttributeDefs.power_source.id: Basic.PowerSource.Battery
+    }
```
TheJulianJES, line 53: Not sure we need this, not an issue for now though. Was it shown wrong on the HA device page?
That data comes from the node descriptor though.

zhaquirks/tuya/ts601_door.py
adam-zlatniczki: As far as I've noticed, the device reports battery state when there's a change in percentage. This doesn't happen too often: either when there's some variability in the measured voltage (e.g. moves from 80% to 81% then back to 80%), or when there's an actual drop. Since this is a door sensor and is supposed to be used frequently, the value will probably get initialized fairly soon, so I'm fine with this decision.

zhaquirks/tuya/ts601_door.py
adam-zlatniczki: Same reasoning applies: the opening sensor will probably get used fairly soon, it doesn't matter much if it doesn't get initialized.

zhaquirks/tuya/ts601_door.py
adam-zlatniczki: This is where I disagree. Significant force is very rarely applied to most doors (in fact, my sensor only detected such change when I was manually testing the quirk, but never since then). This leads to an unitialized entity for quite a long time, which I personally don't really like to see on a UI. Initializing the vibration state to "no vibration" seems to me the lesser evil :)

zhaquirks/tuya/ts601_door.py
adam-zlatniczki: ZHA displayed the power source as "battery or unknown", which was bothering me a little, and in hopes of fixing that I tried setting it this way. However, it didn't really help, so I guess that is solely based on the device signature.

Jimfizz:
Hello, I have the same Tuya vibration sensor (TS0601) but the manufacturer is _TZE200_iba1ckek.
Are you able to deal different sensors (of the same type) in the same quirk or should I create a new topic ?
I use it to detect the sound of a doorbell and I would like to manage the sensitivity of the sensor.

EDIT : ooop, it's not strictly the same sensor, mine is only a vibration one. I will create a new entry.

adam-zlatniczki:
Hi, it really depends on whether the Tuya DataPoints that it reports are
the same or not. Can you do a quick analysis on that?

Jimfizz:
Thank you @adam-zlatniczki : in ZHA the sensor does not report any door state (open/closed), only vibration detection (with reset), battery level (CR2450) & identifier

PR #2602, opened 2023-09-22
Add Owon PC321 support for custom attributes

exelsis423:
## Proposed change

Creating a new quirk

Relevant ZHA PR: https://github.com/home-assistant/core/pull/100735

## Additional information

This quirk was created to fix the work of Owon PC321. At the beginning, zha discovered only current summation and instantaneous demand. With this quirk, all manufacturer datas are available, to work properly zha/sensors needs to be updated too (another PR). Modifications done according to manfucaturer doc which is available there : https://github.com/froggyfly/pc321-zigbee/blob/main/PC321zigbee.docx "Fixes https://github.com/zigpy/zha-device-handlers/issues/1657 https://github.com/zigpy/zha-device-handlers/issues/977".

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

Issue #1657, opened 2022-07-17
[Device Support Request] OWON Technology Inc. PC321,need more information to be display


adi90x:
Hello,

Reoppening #977
At the moment zha will only report one consumption while 3 are available.

Is it possible to do something ?
Signature of the device is available in #977
And manfucaturer doc is available there : https://github.com/froggyfly/pc321-zigbee/blob/main/PC321zigbee.docx

@froggyfly : Did you find another solution ?

Regards

fergalish:
Hi, I'm trying to install one of these OWON 3-phase/triple single phase power meters. It's adds to ZHA with no problem, but the only sensors I'm seeing are instantaneous power and total energy consumed, each a sum of the three physical input sensors.

I'm still learning HA & ZHA, so I have no idea how to edit a device signature etc.

Curious to know if there's a way to proceed.

Thanks.

udivankin:
Seems to be implemented in z2m

BUT

there's two different PC321 zigbee clamp meters - PC321-Z and PC321-Z-TY (Tuya), which are according to @Koenkk two completely different devices.

Here's the long thread PC321-Z
https://github.com/Koenkk/zigbee2mqtt/issues/9525

And here is for PC321-Z-TY (Tuya)
https://github.com/Koenkk/zigbee2mqtt/issues/14790

exelsis423:
Hello,
I reopen this request.
I have this equipment too the OWON PC321 Z.

After having searched a lot, I tried to write a quirk. The first way I tried was following the documentation with using the metering cluster. This the quirk:
`import zigpy.types as t

from zigpy.profiles import zha
from zigpy.quirks import CustomCluster, CustomDevice
from zigpy.zcl.foundation import ZCLAttributeDef, ZCLCommandDef
from zigpy.zcl.clusters.general import (
    Basic,
    Identify,
    )
from zigpy.zcl.clusters.smartenergy import (
    Metering, 
    )
from zhaquirks.const import (
    DEVICE_TYPE,
    ENDPOINTS,
    INPUT_CLUSTERS,
    MODELS_INFO,
    OUTPUT_CLUSTERS,
    PROFILE_ID,
    SKIP_CONFIGURATION,
    )
    
from zhaquirks import LocalDataCluster



class OwonSimpleMetering(CustomCluster):
    cluster_id = 0x0702
    name = "Owon Simple Metering dbl"
    ep_attribute = "electrical_measurement" 

[...]

PR #3019, opened 2024-03-02
Initial quirks v2 documentation thoughts

dmulcahey:
## Checklist

- [ ] The changes are tested and work correctly
- [ ] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

PR #3274, opened 2024-07-29
Add Owon PC321 3-phase energy meter support

### LINKED ISSUE #977: [Device Support Request] OWON Technology Inc. PC321,need more information to be displayed
OWON Technology Inc. PC321, Energy Monitor support needed please. 
Device can be paired @ZHA, but appeared entity shows only shows "Instantaneous Demand" ( 0x0400，Instantaneous energy consumed summation of the 3 phases)
According to the manufacturer's documents, Please help adding 0x0000 current summary delivered (Current summation delivered of the 3 phases);
This data can be now obtained through the following process:
integration-ZHA-Owon Pc321 device-manage clusters- clusters "metering" (0x0702)- current_ summ_ delivered (id: 0x0000)- get zigbee attribute

more info can be shown would be great, thanks!

More monitoring data is needed

**Device signature - this can be acquired by removing the device from ZHA and pairing it again from the add devices screen. Be sure to add the entire content of the log panel after pairing the device to a code block below this line.**
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4412, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.0: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x000d",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0702"
      ],
      "out_clusters": [
        "0x0003"
      ]
    }
  },
  "manufacturer": "OWON Technology Inc.",
  "model": "PC321",
}

**Additional context**

Manufacturer's documents can be obtained here:
[https://github.com/froggyfly/pc321-zigbee/blob/main/PC321zigbee.docx](https://github.com/froggyfly/pc321-zigbee/blob/main/PC321zigbee.docx)

udivankin:
Another attempt to add Owon PC321 3-phase meter support to ZHA. The meter reports a couple of the standard and a list of non-standard attributes to the Metering cluster.

## Proposed change

The proposed quirk maps all reported attributes to OwonManufacturerSpecific custom cluster, and also only supported ones to  Metering and ElectricalMeasurement clusters (a separate set of clusters per each of 3 phase-specific endpoints + totals in the main Metering endpoint).

## Additional information

Since not all sensors are available in Metering and ElectricalMeasurement clusters for reported attributes, the missing ones will be supported later on in HACS plugin, or alternatively - if corresponding clusters eventually get a full set of sensors natively. But as for now, most users will probably be fine as it is (see screenshot attached)

Fixes #977 #1657 
Also, see #2602 for the same device but with different approach (never merged)

## Checklist

Screenshot 2024-07-31 at 11 07 41


zhaquirks/owon/pc321.py
```diff
@@ -0,0 +1,708 @@
...
+}
+
+
+class OwonMetering(CustomCluster, Metering):
+    """Owon non-standard Metering cluster attributes, to be mapped into bus for later use in another clusters."""
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        self._current_state = {}
+        super().__init__(*args, **kwargs)
```
TheJulianJES, line 81: `_current_state` doesn't seem to be used. If so, this can be removed completely.

TheJulianJES (commented):
The `Bus` stuff isn't needed anymore. You can directly use something like this
```python
self.endpoint.device.endpoints[1].smartenergy_metering.update_attribute(...)
```
to update attribute cace on another endpoint.

TheJulianJES (commented):
Also, thanks for the PR, of course.

I think the solution in https://github.com/zigpy/zha-device-handlers/pull/2602 might be preferable.
However, it should be converted to a v2 quirk and add custom sensors for the custom attributes (using quirks v2 "exposes").
The entities would also have proper default names in HA then.

(Outdated docs draft PR, but still mostly valid for v2 quirks: https://github.com/zigpy/zha-device-handlers/pull/3019)

udivankin:
@TheJulianJES Could you please clarify if v2 quirks are already supported? I mean shall I update the current PR with both v1 and v2 quirks or v2 only?

zoic21:
Hello,

Any news on that ? I just add this own to my zigbee network but only have 1 on 3 phases.... 

Thanks in advance

abmantis:
You can check [this PR](https://github.com/zigpy/zha-device-handlers/pull/3571) for an example of a 3phase meter using quirks v2. It converts the tuya DPs to standard zigbee attributes (except for Energy since those are not on the standard), to reduce the need of custom attributes. The downside is that currently there is no distinction between phase A and single phase, so both the total and phase A will be named Current, for example.

When [this proposal](https://github.com/zigpy/zha/issues/311#issuecomment-2532109597) gets implemented the naming could be properly defined by having a `feature_3_phase` on the quirk. 

EDIT: I just noticed that this quirk is not for a tuya device, so the PR I mentioned may not be that helpful.

PR #3276, opened 2024-08-01
Add custom attributes for ThirdReality nightlight `3RSNL02043Z`

llhappier:
Change-Id: I136b856588fb42e69274b899cdafa3c5e9ce819f

## Checklist

### DIFF
--- a/zhaquirks/thirdreality/night_light.py
+++ b/zhaquirks/thirdreality/night_light.py
@@ -2,6 +2,7 @@
 
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice
+import zigpy.types as t
 from zigpy.zcl.clusters.general import (
     Basic,
     Groups,
@@ -16,6 +17,7 @@ from zigpy.zcl.clusters.lighting import Color
 from zigpy.zcl.clusters.lightlink import LightLink
 from zigpy.zcl.clusters.measurement import IlluminanceMeasurement
 from zigpy.zcl.clusters.security import IasZone
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 from zhaquirks import LocalDataCluster
 from zhaquirks.const import (
@@ -36,6 +38,25 @@ class ThirdRealitySpecificCluster(CustomCluster):
     """Manufacturer specific cluster to relay motion event to IAS Zone cluster."""
 
     cluster_id = THIRD_REALITY_CLUSTER_ID
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        cooldown_time = ZCLAttributeDef(
+            id=0x0003,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+        local_routine_time = ZCLAttributeDef(
+            id=0x0004,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+        lux_threshold = ZCLAttributeDef(
+            id=0x0005,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
 
     def _update_attribute(self, attrid, value):
         super()._update_attribute(attrid, value)


puddly:
Can you describe what these attributes do?

llhappier:
Hi,
Through this private cluster, these attributes are used to set the coldown time for pir sensor, the duration of lthe ocal routines, and the threshold for reporting changes in lux sensor values.

llhappier:
Hi,
Through this private cluster, these attributes are used to set the coldown time for pir sensor, the duration of lthe ocal routines, and the threshold for reporting changes in lux sensor values.

puddly:
Right, that's clear from the names of the attributes.

- What are local routines?
- PIR cool down time can be set using the PIR Configuration Set attributes from the ZCL: `PIROCcupiedToUnoccupiedTime`, etc. Does this need to be a manufacturer specific attribute?
- Why does the lux sensor have a separate attribute for configuring attribute reporting? This is normally done using ZCL attribute reporting config, which sets the min/max reporting interval and the minimum change.

llhappier:
Hi, puddly. This is a multifunction device and It contains light, pir sensor, and illuminance sensor.

1. This local routine is performs the actions of turning on and off the lights based on the status of the pir sensor and illuminance sensor and does not require sending Zigbee commands.  The modification this time is the interface for the duration of the light on.
2. Yes, the IAS zone cluster can be used for pir sensor. However, if the IAS zone cluster is registered, this device may be recognized as a sensor on some platforms instead of the primary feature light, such as on Amazon's Echo v4 etc.
3. We want freely set the reporting sensitivity of the illuminance sensor.
By the way，we can also use service: zhatoolkit.attr_write to write any attribute of any cluster and config report, it's just that installing this ZHA toolkit is inconvenient, so we changed this script.
Thank you.

zhaquirks/thirdreality/night_light.py
```diff
@@ -36,6 +37,12 @@ class ThirdRealitySpecificCluster(CustomCluster):
     """Manufacturer specific cluster to relay motion event to IAS Zone cluster."""
 
     cluster_id = THIRD_REALITY_CLUSTER_ID
+    # Add read-write interfaces for attributes in private cluster
+    attributes = {
```
puddly: Since you're adding new attributes, please use the new syntax for defining them. Here is an example: https://github.com/zigpy/zha-device-handlers/blob/915ff9b41e7393e1f2eaa449bd6a09267fa84277/zhaquirks/legrand/cable_outlet.py#L51-L68

llhappier:
Copy that, thank you.

zhaquirks/thirdreality/night_light.py
```diff
@@ -37,6 +39,25 @@ class ThirdRealitySpecificCluster(CustomCluster):
...
+            id=0x0003,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+        local_routine_time = ZCLAttributeDef(
+            id=0x0004,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+        lux_thredhold = ZCLAttributeDef(
```
puddly:
```suggestion
        lux_threshold = ZCLAttributeDef(
```

llhappier:
OK. Got it. I'm sorrry about this mistake and will definitely pay attention next time. Thanks.

TheJulianJES:
All good! By the way, you don't need to squash and force push your commits every time. You can just add another commit changing things. The whole PR will be squashed automatically when merging, so no need to do it manually before.

llhappier:
Okay, I will understand and use it next time. Thank you

TheJulianJES (approved):
I think this should be fine for now, since it just adds the custom attribute definitions.
In the future, we can think about replacing the `Nightlight`/`CustomDevice` v1 quirk with a v2 quirk that could look something like this:

```python
(
    QuirkBuilder(THIRD_REALITY, "3RSNL02043Z")
    .replaces(LocalIasZone)
    .replaces(ThirdRealitySpecificCluster)
    .number(
        ThirdRealitySpecificCluster.AttributeDefs.cooldown_time.name,
        ThirdRealitySpecificCluster.cluster_id,
        min_value=0,  # change
        max_value=100,  # change
        step=1,  # change
        unit="s",  # change?
    )
    .number(
        ThirdRealitySpecificCluster.AttributeDefs.local_routine_time.name,
        ThirdRealitySpecificCluster.cluster_id,
        min_value=0,  # change
        max_value=100,  # change
        step=1,  # change
        unit="s",  # change
    )
    .number(
        ThirdRealitySpecificCluster.AttributeDefs.lux_threshold.name,
        ThirdRealitySpecificCluster.cluster_id,
        min_value=0,  # change
        max_value=100,  # change
        step=1,  # change
        unit="s",  # change
    )
    .add_to_registry()
)
```
(untested and unfinished code above, just a quick draft)

It would allow users to configure the custom attributes from entities in HA. Translation keys will also need to be added to HA, but we should do that in the quirks/ZHA bump PR.

To confirm, @puddly, you're also good with this PR for now?

puddly:
Yeah, there's nothing new, it's just mapping attributes. A followup PR should expose them as entities.

PR #3318, opened 2024-08-26
Add Tuya temperature sensor `_TZE204_yjjdcqsq`, fix battery reporting half

GeoffWy:
Add quirk for  Temperature/humidity sensor model "_TZE204_yjjdcqsq", "TS0601".  

## Proposed change

 Add quirk for  Temperature/humidity sensor model "_TZE204_yjjdcqsq", "TS0601".  This is added as class TuyaTempHumiditySensorVar05(CustomDevice)Var02, and references TemperatureHumidityBatteryStatesManufClusterVar02. The only difference from the original is that TemperatureHumidityBatteryStatesManufClusterVar02 has double the values for battery states, otherwise the sensor appears in HA with half the battery remaining %. It seems very likely that the original TemperatureHumidityBatteryStatesManufCluster should also have the same correction, but I don't have those sensors to test.

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_sensor.py
+++ b/zhaquirks/tuya/ts0601_sensor.py
@@ -95,6 +95,27 @@ class TemperatureHumidityBatteryStatesManufCluster(TuyaMCUCluster):
     }
 
 
+class TemperatureHumidityBatteryStatesManufClusterVar02(TuyaMCUCluster):
+    """Tuya Manufacturer Cluster with Temperature and Humidity data points. Battery states 20, 50 and 100%."""
+
+    dp_to_attribute: dict[int, DPToAttributeMapping] = {
+        1: TemperatureHumidityManufCluster.dp_to_attribute[1],
+        2: TemperatureHumidityManufCluster.dp_to_attribute[2],
+        3: DPToAttributeMapping(
+            TuyaPowerConfigurationCluster2AAA.ep_attribute,
+            "battery_percentage_remaining",
+            converter=lambda x: {0: 40, 1: 100, 2: 200}[x],  # double reported percentage
+            #for some reason values are halved somewhere downstream
+        ),
+    }
+
+    data_point_handlers = {
+        1: "_dp_2_attr_update",
+        2: "_dp_2_attr_update",
+        3: "_dp_2_attr_update",
+    }
+
+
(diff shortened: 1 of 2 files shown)


zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -95,6 +95,29 @@ class TemperatureHumidityBatteryStatesManufCluster(TuyaMCUCluster):
...
+    dp_to_attribute: dict[int, DPToAttributeMapping] = {
+        1: TemperatureHumidityManufCluster.dp_to_attribute[1],
+        2: TemperatureHumidityManufCluster.dp_to_attribute[2],
+        3: DPToAttributeMapping(
+            TuyaPowerConfigurationCluster2AAA.ep_attribute,
+            "battery_percentage_remaining",
+            converter=lambda x: {0: 40, 1: 100, 2: 200}[
+                x
+            ],  # double reported percentage
+            # for some reason values are halved somewhere downstream
```
TheJulianJES: This is according to ZCL specs. `200` means 100%. It's done, so it can be reported in 0.5 increments.
Let's remove that comment. The "double reported percentage" already explains it.
```suggestion
```

TheJulianJES (commented):
Hmm, yeah. It does seem like the other quirk also needs to be corrected.
Some Tuya power config clusters already double it in the cluster `update_attribute`, but the `TuyaPowerConfigurationCluster2AAA` does not seem to do that.

TheJulianJES (commented):
Hmm, the Tuya cluster code could use some some cleanup.
Notice how the AA vs AAA clusters inherit from different clusters:
https://github.com/GeoffWy/zha-device-handlers/blob/14b87ed7a07886e00ba4dd2b2d60f395c870f3db/zhaquirks/tuya/__init__.py#L935-L974

The `TuyaPowerConfigurationCluster` cluster doubles it there, the other does not.
I'll try and clean this up and push the changes to your PR.

Actually no, that cluster also doesn't update it in `update_attribute`. Other Tuya power config clusters do though, it's a bit of a mess...

TheJulianJES (approved):
I've updated it to report double battery percentage for all sensors now. I *think* that should be correct. Hopefully didn't break anything else.

Thanks for the PR!

PR #3319, opened 2024-08-26
Add Tuya PowerConfiguration cluster for 3 AAA devices

cecinitmortem:
Update Tuya __init__.py with class for 3 AAA devices like TZE200_vvmbj46n or TZE200_locansqn.

I have the two devices mentioned above with 3 AAA batteries. Currently the custom quirk for them is not in ZHA. Both devices work fine with just using the 2AAA battery class, however I think it makes sense to implement this natively. 

The class is just copied from the currently existing 2 AAA class and the BATTERY_QUANTITY is set to 3.


zhaquirks/tuya/__init__.py
```diff
@@ -954,6 +954,20 @@ class TuyaPowerConfigurationCluster2AAA(PowerConfiguration, TuyaLocalCluster):
...
+
+    BATTERY_SIZES = 0x0031
+    BATTERY_QUANTITY = 0x0033
+    BATTERY_RATED_VOLTAGE = 0x0034
+
+    _CONSTANT_ATTRIBUTES = {
+        BATTERY_SIZES: 4,
+        BATTERY_QUANTITY: 3,
+        BATTERY_RATED_VOLTAGE: 15,
+    }
```
TheJulianJES, line 968: Take a look at `TuyaPowerConfigurationCluster2AAA`, it grabs the ZCL IDs from zigpy instead of needing to specify them again. Let's also do that here.

TheJulianJES (commented):
Ideally, we'd add this with an official quirk actually using it.

PR #3329, opened 2024-08-28
Add Xiaomi LYWSD03MMC temp sensor with custom Zigbee firmware

### LINKED ISSUE #153: Device showing up weird cluster endpoints on ZHA
Hello, 
I'm using the latest firmware (0x11003001) on HA 2024.5.5 with ZHA

When trying to use the attributes of the User interface cluster, I see this attributes:

image

Which do not match the description provided in the repository.

However, I've been able to successfully use them with ZHA_Toolkit.

Also the thermometers show this entities when paired, and I think there might be something wrong:
image

gigamonster256:
## Proposed change
[This](https://github.com/devbis/z03mmc) repo has published a custom firmware for converting the Xiaomi LYWSD03MMC to use Zigbee instead of Bluetooth. It implements a few custom attributes in order to control the built in display that require a quirk to work correctly.

## Additional information
  Conversion Guide: https://smarthomescene.com/guides/convert-xiaomi-lywsd03mmc-from-bluetooth-to-zigbee/
Fixes devbis/z03mmc#153

## Checklist


zhaquirks/xiaomi/custom/z03mmc.py
```diff
@@ -0,0 +1,156 @@
+"""Xiaomi LYWSD03MMC Bluetooth temperature and humidity sensor."""
+
+from zigpy.profiles import zha
+from zigpy.types import Bool, int16s, uint16_t
+from zigpy.zcl.clusters.general import Basic, Identify, Ota, PowerConfiguration
+from zigpy.zcl.clusters.hvac import UserInterface as UserInterfaceBase
+from zigpy.zcl.clusters.measurement import (
+    RelativeHumidity as RelativeHumidityBase,
+    TemperatureMeasurement as TemperatureMeasurementBase,
+)
```
TheJulianJES: Can we keep the original name for the ZCL imports and just name the custom clusters something like `RelativeHumidityCustom`?

gigamonster256: Sure, that works. I named the derived class so that it matches the base class name and is transparent in the homeassistant UI, but if this naming scheme is preferred then that works for me.

gigamonster256:
Test failed due to venv cache miss, but should pass.

tobiasb80:
@gigamonster256 I have adapted your Quirk to ZHA-Quirks V2 so that entities are created for the setting options. Maybe someone is interested: https://github.com/tobiasb80/zha-device-handlers-LYWSD03MMC-/blob/dev/zhaquirks/custom/z03mmc.py

k0stett:
Which ZHA version will contain this sensor support? 6 months since review request...

puddly (approved):
This looks fine to me to define the manufacturer-specific attributes.

@tobiasb80 Could you make a PR with your quirk once this one is merged? New entities can only really be added with the quirks v2 API that you're using and your quirk looks complete to me.

zhaquirks/xiaomi/custom/z03mmc.py
```diff
@@ -0,0 +1,153 @@
...
+            type=uint16_t,
+            access="rw",
+            is_manufacturer_specific=True,
+        )
+
+
+# https://github.com/devbis/z03mmc
+# defined by 1.1.0 firmware (0x11003001)
+# see README.md in the repo for more info
+class LYWSD03MMC_devbis(XiaomiCustomDevice):
```
TheJulianJES: `XiaomiCustomDevice` [sets an (internal) attribute](https://github.com/zigpy/zha-device-handlers/blob/b5f2f93a3f3a151471fc358f24ba71aa13a82d71/zhaquirks/xiaomi/__init__.py#L115-L122) for the battery size.
However, since we don't use a `XiaomiPowerConfiguration` in the `replacement`, but just leave the `PowerConfiguration` cluster untouched—which is fine—that has no effect and we should just use `CustomDevice` here. It's also not "real Xiaomi firmware".

TheJulianJES: Also, we generally don't want to use underscores in class names. I see why you did it here, but maybe there could be a better name.

TheJulianJES (commented):
Yeah, this looks good.

If the v2 quirk is PR'd and merged shortly after this, my notes below don't really mean much. The signature test also kind of becomes irrelevant then and can just be removed.
I'll still just quickly address my two comments.

TheJulianJES:
Thanks for the PR, @gigamonster256!

@tobiasb80 Like puddly mentioned, please PR your v2 quirk implementation after this is merged. It does look basically finished. Some comments on it already:
- `mode="box"` does nothing for now. We'll add support for that in ZHA in the future. You can keep it as an argument though. See https://github.com/zigpy/zha/issues/364 for the feature request
- Please move `translation_key` and `fallback_name` to the bottom, so they're the last arguments in the methods.
- And just a note: `endpoint_id=1` is already the default for basically all methods, but it's fine to keep it to be more explicit.

tobiasb80:
@TheJulianJES I have created a [pull request](https://github.com/zigpy/zha-device-handlers/pull/4179)

Issue #3298, opened 2024-08-11
[Device Support Request] Sonoff SWV-BSP Smart Water Valve


andy-81:
### Problem description

Currently only opening and closing the valve is possible however the Sonoff hub and Zigbee2MQTT support pulling in if there is an issue with the valve and the flow data from the valve which would be fantastic to be able to use in ZHA.

### Solution description

It would be great if a custom quirk could be created however I have struggled to follow along with the details of how to create the custom device handlers. I was hoping the work from Z2M may help point to how to pull the information but the information does not look similar to the Zwave standard where you require the address details which has further complicated me understanding how to write the custom driver. Any help in someone creating this custom handler would be appreciated.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

https://youtu.be/ofzIoxliNJo?si=9JpD2oPFi9j76muC

this video shows the details of what is shown in ZHA vs what is shown in Zigbee2MQTT

</details>

### Device signature

<details><summary>Device signature</summary>

```json
[Paste the device signature here]
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
[Paste the diagnostic information here]
```

</details>

### Logs

<details><summary>Logs</summary>

```python
[Paste the logs here]
```

</details>

### Custom quirk

<details><summary>Custom quirk</summary>

```python
[Paste your custom quirk here]
```

</details>

AKTheKnight:
Just got one of these and can provide some of the info missing above:

Signature:
```
{
  "node_descriptor": {
    "logical_type": 2,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4742,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 255,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 255,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0002",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0006",
        "0x0020",
        "0x0404",
        "0x0b05",
        "0xfc11",
        "0xfc57"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
[...]

PR #3340, opened 2024-08-31
Add quirk for Sonoff SWV Smart Water Valve

benbancroft:
## Proposed change

Adds support for Sonoff SWV Smart Water Valve custom cluster, which exposes a new attribute called valve_status for querying if the valve is leaking or if the water supply to the valve has stopped.

Code should hopefully should be self explanatory, the only complication is the overridden `_is_manuf_specific()` property in `SonoffFC11Cluster` class. This is required, as if the device receives an attribute read or notify zigbee command which has manufacturer_specific = True set (default if cluster's ID is within manufacturer specific range which this is), it will return `UNSUPPORTED_ATTRIBUTE`.

This feature was requested as part of https://github.com/zigpy/zha-device-handlers/issues/3298

## Additional information

- zha pull request for Home Assistant Sensors: https://github.com/zigpy/zha/pull/189
- Home Assistant core pull request for language strings for sensor: https://github.com/home-assistant/core/pull/124989

## Checklist

### DIFF
--- a/zhaquirks/sonoff/swv.py
+++ b/zhaquirks/sonoff/swv.py
@@ -1,0 +1,47 @@
+"""Sonoff SWV - Zigbee smart water valve."""
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import EntityPlatform, EntityType, QuirkBuilder
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ValveState(t.enum8):
+    """Water valve state."""
+
+    Normal = 0
+    Water_Shortage = 1
+    Water_Leakage = 2
+    Water_Shortage_And_Leakage = 3
+
+
+class EwelinkCluster(CustomCluster):
+    """Ewelink specific cluster."""
+
+    cluster_id = 0xFC11
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        water_valve_state = ZCLAttributeDef(
+            id=0x500C,
+            type=ValveState,
+        )
+
+    @property
+    def _is_manuf_specific(self):
+        return False
+
+
+(
+    QuirkBuilder("SONOFF", "SWV")
+    .replaces(EwelinkCluster)
+    .enum(
+        EwelinkCluster.AttributeDefs.water_valve_state.name,
+        ValveState,
+        EwelinkCluster.cluster_id,
+        entity_platform=EntityPlatform.SENSOR,
+        entity_type=EntityType.DIAGNOSTIC,
+    )
+    .add_to_registry()
+)
--- a/zhaquirks/sonoff/swv.py
+++ b/zhaquirks/sonoff/swv.py
@@ -1,7 +1,10 @@
 """Sonoff SWV - Zigbee smart water valve."""
 
+import typing
+
 from zigpy.quirks import CustomCluster
-from zigpy.quirks.v2 import EntityPlatform, EntityType, QuirkBuilder
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
 import zigpy.types as t
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
@@ -9,10 +12,8 @@ from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 class ValveState(t.enum8):
     """Water valve state."""
 
-    Normal = 0
     Water_Shortage = 1
     Water_Leakage = 2
-    Water_Shortage_And_Leakage = 3
 
 
 class EwelinkCluster(CustomCluster):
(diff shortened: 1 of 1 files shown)


zhaquirks/sonoff/swv.py
```diff
@@ -0,0 +1,112 @@
...
+SONOFF_CLUSTER_FC57_ID = 0xFC57
+ATTR_SONOFF_VALVE_STATUS = 0x500C
+
+
+class ValveStatusBitmap(t.bitmap8):
+    """Valve status value enum."""
+
+    Water_Shortage = 0x1
+    Water_Leakage = 0x2
+
```
fgsch: I think you want Normal (0) and probably Leakage and Shortage (3).

zhaquirks/sonoff/swv.py
```diff
@@ -0,0 +1,112 @@
...
+        valve_status return UNSUPPORTED_ATTRIBUTE if manufacturer_specific is set on report command.
+        We therefore need to treat cluster as if it doesn't have ID within manufacturer specific range
+        and explicitly set if each attribute is manufacturer_specific or not instead in attribute list.
+        """
+        return False
+
+    cluster_id = SONOFF_CLUSTER_FC11_ID
+    ep_attribute = "sonoff_manufacturer"
+    attributes = {ATTR_SONOFF_VALVE_STATUS: ("valve_status", ValveStatusBitmap, False)}
```
fgsch: I believe attributes are deprecated. It might be better to use `ZCLAttributeDef` to define this.

fgsch:
Nice! I was testing a similar change using the Quirks Builder v2 and exposing the state as a sensor.

I'll probably open a PR on Monday, as I wrote most of the code already, but I've left some comments in this PR.

zhaquirks/sonoff/swv.py
benbancroft: So the absence of Normal (0) and Leakage and Shortage (3) was intentional - if you look at class you will see that this is a bitfield not an enum. Leakage and Shortage (3) would just be Water_Shortage (1) & Water_Leakage (2)

If you look what I've done in ZHR PR (https://github.com/zigpy/zha/pull/189), I'm actually exporting two binary sensors, one for each bit. I personally think this is a lot nicer than how they did it in Zigbee2MQTT.

Screenshot 2024-08-31 201932

zhaquirks/sonoff/swv.py
benbancroft: Thanks - I based this off the other Sonoff quirk, with no prior experience writing quirks. Let me look into `ZCLAttributeDef` and see how easy it is to convert over.

benbancroft:
Hi @fgsch 

Thanks for your review comments, much appreciated! Given you've gone to the work of writing yours, I would definetly push it up. I wasn't aware of Quirks Builder v2 (with this being my first quirk), but looking at API it looks very nice and I guess it allows you to avoid a ZHA change also? What might be a pain is exporting two not one sensors like I have (which in my personal opinion makes more sense for these sensors).

zhaquirks/sonoff/swv.py
fgsch: Ah, I see, I missed the bitmap8.  
Two related comments: 

- The docstrings say enum :)
- While this is fine for the sensor, the other part is the response when reading the attribute directly via Manage Zigbee Device interface. This is not a big issue, but it'd be nice to have the full range of values IMO.

fgsch:
> Hi @fgsch
> 
> Thanks for your review comments, much appreciated! Given you've gone to the work of writing yours, I would definetly push it up. I wasn't aware of Quirks Builder v2 (with this being my first quirk), but looking at API it looks very nice and I guess it allows you to avoid a ZHA change also? What might be a pain is exporting two not one sensors like I have (which in my personal opinion makes more sense for these sensors).

My original plan was to expose the valve status as a sensor, but I've been struggling to do so as I envisioned.
I don't know if it's me or if it's just not possible, but unfortunately, no other zha quirks are using this.
Might be a good time to ask around. 

As for my PR, I sort of ran out of time today. I will try to do it tomorrow.

fgsch:
And finally, here's my PR: https://github.com/zigpy/zha-device-handlers/pull/3346

I managed to get the sensor working as I wanted with help from some devs in discord. Unfortunately, it is not yet possible to do the same with binary sensors.

zhaquirks/sonoff/swv.py
fgsch: Actually, is there any reason you cannot use an enum here and check for the specific values on the other PR?

RogerSero:
Hi, does this feature work? Where can I get the full file?

benbancroft:
Hi @fgsch 

Apologies for the big delay in getting back to you since you posted your pull request. I was finally able to try it out locally, V2 quirk interface is certainly a big improvement.
Based on your commit, I think I've come up with a clean way of handling the binary sensors instead of enum approach. Please let me know what you think.

Looks like there are still some issues with test line coverage - need to work out how to write a test for V2 quirk interface next.

zhaquirks/sonoff/swv.py
```diff
@@ -0,0 +1,71 @@
...
+from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ValveState(t.enum8):
+    """Water valve state."""
+
+    Water_Shortage = 1
+    Water_Leakage = 2
```
fgsch, line 16: To show all the possible values if querying the cluster directly, can you also add the Normal (0) and Water leakage and shortage (3) values?

zhaquirks/sonoff/swv.py
```diff
@@ -0,0 +1,71 @@
...
+    def get(self, key: int | str, default: typing.Any | None = None) -> typing.Any:
+        """Map binary sensors onto bits of shared attribute."""
+
+        value = super().get(key, default)
+        match key:
+            case self.AttributeDefs.water_supply_state.name:
+                return value & ValveState.Water_Shortage
+            case self.AttributeDefs.valve_leak_state.name:
+                return value & ValveState.Water_Leakage
+        raise ValueError(f"Unknown attribute {key}")
```
fgsch, line 50: Q: I suppose key would only be one of the supported attributes but perhaps for unknown keys this should return the value?

fgsch:
I left some comments, but it looks good. I won't be able to test it until this weekend, though.

RogerSero:
I'm doing something wrong, because with the quirk file, it doesn't show me the water flow. What could it be?
Captura de pantalla 2024-10-16 104233

benbancroft:
Hi @RogerSero 

To clarify, this PR/quirk only adds the sensors for diagnostic of the valve (e.g. water supply or leak). I can see from your screenshot that this appears to be working. If you want to see the sensor names to appear correctly, then https://github.com/zigpy/zha/pull/234 will also be required.

As it sounds like you are looking for water flow rate in particular, there is a separate PR implementing that on ZHA repository, please see here: https://github.com/zigpy/zha/pull/187. I likely confused you with my screenshot above - as I had both patches locally when testing this. This implements the whole Flow cluster (with it being a Zigbee standard cluster) and therefore we don't require any special device side quirk like this patch does for the additional sensors.

benbancroft:
Much appreciated @fgsch - let me also go over your comments in more detail (hopefully this week) and get back to you.

fgsch:
I've tested this quickly, and it works as expected. 

If you could address https://github.com/zigpy/zha-device-handlers/pull/3340#discussion_r1802032875 I think I'd be ready to go from my POV.

fgsch:
This can be closed now.

TheJulianJES:
Thanks for the PR! This was implemented with:
- https://github.com/zigpy/zha-device-handlers/pull/3346

PR #3357, opened 2024-09-10
Add Aqara H1 Wireless Remote Double Rocker variant

tfranzel:
## Proposed change

This is a variation for the Aqara H1 Double Rocker. 

Search led me here: https://github.com/zigpy/zha-device-handlers/issues/940#issuecomment-1001107154
and here https://gist.github.com/xuxiaoqiao/6ed8dcb950f26337218d0507733970a0#file-aqara_h1_remote-py

Unfortunately it didn't work because my device quirk somehow wasn't recognized properly.

By looking at the debug messages and tinkering a bit, I found that my device's `INPUT_CLUSTERS` are slightly different and a single line flip in **Variant 3** did the trick for me:

```diff
-                INPUT_CLUSTERS: [Basic.cluster_id, PowerConfiguration.cluster_id, Identify.cluster_id,], 
+                INPUT_CLUSTERS: [Basic.cluster_id, Identify.cluster_id, PowerConfiguration.cluster_id,], 
```

This has been working flawlessly for several months for me.

Sorry if this is not done properly, but I only have a limited understanding what is going on here 😅 

### DIFF
--- a/zhaquirks/xiaomi/aqara/remote_h1.py
+++ b/zhaquirks/xiaomi/aqara/remote_h1.py
@@ -393,3 +393,88 @@ class RemoteH1DoubleRocker4(RemoteH1DoubleRocker1):
             },
         },
     }
+
+
+class RemoteH1DoubleRocker5(RemoteH1DoubleRocker1):
+    """Aqara H1 Wireless Remote Double Rocker Version WRS-R02, variant 5."""
+
+    signature = {
+        MODELS_INFO: [(LUMI, "lumi.remote.b28ac1")],
+        ENDPOINTS: {
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.COLOR_DIMMER_SWITCH,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                    PowerConfiguration.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [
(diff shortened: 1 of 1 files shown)


TheJulianJES:
The list order doesn't matter, but variant 3 has these empty clusters at the end:
```python
            4: {},
            5: {},
            6: {},
```
That's probably why it doesn't match. Adding another variant should be fine though.

tfranzel:
Hey @TheJulianJES,

I have no idea how the signature matching works, but your statement does not really match my experience here. 

I tried those quirks from the gist above and it **did not work** right away. Only after flipping the `INPUT_CLUSTER` order, it started working. I basically adapted the "shape" from the HA device info dump into the quirk.

As a matter of fact, I bought the same device again a couple of days ago. Turns out the new device is again different and does indeed have empty clusters 4/5/6. Had to create another one as it was not matching out-of-the box. I should probably add that one also.

Is there by any chance some doc on how all of this zigbee stuff works (without reading some 1000 page specification)?

bergerx:
possibly related issue https://github.com/home-assistant/core/issues/122689

tfranzel:
there are the changes I successfully been using for a year. imho should be merged.

TheJulianJES:
I'll leave this PR open to come back to it, but my comment above still holds true. The order of clusters is completely irrelevant for quirk matching. It's the empty endpoints you've added that allows a match with your devices.

We might even be able to convert this to a v2 quirk in the future and avoid the signature matching completely. Again, I'll leave the PR open to come back to it eventually.

Issue #2777, opened 2023-11-27
[Device Support Request] - SIN-4-FP-21_EQU


APCM-1:
### Problem description

I have a Pilot wire heating module (Leroy Merlin version) and I want to control it via ZHA. 

Brand : ADEO
Model : SIN-4-FP-21_EQU

![3624961](https://github.com/zigpy/zha-device-handlers/assets/141181269/24e49ef4-11e1-4a62-9491-ea704c402837)

At the moment we don't have mode handling, only On and Off (Hors Gel) with power measurement.

Some people seems to have handle this module with Z2M but not ZHA. 

[https://www.zigbee2mqtt.io/devices/SIN-4-FP-21_EQU.html](url)

[https://github.com/Koenkk/zigbee2mqtt/issues/19169](url)

### Solution description

We need to have mode handling. There are 6 modes : confort, confort -1, confort -2, eco, Hors Gel, Off.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

![Capture d'écran 2023-11-27 161003](https://github.com/zigpy/zha-device-handlers/assets/141181269/6c43cb65-fdf5-4d3f-8730-e81db0959d95)

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.FullFunctionDevice|MainsPowered|RxOnWhenIdle|AllocateAddress: 142>, manufacturer_code=4727, maximum_buffer_size=82, maximum_incoming_transfer_size=500, server_mask=11264, maximum_outgoing_transfer_size=500, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0702",
        "0x1000",
        "0xfc00"
      ],
[...]

PR #3364, opened 2024-09-15
Add initial NodOn SIN-4-FP-21 (Adeo SIN-4-FP-21_EQU) quirk

ikruglov:
## Proposed change
This PR adds a quirk for SIN-4-FP-21 (PilotWire). The device reports itself as a SmartPlug but it has special manufacturer-specific cluster which allows setting the following modes:
```
        Off = 0x00
        Comfort = 0x01
        Eco = 0x02
        FrostProtection = 0x03
        ComfortMinus1 = 0x04
        ComfortMinus2 = 0x05
```

## Additional information

> [!IMPORTANT]  
> The HA entity shown below is not yet implemented with this version.
> See this comment for more information: https://github.com/zigpy/zha-device-handlers/pull/3364#issuecomment-2502032896

The result look like this in Home Assitant:
Screenshot 2024-09-22 at 22 16 55

The PR also contains a solution for Adeo SIN-4-FP-21_EQU, which seems to be an exact clone of NodOn. However, it requires https://github.com/zigpy/zigpy/pull/1505 in HA to work. https://github.com/zigpy/zigpy/pull/1505 has not yet been released as part of HA as of writing.

## Checklist

### DIFF
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -3,13 +3,7 @@
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.v2 import EntityType, QuirkBuilder
 import zigpy.types as t
-from zigpy.zcl.foundation import (
-    BaseAttributeDefs,
-    DataTypeId,
-    Direction,
-    ZCLAttributeDef,
-    ZCLCommandDef,
-)
+from zigpy.zcl.foundation import BaseAttributeDefs, DataTypeId, ZCLAttributeDef
 
 NODON = "NodOn"
 NODON_MANUFACTURER_ID = 4747
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -13,7 +13,6 @@ from zigpy.zcl.foundation import (
 )
 from zigpy.zdo.types import NodeDescriptor
 
-
 NODON = "NodOn"
 NODON_MANUFACTURER_ID = 4747
 NODON_PILOT_WIRE_CLUSTER_ID = 0xFC00  # 64512
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -5,7 +5,6 @@ from zigpy.quirks.v2 import EntityType, QuirkBuilder
 import zigpy.types as t
 from zigpy.zcl.foundation import (
     BaseAttributeDefs,
-    BaseCommandDefs,
     DataTypeId,
     Direction,
     ZCLAttributeDef,
@@ -32,8 +31,8 @@ class NodOnPilotWireMode(t.enum8):
     ComfortMinus2 = 0x05
 
 
-class BasePilotWireCluster(CustomCluster):
-    """Base cluster to set Pilot Wire mode."""
+class NodOnPilotWireCluster(CustomCluster):
+    """NodOn manufacturer specific cluster to set Pilot Wire mode."""
 
     name: str = "PilotWireCluster"
     cluster_id: t.uint16_t = NODON_PILOT_WIRE_CLUSTER_ID
@@ -51,24 +50,8 @@ class BasePilotWireCluster(CustomCluster):
             is_manufacturer_specific=True,
         )
 
-    class ServerCommandDefs(BaseCommandDefs):
-        """Server command definitions."""
 
-        set_pilot_wire_mode = ZCLCommandDef(
-            id=0x00,
-            schema={"mode": NodOnPilotWireMode},
-            direction=Direction.Client_to_Server,
-            is_manufacturer_specific=True,
-        )
-
-
-class NodOnPilotWireCluster(BasePilotWireCluster):
-    """NodOn manufacturer specific cluster to set Pilot Wire mode."""
-
-    pass
-
-
-class AdeoPilotWireCluster(BasePilotWireCluster):
+class AdeoPilotWireCluster(NodOnPilotWireCluster):
     """Adeo manufacturer specific cluster to set Pilot Wire mode."""
 
     # Adeo SIN-4-FP-21_EQU has a weird setup where it reports 4727
@@ -86,7 +69,7 @@ nodon = (
         cluster_id=NodOnPilotWireCluster.cluster_id,
         entity_type=EntityType.STANDARD,
         translation_key="pilot_wire",
-        fallback_name="Pilot Wire",
+        fallback_name="Pilot wire",
     )
 )
 
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -53,7 +53,10 @@ class AdeoPilotWireCluster(NodOnPilotWireCluster):
     manufacturer_id_override: t.uint16_t = NODON_MANUFACTURER_ID
 
 
-nodon = QuirkBuilder(NODON, "SIN-4-FP-21").replaces(NodOnPilotWireCluster)
+nodon = (
+    QuirkBuilder(NODON, "SIN-4-FP-21")
+    .replaces(NodOnPilotWireCluster)
+)  # fmt: skip
 
 adeo = (
     nodon.clone(omit_man_model_data=True)
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -14,20 +14,6 @@ from zigpy.zcl.foundation import (
 from zigpy.zdo.types import NodeDescriptor
 
 
-class NodOnPilotWireMode(t.enum8):
-    """Pilot wire mode."""
-
-    # Codes taken from
-    # https://github.com/Koenkk/zigbee-herdsman-converters/blob/0f4833340a20db3dae625a61c41d9be0a6f952be/src/converters/fromZigbee.ts#L5285.
-
-    Off = 0x00
-    Comfort = 0x01
-    Eco = 0x02
-    FrostProtection = 0x03
-    ComfortMinus1 = 0x04
-    ComfortMinus2 = 0x05
-
-
 NODON = "NodOn"
 NODON_MANUFACTURER_ID = 4747
 NODON_PILOT_WIRE_CLUSTER_ID = 0xFC00  # 64512
@@ -50,6 +36,20 @@ ADEO_NODE_DESCRIPTION_WITH_CORRECTED_MANUFACTURER_CODE = NodeDescriptor(
     maximum_outgoing_transfer_size=500,
     descriptor_capability_field=0,
 )
+
+
+class NodOnPilotWireMode(t.enum8):
+    """Pilot wire mode."""
+
+    # Codes taken from
+    # https://github.com/Koenkk/zigbee-herdsman-converters/blob/0f4833340a20db3dae625a61c41d9be0a6f952be/src/converters/fromZigbee.ts#L5285.
+
+    Off = 0x00
+    Comfort = 0x01
+    Eco = 0x02
+    FrostProtection = 0x03
+    ComfortMinus1 = 0x04
+    ComfortMinus2 = 0x05
 
 
 class BasePilotWireCluster(CustomCluster):
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -16,25 +16,7 @@ from zigpy.zdo.types import NodeDescriptor
 NODON = "NodOn"
 NODON_MANUFACTURER_ID = 4747
 NODON_PILOT_WIRE_CLUSTER_ID = 0xFC00  # 64512
-
 ADEO = "Adeo"
-ADEO_NODE_DESCRIPTION_WITH_CORRECTED_MANUFACTURER_CODE = NodeDescriptor(
-    # the values are from a real Adeo device
-    logical_type=1,
-    complex_descriptor_available=0,
-    user_descriptor_available=0,
-    reserved=0,
-    aps_flags=0,
-    frequency_band=8,
-    mac_capability_flags=142,
-    # manufacturer_code = 4727,
-    manufacturer_code=NODON_MANUFACTURER_ID,
-    maximum_buffer_size=82,
-    maximum_incoming_transfer_size=500,
-    server_mask=11264,
-    maximum_outgoing_transfer_size=500,
-    descriptor_capability_field=0,
-)
 
 
 class NodOnPilotWireMode(t.enum8):
@@ -93,10 +75,6 @@ class AdeoPilotWireCluster(BasePilotWireCluster):
     # Adeo SIN-4-FP-21_EQU has a weird setup where it reports 4727
     # manufacturer_code in node_descriptor(), but requires NodOn's (4747)
     # manufacturer_id to execute commands and get/set attributes.
-
-    # This seems to have no effect, but I'll still leave it here.
-    # Maybe it will get fixed in future releases. NodeDescriptor magic won't
-    # be necessary then.
     manufacturer_id_override: t.uint16_t = NODON_MANUFACTURER_ID
 
 
@@ -117,8 +95,6 @@ adeo = (
     nodon.clone(omit_man_model_data=True)
     .applies_to(ADEO, "SIN-4-FP-21_EQU")
     .replaces(AdeoPilotWireCluster)
-    # please read the comment in AdeoPilotWireCluster
-    .node_descriptor(ADEO_NODE_DESCRIPTION_WITH_CORRECTED_MANUFACTURER_CODE)
 )
 
 nodon.add_to_registry()
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -38,8 +38,7 @@ class NodOnPilotWireCluster(CustomCluster):
         pilot_wire_mode = ZCLAttributeDef(
             id=0x0000,
             type=NodOnPilotWireMode,
-            # I got the following error without setting zcl_type explicitly to int:
-            # Failed to write attribute pilot_wire_mode=<NodOnPilotWireMode.FrostProtection: 3>: <Status.INVALID_DATA_TYPE: 141>
+            # need to explicitly set ZCL type
             zcl_type=DataTypeId.uint8,
             is_manufacturer_specific=True,
         )
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -56,14 +56,6 @@ class AdeoPilotWireCluster(NodOnPilotWireCluster):
 nodon = (
     QuirkBuilder(NODON, "SIN-4-FP-21")
     .replaces(NodOnPilotWireCluster)
-    .enum(
-        attribute_name=NodOnPilotWireCluster.AttributeDefs.pilot_wire_mode.name,
-        enum_class=NodOnPilotWireMode,
-        cluster_id=NodOnPilotWireCluster.cluster_id,
-        entity_type=EntityType.STANDARD,
-        translation_key="pilot_wire",
-        fallback_name="Pilot wire",
-    )
 )
 
 adeo = (
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -1,7 +1,7 @@
 """NodOn pilot wire heating module."""
 
 from zigpy.quirks import CustomCluster
-from zigpy.quirks.v2 import EntityType, QuirkBuilder
+from zigpy.quirks.v2 import QuirkBuilder
 import zigpy.types as t
 from zigpy.zcl.foundation import BaseAttributeDefs, DataTypeId, ZCLAttributeDef
 
@@ -53,10 +53,7 @@ class AdeoPilotWireCluster(NodOnPilotWireCluster):
     manufacturer_id_override: t.uint16_t = NODON_MANUFACTURER_ID
 
 
-nodon = (
-    QuirkBuilder(NODON, "SIN-4-FP-21")
-    .replaces(NodOnPilotWireCluster)
-)
+nodon = QuirkBuilder(NODON, "SIN-4-FP-21").replaces(NodOnPilotWireCluster)
 
 adeo = (
     nodon.clone(omit_man_model_data=True)
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -11,7 +11,6 @@ from zigpy.zcl.foundation import (
     ZCLAttributeDef,
     ZCLCommandDef,
 )
-from zigpy.zdo.types import NodeDescriptor
 
 NODON = "NodOn"
 NODON_MANUFACTURER_ID = 4747
--- a/zhaquirks/nodon/pilot_wire.py
+++ b/zhaquirks/nodon/pilot_wire.py
@@ -1,0 +1,126 @@
+"""NodOn pilot wire heating module."""
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import EntityType, QuirkBuilder
+import zigpy.types as t
+from zigpy.zcl.foundation import (
+    BaseAttributeDefs,
+    BaseCommandDefs,
+    DataTypeId,
+    Direction,
+    ZCLAttributeDef,
+    ZCLCommandDef,
+)
+from zigpy.zdo.types import NodeDescriptor
+
+
+class NodOnPilotWireMode(t.enum8):
+    """Pilot wire mode."""
+
+    # Codes taken from
+    # https://github.com/Koenkk/zigbee-herdsman-converters/blob/0f4833340a20db3dae625a61c41d9be0a6f952be/src/converters/fromZigbee.ts#L5285.
+
+    Off = 0x00
+    Comfort = 0x01
+    Eco = 0x02
+    FrostProtection = 0x03
+    ComfortMinus1 = 0x04
+    ComfortMinus2 = 0x05
+
+
+NODON = "NodOn"
+NODON_MANUFACTURER_ID = 4747
+NODON_PILOT_WIRE_CLUSTER_ID = 0xFC00  # 64512
+
+ADEO = "Adeo"
+ADEO_NODE_DESCRIPTION_WITH_CORRECTED_MANUFACTURER_CODE = NodeDescriptor(
+    # the values are from a real Adeo device
+    logical_type=1,
+    complex_descriptor_available=0,
+    user_descriptor_available=0,
+    reserved=0,
+    aps_flags=0,
+    frequency_band=8,
+    mac_capability_flags=142,
+    # manufacturer_code = 4727,
+    manufacturer_code=NODON_MANUFACTURER_ID,
+    maximum_buffer_size=82,
+    maximum_incoming_transfer_size=500,
+    server_mask=11264,
+    maximum_outgoing_transfer_size=500,
+    descriptor_capability_field=0,
+)
+
+
+class BasePilotWireCluster(CustomCluster):
+    """Base cluster to set Pilot Wire mode."""
+
+    name: str = "PilotWireCluster"
+    cluster_id: t.uint16_t = NODON_PILOT_WIRE_CLUSTER_ID
+    ep_attribute: str = "pilot_wire_cluster"
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        pilot_wire_mode = ZCLAttributeDef(
+            id=0x0000,
+            type=NodOnPilotWireMode,
+            # I got the following error without setting zcl_type explicitly to int:
+            # Failed to write attribute pilot_wire_mode=<NodOnPilotWireMode.FrostProtection: 3>: <Status.INVALID_DATA_TYPE: 141>
+            zcl_type=DataTypeId.uint8,
+            is_manufacturer_specific=True,
+        )
+
+    class ServerCommandDefs(BaseCommandDefs):
+        """Server command definitions."""
+
+        set_pilot_wire_mode = ZCLCommandDef(
+            id=0x00,
+            schema={"mode": NodOnPilotWireMode},
+            direction=Direction.Client_to_Server,
+            is_manufacturer_specific=True,
+        )
+
+
+class NodOnPilotWireCluster(BasePilotWireCluster):
+    """NodOn manufacturer specific cluster to set Pilot Wire mode."""
+
+    pass
+
+
+class AdeoPilotWireCluster(BasePilotWireCluster):
+    """Adeo manufacturer specific cluster to set Pilot Wire mode."""
+
+    # Adeo SIN-4-FP-21_EQU has a weird setup where it reports 4727
+    # manufacturer_code in node_descriptor(), but requires NodOn's (4747)
+    # manufacturer_id to execute commands and get/set attributes.
+
+    # This seems to have no effect, but I'll still leave it here.
+    # Maybe it will get fixed in future releases. NodeDescriptor magic won't
+    # be necessary then.
+    manufacturer_id_override: t.uint16_t = NODON_MANUFACTURER_ID
+
+
+nodon = (
+    QuirkBuilder(NODON, "SIN-4-FP-21")
+    .replaces(NodOnPilotWireCluster)
+    .enum(
+        attribute_name=NodOnPilotWireCluster.AttributeDefs.pilot_wire_mode.name,
+        enum_class=NodOnPilotWireMode,
+        cluster_id=NodOnPilotWireCluster.cluster_id,
+        entity_type=EntityType.STANDARD,
+        translation_key="pilot_wire",
+        fallback_name="Pilot Wire",
+    )
+)
+
+adeo = (
+    nodon.clone(omit_man_model_data=True)
+    .applies_to(ADEO, "SIN-4-FP-21_EQU")
+    .replaces(AdeoPilotWireCluster)
+    # please read the comment in AdeoPilotWireCluster
+    .node_descriptor(ADEO_NODE_DESCRIPTION_WITH_CORRECTED_MANUFACTURER_CODE)
+)
+
+nodon.add_to_registry()
+adeo.add_to_registry()


zhaquirks/nodon/pilot_wire.py
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    server_commands = {
+        0x0000: foundation.ZCLCommandDef(
+            "setMode",
+            {"mode": PilotWireMode},
+            False,
+            is_manufacturer_specific=True,
+        ),
+    }
```
TheJulianJES: These should be converted to the new-style zigpy `AttributeDefs`, see an example here:
https://github.com/zigpy/zha-device-handlers/blob/18c2292c14c32ea016d883e8228340d46b2e8930/zhaquirks/legrand/cable_outlet.py#L51-L58

The same is also true for server commands, example here:
https://github.com/zigpy/zha-device-handlers/blob/18c2292c14c32ea016d883e8228340d46b2e8930/zhaquirks/legrand/cable_outlet.py#L89-L97

ikruglov: done

ikruglov:
@TheJulianJES I've re-pushed commits to use the v2 interface of `QuirkBuilder()`.
Unfortunately, I didn't find an example of how to write tests for it. The tests that I wrote for the previous version didn't work due to compatibility issues, I believe.

If you think tests are necessary here, please give me an example of any test that uses the v2 interface.

pounard:
Any news on this? I would love this to be merged, I myself own a bunch of those modules.

lboue:
Please validate. I need this quirk as well

ikruglov:
@TheJulianJES, I have a question. I didn't find a better place to ask it.

There is a slight problem with this quirk. An entity added to HA gets the '_none' suffix (see attached picture). I would expect it to get the '_pilot_wire_mode' suffix. Of course, one can rename the HA entity, but getting it right from the start is better.

I did a lot of debugging on my side but got lost, to be honest. I see that the ZHA objects receive the correct  `_unique_id_suffix` value (which was my initial suspicion). But HA's entity still has '_none'.

This problem is not only with an enum but also with a configuration number. I'm [preparing a PR](https://github.com/ikruglov/zha-device-handlers/commit/6453ed6697fd1f66e7ca894888d4f27dd7d2b87b) to allow configuring impulse mode on NodOn switches (similar to [this](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7142/files)). My entity also has the `_none` suffix.

I'd appreciate your help or guidance.

P.S. With no success, I've tried setting `translation_key` and `fallback_name` attributes in quirks.

Screenshot 2024-10-12 at 11 09 06

TheJulianJES:
The `none` suffix is caused by the fact that HA doesn't have a value for the translation key. In future versions, HA will fall back to `fallback_name.
When your PR is merged, the quirks bump merged to ZHA, the ZHA bump merged to HA, we'd need to add the translation key to HA then.

ikruglov:
> The `none` suffix is caused by the fact that HA doesn't have a value for the translation key. In future versions, HA will fall back to `fallback_name.
> 
> When your PR is merged, the quirks bump merged to ZHA, the ZHA bump merged to HA, we'd need to add the translation key to HA then.

Thank you for the answer. Can you point out where I should add it in HA? I want to try it locally. I will change HA's code in my installation.

rungeard:
Hi @ikruglov thank you so much for this merge request. I'm waiting for it !

@TheJulianJES do you have any idea where this translation key is located, so that @ikruglov could test his code ?

Fructose38:
Hi @ikruglov, 
Tried to use the quirk, but got this in my home-assistant journal:
`nov. 08 18:02:05 <host> hass[370001]: 2024-11-08 18:02:05.296 ERROR (SyncWorker_5) [zhaquirks] Unexpected exception importing custom quirk 'nodon.pilot_wire'`
`nov. 08 18:02:05 <host> hass[370001]: Traceback (most recent call last):`
`nov. 08 18:02:05 <host> hass[370001]:   File "/var/lib/hass/.venv/lib/python3.12/site-packages/zhaquirks/__init__.py", line 479, in setup`
`nov. 08 18:02:05 <host> hass[370001]:     spec.loader.exec_module(module)`
`nov. 08 18:02:05 <host> hass[370001]:   File "<frozen importlib._bootstrap_external>", line 995, in exec_module`
`nov. 08 18:02:05 <host> hass[370001]:   File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed`
`nov. 08 18:02:05 <host> hass[370001]:   File "/var/lib/private/hass/zhaquirks/nodon/pilot_wire.py", line 5, in <module>`
`nov. 08 18:02:05 <host> hass[370001]:     from zhaquirks.nodon import (`
`nov. 08 18:02:05 <host> hass[370001]: ImportError: cannot import name 'NodOnPilotWireCluster' from 'zhaquirks.nodon' (/var/lib/hass/.venv/lib/python3.12/site-packages/zhaquirks/nodon/__init__.py)`
Any idea what I might be doing wrong?

Mookunicorn:
> Hi @ikruglov, Tried to use the quirk, but got this in my home-assistant journal: `nov. 08 18:02:05 <host> hass[370001]: 2024-11-08 18:02:05.296 ERROR (SyncWorker_5) [zhaquirks] Unexpected exception importing custom quirk 'nodon.pilot_wire' nov. 08 18:02:05 <host> hass[370001]: Traceback (most recent call last): nov. 08 18:02:05 <host> hass[370001]: File "/var/lib/hass/.venv/lib/python3.12/site-packages/zhaquirks/__init__.py", line 479, in setup nov. 08 18:02:05 <host> hass[370001]: spec.loader.exec_module(module) nov. 08 18:02:05 <host> hass[370001]: File "<frozen importlib._bootstrap_external>", line 995, in exec_module nov. 08 18:02:05 <host> hass[370001]: File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed nov. 08 18:02:05 <host> hass[370001]: File "/var/lib/private/hass/zhaquirks/nodon/pilot_wire.py", line 5, in <module> nov. 08 18:02:05 <host> hass[370001]: from zhaquirks.nodon import ( nov. 08 18:02:05 <host> hass[370001]: ImportError: cannot import name 'NodOnPilotWireCluster' from 'zhaquirks.nodon' (/var/lib/hass/.venv/lib/python3.12/site-packages/zhaquirks/nodon/__init__.py)` Any idea what I might be doing wrong?

Try this :

```Python
"""NODON module for custom device handlers."""
from zigpy.quirks.v2 import EntityType, QuirkBuilder
import zigpy.types as t
from zigpy.zcl.foundation import (
    BaseAttributeDefs,
    BaseCommandDefs,
    DataTypeId,
    Direction,
    ZCLAttributeDef,
    ZCLCommandDef,
)
class NodOnPilotWireMode(t.enum8):
    """Pilot wire mode."""
    OFF = 0x00
    COMFORT = 0x01
    ECO = 0x02
    FROST_PROTECTION = 0x03
    COMFORT_MINUS_1 = 0x04
    COMFORT_MINUS_2 = 0x05
NODON = "NodOn"
NODON_PILOT_WIRE_CLUSTER_ID = 0xFC00  # 64512
class NodOnPilotWireCluster(CustomCluster):
    """NodOn manufacturer specific cluster to set Pilot Wire mode."""
    name: str = "NodOnPilotWireCluster"
    cluster_id: t.uint16_t = NODON_PILOT_WIRE_CLUSTER_ID
    ep_attribute: str = "nodon_pilot_wire_cluster"
        """Attribute definitions."""
        pilot_wire_mode = ZCLAttributeDef(
            id=0x0000,
            type=NodOnPilotWireMode,
            zcl_type=DataTypeId.uint8,
        )
    class ServerCommandDefs(BaseCommandDefs):
        set_pilot_wire_mode = ZCLCommandDef(
            id=0x00,
            schema={"mode": NodOnPilotWireMode},
            direction=Direction.Client_to_Server,
        )
(
    QuirkBuilder(NODON, "SIN-4-FP-21")
    .replaces(NodOnPilotWireCluster)
    .enum(
        attribute_name=NodOnPilotWireCluster.AttributeDefs.pilot_wire_mode.name,
        enum_class=NodOnPilotWireMode,
        cluster_id=NodOnPilotWireCluster.cluster_id,
        translation_key="pilot_wire",
        fallback_name="Pilot Wire",
    )
    .add_to_registry()
)
```

Fructose38:
@Mookunicorn this time, I get no error in the logs, but my SIN-4-FP-21 module still shows a on/off switch, no modes alternative.

Fructose38:
@Mookunicorn Naming the python file `__init__.py` seems to have corrected the issue, thanks!

Fructose38:
But it looks like my heater does not support other modes, or the module fails to interpret the commands sent :(

ikruglov:
> Hi @ikruglov, Tried to use the quirk, but got this in my home-assistant journal: `nov. 08 18:02:05 <host> hass[370001]: 2024-11-08 18:02:05.296 ERROR (SyncWorker_5) [zhaquirks] Unexpected exception importing custom quirk 'nodon.pilot_wire'` `nov. 08 18:02:05 <host> hass[370001]: Traceback (most recent call last):` `nov. 08 18:02:05 <host> hass[370001]: File "/var/lib/hass/.venv/lib/python3.12/site-packages/zhaquirks/__init__.py", line 479, in setup` `nov. 08 18:02:05 <host> hass[370001]: spec.loader.exec_module(module)` `nov. 08 18:02:05 <host> hass[370001]: File "<frozen importlib._bootstrap_external>", line 995, in exec_module` `nov. 08 18:02:05 <host> hass[370001]: File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed` `nov. 08 18:02:05 <host> hass[370001]: File "/var/lib/private/hass/zhaquirks/nodon/pilot_wire.py", line 5, in <module>` `nov. 08 18:02:05 <host> hass[370001]: from zhaquirks.nodon import (` `nov. 08 18:02:05 <host> hass[370001]: ImportError: cannot import name 'NodOnPilotWireCluster' from 'zhaquirks.nodon' (/var/lib/hass/.venv/lib/python3.12/site-packages/zhaquirks/nodon/__init__.py)` Any idea what I might be doing wrong?

Make sure that you got file names right. pilot_wire.py can't import classes from __init__.py (note there is double underscore on each size of 'init').

Fructose38:
Hello @ikruglov,
After a thorough reading on the heater I'm using, it actually supports 6 modes of pilot wire, so every mode should work... But I only get Comfort/Frost protection, though HA shows all the modes... Don't know what is messing up with the mode...
I believe the quirk as you provide it should be placed in /var/lib/hass/.venv/lib/python3.12/site-packages/zhaquirks/nodon/ (on my machine), right? Because placing it in my custom_quirks_path produces the error I shown earlier...

Fructose38:
@ikruglov In fact, placing the quirk in the correct path made everything work as expected. I'm now a happy 6 modes pilot wire user :)

papylhomme:
Hello, I tried this https://github.com/zigpy/zha-device-handlers/pull/3364#issuecomment-2465425127 with a small customization for the Adeo device 

```diff
    QuirkBuilder(NODON, "SIN-4-FP-21")
+   .also_applies_to("Adeo","SIN-4-FP-21_EQU")
    .replaces(NodOnPilotWireCluster)
```

Modes are showing up in the device, and react correctly when toggling the main switch (on => comfort & off=> frost protection)
image

But I get an error when trying to set a mode directly (comfort & frost protection included)

```
home-assistant  |   File "/usr/src/homeassistant/homeassistant/components/select/__init__.py", line 188, in async_handle_select_option
home-assistant  |     await self.async_select_option(option)
home-assistant  |   File "/usr/src/homeassistant/homeassistant/components/zha/helpers.py", line 1288, in handler
home-assistant  |     raise HomeAssistantError(err) from err
home-assistant  | homeassistant.exceptions.HomeAssistantError: Failed to write attribute pilot_wire_mode=<NodOnPilotWireMode.COMFORT: 1>: <Status.UNSUPPORTED_ATTRIBUTE: 134>
```

Running HA 2024.10.2, and I used "Reconfigure" on the device instead of removing / enrolling again as the nodes are already on the wall and I'd prefer not to take the heater off again ;)
I have some modules not mounted so I will probably test with a fresh enrolling to see if it makes any difference.

ikruglov:
> Hello, I tried this [#3364 (comment)](https://github.com/zigpy/zha-device-handlers/pull/3364#issuecomment-2465425127) with a small customization for the Adeo device
> 
> ```diff
>     QuirkBuilder(NODON, "SIN-4-FP-21")
> +   .also_applies_to("Adeo","SIN-4-FP-21_EQU")
>     .replaces(NodOnPilotWireCluster)
> ```
> 
> Modes are showing up in the device, and react correctly when toggling the main switch (on => comfort & off=> frost protection) image
> 
> But I get an error when trying to set a mode directly (comfort & frost protection included)
> 
> ```
> home-assistant  |   File "/usr/src/homeassistant/homeassistant/components/select/__init__.py", line 188, in async_handle_select_option
> home-assistant  |     await self.async_select_option(option)
> home-assistant  |   File "/usr/src/homeassistant/homeassistant/components/zha/helpers.py", line 1288, in handler
> home-assistant  |     raise HomeAssistantError(err) from err
> home-assistant  | homeassistant.exceptions.HomeAssistantError: Failed to write attribute pilot_wire_mode=<NodOnPilotWireMode.COMFORT: 1>: <Status.UNSUPPORTED_ATTRIBUTE: 134>
> ```
> 
> Running HA 2024.10.2, and I used "Reconfigure" on the device instead of removing / enrolling again as the nodes are already on the wall and I'd prefer not to take the heater off again ;) I have some modules not mounted so I will probably test with a fresh enrolling to see if it makes any difference.

Hi. Based on the error' Status. UNSUPPORTED_ATTRIBUTE: 134`, it seems that Adeo has different codes or attributes. I can't comment here or test because I don't have an Adeo device, only NodOn. But I don't think that "Reconfiguring" or re-enroling will help. One needs to find the correct attributes.

papylhomme:
> Hi. Based on the error' Status. UNSUPPORTED_ATTRIBUTE: 134`, it seems that Adeo has different codes or attributes. I can't comment here or test because I don't have an Adeo device, only NodOn. But I don't think that "Reconfiguring" or re-enroling will help. One needs to find the correct attributes.

I got it working "manually" using a manufacturer_id_override as seen here https://github.com/zigpy/zha-device-handlers/issues/2777#issuecomment-2358979497 Is it possible to use the field without overriding write_attributes / read_attributes ?

image

ikruglov:
> ```diff
> SIN-4-FP-21_EQU
> ```

I have a possible solution that should satisfy both NodOn and Adeo devices. But I can't test the latter, so I need to rely on you (or others) who have Adeo.

You can find the code for `__init__.py` and `pilot_wire.py` files below:
https://github.com/ikruglov/zha-device-handlers/blob/adeo-pilot-wire/zhaquirks/nodon/__init__.py
https://github.com/ikruglov/zha-device-handlers/blob/adeo-pilot-wire/zhaquirks/nodon/pilot_wire.py

Let me know if this works. If yes, I'll merge it with this PR.

papylhomme:
> > ```diff
> > SIN-4-FP-21_EQU
> > ```
> 
> I have a possible solution that should satisfy both NodOn and Adeo devices. But I can't test the latter, so I need to rely on you (or others) who have Adeo.
> 
> You can find the code for `__init__.py` and `pilot_wire.py` files below: https://github.com/ikruglov/zha-device-handlers/blob/adeo-pilot-wire/zhaquirks/nodon/__init__.py https://github.com/ikruglov/zha-device-handlers/blob/adeo-pilot-wire/zhaquirks/nodon/pilot_wire.py
> 
> Let me know if this works. If yes, I'll merge it with this PR.

Same error as before `homeassistant.exceptions.HomeAssistantError: Failed to write attribute pilot_wire_mode=<NodOnPilotWireMode.ComfortMinus1: 4>: <Status.UNSUPPORTED_ATTRIBUTE: 134>`

Looking at the code in zigpy, it seems the override is only used if the  manufacturer parameter is not set https://github.com/zigpy/zigpy/blob/d0fa17bfcb5bfbda47146ec7dd6ef49dbedc298b/zigpy/quirks/__init__.py#L249

The device also sends bad information for the manufacturer_id (4727 here, but it responds only with 4747), it may explain why the override is not used ?
```yaml
scan:
  ieee: 08:dd:eb:ff:fe:d7:fd:51
  nwk: "0x784e"
  model: SIN-4-FP-21_EQU
  manufacturer: Adeo
  manufacturer_id: "0x4727"
  ```

I'm willing to debug this but python is not my strong suit, is it possible to edit the files in `/usr/local/lib/python3.12/site-packages/` directly to dump some variable values in the HA logs ? If so, how ?

**Comment by Mysticronos** [comment]
Hi,
Made kind of mix from this thread and #2777 and seems to work without the <Status.UNSUPPORTED_ATTRIBUTE: 134>.
Here is my code, hope this will help :

``` 
    #20241116 - Quirk zha
    #Adeo/NodOn with correction after Python/Quirk regression  

    from typing import Any, Final
    from zigpy.quirks.v2 import QuirkBuilder, EntityType, EntityPlatform
    from zigpy.zcl.foundation import (
        BaseAttributeDefs,
        BaseCommandDefs,
        ZCLAttributeDef,
        ZCLCommandDef,
        WriteAttributesStatusRecord,
        Status
    )

    from zigpy.zcl import ClusterType

    import zigpy.types as t

    import logging
    _LOGGER = logging.getLogger(__name__)

    class NodonPilotWireMode(t.enum16):
        OFF = 0x00
        COMFORT = 0x01
        ECO = 0x02
        FROST_PROTECTION = 0x03
        COMFORT_MINUS_1 = 0x04
        COMFORT_MINUS_2 = 0x05

    NODON_PILOT_WIRE_CLUSTER_ID = 0xfc00  #64512
    NODON = "NodOn"
    ADEO = "Adeo"

    class NodonPilotWireCluster(CustomCluster):
        """NodOn pilot wire manufacturer-specific cluster."""

        cluster_id: Final = NODON_PILOT_WIRE_CLUSTER_ID
        name: Final = "Pilot Wire"
        ep_attribute: Final = "pilot_wire_cluster"
        manufacturer_id_override: Final = 4747
        
            pilot_wire_mode: Final = ZCLAttributeDef(
                id=0x0000, 
                type=NodonPilotWireMode, 
                access="r"
            )

        class ClientCommandDefs(BaseCommandDefs):
            setMode: Final = ZCLCommandDef(
                id=0x0000,
                schema={"mode": NodonPilotWireMode},
                direction=True,
                is_manufacturer_specific=True
            )
        
        async def write_attributes(
            self, attributes: dict[str | int, Any], manufacturer: int | None = None
        ) -> list:
            """Override writes to the pilot_wire_mode attribute."""
            if "pilot_wire_mode" in attributes:
                await self.client_command(0, attributes.get("pilot_wire_mode"), manufacturer=4747)
                await self.read_attributes(attributes)
                return [[WriteAttributesStatusRecord(Status.SUCCESS)]]
            return await super().write_attributes(attributes, manufacturer=4747)

        async def read_attributes(
            self,
            attributes: list[int | str],
            allow_cache: bool = False,
            only_cache: bool = False,
            manufacturer: int | t.uint16_t | None = None,
        ) -> Any:
            return await super().read_attributes(attributes, allow_cache=False, only_cache=False, manufacturer=4747)

    (

        QuirkBuilder(ADEO, "SIN-4-FP-21_EQU")
        .also_applies_to(NODON,"SIN-4-FP-21")
        .replaces(NodonPilotWireCluster)
        .enum(
            attribute_name="pilot_wire_mode", 
            enum_class=NodonPilotWireMode,
            cluster_id=NODON_PILOT_WIRE_CLUSTER_ID,
            translation_key="pilot_wire",
            fallback_name="Pilot Wire",
        )
        .add_to_registry()
    )

ikruglov:
> I'm willing to debug this but python is not my strong suit, is it possible to edit the files in `/usr/local/lib/python3.12/site-packages/` directly to dump some variable values in the HA logs ? If so, how ?

Thanks for the info. I usually mount specific files directly from the host system to freely change them in a comfortable environment. But you can also try modifing files directly inside the container. Just restart HA after each change. It should work.

ikruglov:
> Made kind of mix from this thread and #2777 and seems to work without the <Status.UNSUPPORTED_ATTRIBUTE: 134>.
> Here is my code, hope this will help :

This works because you explicitly specify `manufacturer=4747` for each send and receive command. I'm pretty sure that the entire `write_attributes()` and `read_attributes()` code is unnecessary as long as we find a way to make zigpy to pass `manufacturer=4747` for Adeo and don't pass it for NodOn (or pass a NodOn-specific one). I hoped to override it in the class (via `manufacturer_id_override`), but it was not enough.

ikruglov:
> The device also sends bad information for the manufacturer_id (4727 here, but it responds only with 4747), it may explain why the override is not used ?
> 
> ```yaml
> scan:
>   ieee: 08:dd:eb:ff:fe:d7:fd:51
>   nwk: "0x784e"
>   model: SIN-4-FP-21_EQU
>   manufacturer: Adeo
>   manufacturer_id: "0x4727"
> ```

@papylhomme, so you say that Adeo device has `manufacturer_id: "0x4727"` but only replies to commands and setting/getting attributes when `manufacturer_id=0x4747`?

Also, can you show what's the output of this page for Adeo?
Screenshot 2024-11-18 at 00 07 21

papylhomme:
> > The device also sends bad information for the manufacturer_id (4727 here, but it responds only with 4747), it may explain why the override is not used ?
> > ```yaml
> > scan:
> >   ieee: 08:dd:eb:ff:fe:d7:fd:51
> >   nwk: "0x784e"
> >   model: SIN-4-FP-21_EQU
> >   manufacturer: Adeo
> >   manufacturer_id: "0x4727"
> > ```
> 
> @papylhomme, so you say that Adeo device has `manufacturer_id: "0x4727"` but only replies to commands and setting/getting attributes when `manufacturer_id=0x4747`?

Yes exactly 

```yaml
{
  "node_descriptor": {
    "logical_type": 1,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4727,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 500,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 500,
  },
```

I played with the code, and the `manufacturer_id_override` is not used where we want. It is [used in the Device class when present](https://github.com/zigpy/zigpy/blob/d0fa17bfcb5bfbda47146ec7dd6ef49dbedc298b/zigpy/device.py#L643), but the CustomCluster class reads [directly from the value provided in the endpoint](https://github.com/zigpy/zigpy/blob/d0fa17bfcb5bfbda47146ec7dd6ef49dbedc298b/zigpy/quirks/__init__.py#L378), so I end up using 4727 when sending zigbee messages

ikruglov:
> played with the code, and the `manufacturer_id_override` is not used where we want it to be. It is [used in the Device class when present](https://github.com/zigpy/zigpy/blob/d0fa17bfcb5bfbda47146ec7dd6ef49dbedc298b/zigpy/device.py#L643), but the CustomCluster class reads [directly from the value provided in the endpoint](https://github.com/zigpy/zigpy/blob/d0fa17bfcb5bfbda47146ec7dd6ef49dbedc298b/zigpy/quirks/__init__.py#L378), so I end up using 4727 when sending zigbee messages

Okay. This is what I suspected. We can override node_descriptor completely and set NodOn's 0x4747 manufacturer_code there. QuirkBuilder luckely allows this via `.node_descriptor()`. I'll push a patch later today when I'm done with work.

ikruglov:
@papylhomme, I've pushed the updated fix. Try it out:
https://github.com/ikruglov/zha-device-handlers/blob/adeo-pilot-wire/zhaquirks/nodon/pilot_wire.py

Note two things:
1. There is no longer `__init__.py`. All Pilot Wire-related code is in `pilot_wire.py.` I will also update this PR later so it has all the code in a single file, but after settling on the Adeo code.
2. I'm using features that were added relatively recently. So, you need the most up-to-date version of HA. I'm running 2024.11.2

Updated:
If this doesn't work, then I see two ways out:
1. Fallback to the suggested code where `read/write_attributes` and `commands` are overwritten in the cluster class to pass `manufacturer_id.` In this case, I believe this should be a separate quirk (and PR) because this code is unnecessary for NodOn.
2. Patch zigpy (quite possibly, it's a bug), so `manufacturer_id_override` is used. I'm still deciding whether to invest time in this.

papylhomme:
> @papylhomme, I've pushed the updated fix. Try it out: https://github.com/ikruglov/zha-device-handlers/blob/adeo-pilot-wire/zhaquirks/nodon/pilot_wire.py
> 
> Note two things:
> 
> 1. There is no longer `__init__.py`. All Pilot Wire-related code is in `pilot_wire.py.` I will also update this PR later so it has all the code in a single file, but after settling on the Adeo code.
> 2. I'm using features that were added relatively recently. So, you need the most up-to-date version of HA. I'm running 2024.11.2

Just tested it quickly and still doesn't work :-/ I'll try to debug this further when I have some time, maybe this evening, definitely in the week. I'm running 2024.11.2 too, could you point to the new feature you are using in zigpy/zha, it would help the process.

> Updated: If this doesn't work, then I see two ways out:
> 
> 1. Fallback to the suggested code where `read/write_attributes` and `commands` are overwritten in the cluster class to pass `manufacturer_id.` In this case, I believe this should be a separate quirk (and PR) because this code is unnecessary for NodOn.

:+1: Agree for the separate PR, overriding the internal methods will be a mess to maintain

> 2. Patch zigpy (quite possibly, it's a bug), so `manufacturer_id_override` is used. I'm still deciding whether to invest time in this.

I will open an issue there and see what they think about it.

ikruglov:
> could you point to the new feature you are using in zigpy/zha, it would help the process.

`clone()` method in QuirkBuilder

papylhomme:
@ikruglov it's working with a small fix, manufacturer_id is a decimal not an hexadecimal

```diff
NODON = "NodOn"
+NODON_MANUFACTURER_ID = 4747
-NODON_MANUFACTURER_ID = 0x4747
NODON_PILOT_WIRE_CLUSTER_ID = 0xFC00  # 64512
```

papylhomme:
Both heaters are responding perfectly, it's really nice I will try some clever automation with this :)

ikruglov:
> Both heaters are responding perfectly, it's really nice I will try some clever automation with this :)

Perfect. Can you try commenting `.node_descriptor(ADEO_NODE_DESCRIPTION_WITH_CORRECTED_MANUFACTURER_CODE)`? Maybe decimal vs hexadecimal was the reason that the first approach with `manufacturer_id_override` didn't work.

ikruglov:
I've re-pushed https://github.com/ikruglov/zha-device-handlers/blob/adeo-pilot-wire/zhaquirks/nodon/pilot_wire.py with fixing hexadecimal vs. decimal.

papylhomme:
> > Both heaters are responding perfectly, it's really nice I will try some clever automation with this :)
> 
> Perfect. Can you try commenting `.node_descriptor(ADEO_NODE_DESCRIPTION_WITH_CORRECTED_MANUFACTURER_CODE)`? Maybe decimal vs hexadecimal was the reason that the first approach with `manufacturer_id_override` didn't work.

Sorry it doesn't work without the node_descriptor call

ikruglov:
@papylhomme just to confirm. You've got working the following:
1. Enum in HA interface
2. set_pilot_wire_mode command available in "Manage ZigBee device" -> Cluster -> PilotWireCluster -> Command
3. Manually setting `pilot_wire_mode` attributes in the same window as (2)

Please let me know if all works or not.

papylhomme:
> @papylhomme just to confirm. You've got working the following:
> 
> 1. Enum in HA interface

I didn't try to add this in a dashboard yet, but the enum is working in the device panel

image

> 2. set_pilot_wire_mode command available in "Manage ZigBee device" -> Cluster -> PilotWireCluster -> Command
> 3. Manually setting `pilot_wire_mode` attributes in the same window as (2)

Working without specifying a code

image


> Please let me know if all works or not.

zhaquirks/nodon/__init__.py
```diff
@@ -1 +1,60 @@
...
+    FrostProtection = 0x03
+    ComfortMinus1 = 0x04
+    ComfortMinus2 = 0x05
+
+
+NODON = "NodOn"
+NODON_PILOT_WIRE_CLUSTER_ID = 0xFC00  # 64512
+
+
+class NodOnPilotWireCluster(CustomCluster):
```
TheJulianJES: All of this seems specific to the pilot wire device, so I'd probably just move it into the `pilot_wire.py` file (above the `QuirkBuilder`) and leave this file empty (except for the docstring at the top).

ikruglov: @TheJulianJES, I have prepared a re-newed PR with everything moved to pilot_wire.py. I just didn't have time to push it. I've pushed it now.

TheJulianJES: Thanks! Maybe also move the `NodOnPilotWireMode` class below the constants (so currently to line 54) 😄

TheJulianJES:
Let's wait for the changes in https://github.com/zigpy/zigpy/pull/1505 to see if they fix `manufacturer_id_override` and make the node descriptor override unnecessary.
And btw, no need to squash and force push every time here. All commits from the PR/commit will automatically be squashed when merged anyways.

zhaquirks/nodon/__init__.py
ikruglov: done

ikruglov:
@papylhomme I see that https://github.com/zigpy/zigpy/pull/1505 got merged. Can you test the code with the commented `node_descriptor()` override?

Just comment out this part:
```
    # please read the comment in AdeoPilotWireCluster
    # .node_descriptor(ADEO_NODE_DESCRIPTION_WITH_CORRECTED_MANUFACTURER_CODE)
```

But you also need somehow get the most recent zigpy changes to HA. I guess you can copy-paste the changed file in https://github.com/zigpy/zigpy/pull/1505

papylhomme:
> @papylhomme I see that [zigpy/zigpy#1505](https://github.com/zigpy/zigpy/pull/1505) got merged. Can you test the code with the commented `node_descriptor()` override?

Yes it's working without the `node_descriptor` when I apply the zigpy fix :)

ikruglov:
@TheJulianJES I've removed node_override logic since @papylhomme confirmed that the PR works without it (but with https://github.com/zigpy/zigpy/pull/1505). I I believe this PR is ready for merging.

zhaquirks/nodon/pilot_wire.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Off = 0x00
+    Comfort = 0x01
+    Eco = 0x02
+    FrostProtection = 0x03
+    ComfortMinus1 = 0x04
+    ComfortMinus2 = 0x05
+
+
+class BasePilotWireCluster(CustomCluster):
```
TheJulianJES: What about renaming this to `NodOnPilotWireCluster` and have `AdeoPilotWireCluster` inherit this to just add the `manufacturer_id_override`?
We could remove the empty class then. I don't think it adds to the readability(?)

zhaquirks/nodon/pilot_wire.py
```diff
@@ -0,0 +1,100 @@
...
+nodon = (
+    QuirkBuilder(NODON, "SIN-4-FP-21")
+    .replaces(NodOnPilotWireCluster)
+    .enum(
+        attribute_name=NodOnPilotWireCluster.AttributeDefs.pilot_wire_mode.name,
+        enum_class=NodOnPilotWireMode,
+        cluster_id=NodOnPilotWireCluster.cluster_id,
+        entity_type=EntityType.STANDARD,
+        translation_key="pilot_wire",
+        fallback_name="Pilot Wire",
```
TheJulianJES: Per HA specifications, only the first letter should be capitalized: 
```suggestion
        fallback_name="Pilot wire",
```

zhaquirks/nodon/pilot_wire.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # Codes taken from
+    # https://github.com/Koenkk/zigbee-herdsman-converters/blob/0f4833340a20db3dae625a61c41d9be0a6f952be/src/converters/fromZigbee.ts#L5285.
+
+    Off = 0x00
+    Comfort = 0x01
+    Eco = 0x02
+    FrostProtection = 0x03
+    ComfortMinus1 = 0x04
+    ComfortMinus2 = 0x05
```
TheJulianJES, line 25: I just remembered that https://github.com/zigpy/zha-device-handlers/pull/3031 has the same enum.
In that PR, we wanted to implement it as additional modes for the climate entity (so partly in ZHA).

I'll have to think about what we want to do here now.

zhaquirks/nodon/pilot_wire.py
```diff
@@ -0,0 +1,100 @@
...
+
+    class ServerCommandDefs(BaseCommandDefs):
+        """Server command definitions."""
+
+        set_pilot_wire_mode = ZCLCommandDef(
+            id=0x00,
+            schema={"mode": NodOnPilotWireMode},
+            direction=Direction.Client_to_Server,
+            is_manufacturer_specific=True,
+        )
```
TheJulianJES: This command is now unused, right? Is there a difference to directly setting the `pilot_wire_mode` attribute vs using the command?

zhaquirks/nodon/pilot_wire.py
TheJulianJES: ~~Tbh, we can probably just add it like this for now (select entity for the enum)..~~
~~If we ever get to implementing it as a climate entity in HA, there might be a breaking change though, as this would eventually be removed (and/or disabled by default).~~

~~But I guess having this would be better than not having it at all for a while.~~

EDIT: Have to ask if we can do this. We might have to implement it in ZHA otherwise.

zhaquirks/nodon/pilot_wire.py
ikruglov: ahm... yes, it's unused. I'll remove it

TheJulianJES: You can leave it in if it's present on the device. Might be useful for debugging or something. It can still be accessed through the "Manage Zigbee device" UI, even if not directly used by the quirk.
I was just curious if there's a difference to using that vs setting the attribute directly.

ikruglov:
@TheJulianJES I've addressed your suggestions except for the enum. Let me know what you want to see there.

TheJulianJES:
After some discussion, we might need to implement it as a climate entity, similar to the screenshots seen in the description of https://github.com/zigpy/zha-device-handlers/pull/3031. Since we can't expose climate devices via quirks v2, we'd need to partly implement this in ZHA.

I'm aware this is annoying, but can we remove the enum/select entity for now?
We could merge this and make the HA 2024.12 release then. With that, you could at least write the attribute manually or using some third-party components (like zha-toolkit), until support is properly implemented (or there's a decision to just implement it as select entity, like Z2M does).

zhaquirks/nodon/pilot_wire.py
```diff
@@ -0,0 +1,77 @@
...
+    QuirkBuilder(NODON, "SIN-4-FP-21")
+    .replaces(NodOnPilotWireCluster)
+    .enum(
+        attribute_name=NodOnPilotWireCluster.AttributeDefs.pilot_wire_mode.name,
+        enum_class=NodOnPilotWireMode,
+        cluster_id=NodOnPilotWireCluster.cluster_id,
+        entity_type=EntityType.STANDARD,
+        translation_key="pilot_wire",
+        fallback_name="Pilot wire",
+    )
```
TheJulianJES: Per the [above comment](https://github.com/zigpy/zha-device-handlers/pull/3364#issuecomment-2502032896), it would mean removing this. No entity will then be shown in HA (for now).
```suggestion
```

zhaquirks/nodon/pilot_wire.py
```diff
@@ -0,0 +1,77 @@
...
+    ep_attribute: str = "pilot_wire_cluster"
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        pilot_wire_mode = ZCLAttributeDef(
+            id=0x0000,
+            type=NodOnPilotWireMode,
+            # I got the following error without setting zcl_type explicitly to int:
+            # Failed to write attribute pilot_wire_mode=<NodOnPilotWireMode.FrostProtection: 3>: <Status.INVALID_DATA_TYPE: 141>
```
TheJulianJES: Let's remove or shorten this comment
```suggestion
            # need to explicitly set ZCL type
```

zhaquirks/nodon/pilot_wire.py
```diff
@@ -0,0 +1,65 @@
...
+class AdeoPilotWireCluster(NodOnPilotWireCluster):
+    """Adeo manufacturer specific cluster to set Pilot Wire mode."""
+
+    # Adeo SIN-4-FP-21_EQU has a weird setup where it reports 4727
+    # manufacturer_code in node_descriptor(), but requires NodOn's (4747)
+    # manufacturer_id to execute commands and get/set attributes.
+    manufacturer_id_override: t.uint16_t = NODON_MANUFACTURER_ID
+
+
+nodon = QuirkBuilder(NODON, "SIN-4-FP-21").replaces(NodOnPilotWireCluster)
```
TheJulianJES: We need to add `# fmt: skip` now, since ruff wants to reformat this otherwise, since it's now shorter because of no enum entity being exposed here.

```suggestion
nodon = (
    QuirkBuilder(NODON, "SIN-4-FP-21")
    .replaces(NodOnPilotWireCluster)
)  # fmt: skip
```

TheJulianJES:
Per my [previous comment](https://github.com/zigpy/zha-device-handlers/pull/3364#issuecomment-2502032896), I've removed the enum entity for now.
We can at least expose the additional Zigbee attributes by merging this initial version of a quirk then.

It needs to be discussed how the heating mode is handled: presets in a HA climate entity or just a select entity like Zigbee2MQTT provides for devices like this.

TheJulianJES (approved):
And of course, thanks for the PR!
I hope we can implement an entity for the wire mode soon. For now, you'll at least have some initial support by directly writing to the attribute.

TheJulianJES:
Opened the following issue for tracking and so this is not forgotten about:
- https://github.com/zigpy/zha-device-handlers/issues/3553

ikruglov:
@TheJulianJES. Thanks for merging, but it's unfortunate that you removed the enum as it makes automation harder.

I'm willing to investigate necessary changes in HA if you can tell me what is needed and give me some guidance.

TheJulianJES:
It also really wasn't what I wanted, but we have to make sure to adhere to Home Assistant standards. We'll follow up in the issue on what to do next.

Just thought it would be useful to get some initial support for the attribute at least. Quirks v2 adding entities is relatively new and it's uncommon for something like this to work with HA. We just have to double check we're doing everything correctly.
In this case, we might need to implement the heating modes as present on a climate entity that's created in ZHA. Again, I'll follow up in the issue, hopefully relatively soon.

ofthesun9:
many thanks @ikruglov for proposing this PR. I am currently using the quirk defined in this [comment](https://github.com/zigpy/zha-device-handlers/pull/3364#issuecomment-2465425127).

I don't get what will happen when my home-assistant container will be upgraded to the december version:
can I keep the quirk and still get the enum definition and have the "select" available and working  ?

ikruglov:
@ofthesun9, it depends on how you set it up. I run HA in a docker container and mount modified files on top of original ones. So, in this case, there will be no change. If your setup is different I can't comment then.

TheJulianJES:
You can keep using the custom quirk with the select entity.
However, due to a bug in zigpy in connection with v2 quirks, you'll need to add the following lines to the custom quirk when using Home Assistant Core 2024.12:
```python
from zigpy.quirks import DEVICE_REGISTRY
# other existing imports from custom quirk

# remove upstream quirk from registry
del DEVICE_REGISTRY._registry_v2[("NodOn", "SIN-4-FP-21")]

# original custom quirk code here
```
For more details on the issue, see: https://github.com/zigpy/zigpy/issues/1508
Custom quirks should always take priority over built-in ones. They do for "v1" quirks, but it doesn't properly work with v2 quirks, yet.

(This is not required if the original library files are modified, but that's not as easy to do and could break things. The custom quirks config + path is recommended in general.)

ofthesun9:
thanks @TheJulianJES and @ikruglov for your help.
I have now Home Assistant Core 2024.12.0 running with a custom quirk.
For reference, hereafter the main steps:

Enabling custom quirks in config/configuration.yaml

```
zha:
  enable_quirks: true
```
Adding two files in config/custom_zha_quirks/nodon
```$ cat config/custom_zha_quirks/nodon/__init__.py```
```"""NODON module for custom device handlers."""```

```$ cat config/custom_zha_quirks/nodon/pilot_wire.py```
```
"""NodOn pilot wire heating module."""
from zigpy.quirks import DEVICE_REGISTRY
# other existing imports from custom quirk

# remove upstream quirk from registry
del DEVICE_REGISTRY._registry_v2[("NodOn", "SIN-4-FP-21")]

from zigpy.quirks.v2 import EntityType, QuirkBuilder
import zigpy.types as t
from zigpy.zcl.foundation import BaseAttributeDefs, DataTypeId, ZCLAttributeDef

NODON = "NodOn"
NODON_MANUFACTURER_ID = 4747
NODON_PILOT_WIRE_CLUSTER_ID = 0xFC00  # 64512
ADEO = "Adeo"


class NodOnPilotWireMode(t.enum8):
    """Pilot wire mode."""

    # Codes taken from
    # https://github.com/Koenkk/zigbee-herdsman-converters/blob/0f4833340a20db3dae625a61c41d9be0a6f952be/src/converters/fromZigbee.ts#L5285.

    Off = 0x00
    Comfort = 0x01
    Eco = 0x02
    FrostProtection = 0x03
    ComfortMinus1 = 0x04
    ComfortMinus2 = 0x05


class NodOnPilotWireCluster(CustomCluster):
    """NodOn manufacturer specific cluster to set Pilot Wire mode."""

    name: str = "PilotWireCluster"
    cluster_id: t.uint16_t = NODON_PILOT_WIRE_CLUSTER_ID
    ep_attribute: str = "pilot_wire_cluster"

        """Attribute definitions."""

        pilot_wire_mode = ZCLAttributeDef(
            id=0x0000,
            type=NodOnPilotWireMode,
            # need to explicitly set ZCL type
            zcl_type=DataTypeId.uint8,
        )


class AdeoPilotWireCluster(NodOnPilotWireCluster):
    """Adeo manufacturer specific cluster to set Pilot Wire mode."""

    # Adeo SIN-4-FP-21_EQU has a weird setup where it reports 4727
    # manufacturer_code in node_descriptor(), but requires NodOn's (4747)
    # manufacturer_id to execute commands and get/set attributes.
    manufacturer_id_override: t.uint16_t = NODON_MANUFACTURER_ID


nodon = (
    QuirkBuilder(NODON, "SIN-4-FP-21")
    .replaces(NodOnPilotWireCluster)
    .enum(
        attribute_name=NodOnPilotWireCluster.AttributeDefs.pilot_wire_mode.name,
        enum_class=NodOnPilotWireMode,
        cluster_id=NodOnPilotWireCluster.cluster_id,
        translation_key="pilot_wire",
        fallback_name="Pilot wire",
    )
)  # fmt: skip

adeo = (
    nodon.clone(omit_man_model_data=True)
    .applies_to(ADEO, "SIN-4-FP-21_EQU")
    .replaces(AdeoPilotWireCluster)
)

nodon.add_to_registry()
adeo.add_to_registry()
```

Skunnyk:
Thank you, nice and clean workaround waiting for full upstream integration :)

PR #3378, opened 2024-09-21
Add GIEX smart valve `_TZE284_7ytb3h8u` variant

prairiesnpr:
## Proposed change
Add GEIX Smart Valve Variant _TZE284_7ytb3h8u.

## Additional information
Newer TUYA devices are adding an additional cluster ```0xed00``` this adds that variant. [Manufacturer confirmed that this is working.](https://community.home-assistant.io/t/giex-smart-product-manufacturer/772792/10)

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_valve.py
+++ b/zhaquirks/tuya/ts0601_valve.py
@@ -16,7 +16,7 @@ from zhaquirks.const import (
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-from zhaquirks.tuya import EnchantedDevice, TuyaLocalCluster
+from zhaquirks.tuya import TUYA_CLUSTER_ED00_ID, EnchantedDevice, TuyaLocalCluster
 from zhaquirks.tuya.mcu import (
     DPToAttributeMapping,
     TuyaMCUCluster,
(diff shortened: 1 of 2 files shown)


zhaquirks/tuya/ts0601_valve.py
```diff
@@ -422,3 +422,48 @@ class GiexValve(CustomDevice):
...
+            # input_clusters=[0x0000, 0x0004, 0x0005, 0xed00, 0xef00]
+            # output_clusters=[0x000a, 0x0019]>
+            1: {
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.SMART_PLUG,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    TUYA_CLUSTER_ED00_ID,
```
TheJulianJES, line 445: As a note, this is not in the `replacement`. It doesn't really matter though, as there are no attributes on it (that we use) and I think that's in the manufacturer cluster range, so we'd need an (empty) CustomCluster.

TheJulianJES (approved):
Thanks!

PR #3379, opened 2024-09-22
Remove LevelControl cluster for IKEA INSPELNING plugs

folfy:
The new smart plug suffers from the same issue like all IKEA plugs, unnecessary level controls in the configuration panel:
inspelning

## Proposed change

Expanded Quirk to be applied to new plug as well.

## Checklist

### DIFF
--- a/zhaquirks/ikea/plug.py
+++ b/zhaquirks/ikea/plug.py
@@ -9,6 +9,7 @@ from zhaquirks.ikea import IKEA
 (
     QuirkBuilder(IKEA, "TRADFRI control outlet")
     .also_applies_to(IKEA, "TRETAKT Smart plug")
+    .also_applies_to(IKEA, "INSPELNING Smart plug")
     .removes(LevelControl.cluster_id)
     .add_to_registry()
 )


MattWestb:
Its the "personality" of all IKEA Zigbee 3.0+ devices then they is using remotes with level with On / Off then being direct paired with devices (tush link).
Funny is that Tretakt is also exposing the light level attributes with Dirigera and Matter integration to HA so very likely Inspelning is doing the same.

By the way where did you getting the plug ?
I have not seen it in Vienna but its in there the system but hidden and not in the store shelf.

MattWestb (approved):
Looks very OK  for me.

folfy:
@MattWestb Yeah, it's not on the website or in their "normal" system yet, but I asked at Wien Westbahnhof on Thursday, and they told me they get a shipment of 30pcs on Friday to be put into the shelves the same day. Sure enough they were there in the afternoon, as promised, so feel free to grab one while they last. :D

Also, there's issue #3374 - Looks like we need to update the ac_power_divisor and consider it on power-value updates, see my [last comment](https://github.com/zigpy/zha-device-handlers/issues/3374#issuecomment-2366735032)

MattWestb:
I was also at Wbh on Thursday but was not asking and need going on Monday if i can get time for getting some of them.
Its very likely the power is not OK set in the device but also if it the original firmware or the latest production ?
If its only the factory one then leaving it and update the firmware to the latest production one and dont need struggling with quirk.

MattWestb:
Interesting testing you have done !!
I was not knowing it  was allowed changing this attribute "on the fly".
First i was thinking they way using wrong datatypes and must being corrected for getting the full range but with your testing i think its one new way of "extending the Zigbee specification" but i think its need being implanted i ZHA /zigpy.

Ping our @puddly and @TheJulianJES  !!!
Some thinking of this problem ?
If IKEA is using "dynamic changing attribute" on the Measurement Cluster we must pulling it or catching if getting it reprtebal in the device so the system can using it.

puddly:
Is this attribute reportable? I wonder how the IKEA hubs avoids polling the attribute.

folfy:
@puddly Unfortunately don't have an IKEA hub, but tomorrow I'll try to check in the ZHA logs if I maybe can see the attribute being pushed by the plug automatically, where it maybe/hopefully just isn't evaluted by us yet. So far just been manually fetching/checking it for the first analysis of the issue:
inspelning_cluster
If not, I'll consider to get some IKEA hub, just to find out what's going on there xP

puddly:
The attribute is apparently reportable according to the spec, we should just add it to the list of attributes to set up for reporting.

I hope the device sends updates to that attribute first. Otherwise, the quirk will need logic to handle it.

MattWestb:
Thanks our @puddly for taking on look !!
You have  getting one email with  my first paring with DIRIGERA  but its on the factors firmware so perhaps not OK.
I do one new sniff then its updated and sending it to you.

Thanks in advance !!!

MattWestb:
With the latest firmware is indeed reporting the `AC Power Divisor (0x0605)` if the power his higher then the lowerst setting and also updating it later if its needed.
But is normally not sending one 0 if the current  power is zero in the  last report.
So some logic  need being  implemented here.
Here is one sniff with one toaster  with 1300W and one light of 0.5 W and  some zero loaded between.

[INSPPELNING03.pcapng.gz](https://github.com/user-attachments/files/17096674/INSPPELNING03.pcapng.gz)

(PS: only Puddly have my network key)

MattWestb:
PS: I think we shall moving  the power factor out of this PR then its good ass it it and making one new PR / issue for the "dynamic power factor" implementation.

TheJulianJES (approved):
Thanks!
(The `ac_power_divisor` attribute stuff needs to be addressed separately in ZHA.)

PR #3380, opened 2024-09-22
Add Tuya TS130F `_TZ3000_qqdbccb3` curtain roller

ethervoid:
## Proposed change

This PR includes a new signature for the Zemismart curtain switch with reference WN2-EC1E because the current one doesn't work for this device.

Basically is a new signature for the device already present but this one has a different signature.

## Additional information

signal-2024-09-22-172942_002

signal-2024-09-22-172942_003

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts130f.py
+++ b/zhaquirks/tuya/ts130f.py
@@ -196,6 +196,7 @@ class TuyaZemismartWNEC1ETS130F(CustomDevice):
                 DEVICE_TYPE: zha.DeviceType.WINDOW_COVERING_DEVICE,
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
+                    Identify.cluster_id,
                     Groups.cluster_id,
                     Scenes.cluster_id,
                     TuyaWithBacklightOnOffCluster,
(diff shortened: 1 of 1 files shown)


javicalle (changes requested):
It's needed to remove the `Identify` cluster in the `replacement` block?
I would suggest not to remove.

ethervoid:
Yes! You're completely right, I added it to the signature but not to the replacement part.

javicalle (approved):
LGTM 

Thanks.

zhaquirks/tuya/ts130f.py
```diff
@@ -164,6 +164,53 @@ class TuyaZemismartTS130F(CustomDevice):
     }
 
 
+class TuyaZemismartWNEC1ETS130F(CustomDevice):
+    """Tuya ZemiSmart smart curtain roller shutter. WN-EC1E version."""
+
+    signature = {
+        # SizePrefixedSimpleDescriptor(endpoint=1, profile=104, device_type=0x0202, device_version=1, input_clusters=[0x0000, 0x0003, 0x0004, 0x0005, 0x0102], output_clusters=[0x0019]))
```
TheJulianJES: This doesn't seem to be accurate. We can just remove it.
```suggestion
```

TheJulianJES (approved):
Thanks, @ethervoid and @javicalle!

PR #3408, opened 2024-10-08
Add logging to LocalDataCluster

dmulcahey:
## Proposed change

This PR adds logging to LocalDataCluster methods.

## Checklist

### DIFF
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -76,8 +76,7 @@ class LocalDataCluster(CustomCluster):
 
     async def _configure_reporting(self, *args, **kwargs):  # pylint: disable=W0221
         """Prevent remote configure reporting."""
-        self.log(
-            logging.DEBUG,
+        self.debug(
             f"configuring reporting for LocalDataCluster with cluster_id: {self.cluster_id}",
             *args,
             **kwargs,
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -76,13 +76,13 @@ class LocalDataCluster(CustomCluster):
 
     async def _configure_reporting(self, *args, **kwargs):  # pylint: disable=W0221
         """Prevent remote configure reporting."""
-        self.debug("configuring reporting for LocalDataCluster", *args, **kwargs)
+        self.debug("configuring reporting for LocalDataCluster")
         return (foundation.ConfigureReportingResponse.deserialize(b"\x00")[0],)
 
     async def read_attributes_raw(self, attributes, manufacturer=None):
         """Prevent remote reads."""
         msg = "reading attributes for LocalDataCluster"
-        self.debug(f"{msg} attributes: {attributes} manufacturer: {manufacturer}")
+        self.debug(f"{msg}: attributes={attributes} manufacturer={manufacturer}")
         records = [
             foundation.ReadAttributeRecord(
                 attr, foundation.Status.UNSUPPORTED_ATTRIBUTE, foundation.TypeValue()
@@ -101,7 +101,7 @@ class LocalDataCluster(CustomCluster):
     async def write_attributes(self, attributes, manufacturer=None):
         """Prevent remote writes."""
         msg = "writing attributes for LocalDataCluster"
-        self.debug(f"{msg} attributes: {attributes} manufacturer: {manufacturer}")
+        self.debug(f"{msg}: attributes={attributes} manufacturer={manufacturer}")
         for attrid, value in attributes.items():
             if isinstance(attrid, str):
                 attrid = self.attributes_by_name[attrid].id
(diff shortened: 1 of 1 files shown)


zhaquirks/__init__.py
```diff
@@ -66,18 +66,37 @@ class LocalDataCluster(CustomCluster):
...
+            logging.DEBUG,
+            f"configuring reporting for LocalDataCluster with cluster_id: {self.cluster_id}",
+            *args,
+            **kwargs,
+        )
         return (foundation.ConfigureReportingResponse.deserialize(b"\x00")[0],)
 
     async def read_attributes_raw(self, attributes, manufacturer=None):
         """Prevent remote reads."""
+        msg = f"reading attributes for LocalDataCluster with cluster_id: {self.cluster_id}"
```
TheJulianJES: It's probably fine, but this might be logged a lot (in deubg). Like for Xiaomi plugs with multiple `LocalDataCluster`s when they're polled.

zhaquirks/__init__.py
```diff
@@ -66,18 +66,37 @@ class LocalDataCluster(CustomCluster):
 
     async def bind(self):
         """Prevent bind."""
+        self.log(
+            logging.DEBUG,
```
TheJulianJES: I think just using `self.debug` is preferred, as that's what is used in zigpy all the time.

TheJulianJES: Also, do we want to use the zigpy logger here or create a quirks logger? When we use the zigpy one, I think we always have a zigpy prefix in logs, not a zhaquirks one, so it's sometimes a bit hard to differentiate what's coming from where.
It's not a huge issue here, as we can clearly identify it as coming from quirks because of the message, but it's an issue with other quirks as well (that use the zigpy logger with generic messages, so it's unclear where they're coming from).
We may want to standardize our approach to this at some point.

dmulcahey: I didn’t see the log mixin in the class hierarchy. I’ll look again.

zhaquirks/__init__.py
dmulcahey: At debug we log all the traffic anyway so this is an outlier…

zhaquirks/__init__.py
dmulcahey: IMO the log is from a cluster impl so it should use the cluster logger but I am open to discussing some sort of standardization

dmulcahey: Ugh it is pulled in by CatchingTaskMixin

dmulcahey: Updated

zhaquirks/__init__.py
```diff
@@ -66,26 +66,22 @@ class LocalDataCluster(CustomCluster):
...
         return (foundation.Status.SUCCESS,)
 
     async def _configure_reporting(self, *args, **kwargs):  # pylint: disable=W0221
         """Prevent remote configure reporting."""
-        self.debug(
-            f"configuring reporting for LocalDataCluster with cluster_id: {self.cluster_id}",
-            *args,
-            **kwargs,
-        )
+        self.debug("configuring reporting for LocalDataCluster", *args, **kwargs)
```
TheJulianJES: Does this work with `*args` and `**kwargs`?

dmulcahey: image

TheJulianJES: I don't think we can just pass additional arguments and not do anything with them though. How are they expected to appear in a log message?

Did a quick test now and the added log message just seems to break attribute reporting configuration completely:
```python
self = <Logger zigpy.zcl (DEBUG)>, level = 10, msg = '[%s:%s:0x%04x] configuring reporting for LocalDataCluster'
args = ('0x1234', 1, 2820, [AttributeReportingConfig(direction=0, attrid=0x050B, datatype=<DataTypeId.int16: 41>, min_interval=10, max_interval=20, reportable_change=1)]), kwargs = {'manufacturer': None, 'stacklevel': 4}

    def log(self, level, msg, *args, **kwargs):
        """
        Log 'msg % args' with the integer severity 'level'.
    
        To pass exception information, use the keyword argument exc_info with
        a true value, e.g.
    
        logger.log(level, "We have a %s", "mysterious problem", exc_info=True)
        """
        if not isinstance(level, int):
            if raiseExceptions:
                raise TypeError("level must be an integer")
            else:
                return
        if self.isEnabledFor(level):
>           self._log(level, msg, args, **kwargs)
E           TypeError: Logger._log() got an unexpected keyword argument 'manufacturer'

/opt/homebrew/Cellar/python@3.12/3.12.7/Frameworks/Python.framework/Versions/3.12/lib/python3.12/logging/__init__.py:1609: TypeError
```

dmulcahey: Thanks for saving me :) I am in the office today so I couldn't test it directly.

TheJulianJES (approved):

PR #3412, opened 2024-10-10
Add Giex GX04 Soil Sensor

prairiesnpr:
## Proposed change

Add Giex GX04 Soil Moisture Probe

image

## Checklist


dmulcahey (approved):

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -485,3 +487,84 @@ class TuyaSoilSensorVar02(CustomDevice):
...
+        Celsius = 0x00
+        Fahrenheit = 0x01
+
+    attributes = TuyaMCUCluster.attributes.copy()
+    attributes.update(
+        {
+            BATTERY_STAT_ATTR: ("battery_status", GiexBatteryStatus, True),
+            TEMP_SW_ATTR: ("temp_cf_switch", GiexTempEnum, True),
+        }
+    )
```
TheJulianJES: Ideally, we'd use the new `AttributeDefs`, but I do see that the `TuyaMCUCluster` isn't converted yet. Are there any attributes on that cluster?
EDIT: It's converted to `AttributeDefs` now 😄

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -485,3 +488,31 @@ class TuyaSoilSensorVar02(CustomDevice):
...
+    TuyaQuirkBuilder("_TZE284_nhgdf6qr", "TS0601")  # Giex GX04
+    .tuya_battery(dp_id=15)
+    .tuya_temperature(dp_id=5, scale=10)
+    .tuya_soil_moisture(dp_id=3)
+    .tuya_enum(
+        dp_id=14,
+        attribute_name="battery_status",
+        enum_class=GiexBatteryStatus,
+        translation_key="battery_status",
+        fallback_name="Battery Status",
```
TheJulianJES: Due to [Home Assistant naming](https://developers.home-assistant.io/blog/2022/07/10/entity_naming/#:~:text=the%20rest%20of%20the%20words%20are%20lower%20case), this should be "Battery status".
```suggestion
        fallback_name="Battery status",
```

TheJulianJES:
I think we're fine to add it, but what's the difference between the actual battery entity and the battery status? (other than the latter just having three states)

prairiesnpr:
> I think we're fine to add it, but what's the difference between the actual battery entity and the battery status? (other than the latter just having three states)

Really no reason for it IMO, it's just another way to indicate the battery state. I left it since I was trying to show various uses of the TuyaQuirkBuilder, but set it initially disabled since it's not all that useful. I'm fine with removing it, if you don't want it.

TheJulianJES (approved):

PR #3414, opened 2024-10-11
Refactor Tuya `EnchantedDevice` to use custom configuration

TheJulianJES:
## Proposed change
This PR refactors the Tuya spell code in `EnchantedDevice` to use the `apply_custom_configuration` hook in `BaseCustomDevice` (introduced in recent zha/zigpy versions).

This allows us to remove the workarounds of "enchantable" clusters where `bind()` was hijacked and called the Tuya spells.
With this PR, we can call the Tuya spell code way easier (and slightly earlier).

A bunch of now irrelevant documentation and tests are also removed. The tests can likely be reduced a bit more in future PRs, as we no longer need to test that every device/quirk correctly casts the spell, but it also doesn't hurt anything.
So, if we want to change the "spell tests" further, I'd push that to a future PR.

## Additional information
Requires:
- https://github.com/zigpy/zha/pull/231
- https://github.com/zigpy/zha/pull/239 (kind of)

## Checklist

### DIFF
--- a/tests/test_tuya_spells.py
+++ b/tests/test_tuya_spells.py
@@ -135,3 +135,25 @@ def test_tuya_spell_devices_valid():
                 f"Enable at least one Tuya spell by setting `TUYA_SPELL_READ_ATTRIBUTES` or `TUYA_SPELL_DATA_QUERY` "
                 f"or inherit CustomDevice rather than EnchantedDevice."
             )
+
+        tuya_cluster_exists = False  # cluster subclassing TuyaNewManufCluster existing
+
+        # iterate over all clusters in the replacement
+        for endpoint_id, endpoint in quirk.replacement[ENDPOINTS].items():
+            if endpoint_id != 1:  # cluster with data query spell command on endpoint 1
+                continue
+            for cluster in endpoint[INPUT_CLUSTERS] + endpoint[OUTPUT_CLUSTERS]:
+                if not isinstance(cluster, int):
+                    # check if there's a valid Tuya cluster where the id wasn't modified
+                    if (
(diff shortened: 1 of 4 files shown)


dmulcahey (approved):

tests/test_tuya_spells.py
```diff
@@ -186,39 +135,3 @@ def test_tuya_spell_devices_valid():
...
-        elif enchanted_clusters > 1:
-            pytest.fail(
-                f"{quirk} has more than one cluster subclassing `TuyaEnchantableCluster` on endpoint 1"
-            )
-
-        # an EnchantedDevice with the data query spell must also have a cluster subclassing TuyaNewManufCluster
-        if quirk.tuya_spell_data_query and not tuya_cluster_exists:
-            pytest.fail(
-                f"{quirk} set Tuya data query spell but has no cluster subclassing `TuyaNewManufCluster` on endpoint 1"
-            )
```
TheJulianJES: Just noticed that this small part is still required, so I'll add that back.
That's the "second spell" for which the command is on the `TuyaNewManufCluster`.

TheJulianJES: https://github.com/zigpy/zha-device-handlers/pull/3414/commits/23f47f07b90a485aa85e3674c842b30550d8310a

TheJulianJES:
Just realized that we do the device specific configuration in ZHA after configuring all clusters.
We may want to move this a couple of lines up: https://github.com/zigpy/zha/blob/4c5fc061374aeaae5a80769040d552177dc8c6d0/zha/zigbee/device.py#L725-L727

PR (with more info):
- https://github.com/zigpy/zha/pull/239

PR #3415, opened 2024-10-11
Add `_VALID_ATTRIBUTES` to LocalDataCluster, fix Tuya datapoint mappings on LocalDataClusters

TheJulianJES:
### UNTESTED
This should probably be tested with some actual devices. Unit tests pass already.

## Proposed change
- This adds `_VALID_ATTRIBUTES` to `LocalDataCluster`.
This allows entities in ZHA to be created*, as the initial attribute reads will return `None` with success status, instead of "unsupported attribute" for all attributes in the `_VALID_ATTRIBUTES` list.

- The PR also automatically adds all mapped to attributes from Tuya datapoints on `LocalDataClusters` to `_VALID_ATTRIBUTES` to ensure entities in ZHA can be created on pairing.

- A basic test checking that attribute reads work as expected on a `LocalDataCluster` is also added.

---

*: Do note that some entities in ZHA will not be created for `None` as a value in the attribute cache. This is only the case for configuration switch, configuration select, and configuration number entities IIRC. This should not be an issue, as this PR is mainly intended to fix the "actual ZCL entities" like temperature sensors and so on. Those are also created with `None` values on pairing. 

I explicitly chose to not populate the attribute cache with "valid" numbers like `0`, as the data type can vary and that's somewhat unclean in general. If needed, we can add an `initial_value` to `DPToAttributeMapping` in the future, but I do not think this will be an issue at all.
For ZCL devices, the attribute cache will also contain `None` at first. We just follow that behavior for Tuya devices that use a `LocalDataCluster` with an attribute being mapped to from a Tuya datapoint.

## Additional information
Other non-Tuya quirks should likely also utilize `_VALID_ATTRIBUTES` if they are currently using a `LocalDataCluster` where some attributes are only updated later and not when a device is initially paired.

I've also noticed that a bunch of Tuya mappings do not exist on some quirks. It's not really an issue though, but I needed to account for that.

~~If the PR approach is good, then I'll probably split the PR to only include the `_VALID_ATTRIBUTES` part in this PR and have a second PR with the third commit for the Tuya mapped attributes being initialized as "valid".~~
EDIT: It was both done in this PR now.

## Checklist


TheJulianJES:
Let’s try this on HA `dev`. Should work fine.

PR #3411, opened 2024-10-09
Add Giex GX03 Valve

prairiesnpr:
## Proposed change

Add GIEX GX03 Quirk.

image

## Checklist

- [X] The changes are tested and work correctly
- [X] `pre-commit` checks pass / the code has been formatted using Black
- [X] Tests have been added to verify that the new code works

PR #3417, opened 2024-10-11
Add Tuya v2 quirk builder

prairiesnpr:
## Proposed change

This extends QuirkBuilder to add `add_tuya_dp` this allows DPs and converters to be added in the following manner.

```python
.tuya_dp(
  3,
  TuyaSoilMoisture.ep_attribute,
  "measured_value",
  converter=lambda x: x * 100,
)
```

Or if the if it's known attribute, such as battery, temperature, or soil_moisture.

`.tuya_temperature(dp_id=5, scale=10)`

I've implemented replacement V2 Quirks for the existing soil quirks and tested with a `_TZE284_aao3yzhs`.

## Additional information

Full Quirk Example:
Orginal V2: https://github.com/zigpy/zha-device-handlers/pull/3411 
With TuyaQuirkBuilder: https://gist.github.com/prairiesnpr/e32e10b86a0736f5ff09693f6006ab3b

## Checklist


TheJulianJES:
Nice work!

Currently, we have a bit of an issue with duplicating parts of a v2 quirk. There's no proper way to share them, so a downside with this approach is that there'll be somewhat of a duplication for basic datapoints like temperature and so on.
Personally, I'd probably keep the Tuya datapoint mappings in the temperature/soil clusters and rather only define Tuya datapoints in a v2 quirk (with this approach) for specific/custom entities that are unlikely to be common across devices.

That would allow us to easily re-use common clusters like temperature/soil (in v1 and v2 quirks), but adds the flexibility to just tack on custom v2 entities with Tuya datapoints in a v2 quirk if needed.

In the long term, we should still take a look at how we can better share parts of a v2 quirk across different quirks, but I think common datapoint mappings just make more sense to have in a cluster compared to a in a device/quirk. What do you think?
EDIT: See my [newer comment](https://github.com/zigpy/zha-device-handlers/pull/3417#issuecomment-2407813888).

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -334,154 +334,59 @@ class TuyaTempHumiditySensorVar04(CustomDevice):
...
+(
+    TuyaQuirkBuilder("_TZE200_myd45weu", "TS0601")
+    .also_applies_to("_TZE200_ga1maeof", "TS0601")
+    .also_applies_to("_TZE200_9cqcpkgb", "TS0601")
+    .also_applies_to("_TZE204_myd45weu", "TS0601")
+    .add_tuya_dp(
+        5,
+        TuyaTemperatureMeasurement.ep_attribute,
+        "measured_value",
+        converter=lambda x: x * 100,
```
TheJulianJES: Hmm, although we need to treat the temperature datapoint slightly differently in these quirks (`* 10` vs `* 100`).
Maybe it does make sense to use `TuyaQuirkBuilder` for this then, since we'd need two separate temperature clusters otherwise (although one could extend the other one there and just override a value like `multiplication = 10` that's used in the base (with a default).

dmulcahey:
Could we add methods to this for the known datapoints / clusters that are shared? Either by device type (I think they have required dp’s for each device type) or by functionality? Maybe something like .temperature, .onoff, .fan etc or if device type specific something like .smart_switch, .relay, .plug, etc? And have that pull in the dp’s (or clusters) depending on which way we decide to go?

TheJulianJES:
Ah, regarding my previous comment. I somewhat forgot that the Tuya datapoints are obviously defined on the Tuya manufacturer cluster and not on the virtual "temperature cluster". I had in my mind that you just do `.replaces(TuyaTempCluster)` (and have the datapoints defined in that cluster), but that's obviously not the case.
So, you can kind of disregard my previous comment (except the duplication part).

Some methods like `.tuya_temperature` (with optional multiplication and optional other datapoint kwargs) in the `TuyaQuirkBuilder` would make sense. It could add a virtual ZCL temperature cluster and create the datapoint mapping.
I think I really like that approach. Should be easy to implement as well with the current `TuyaQuirkBuilder`.

This would also cause a lot less duplication across devices, as it's only like a single line (maybe with a differing datapoint or multiplication value). I *really* like this idea 😄

prairiesnpr:
I think I prefer methods by functionality but still pass the DP. We could even have them go ahead and add the appropriate entity, a switch if it's onoff etc. 

I'd like to be a bit careful that we don't abstract too much. A typical user building a quirk will go pull a list of DPs using something like https://www.zigbee2mqtt.io/advanced/support-new-devices/03_find_tuya_data_points.html. I'd like to keep it straightforward so that they can quickly scan a quirk and see that the quirk matches their device and that they can just add the manufacturer and model. I think having the list of DPs in the quirk will help with that even at the cost of some duplication.

dmulcahey:
> I think I prefer methods by functionality but still pass the DP. We could even have them go ahead and add the appropriate entity, a switch if it's onoff etc.
> 
> I'd like to be a bit careful that we don't abstract too much. A typical user building a quirk will go pull a list of DPs using something like https://www.zigbee2mqtt.io/advanced/support-new-devices/03_find_tuya_data_points.html. I'd like to keep it straightforward so that they can quickly scan a quirk and see that the quirk matches their device and that they can just add the manufacturer and model. I think having the list of DPs in the quirk will help with that even at the cost of some duplication.

Couldn’t we have our own version of this that also determines if the quirk exists and maybe even creates the PR if it doesn’t 🤔

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -334,154 +327,23 @@ class TuyaTempHumiditySensorVar04(CustomDevice):
...
-
-
-class TuyaSoilSensorVar02(CustomDevice):
-    """Tuya temp and humidity sensor (variation 05)."""
+(
+    TuyaQuirkBuilder("_TZE284_aao3yzhs", "TS0601")
+    .also_applies_to("_TZE284_sgabhwa6", "TS0601")
+    .temperature(5, converter=lambda x: x * 10)
+    .battery(15)
+    .soil_moisture(3)
```
TheJulianJES: Would it be nicer to always pass the datapoint ids via keyword, so e.g. `dp_id=15`?
It would be slightly clearer what that value means then IMO.

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -334,154 +327,23 @@ class TuyaTempHumiditySensorVar04(CustomDevice):
...
-        15: "_dp_2_attr_update",
-    }
-
-
-class TuyaSoilSensorVar02(CustomDevice):
-    """Tuya temp and humidity sensor (variation 05)."""
+(
+    TuyaQuirkBuilder("_TZE284_aao3yzhs", "TS0601")
+    .also_applies_to("_TZE284_sgabhwa6", "TS0601")
+    .temperature(5, converter=lambda x: x * 10)
```
TheJulianJES: Also, would an extra argument like `multiplication` make sense here (with maybe `10` as default `.temperature`?
So, we could just pass `multiplication=100` if we want that. I think it's rare we need to do more complex calculations for basic entities (like temperature)(?)

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -753,7 +753,7 @@ def __init__(
 
     def battery(
         self,
-        dp_id: int,
+        dp_id: int = 15,
```
TheJulianJES: I think we don't even need a default for this, even if we always pass it in as a kwarg.
That said, if the datapoints for battery/soil/temp are mostly/always the same, the defaults might still make sense. But I also see why you don't want to hide everything in defaults.

zhaquirks/tuya/ts0601_sensor.py
prairiesnpr: I had the same thought, the converter was the safe option, are we good with something like scale=10, then if it's something else they can add it using `.tuya_dp`?

TheJulianJES: I think that should be fine.

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -724,3 +730,122 @@ class TuyaLevelControlManufCluster(TuyaMCUCluster):
...
+            ]
+        ] = None,
+        dp_converter: Optional[
+            Callable[
+                [
+                    Any,
+                ],
+                Any,
+            ]
+        ] = None,
```
TheJulianJES: Not sure if this works, can we format `converter` and `dp_converter` typing like this?
```python
        converter: Optional[Callable[[Any], Any]] = None,
        dp_converter: Optional[Callable[[Any], Any]] = None,
```

prairiesnpr: Ruff is reformatting it to look like that during pre-commit

prairiesnpr: Went ahead and set ruff to skip that line, I like it better compact.

zhaquirks/tuya/ts0601_sensor.py
dmulcahey: Should we just stick w/ the common multiplier and divisor like we have for other clusters?

javicalle: I agree to keep as simple as possible, but there are some really exotic converters around:
https://github.com/zigpy/zha-device-handlers/blob/7d81ef9e8f112de3fc76934ec0dc885aacde07a7/zhaquirks/tuya/ts0601_illuminance.py#L57

TheJulianJES: Hmm, that's only illuminance, right? We can just put the "exotic part" of the converter in the `.illuminance()` method.

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -724,3 +730,108 @@ class TuyaLevelControlManufCluster(TuyaMCUCluster):
...
+        self.tuya_dp(
+            dp_id,
+            soil_cfg.ep_attribute,
+            "measured_value",
+            converter=lambda x: x * scale,
+        )
+        self.adds(soil_cfg)
+        return self
+
+    def temperature(
```
TheJulianJES: I'm not set on this yet, but would it make sense to append a `tuya_` prefix to all Tuya related methods?
Just so there aren't any conflicts down the road and it's clear this is for Tuya datapoints?

prairiesnpr: I'm leaning towards adding the tuya prefix. I'm working on adding support for adding attributes, switches, and enums now, and if we add the prefix, then you can still call the original methods if needed. I think it adds clarity and keeps functionality that we may need in the future if we run into something that doesn't fit the normal approach.

prairiesnpr:
https://github.com/zigpy/zha-device-handlers/pull/3417/commits/b83c29fc50a895d1e4a19efb7040b190de820fc8 adds methods corresponding to switch, sensor, binary_sensor, enum, and number from QuirkBuilder. This allows for a quirks such as this https://github.com/zigpy/zha-device-handlers/pull/3411 to be simplified to https://gist.github.com/prairiesnpr/e32e10b86a0736f5ff09693f6006ab3b

prairiesnpr:
Will need a zigpy release for tests to pass, https://github.com/zigpy/zigpy/commit/0707ff314c6975e33e61d29fadbe774e775d8a83 added `entity_type` for switches.

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -740,6 +750,24 @@ class TuyaSoilMoisture(SoilMoisture, TuyaLocalCluster):
...
+    """Tuya Valve Water consumed cluster."""
+
+    VOLUME_LITERS = 0x0007
+    WATER_METERING = 0x02
+
+    """Setting unit of measurement."""
+    _CONSTANT_ATTRIBUTES = {
+        0x0300: VOLUME_LITERS,
+        0x0306: WATER_METERING,
+    }
```
TheJulianJES: We should be able to use zigpy `AttributeDefs` for this. Example:
```python
    _CONSTANT_ATTRIBUTES = {
        Metering.AttributeDefs.unit_of_measure.id: attr_value,
        Metering.AttributeDefs.metering_device_type.id: attr_value,
    }
```
It looks like "units of measure" and "metering device types" do not have their own classes in zigpy at the moment, so the constants for `VOLUME_LITERS` and `WATER_METERING` still make sense there.

TheJulianJES (commented):
Also, would it make sense to move everything Tuya quirk builder related (including the small local clusters) into something like `tuya/builder.py`, `tuya/mcu/builder.py`, or `tuya/builder/__init__.py`?

prairiesnpr:
Do we have any other Tuya methods we would want added to this initial PR?

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+            invert_attribute_name=invert_attribute_name,
+            off_value=off_value,
+            on_value=on_value,
+            entity_platform=entity_platform,
+            entity_type=entity_type,
+            initially_disabled=initially_disabled,
+            attribute_initialized_from_cache=attribute_initialized_from_cache,
+            translation_key=translation_key,
+            fallback_name=fallback_name,
+        )
```
TheJulianJES, line 287: Not sure we want to do this, but I think we could replace all default arguments (passed down to the normal `.switch()` method) with `*args, **kwargs`. Then, just pass those in.
It should contain all arguments that we did not explicitly define in our method definition.

This means we could introduce additional parameters in zigpy, without having to modify this code.

prairiesnpr: I have no strong preference on it, I had started with *args and **kwargs, but looking through the repo it looked like that wasn't the typical pattern, so I went the other way. Happy to change it, if that's the consensus.

TheJulianJES: Personally, I think I'd rather go with `**kwargs` for all arguments just passed to the underlying v2 builder methods.
I feel like it's possible that we're adding new arguments to these methods in zigpy and we'd always have to update quirks too then. If we do a similar thing to the Tuya builder with Aqara, then we'd have to update that too.

prairiesnpr: One issue with **kwargs, we would lose type hinting. Might be worth adding appropriate TypedDicts for each function in zigpy and then importing them here.

dmulcahey: So I think we need to change Zigpy to do this cleanly. We would have to add **kwargs to the method signatures and then here we would use **locals() to call the methods or do something like this hack…

```
import inspect

def foo(bar=baz, test=blah, **kwargs):
    # Gather all local variables (including bar, test, and kwargs)
    all_args = {**locals(), **kwargs}

    # Get the signature of the inner function (another_method)
    sig = inspect.signature(another_method)

    # Filter all_args to include only parameters that are explicitly defined in another_method
    filtered_args = {k: v for k, v in all_args.items() if k in sig.parameters}

    # Pass only filtered arguments to another method
    another_method(**filtered_args)

def another_method(bar, test):
    print(f"bar: {bar}, test: {test}")

# Example usage:
foo(bar=1, test=2, extra_arg="extra")
```

why don’t we leave it as is for now and we can address it later in a streamlining PR? (Unless I am misunderstanding or missing something)

TheJulianJES: Ok, let's leave it how it is at the moment. We can reconsider this in the future.

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+            "battery_percentage_remaining",
+            converter=lambda x: x * scale,
+        )
+        self.adds(power_cfg)
+        return self
+
+    def tuya_metering(
+        self,
+        dp_id: int,
+        metering_cfg: TuyaLocalCluster = TuyaValveWaterConsumed,
```
TheJulianJES, line 93: The `Metering` cluster is most often used for tracking energy consumption in devices/quirks.
Most Tuya devices seem to support the ZCL cluster for that at least (just with a manual/different multiplier/divisor), so they do not send it via Tuya datapoints.
I thought about renaming this to `tuya_water_metering`, but that doesn't make a lot of sense, as we can change the `metering_cfg`. So, this is just a note.. It might not be obvious that `tuya_metering` is used for tracking water consumption by default.

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+    def tuya_enum(
+        self,
+        dp_id: int,
+        attribute_name: str,
+        enum_class: type[Enum],
+        access: foundation.ZCLAttributeAccess = foundation.ZCLAttributeAccess.Read
+        | foundation.ZCLAttributeAccess.Write,
+        endpoint_id: int = 1,
+        entity_type: EntityType = EntityType.CONFIG,
+        entity_platform: EntityPlatform = EntityPlatform.SELECT,
```
TheJulianJES: Doesn't really matter, but let's swap these around (like in https://github.com/zigpy/zigpy/pull/1487), so `entity_platform` is always before `entity_type`.
```suggestion
        entity_platform: EntityPlatform = EntityPlatform.SELECT,
        entity_type: EntityType = EntityType.CONFIG,
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+            type=enum_class,
+            access=access,
+        )
+        self.enum(
+            attribute_name=attribute_name,
+            enum_class=enum_class,
+            cluster_id=TUYA_CLUSTER_ID,
+            endpoint_id=endpoint_id,
+            entity_type=entity_type,
+            entity_platform=entity_platform,
```
TheJulianJES: and here too:
```suggestion
            entity_platform=entity_platform,
            entity_type=entity_type,
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+        attribute_name: str,
+        access: foundation.ZCLAttributeAccess = foundation.ZCLAttributeAccess.Read
+        | foundation.ZCLAttributeAccess.Write,
+        endpoint_id: int = 1,
+        min_value: float | None = None,
+        max_value: float | None = None,
+        step: float | None = None,
+        unit: str | None = None,
+        mode: str | None = None,
+        multiplier: float | None = None,
```
TheJulianJES, line 343: https://github.com/zigpy/zigpy/pull/1487/ added `entity_type` for `number`. Let's also add it here:
```suggestion
        multiplier: float | None = None,
        entity_type: EntityType = EntityType.CONFIG,
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+            attribute_name=attribute_name,
+            cluster_id=TUYA_CLUSTER_ID,
+            endpoint_id=endpoint_id,
+            min_value=min_value,
+            max_value=max_value,
+            step=step,
+            unit=unit,
+            mode=mode,
+            multiplier=multiplier,
+            device_class=device_class,
```
TheJulianJES, line 372: and here:
```suggestion
            device_class=device_class,
            entity_type=entity_type,
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+
+        return self
+
+    def tuya_dp(
+        self,
+        dp_id: int,
+        ep_attribute: str,
+        attribute_name: str,
+        converter: Optional[Callable[[Any,], Any,]] = None,
+        dp_converter: Optional[Callable[[Any,], Any,]] = None,
```
TheJulianJES: I'm aware formatting is disabled for this, but I think we can remove the trailing commas inside the typing at the end here, so it looks slightly nicer.
```suggestion
        converter: Optional[Callable[[Any], Any]] = None,
        dp_converter: Optional[Callable[[Any], Any]] = None,
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+        self.tuya_data_point_handlers.update({dp_id: dp_handler})
+        return self
+
+    def tuya_dp_attribute(
+        self,
+        dp_id: int,
+        attribute_name: str,
+        ep_attribute: str = TuyaMCUCluster.ep_attribute,
+        converter: Optional[Callable[[Any,], Any,]] = None,
+        dp_converter: Optional[Callable[[Any,], Any,]] = None,
```
TheJulianJES: and here:
```suggestion
        converter: Optional[Callable[[Any], Any]] = None,
        dp_converter: Optional[Callable[[Any], Any]] = None,
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+            fallback_name=fallback_name,
+        )
+
+        return self
+
+    def tuya_binary_sensor(
+        self,
+        dp_id: int,
+        attribute_name: str,
+        endpoint_id: int = 1,
```
TheJulianJES, line 385: We also added `entity_type` to `binary_sensor` in https://github.com/zigpy/zigpy/pull/1487/. So, let's also add it here:
```suggestion
        endpoint_id: int = 1,
        entity_type: EntityType = EntityType.DIAGNOSTIC,
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+            attribute_name=attribute_name,
+            type=t.Bool,
+            access=foundation.ZCLAttributeAccess.Read
+            | foundation.ZCLAttributeAccess.Report,
+        )
+        self.binary_sensor(
+            attribute_name=attribute_name,
+            cluster_id=TUYA_CLUSTER_ID,
+            endpoint_id=endpoint_id,
+            device_class=device_class,
```
TheJulianJES, line 409: and here:
```suggestion
            device_class=device_class,
            entity_type=entity_type,
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+
+        attr_dict: dict[int, foundation.ZCLAttributeDef] = self.new_attributes
+
+        class TuyaReplacementCluster(TuyaMCUCluster):
+            """Replacement Tuya Cluster."""
+
+            data_point_handlers: dict[int, str]
+            dp_to_attribute: dict[int, DPToAttributeMapping]
+            attributes = TuyaMCUCluster.attributes.copy()
+            attributes.update(attr_dict)
```
TheJulianJES: Assuming the `TuyaMCUCluster` and underlying clusters finally get updated to use `AttributeDefs`, this won't work anymore.
As far as I can see, `TuyaMCUCluster` (and the underlying clusters) do not actually have any attributes defined, only server and client commands that need to be updated.

Ideally, we'd already use the new zigpy-style `AttributeDefs` here. It might not be as clean though..

Just for reference, here's the "programmatic creation and populating" of `AttributeDefs` in zigpy:
https://github.com/zigpy/zigpy/blob/dfeb958b1f737b0a6c9dedbdb7fc169a188a516e/zigpy/zcl/__init__.py#L135-L143
(and the whole conversion from old-style to new-style attributes definition:)
https://github.com/zigpy/zigpy/blob/dfeb958b1f737b0a6c9dedbdb7fc169a188a516e/zigpy/zcl/__init__.py#L95-L211

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+
+            async def write_attributes(self, attributes, manufacturer=None):
+                """Overwrite to force manufacturer code."""
+
+                return await super().write_attributes(
+                    attributes, manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID
+                )
```
TheJulianJES, line 492: If we always want this, shouldn't this be in `TuyaMCUCluster` or another subclass of it? (and not nested in `add_to_registry` of the Tuya quirk builder?)

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+
+        TuyaReplacementCluster.data_point_handlers = (
+            TuyaMCUCluster.data_point_handlers.copy()
+        )
+        TuyaReplacementCluster.data_point_handlers.update(self.tuya_data_point_handlers)
```
TheJulianJES: All quirks seem to just redefine `data_point_handlers`. I'd also do it here. There's probably not really a point to always copying an empty dictionary.

zhaquirks/tuya/builder/__init__.py
TheJulianJES: If you don't want to change this at the moment, we could also leave it like this for a first version, but we generally don't want to add "old-style zigpy attributes" at this point, though this is a somewhat different case compared to most other quirks, so it's fine for now IMO.

tests/async_mock.py
```diff
@@ -0,0 +1,43 @@
+"""Mock utilities that are async aware."""
+
+from unittest.mock import *  # noqa: F401, F403
+
+
+class _IntSentinelObject(int):
```
TheJulianJES, line 6: Hmm, I don't really like that we need to port this over from zigpy, but it's fine..
We might also use the `device_mock` fixture in other tests. We can move that part then.

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+    """Setting unit of measurement."""
+    _CONSTANT_ATTRIBUTES = {
+        Metering.AttributeDefs.unit_of_measure.id: VOLUME_LITERS,
+        Metering.AttributeDefs.metering_device_type.id: WATER_METERING,
+    }
+
+    def __init__(self, *args, **kwargs):
+        """Init a TuyaValveWaterConsumed cluster."""
+        super().__init__(*args, **kwargs)
+        self.add_unsupported_attribute(Metering.AttributeDefs.instantaneous_demand.id)
```
TheJulianJES, line 59: The `LocalDataCluster` should return this as unsupported by default, right? (and thus prevent entity creation for that)
I mean, it doesn't hurt to explicitly add it as unsupported again, but is there a reason I'm missing?

TheJulianJES (commented):
I think this is almost good to go. Nice work 😄

zhaquirks/tuya/builder/__init__.py
prairiesnpr: The issues here is, I wasn't certain we always wanted to clear the manufacturer id. I was concerned about changing it without being certain that was the case. If we are certain, then yes, it should be in `TuyaMCUCluster` I just wasn't and only had a couple devices to test with in the first place.

zhaquirks/tuya/builder/__init__.py
prairiesnpr: Maybe this is a bug upstream? If you don't add it to unsupported attributes, HA will create the entity.

TheJulianJES: Hmm, I'll try and check on this in the future. The attribute is in the `REPORT_CONFIG` for the smart energy cluster, so it should be read upon pairing: https://github.com/zigpy/zha/blob/3d34a30499a80ad1802fbc72d36ad5cd5dc89ac0/zha/zigbee/cluster_handlers/smartenergy.py#L80-L84
When the attribute is read from the `LocalDataCluster`, it *should*  return "unsupported attribute", as there's no value in the attribute cache for it.
Let's leave this in then.

zhaquirks/tuya/builder/__init__.py
TheJulianJES: Ok, we can consider adding a builder method in the future if some devices explicitly do not want it.

zhaquirks/tuya/__init__.py
```diff
@@ -1507,6 +1507,9 @@ class TuyaNewManufCluster(CustomCluster):
     cluster_id: t.uint16_t = TUYA_CLUSTER_ID
     ep_attribute: str = "tuya_manufacturer"
 
+    class AttributeDefs(CustomCluster.AttributeDefs):
```
TheJulianJES: Ok, I guess `CustomCluster.AttributeDefs` works.
Just to stay consistent with other quirks, I'd also use `BaseAttributeDefs` here.
Yeah, it's basically the same, but.. 😄 

```suggestion
    class AttributeDefs(BaseAttributeDefs):
```
(might also need to import `BaseAttributeDefs`)

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,513 @@
...
+        """Attribute Definitions."""
+
+    def __init__(
+        self, manufacturer: str, model: str, registry: DeviceRegistry = _DEVICE_REGISTRY
+    ) -> None:
+        """Init the TuyaQuirkBuilder."""
+        self.tuya_data_point_handlers: dict[int, str] = {}
+        self.tuya_dp_to_attribute: dict[int, DPToAttributeMapping] = {}
+        # self.new_attributes: dict[int, foundation.ZCLAttributeDef] = {}
+        # self.new_attributes: set[foundation.ZCLAttributeDef] = set()
```
TheJulianJES: If the `AttributeDefs` stuff works now, can we remove the commented out code?
```suggestion
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,513 @@
...
+        #         #attr_id:
+        #         foundation.ZCLAttributeDef(
+        #             id=attr_id,
+        #             type=type,
+        #             access=access,
+        #             is_manufacturer_specific=is_manufacturer_specific,
+        #             name=attribute_name,
+        #         )
+        #     #}
+        # )
```
TheJulianJES: Can we also remove this (if the `AttributeDefs` stuff works)?

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,513 @@
...
+
+        new_attr = foundation.ZCLAttributeDef(
+            id=attr_id,
+            type=type,
+            access=access,
+            is_manufacturer_specific=is_manufacturer_specific,
+            name=attribute_name,
+        )
+
+        setattr(self.AttributeDefs, new_attr.name, new_attr)
```
TheJulianJES: Adding these to `AttributeDefs` here is nice. Didn't think of that 😄

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -161,13 +161,15 @@ class TuyaConnectionStatus(t.Struct):
...
-    )
+    class AttributeDefs(TuyaNewManufCluster.AttributeDefs):
+        """Attribute Definitions."""
+
+        mcu_version = foundation.ZCLAttributeDef(
+            id=ATTR_MCU_VERSION,
+            type=t.uint48_t,
+            access=foundation.ZCLAttributeAccess.Read,
+            is_manufacturer_specific=True,
+        )
```
TheJulianJES: I completely missed that there were some `attributes` in the `TuyaMCUCluster` before.
Can we move this up to the top of the cluster class, like it is in other custom clusters?

zhaquirks/tuya/builder/__init__.py
```diff
@@ -0,0 +1,513 @@
...
+            setattr(NewAttributeDefs, attr.name, attr)
+
+        class TuyaReplacementCluster(TuyaMCUCluster):
+            """Replacement Tuya Cluster."""
+
+            data_point_handlers: dict[int, str]
+            dp_to_attribute: dict[int, DPToAttributeMapping]
+
+            class AttributeDefs(NewAttributeDefs):
+                """Attribute Definitions."""
```
TheJulianJES: Is the `AttributeDefs` stuff done like this, so we don't accidentally modify the "global class"?
If so, couldn't we create an `AttributeDefs` ~~instance~~ class in the ``__init__`` of the quirk builder and just set the attributes there? Or am I missing something?

dmulcahey: AttributeDefs are just classes… there are no instances

TheJulianJES: We're "dynamically modifying" `AttributeDefs` in the `tuya_attribute` method. I was trying to say/ask if this affects "all `AttributeDefs`" used in quirks created with the builder (which it should not) or if there's a specific `AttributeDefs` class (which is what I meant with "instance" above) assigned to every quirk built with the builder (which should be the case).

TheJulianJES: I was thinking if we need something similar to where `types.new_class` is used here in zigpy, as we don't create an `AttributeDefs` class with attributes that are the same for all quirks created with the Tuya builder: https://github.com/zigpy/zigpy/blob/dfeb958b1f737b0a6c9dedbdb7fc169a188a516e/zigpy/zcl/__init__.py#L136-L143

dmulcahey: > I was thinking if we need something similar to where `types.new_class` is used here in zigpy, as we don't create an `AttributeDefs` class with attributes that are the same for all quirks created with the Tuya builder: https://github.com/zigpy/zigpy/blob/dfeb958b1f737b0a6c9dedbdb7fc169a188a516e/zigpy/zcl/__init__.py#L136-L143

I think this is the correct way to do it

dmulcahey: That said… these really are for making code clearer / easier to write…. Do we need the AttributeDefs here or is the dict syntax sufficient?

prairiesnpr: I was in a hurry and missed a commit, this should be resolved now. While this seems odd, it's the only way I found that actually worked. Setting the attribute on `TuyaReplacementCluster` directly or replacing `AttributeDefs` similar to zigpy results in the attributes being dropped when passed to `replace`.

zhaquirks/tuya/builder/__init__.py
TheJulianJES: I think this still needs to be done.

zhaquirks/tuya/builder/__init__.py
TheJulianJES: I think this still needs to be done.

zhaquirks/tuya/builder/__init__.py
TheJulianJES: I think this still needs to be done.

zhaquirks/tuya/builder/__init__.py
TheJulianJES: I think this still needs to be done.

zhaquirks/tuya/builder/__init__.py
TheJulianJES: > Do we need the AttributeDefs here or is the dict syntax sufficient?

(assuming "dict syntax" is just talking about the old-style zigpy `attributes` definition with a dict)
I was under the impression we no longer wanted to include them if possible. Although it's unlikely zigpy will drop them in the near future.. 😄 

If it's easily possible, I'd prefer to use the new-style `AttributeDefs`. We can also fall back to the old-style `attributes` otherwise (at least with the new `ZCLAttributeDef` objects).
I guess for a case like this, using `AttributeDefs` is more difficult than just using the dict style.

---

If we want to use new-style `AttributeDefs`, mind trying to see if something like this works? (replace the whole `add_to_registry` registry method with below).
Not saying this is the best solution, just curious if this works..

```python
    def add_to_registry(self) -> QuirksV2RegistryEntry:
        """Build the quirks v2 registry entry."""

        class TuyaReplacementCluster(TuyaMCUCluster):
            """Replacement Tuya Cluster."""

            data_point_handlers: dict[int, str]
            dp_to_attribute: dict[int, DPToAttributeMapping]

            def __init_subclass__(cls) -> None:
                cls.AttributeDefs = types.new_class(
                    name="AttributeDefs",
                    bases=(TuyaMCUCluster.AttributeDefs,),
                )
    
                for attr in self.new_attributes:
                    setattr(cls.AttributeDefs, attr.name, attr)

                super().__init_subclass__()

                """Overwrite to force manufacturer code."""
                return await super().write_attributes(
                    attributes, manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID
                )
```

TheJulianJES: Ok, so, apparently that doesn't seem to work. The current way in this PR works at least.

I've thought about it again and it's preferable to use `AttributeDefs` if possible.
If the base clusters (`TuyaMCUCluster` and so on) use `AttributeDefs`, we also have to. If they use the `attributes` dict, we also have to. 
Since `AttributeDefs` is the preferred way to go, we should use that here and convert the other clusters to `AttributeDefs`.

(EDIT: Actually, it looks like zigpy's conversion might accept both `AttributeDefs` and `attributes` dict in the same cluster at the same time?)

I'll try and take another stab at maybe slightly changing/improving what we have now (regarding `AttributeDefs`), but the current implementation works already..

TheJulianJES (approved):
Looks good 😄

PR #3425, opened 2024-10-15
Convert Tuya Sensors to TuyaQuirkBuilder format

prairiesnpr:
## Proposed change

I have a single soil sensor `_TZE284_aao3yzhs`, I was able to test physically with.

## Checklist


zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -1,487 +1,67 @@
...
-                    Scenes.cluster_id,
-                    TemperatureHumidityManufCluster.cluster_id,
-                ],
-                OUTPUT_CLUSTERS: [Ota.cluster_id, Time.cluster_id],
-            }
-        },
-    }
-
-    replacement = {
-        SKIP_CONFIGURATION: True,
```
TheJulianJES, line 129: Do we want to include `.skip_configuration()`?
By default, ZHA binds some default clusters like `PowerConfiguration` and `Scenes`(?) and sets up reporting (if they have a `REPORT_CONFIG`.
Most of our replacement clusters are `LocalDataCluster`s anyway, so it doesn't matter for them, as they ignore the calls.
By including `.skip_configuration()`, we'd make sure that ZHA doesn't even try to bind any clusters. Do we want that? The v1 quirks seem to have it.

TheJulianJES (approved):
Looks good

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -1,487 +1,67 @@
...
-        },
-    }
+(
+    TuyaQuirkBuilder("_TZE200_myd45weu", "TS0601")
+    .also_applies_to("_TZE200_ga1maeof", "TS0601")
+    .also_applies_to("_TZE200_9cqcpkgb", "TS0601")
+    .also_applies_to("_TZE204_myd45weu", "TS0601")
+    .tuya_temperature(dp_id=5)
+    .tuya_battery(dp_id=15)
+    .tuya_soil_moisture(dp_id=3)
```
TheJulianJES, line 69: Actually, is the scale correct on these?

prairiesnpr: Temp and soil both default to a scale of 100, which matches the original quirk. Do we want to explicitly set the scale to avoid confusion?

zhaquirks/tuya/ts0601_sensor.py
prairiesnpr: I'll add it, I never saw anything in the logs, so I don't think it matters, but makes sense to be clear.

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -1,487 +1,72 @@
...
-    PROFILE_ID,
-    SKIP_CONFIGURATION,
+(
+    TuyaQuirkBuilder("_TZE200_a8sdabtg", "TS0601")  # Variant without screen, round
+    .also_applies_to("_TZE200_qoy0ekbd", "TS0601")
+    .also_applies_to("_TZE200_znbl8dj5", "TS0601")
+    .also_applies_to("_TZE200_qyflbnbj", "TS0601")
+    .also_applies_to("_TZE200_zppcgbdj", "TS0601")
+    .tuya_temperature(dp_id=1)
+    .tuya_humidity(dp_id=2)
```
TheJulianJES: Unless I'm missing something, these were scaled by 10 before, right?

zhaquirks/tuya/ts0601_sensor.py
TheJulianJES: Ah, right. I got a bit confused looking at everything.
> Do we want to explicitly set the scale to avoid confusion?

Hmm, let's try not to for now (most seem to use `100`). If it causes more issues in the future, we can do that though.

zhaquirks/tuya/ts0601_sensor.py
prairiesnpr: Ok, going back through the old quirk, the two classes that were scaled by 10 were, `SoilManufClusterVar02` and `TuyaTempHumiditySensor` that corresponds to `_TZE284_aao3yzhs`, `_TZE284_sgabhwa6`, `_TZE200_bjawzodf`, and `_TZE200_zl1kmjqx`. See line 6 and 52. These all should be scaled to 100, so they are good.

TheJulianJES: I was also looking at the scaling in the cluster itself:
image

I'll double check again 😄

prairiesnpr: Yep, missed that. Checking now

prairiesnpr: Ok, that should capture `TemperatureHumidityManuCluster`, looks like I split them out then didn't add the scale, great catch, thanks.

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -1,487 +1,72 @@
...
-    }
+(
+    TuyaQuirkBuilder("_TZE200_yjjdcqsq", "TS0601")
+    .applies_to("_TZE200_9yapgbuv", "TS0601")
+    .applies_to("_TZE204_yjjdcqsq", "TS0601")
+    .applies_to("_TZE200_utkemkbs", "TS0601")
+    .applies_to("_TZE204_utkemkbs", "TS0601")
+    .applies_to("_TZE204_yjjdcqsq", "TS0601")
+    .applies_to("_TZE204_ksz749x8", "TS0601")
+    .tuya_temperature(dp_id=1)
```
TheJulianJES: Mind checking this again before/after? Just the temp scale. It seems like it was `10` before, but it would now be `100`?

TheJulianJES (commented):
Ok, all other scales seem fine (just looking through them).
Only for `_TZE200_yjjdcqsq` (and the others there) might `scale=10` be needed for temperature.

zhaquirks/tuya/ts0601_sensor.py
prairiesnpr: Agreed, `TemperatureHumidityManufCluster.dp_to_attribute[1],` threw me off here, fixed.

TheJulianJES (approved):

PR #3432, opened 2024-10-15
Add Philips Hue contact sensor quirk

### LINKED ISSUE #3314: [Device Support Request] SOC001 - Philips Contact sensor
### Problem description

Signify released a while a go a contact sensor to extend their security product portfolio.
Unfortunately they almost exclusively putted all relevant attributes in their manufacturer cluster.
Good news is that the device seems to overall have a good behaviour continuing the reputation of Signify.

### Solution description

1. Implement the manufacturer specific cluster :
- PhilipsContactCluster : 0xFC06

2. Implements the associated attributes :
- Contact
Attribute: 0x0100
Data Type: 8-Bit Enumeration (0x30)
Uint8: 1 (0x01)

- Last contact change (10th of a second)
Attribute: 0x0101
Data Type: 32-Bit Unsigned Integer (0x23)
Uint32: 0 (0x00000000)

- Tamper
Attribute: 0x0102
Data Type: 8-Bit Enumeration (0x30)
Uint8: 0 (0x00)

- Last tamper change (10th of a second)
Attribute: 0x0103
Data Type: 32-Bit Unsigned Integer (0x23)
Uint32: 12425 (0x00003089)

3. Handle the on_off cluster as it is stateless and only meant for biding with lights.

### Screenshots/Video

</details>

### Device signature

```json
{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4107,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 128,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 128,
  },
  "endpoints": {
    "2": {
      "profile_id": "0x0104",
      "device_type": "0x0402",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0xfc06"
      ],
      "output_clusters": [
        "0x0000",
        "0x0003",
        "0x0006",
        "0x0019"
      ]
    }
  },
  "manufacturer": "Signify Netherlands B.V.",
  "model": "SOC001",
  "class": "soc001.SignifyContact"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
None
```

</details>

### Logs

```python
None
```

</details>

### Custom quirk

```python
"""Signify SOC001 device."""

from zigpy.quirks.v2 import QuirkBuilder,ClusterType,BinarySensorDeviceClass
import zigpy.types as types

class PhilipsManufacturerSpecific(CustomCluster):
    """Philips Contact manufacturer cluster."""

    cluster_id: types.uint16_t = 0xFC06
    name = "Philips Manufacturer specific"
    ep_attribute = "philips_manufacturer_specific"
    attributes = {
        0x0100: ("contact", types.enum8, True),
        0x0101: ("last_contact_change", types.uint32_t, True),
        0x0102: ("tamper", types.enum8, True),
        0x0103: ("last_tamper_change", types.uint32_t, True),
    }

(
    QuirkBuilder("Signify Netherlands B.V.", "SOC001")
    .removes(
        cluster_id=6,
        endpoint_id=2,
        cluster_type=ClusterType.Client
    )
    .replaces(
        replacement_cluster_class=PhilipsManufacturerSpecific,
        cluster_id=64518,
        endpoint_id=2
    )
    .binary_sensor(
        attribute_name="contact",
        cluster_id=64518, #0xfc06,
        endpoint_id=2,
        device_class=BinarySensorDeviceClass.OPENING
    )
    .binary_sensor(
        attribute_name="tamper",
        cluster_id=64518, #0xfc06,
        endpoint_id=2,
        device_class=BinarySensorDeviceClass.TAMPER
    )
    .add_to_registry()
)
```

</details>

### Additional information

Unfortunately the device does not respond it's attributes whatsoever, probably to make our life a little bit harder.
To discover the attributes, I iterated all attribute addresses and read them to get the following data :

-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
2024-08-19T04 | 37 | 20.475616+00 | 0 | Manufacturer Specific | 256 | enum8.undefined_0x00 | 0x0100 | 0xFC06 | 2 | 0 | 17 | 88 | 1 | 0d | ce | f9 | 51 | 0x100B | 0x30
2024-08-19T04 | 37 | 25.483887+00 | 0 | Manufacturer Specific | 257 | 47103 | 0x0101 | 0xFC06 | 2 | 0 | 17 | 88 | 1 | 0d | ce | f9 | 51 | 0x100B | 0x23
2024-08-19T04 | 37 | 30.487990+00 | 0 | Manufacturer Specific | 258 | enum8.undefined_0x00 | 0x0102 | 0xFC06 | 2 | 0 | 17 | 88 | 1 | 0d | ce | f9 | 51 | 0x100B | 0x30
2024-08-19T04 | 37 | 35.495690+00 | 0 | Manufacturer Specific | 259 | 172317 | 0x0103 | 0xFC06 | 2 | 0 | 17 | 88 | 1 | 0d | ce | f9 | 51 | 0x100B | 0x23

But realised afterward that ebaauw did all the work here : https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7226

mguaylam:
## Proposed change

This adds the custom Philips Hue contact cluster with all its custom attributes.
A v2 quirk entity for the binary temper sensor is also added.

## Additional information

Needed by this PR: https://github.com/zigpy/zha/pull/238
Fixes: #3314 

## Checklist


zhaquirks/philips/soc001.py
```diff
@@ -0,0 +1,45 @@
...
+
+    cluster_id = 64518  # 0xfc06
+    name = "Philips contact cluster"
+    ep_attribute = "philips_contact_cluster"
+    attributes = {
+        0x0100: ("contact", types.enum8, True),
+        0x0101: ("last_contact_change", types.uint32_t, True),
+        0x0102: ("tamper", types.enum8, True),
+        0x0103: ("last_tamper_change", types.uint32_t, True),
+    }
```
TheJulianJES: Can we use the new-style zigpy `AttributeDefs` here?

```python
        """Attribute definitions."""

        contact = ZCLAttributeDef(
            id=0x0100,
            type=types.enum8,
        )
        # ... (and so on) 
```

(There are some examples if you search for `AttributeDefs` in this or the zigpy repo)

zhaquirks/philips/soc001.py
```diff
@@ -0,0 +1,45 @@
...
+
+(
+    #  <SimpleDescriptor endpoint=2 profile=260 device_type=1026
+    #  device_version=0
+    #  input_clusters=[0, 1, 3, 64518]
+    #  output_clusters=[0, 3, 6, 25]>
+    QuirkBuilder("Signify Netherlands B.V.", "SOC001")
+    .removes(cluster_id=6, endpoint_id=2, cluster_type=ClusterType.Client)
+    .replaces(
+        replacement_cluster_class=PhilipsContactCluster, cluster_id=64518, endpoint_id=2
```
TheJulianJES: We do not need to specify `cluster_id` here, as it's the same as in the custom `PhilipsContactCluster`.
(It removes the cluster with the `cluster_id` specified here and then adds the new custom cluster with the `cluster_id` specified in it. Since it's the same for both, we can remove the `cluster_id` here.)

We can probably also drop the first argument being a keyword one. So something like this should be enough:
```python
    .replaces(PhilipsContactCluster, endpoint_id=2)
```

zhaquirks/philips/soc001.py
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+
+(
+    #  <SimpleDescriptor endpoint=2 profile=260 device_type=1026
+    #  device_version=0
+    #  input_clusters=[0, 1, 3, 64518]
+    #  output_clusters=[0, 3, 6, 25]>
+    QuirkBuilder("Signify Netherlands B.V.", "SOC001")
+    .removes(cluster_id=6, endpoint_id=2, cluster_type=ClusterType.Client)
```
TheJulianJES: Just noting:
Since the `OnOff` cluster is removed here, this would be a breaking change (and also makes it impossible to configure binding via the UI).

zhaquirks/philips/soc001.py
```diff
@@ -0,0 +1,45 @@
...
+        cluster_id=64518,  # 0xfc06
+        endpoint_id=2,
+        device_class=BinarySensorDeviceClass.OPENING,
+    )
+    .binary_sensor(
+        attribute_name="tamper",
+        cluster_id=64518,  # 0xfc06
+        endpoint_id=2,
+        device_class=BinarySensorDeviceClass.TAMPER,
+    )
```
TheJulianJES: We may also want to include `fallback_name`. It will be supported in future HA versions. If the translation key isn't in HA yet (probably not the case until a ZHA bump with the new quirk is merged to HA), it will fall back to that name and ignore the translation key.

And I'm not sure what the current state on this is:
Currently, `translation_key` defaults to `attribute_name` if not explicitly provided, but it might be required in future versions(?) Again, not sure what we set on. Both might still work.

TheJulianJES: Also, we could also drop `attribute_name` and `cluster_id` here, since they're always the first two arguments:
```python
    .binary_sensor(
        "tamper",
        PhilipsContactCluster.cluster_id,
        endpoint_id=2,
        device_class=BinarySensorDeviceClass.TAMPER,
    )
```

zhaquirks/philips/soc001.py
mguaylam: Done here : https://github.com/zigpy/zha-device-handlers/pull/3432/commits/a12a3e08b45e5c7123965f3ea0d3ece9fd98ee4d

zhaquirks/philips/soc001.py
mguaylam: Changed in : https://github.com/zigpy/zha-device-handlers/commit/5cd4857d73077c6bb0cb4ac1009533690db10919

zhaquirks/philips/soc001.py
mguaylam: Changed in : https://github.com/zigpy/zha-device-handlers/commit/e101ca095ae4a7d6e21f34bd49d587de7c575ea2

zhaquirks/philips/soc001.py
mguaylam: Changed in : https://github.com/zigpy/zha-device-handlers/commit/475529201f9e41abce020b10722a3df2549b9d7c

mguaylam:
As discussed here and on the HA Discord, I am advocating in handling the functions of the sensor entirely based on the stateful manufacturer specific cluster as it was intended by Signify.

It is a breaking change.

### What the device-handler does

- Implement the custom cluster with the contact and tamper attributes. I did not implement their timers as it has no real world usage to HA. Could be considered in the future.
- Implement the On/Off cluster in a custom cluster to make it still bindable.

### Reasons supporting this
- It's what the manufacturer intended. The On/Off cluster even if usable, is only meant for binding.
- Any state change during a network shutdown or failure will fail to be reported. Being a security device this feature is important. I have lived and still live with not knowing if my doors are closed or opened after a power outage and often then not, forced to cycle the state of the sensor to get it back to a known state (Xiaomi example here).
- Failure of the network to deliver the message may cause an out of sync state of the attribute (I believe, if it is not corrected by the maximum report period).

### Other reasons supporting my decision
- I tried several contact sensor (IKEA, Xiaomi, Konke and Sonoff) and got tired of their poor reliability : broken reporting, fast battery drain, out of sync attribute, dropping off the network.. My most reliable devices are Philips Hue's so I decided to spend the money to purchase this device and until now, it does deliver on the promise.
- dresden-elektronik did the integration similar as me and came to a similar conclusion : https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7226. I know ZigBee2MQTT did not implement this cluster at all https://github.com/Koenkk/zigbee2mqtt/issues/19054 but I think it was due to just not knowing what it incorporate as the device does not reply with a list of attributes.
- I still think it's the right time to implement this breaking change as the device is brand new on the market. Only purchasable in Canada for 2 months now. Combined to the very high price, I don't think many people have it in their hands and I'm the first one to ask and propose a device handler to enable the tamper feature demonstrating the still low usage of the device.

**I hope this proposal will convince to consider this pull request.**

### Other notes
- This device-handler v2 depends on https://github.com/zigpy/zigpy/pull/1487
- [https://github.com/zigpy/zha/pull/238](https://github.com/zigpy/zha/pull/238) depends on this pull request.

zhaquirks/philips/soc001.py
```diff
@@ -0,0 +1,93 @@
...
+    ep_attribute = "on_off_cluster"
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        on_off = ZCLAttributeDef(
+            id=0x0000,
+            type=types.Bool,
+            is_manufacturer_specific=False,
+        )
```
TheJulianJES: I don't think we should/can do this. We're replacing a ZCL cluster and are also changing the `ep_attribute` here.
IIRC there are places where ZHA expects the `ep_attribute` to be unchanged for a standard ZCL cluster.
Not subclassing `OnOff` (and its `AttributeDefs`) is also not good.

TheJulianJES:
Like I mentioned [on Discord](https://discord.com/channels/330944238910963714/1292736317569896591/1295901260985143329), I'd remove the custom `OnOff` cluster and custom "contact" binary sensor for now. The attribute for it can stay.
Then, we could merge this PR already and also the [ZHA PR](https://github.com/zigpy/zha/pull/238) afterwards that binds the custom Philips contact cluster.

We can still consider on how we want to deal with the "contact" binary sensor then. It's not trivial, as breaking changes are always bad (especially for a security device where the behavior might go unnoticed in automations until it's too late).
We need to find a good solution here, because we could theoretically avoid a breaking change. It's not easy though.

The parts where the change would be breaking:
- replacing the existing contact binary sensor
- re-pairing or reconfiguration of the sensor is required to bind the new Philips contact cluster to the coordinator

TheJulianJES:
Also, to re-iterate some of the points I mentioned on Discord:

> It's what the manufacturer intended. 

Tuya intends that we shouldn't work with its devices all.. 😄 
 
> The On/Off cluster even if usable, is only meant for binding.

Well, we're doing exactly that. ZHA requests the `OnOff` cluster to be bound to the coordinator. We're receiving the "on"/"off" commands then and set the entity state in HA based on that.

> Any state change during a network shutdown or failure will fail to be reported. 

Even for a binary sensor based on the custom attribute, it wouldn't fix this. If ZHA is restarted after a power outage, we only poll some attributes of mains-powered devices (based on what's defined in the ZHA cluster handler).
You'd need to run `homeassistant.update_entity` for the new binary sensor entity to (hopefully) poll new state once (after ZHA has started and the network has settled down).

> Failure of the network to deliver the message may cause an out of sync state of the attribute

But the same behavior would happen for the binary sensor based on the custom attribute. ZHA doesn't poll entities by default, especially not ones on battery-powered devices.
You'd need to regularly run `homeassistant.update_entity` to poll the device then.

---

The only benefit that I'm seeing for such a breaking charge (or large change if we make it not breaking) is that you can poll the entity state (assuming the device is always somewhat awake, but that does seem to be the case according to what you mentioned).
The majority of people isn't going to use `homeassistant.update_entity` to poll the entity state regularly.

So just to repeat myself, I'd change the scope of this PR to only include the tamper entity for now (and the custom contact cluster with both custom attributes defined). We can also merge the ZHA PR then. Everyone who installs/pairs a Hue sensor after a release with those changes will at least have the custom Hue cluster bound (and a tamper entity), so if we make the binary sensor switch in the future (whatever that may look like), they won't have to reconfigure or re-pair the device at least.

TheJulianJES:
Another thought I'm not sure works (ideally for a future PR):
We might be able to just catch attributes reads on the output `OnOff` cluster, redirect them to custom contact cluster, then redirect the result to the `on_off` attribute.
It's not really a clean way, but it would keep using the ZHA entity for the `OnOff` cluster, whilst allowing to poll the sensor state.

This won't solve the issue of needing to re-pair the sensor though, unless we still listen to the actual `OnOff` on/off commands and only use the custom contact cluster when the `OnOff` `on_off` attribute is trying to be read/polled.

mguaylam:
I just want to say first, thank you so much for following up. I recognise the quality of your free time spent on this project and the effort you took to assist me.

For this reason I was about to submit as you proposed because I don't want to drag you forever for something with mild consequences.

But I wanted to make sure the consequences we're as you described so it can be at worst, left like this.

I did 2 tests :

1. No quirk, left as is using the on_off attribute : I produce a network failure and change the contact state while the network is offline.
Result : After an hour, there never was a report back of the open state, for Home Assistant it would remain closed based on the cached attribute state forever until the sensor is interacted with.

2. With the quirk, using the custom cluster and attribute reporting configuration: I produce a network failure and change the contact state while the network is offline.
Result : After 15 minutes since the last tried report, it produce a new one updating the attribute and now display the correct state by itself in Home Assistant.

This also apply to a transmission failure, I've seen the sensor fail 1 or 2 times to transmit on the network but tried again right after.

And the reason is simple enough :
There is no reporting on the on_off attribute, only a binding on the cluster :
(I did the report without the quirk but forgot to write it down, the result was the same)
```YAML
result_conf:
  - cluster: On/Off
    cluster_id: "0x0006"
    ep: 2
    attr_id: "0x0000"
    direction: 0
    status: 134 #(not supported)
    attr: on_off
```

Where on the custom cluster :
```YAML
result_conf:
  - cluster: Philips contact cluster
    cluster_id: "0xFC06"
    ep: 2
    attr_id: "0x0100"
    direction: 0
    status: 0
    type: "0x30"
    min_interval:
      - 0
    max_interval:
      - 900
    reportable_change:
      - null
    attr: contact
```

The on_off cluster in this case is a client type and does not seem to support reporting configuration but will report a change on the bind device. I don't know if it's per specification but seems logical to not report a state multiple times if it's to command a light as you might have turned off the light with a remote meanwhile.

Also it's a light sleeper, it does respond to commands within the timeout of the network.

You're the maintainer and I want to respect that. If the breaking change is unacceptable I'll submit without it.
But I want to convince you one last time as I really believe having a device with proper reporting is more important and would make the device more reliable. A change of state during ZHA being offline in the current version will never be known to ZHA.

TheJulianJES:
> With the quirk, using the custom cluster and attribute reporting configuration: I produce a network failure and change the contact state while the network is offline.
Result : After 15 minutes since the last tried report, it produce a new one updating the attribute and now display the correct state by itself in Home Assistant.

Ok, this is good to know.

> But I want to convince you one last time as I really believe having a device with proper reporting is more important and would make the device more reliable.

Yeah. Using the `OnOff` cluster like done in this PR to block the entity from generating isn't ideal. I have an idea on how to make this potentially non-breaking, but if it needs to be a breaking change, we should consider how we want to block the `OnOff` entity from being created: directly in ZHA? allow quirks to properly disable individual entity generation? just use a `LocalDataCluster` and make the `on_off` attribute unsupported, so ZHA doesn't create? I believe that would make binds non-functional though.. But marking just the `on_off` attribute as unsupported *should* be fine.
... or have both entities (temporarily)?

Just redirecting attribute reports from the custom Hue cluster to the OnOff cluster would get those 15 minute interval reports to come through. Depending on the implementation, it may also allow for manual polling.
This is something I want to explore soon-ish.

Anyways, like mentioned before: We can basically instantly merge this PR for the tamper entity if the contact entity and custom `OnOff` cluster is taken out for now. At least that *should* land in the next major HA release and so people will already have their Hue contact cluster bound when pairing such a device (since we can also merge the ZHA PR then).
If you want, you can then open another PR after this one is merged that adds the custom contact sensor entity back.
There will be some further discussion and we can see how to best make that change. That very likely won't be in the next major HA release though.

mguaylam:
> Using the OnOff cluster like done in this PR to block the entity from generating isn't ideal.

Absolutely. It was a crude way to get to the result. There's surely a proper way to do it.

> I have an idea on how to make this potentially non-breaking, but if it needs to be a breaking change, we should consider how we want to block the OnOff entity from being created: directly in ZHA? allow quirks to properly disable individual entity generation? just use a LocalDataCluster and make the on_off attribute unsupported, so ZHA doesn't create? I believe that would make binds non-functional though.. But marking just the on_off attribute as unsupported should be fine.
> ... or have both entities (temporarily)?

I agree to look at that in another PR so we at least start to bind and report as soon as possible.

> That very likely won't be in the next major HA release though.

I'm ok with that. As long as we have a reliable sensor at the end.

Let's get that merge and I could learn how to do tests at some point.

zhaquirks/philips/soc001.py
```diff
@@ -0,0 +1,57 @@
...
+    #  output_clusters=[0, 3, 6, 25]>
+    QuirkBuilder("Signify Netherlands B.V.", "SOC001")
+    .replaces(PhilipsContactCluster, endpoint_id=2)
+    .binary_sensor(
+        "tamper",
+        PhilipsContactCluster.cluster_id,
+        endpoint_id=2,
+        device_class=BinarySensorDeviceClass.TAMPER,
+        fallback_name="Tamper",
+        entity_type=EntityType.DIAGNOSTIC,
```
TheJulianJES: This was fine before, but I'll just explicitly add the `translation_key` here. It'll be the `attribute_name` by default, but we've explicitly set it again in all recent Tuya quirks v2 PRs. (It might also explicitly be required in the future?)
```suggestion
        entity_type=EntityType.DIAGNOSTIC,
        translation_key="tamper",
        fallback_name="Tamper",
```

TheJulianJES (approved):
This should be good. Thanks!

TheJulianJES:
> Let's get that merge and I could learn how to do tests at some point.

We can't really do a lot in terms of unit tests for this quirk. Entity creation only happens in ZHA/HA, so there's no real code to test here.

PR #3438, opened 2024-10-17
Send all button press events for Sinope SW2500ZB/DM2500ZB

bassdr:
## Proposed change

Added support for both press and release button events in order to properly support short and long press events.
Also added the pressed button and text description in the event arguments to simplify writing automations processing it.

ZHA now send both when we start pressing the Switch's up or down button, and when we release. Before, we were only sending events when a button was released after being pressed for a long time, or if double pressed, limiting possibilities. Now all events are sent: pressed, short released, long released and double pressed.

Note that I also added two extra arguments the the send `zha_event` because it was a bit hard to extract which button was pressed (up or down). It's now a new attribute `{button: turn_on}` or `{button: turn_off}` and a `description` that contains the `ActionButton` name (Pressed_on, Released_off, Double_on, etc.)

## Additional information

I have been using this as a custom quirks for a while. I have two dimmer switches DM2500ZB dimmer switches and I'm emulating a 3-way dimmers with the downstairs dimmer using an HA automation to control upstairs dimmer that has the stairs lights connect to it. It works fine.

I also have two SW2500ZB, the events are also sent as expected.

## Checklist


bassdr:
@claudegel FYI.

zhaquirks/sinope/__init__.py
```diff
@@ -29,38 +32,90 @@
...
+        COMMAND: COMMAND_M_MULTI_PRESS_COMPLETE,
+        ARGS: {
+            ATTRIBUTE_ID: 84,
+            ATTRIBUTE_NAME: ATTRIBUTE_ACTION,
+            BUTTON: TURN_OFF,
+            VALUE: 20,
+        },
     },
-    (LONG_PRESS, TURN_ON): {
+    (LONG_RELEASE, TURN_ON): {
```
TheJulianJES: This would break existing automations with the "long press" trigger, right?

dmulcahey: Yes

bassdr: Indeed. The automation using them will show an error. Easy fix tho.

Is there a deprecation mechanism in place in ha? Or worse case I guess we can alias? It's really when we release the button that it sends either a long or short press event :/

TheJulianJES: There is not, unfortunately. What do you think about changing this back for now? Only real difference between them is another UI text for the end-user.

bassdr: If it's the only thing that breaks backward compatibility, I'm fine. If someone was  directly listening too zha_event, there are more breaking changes tho... Since the command changed too. Not sure if it's acceptable?

bassdr: I think my last commit fixes this concern. I'll leave you the final word.

bassdr:
I've tested my latest changes in HA and it works fine.

zhaquirks/sinope/sensor.py
```diff
@@ -30,7 +30,8 @@
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-from zhaquirks.sinope import SINOPE, SINOPE_MANUFACTURER_CLUSTER_ID
+
+from . import SINOPE, SINOPE_MANUFACTURER_CLUSTER_ID
```
TheJulianJES: Please revert to using absolute imports everywhere. It's easier to use them as custom quirks then.
Background: https://github.com/zigpy/zha-device-handlers/pull/827

bassdr: It's quite the opposite. Without them I can't use this as a quirk. Since there are modifications to the `__init__`. Without them you'll be using the wrong init file. I definitely challenge the call to use absolute imports for current module. I don't understand what's easier either. My setup to include this as a quirks is very simple.

TheJulianJES: Ah, so you're importing all these via the "custom quirks" feature.
If you do bigger modifications (involving multiple classes/files), it's ideal to directly modify the `zhaquirks` package.

We need absolute imports, as we want users to be able to easily modify a single quirk file (that imports from other files), without needing to load everything as custom quirks. For example, to easily add a single model to `MODELS_INFO`.
So, these changes need to be reverted before we can merge it.

bassdr: done

TheJulianJES (approved):
Looks good. Thanks.

PR #3451, opened 2024-10-22
Add Immax NEO Smart radiator valve `_TZE200_rufdtfyv`

markyxl:
## Proposed change
This PR is to add support for an updated model of [Immax NEO Smart Thermostatic head](https://www.immax.eu/immax-neo-smart-thermostatic-head-zigbee-p13447/). The retailers keep selling it as a single model, but the updated one has different signature and doesn't work with ZHA yet. As far as I can tell it's the same hardware, just with updated firmware with 2 new configruation options in the menu on the device itself.

## Additional information
Here is how it now reports in HA after including the quirk:
image

## Checklist


TheJulianJES (approved):
Thanks!

Issue #3095, opened 2024-04-08
[Device Support Request] Nimly doorlock (Touch, Touch Pro etc)


Fredrik81:
### Problem description

Add support for Nimly doorlocks.
According to support this should be possible from ZHA/z2m.
- Add remove user
- Open/Lock
- Change sound volume
- Activate autolock
- Se who unlocked and what method (code, finger, nfc etc)

In ZHA today only lock/unlock works. z2m seems to have more and user report they can do all except see who unlocked today.
Would be nice to get more features and at least battery level.

Here is link to z2m support for Nimly: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/onesti.ts

### Solution description

At least be able to see battery status but support for all the listed features would be great!

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.MainsPowered|RxOnWhenIdle|AllocateAddress: 140>, manufacturer_code=4660, maximum_buffer_size=108, maximum_incoming_transfer_size=127, server_mask=11264, maximum_outgoing_transfer_size=127, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "11": {
      "profile_id": "0x0104",
      "device_type": "0x000a",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0101",
        "0xfea2"
      ],
      "output_clusters": [
        "0x0019"
      ]
    }
  },
  "manufacturer": "Onesti Products AS",
  "model": "NimlyPRO",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json

```

</details>

### Logs

<details><summary>Logs</summary>

```python
[...]

Issue #2354, opened 2023-04-24
[Device Support Request] Onesti EasyCodeTouch doorlock


1bard:
This works as a lock, but I'm missing battery status.

The device is powered by 3xAA batteries, and as far as I know, mains power is not an option on this model.
The correct battery information seems to be available, but not used.
I believe the issue is that the device reports power source as mains/DC.

<details>

<summary>Device signature</summary>

```{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered: 140>, manufacturer_code=4660, maximum_buffer_size=108, maximum_incoming_transfer_size=127, server_mask=11264, maximum_outgoing_transfer_size=127, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "11": {
      "profile_id": 260,
      "device_type": "0x000a",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0101",
        "0xfea2"
      ],
      "out_clusters": [
        "0x0019"
      ]
    }
  },
  "manufacturer": "Onesti Products AS",
  "model": "EasyCodeTouch",
  "class": "zigpy.device.Device"
}
```

</details>

<details>
<summary>Diagnostic information</summary>

```"data": {
    "ieee": "**REDACTED**",
    "nwk": 34502,
    "manufacturer": "Onesti Products AS",
    "model": "EasyCodeTouch",
    "name": "Onesti Products AS EasyCodeTouch",
    "quirk_applied": false,
    "quirk_class": "zigpy.device.Device",
    "manufacturer_code": 4660,
    "power_source": "Mains",
    "lqi": 255,
    "rssi": -81,
    "last_seen": "2023-04-24T20:21:03",
    "available": true,
    "device_type": "EndDevice",
    "signature": {
[...]

PR #3457, opened 2024-10-25
Fix Nimly smart lock mains-powered capability

uvNikita:
## Proposed change
Add nimly smart lock quirk
Clears the mains powered mac capability flag since lock is battery powered.

Note that after quirk is applied, ZHA reports that my lock has 20% remaining battery, but when I query PowerConfiguration/battery_percentage_remaining attribute it reports 41%.
Not sure where ZHA gets 20% from.

See picture below:

## Additional information
Tested on NimlyPro model.

Ref
#3095
#2354

## Checklist

### DIFF
--- a/zhaquirks/nimly/__init__.py
+++ b/zhaquirks/nimly/__init__.py
@@ -1,0 +1,3 @@
+"""Nimly module."""
+
+NIMLY = "Onesti Products AS"
--- a/zhaquirks/nimly/lock.py
+++ b/zhaquirks/nimly/lock.py
@@ -1,0 +1,36 @@
+"""Device handler for Nimly Smart Locks."""
+
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.zdo.types import NodeDescriptor
+
+from zhaquirks.nimly import NIMLY
+
+# clears the mains powered mac capability flag
+NIMLY_LOCK_NODE_DESCRIPTOR = NodeDescriptor(
+    logical_type=2,
+    complex_descriptor_available=0,
+    user_descriptor_available=0,
+    reserved=0,
+    aps_flags=0,
+    frequency_band=8,
+    manufacturer_code=4660,
+    maximum_buffer_size=108,
+    maximum_incoming_transfer_size=127,
+    server_mask=11264,
+    maximum_outgoing_transfer_size=127,
+    descriptor_capability_field=0,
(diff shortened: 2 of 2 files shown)


uvNikita:
I didn't add tests since I'm not sure how to test quirks v2, e.g. `assert_signature_matches_quirk`. Also not sure how to test node descriptor and power source specifically.

So it would be great with some direction on this if tests are required :)

TheJulianJES:
Tests aren't really required for this. All good.

TheJulianJES:
> when I query PowerConfiguration/battery_percentage_remaining attribute it reports 41%.
Not sure where ZHA gets 20% from.

`battery_percentage_remaining` goes from 0 to 200. 200 means 100%. ZHA remaps it.
Some manufacturers also miss that and use 100 for 100%, which is wrong. There's a doubling power configuration cluster which doubles that attribute value before it reaches ZHA. We can also use that if the manufacturer made the same mistake.

TheJulianJES (approved):
Thanks!

uvNikita:
Thanks for the review and the explanation about battery levels. I'll try to replace the battery and see if it will report the correct value. If not, I'll open another PR 🙂

jb3112:
How to add it, im not very expirenced with home assistant

uvNikita:
@jb3112 this quirk is available since Home Assistant 2024.11, so just make sure you are running the latest version.
Reported battery level will be halved until you will update to Home Assistant 2024.12 where fix for that was pulled.

jb3112:
how to make quirk

uvNikita:
@jb3112 The quirk from this PR is applied automatically if your are using ZHA, the device producer is "Onesti Products AS", model name is "NimlyPRO", "NimlyCode", "NimlyTouch", "NimlyIn", "EasyFingerTouch, or "EasyCodeTouch". Nothing that you should be doing other than update Home Assistant.

jb3112:
I have it here 
image

uvNikita:
Oh, I see the issue. Your model name is `easyCodeTouch_v1` which is not on the list.

I'll create another PR to add it. Unfortunately you will have to wait for a new release to get the battery readings.

jb3112:
Great, is there other thinks, like who has unlocked
Had it on smarthings before there i could see who unlocked

uvNikita:
Not yet, but I'm planning to add this information too. Here is my comment for the reference: https://github.com/zigpy/zha-device-handlers/pull/3465#issuecomment-2439976447

PR #3456, opened 2024-10-25
Zlinky_TIC conversion to quirk v2

blauret:
## Proposed change

The quirk for is converted to v2 and adds more sensor to the use in HA

## Checklist

- [x] The changes are tested and work correctly
- [X] `pre-commit` checks pass / the code has been formatted using Black
- [x] Tests have been added to verify that the new code works

PR #3465, opened 2024-10-27
Fix battery reporting half on some Nimly locks

uvNikita:
## Proposed change

Add `DoublingPowerConfigurationCluster` for some nimly locks.
Following the previous PR #3457.

According to this conversation over at Z2M, only Nimly new locks require this fix, not their old EasyTouch/EasyFinger locks: https://github.com/Koenkk/zigbee-herdsman-converters/pull/6940.

## Checklist

### DIFF
--- a/zhaquirks/nimly/lock.py
+++ b/zhaquirks/nimly/lock.py
@@ -3,6 +3,7 @@
 from zigpy.quirks.v2 import QuirkBuilder
 from zigpy.zdo.types import NodeDescriptor
 
+from zhaquirks import DoublingPowerConfigurationCluster
 from zhaquirks.nimly import NIMLY
 
 # clears the mains powered mac capability flag
@@ -25,12 +26,18 @@ NIMLY_LOCK_NODE_DESCRIPTOR = NodeDescriptor(
 
 
 (
+    QuirkBuilder(NIMLY, "EasyFingerTouch")
+    .also_applies_to(NIMLY, "EasyCodeTouch")
+    .node_descriptor(NIMLY_LOCK_NODE_DESCRIPTOR)
+    .add_to_registry()
+)
+
+(
     QuirkBuilder(NIMLY, "NimlyPRO")
     .also_applies_to(NIMLY, "NimlyCode")
     .also_applies_to(NIMLY, "NimlyTouch")
     .also_applies_to(NIMLY, "NimlyIn")
-    .also_applies_to(NIMLY, "EasyFingerTouch")
-    .also_applies_to(NIMLY, "EasyCodeTouch")
     .node_descriptor(NIMLY_LOCK_NODE_DESCRIPTOR)
+    .replaces(DoublingPowerConfigurationCluster, endpoint_id=11)
     .add_to_registry()
 )


uvNikita:
@TheJulianJES I aslo would like to add some other attributes exposed by this lock that are already available at Z2M (see code [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3e3453fdd8c0def64232fed4973b58537a6f531c/src/devices/onesti.ts#L12)):

- last_used_pin_code
- last_action_source
- last_action_user
- last_lock_user
- last_lock_source
- last_unlock_user
- last_unlock_source

Could you give me some pointers on how I can implement this? From what I understand looking at Z2M code, the information can be parsed from attribute 256 and 257.
Maybe you could point to an example quirk from this module that is doing something similar?

blauret:
@uvNikita, maybe you can find inspiration in #3456

uvNikita:
@blauret thanks, that might be what I'm looking for!

I think one puzzle piece I'm still missing is how can I modify published sensor to not just publish raw value, but parse the attribute first and extract information similar to Z2M.

TheJulianJES:
> how can I modify published sensor to not just publish raw value, but parse the attribute first

You can't, yet. We're looking into adding this.

TheJulianJES (approved):
Thanks!

Eriond:
Sorry for this notification being just an observation, instead of a solution proposal, but after updating to HA 2024.11.0 as of today, my ZHA wouldn't start unless I commented out the quirk that adds the battery percentage doubling. I figured that I had to let this excellent solution go in favor of having the rest of my home working 🙂
It's kinda dark now most of the day where I live, so controlling all my lights have precedence over correct battery readings 😄 
Perhaps You've had a similar experience? I think it has to do with the quirk being detected as doubled or redundant, from this release?

TheJulianJES:
@Eriond Please open an issue, tag me there, and attach logs of the issue.

uvNikita:
@Eriond it's probably because the original PR which added support for this lock got into the new home assistant release, but not the doubling fix. And your custom quirk conflicts with the new quirk added.

Eriond:
Yes, that was my conclusion too.
I have now commented out the quirk, and so my logs no longer have the error message in them.
Are you still helped by me recreating the fault (and providing you with the relevant error message), or can you just assume that your assumption is right?

Edit:
I tried to restore that quirk (by un-commenting it), but I don't get the error message anymore. Sorry.

uvNikita:
@TheJulianJES is there a way to override an existing quirk in zha-device-handlers with the one from `custom_quirks_path`?

As of now, I'm getting 

> zigpy.exceptions.MultipleQuirksMatchException: Multiple matches found for device \<Device model='NimlyPRO' manuf='Onesti Products AS' nwk=0xF84D ieee=f4:ce:36:51:39:99:b3:30 is_initialized=True>
....

uvNikita:
@Eriond I managed to override the quirk with the custom one using this workaround:


```
from zigpy.quirks import DEVICE_REGISTRY
....

# remove upstream quirk from registry
del DEVICE_REGISTRY._registry_v2[(NIMLY, "NimlyPRO")]

# add custom quirk
(
    QuirkBuilder(NIMLY, "NimlyPRO")
    .node_descriptor(NIMLY_LOCK_NODE_DESCRIPTOR)
    .replaces(DoublingPowerConfigurationCluster, endpoint_id=11)
    .add_to_registry()
)
```

Eriond:
Thanks for investigating this further.
However, changing the quirk according to your instructions above, doesn't have any effect on the battery sensor. It is still at a steady 50 (%) after replacing with fresh batteries a couple of weeks ago.

This is my current card code:
```
type: entities
entities:
  - entity: lock.ytterdorr_door_lock
    icon: mdi:door-closed-lock
  - type: custom:template-entity-row
    icon: mdi:battery
    name: Battery remaining {{ states('sensor.ytterdorr_battery') }}%
    state: >-
      {% if is_state('lock.ytterdorr_door_lock', 'locked')%} Locked {% else %}
      Open {% endif %}
    secondary: >-
      Last change {{as_timestamp(states.lock.ytterdorr_door_lock.last_changed) |
      timestamp_custom('%X, %a %d %b')}}
```

uvNikita:
@Eriond 
My code above will work only for Nimly Pro lock, so make sure you have that one or adjust the code for your model.

Also, make sure to re-read the value from Manage Zigbee device → DoublingPowerConfigurationCluster → battery_percentage_remaining.
Might also need to click on “Reconfigure” button for the device.

Eriond:
Yep, that was the trick! Re-reading the value fixed it. Thank you so much.
I thought that restarting the whole machine on which HA is running, should be enough, but apparently not.
Excellent work. I hope it will find it's way into the regular releases.

gustavostmo:
@TheJulianJES I am also very interested in getting the same functionality as z2m as @uvNikita wrote higher up. I don't have much experience with Quirk so I can't do it myself unfortunately

uvNikita:
> @TheJulianJES I aslo would like to add some other attributes exposed by this lock that are already available at Z2M (see code [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3e3453fdd8c0def64232fed4973b58537a6f531c/src/devices/onesti.ts#L12)):
> 
>     * last_used_pin_code
> 
>     * last_action_source
> 
>     * last_action_user
> 
>     * last_lock_user
> 
>     * last_lock_source
> 
>     * last_unlock_user
> 
>     * last_unlock_source
> 
> 
> Could you give me some pointers on how I can implement this? From what I understand looking at Z2M code, the information can be parsed from attribute 256 and 257. Maybe you could point to an example quirk from this module that is doing something similar?

@TheJulianJES  I see that you [added](https://github.com/zigpy/zigpy/pull/1540) `attribute_converter` to sensor and binary sensor entities. I presume those can be used to parse out the data we want similar to how Z2M does it. It would solve #3095 

If that's the case, I can try to add missing sensors.

PR #3473, opened 2024-10-29
Add Schneider Electric Thermostats

uvNikita:
## Proposed change
Adds support to some Electric Thermostats based on the zigbee specs published here: https://github.com/Koenkk/zigbee2mqtt/issues/23739

Specs: [RTC-6754012INT-cFMT_16A_F_Zigbee-29.pdf](https://github.com/user-attachments/files/17563282/RTC-6754012INT-cFMT_16A_F_Zigbee-29.pdf)

For now, I added two most interesting attributes for me:

- `Metering/fixed_load_demand`. It can be set to a nominal consumption of the heating. After setting this value, thermostat starts to output summation delivered estimated based on specified load.
- `TemperatureMeasurement/temperature_sensor_type`. Can be used to modify external (floor) sensor type specified during initial installation.

Reference to the original issue https://github.com/zigpy/zha-device-handlers/pull/1705

## Checklist


uvNikita:
I wasn't sure if I should add it under `zhaquirks/schneiderelectric` or `zhaquirks/schneider`

uvNikita:
I went ahead and added the rest of the clusters, custom attributes and commands from the spec file.

uvNikita:
@TheJulianJES I think this one should be ready for review :)

I think everything from spec is implemented. The only thing to add I can think of is home assistant numbers, sensors etc. For now, whatever I add gets `_none` suffix in it's entity id, not sure how to fix it yet.

TheJulianJES:
> I think everything from spec is implemented. The only thing to add I can think of is home assistant numbers, sensors etc. For now, whatever I add gets `_none` suffix in it's entity id, not sure how to fix it yet.

With HA Core 2024.11, ZHA should fall back to the `fallback_name` if it cannot find the value for a translation key.
We'll add the translation key values to Home Assistant Core in the ZHA dependency bump, based on the `fallback_name` in the quirk.

zhaquirks/schneiderelectric/thermostat.py
```diff
@@ -0,0 +1,576 @@
...
+    .replaces(SEBasic, endpoint_id=2)
+    .replaces(SEBasic, endpoint_id=3)
+    .replaces(SEBasic, endpoint_id=5)
+    .replaces(SEThermostat)
+    .replaces(SEUserInterface)
+    .replaces(SETemperatureMeasurement, endpoint_id=2)
+    .replaces(SETemperatureMeasurementExternal, endpoint_id=3)
+    .replaces(SEMetering, endpoint_id=5)
+    .adds(SECycleTime)
+    .adds(SEHeatingCoolingOutput)
```
TheJulianJES, line 574: If these are also in the original signature (with no quirk), we should use `replaces` here too.

TheJulianJES: If these are also in the original signature (with no quirk), we should use `replaces` here too.
```suggestion
    .replaces(SECycleTime)
    .replaces(SEHeatingCoolingOutput)
```

uvNikita: You are right, should be `replaces`

TheJulianJES (approved):
LGTM, thanks!

iamdabe:
Can this be extended to handle the Image CCTFR6700 Schneider Electric's Wiser - Electric heating actuator- 16A - Pilot wire CCTFR6700 too?

https://www.zigbee2mqtt.io/devices/CCTFR6700.html

The device requires pilot wire setting and method to set temperature to work correctly. Currently it is detected but does not work correctly. 

https://community.home-assistant.io/t/schneider-electric-cctfr6700-16a-relay-external-temperature-sensor/341537/9?u=iamdabe

PR #3481, opened 2024-11-03
Add UseeLink 302EU-Zigbee power strip (_TZ3000_cfnprab5)

hangy:
Adds support for UseeLink 302EU-Zigbee, based on https://github.com/zigpy/zha-device-handlers/issues/1632#issuecomment-1627072807

### DIFF
--- a/zhaquirks/tuya/ts011f_plug.py
+++ b/zhaquirks/tuya/ts011f_plug.py
@@ -1838,3 +1838,162 @@ class Plug_v7(EnchantedDevice):
             },
         },
     }
+
+
+class Plug_4AC_2USB_TZ3000_cfnprab5(CustomDevice):
+    """Tuya 4 outlet + 4 USB without any metering or restore."""
+
+    signature = {
+        MODEL: "TS011F",
+        ENDPOINTS: {
+            1: {
+                # <SimpleDescriptor endpoint=1 profile=260 device_type=0x0100
+                # input_clusters=[0, 3, 4, 5, 6, 0xe001]
+                # output_clusters=[]>
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    TuyaZBE000Cluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [Time.cluster_id],
+            },
+            2: {
+                # <SimpleDescriptor endpoint=1 profile=260 device_type=0x0100
+                # input_clusters=[0, 3, 4, 5, 6, 0xe001]
+                # output_clusters=[]>
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    TuyaZBE000Cluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
+            3: {
+                # <SimpleDescriptor endpoint=1 profile=260 device_type=0x0100
+                # input_clusters=[0, 3, 4, 5, 6, 0xe001]
+                # output_clusters=[]>
+                PROFILE_ID: zha.PROFILE_ID,
+                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT,
+                INPUT_CLUSTERS: [
+                    Basic.cluster_id,
+                    Identify.cluster_id,
+                    Groups.cluster_id,
+                    Scenes.cluster_id,
+                    OnOff.cluster_id,
+                    TuyaZBE000Cluster.cluster_id,
+                ],
+                OUTPUT_CLUSTERS: [],
+            },
+            4: {
+                # <SimpleDescriptor endpoint=1 profile=260 device_type=0x0100
+                # input_clusters=[0, 3, 4, 5, 6, 0xe001]
+                # output_clusters=[]>
+                PROFILE_ID: zha.PROFILE_ID,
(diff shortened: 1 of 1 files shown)


kender65:
Hi!
Based on the description "UseeLink 302EU-Zigbee" I guess [this](https://www.amazon.de/Steckdosenleiste-UseeLink-Mehrfachsteckdose-Unabh%C3%A4ngig-Anschl%C3%BCsse-302EU-Zigbee/dp/B0CTMXYQLZ?th=1) is the device in question.
Wouldn't it be better to name the class "Plug_4AC_**4**USB_TZ3000_cfnprab5" to make it clear, that it has 4 USB (3 x USB-A, 1 x USB-C) ports?
I got this [SM-302Ez-PA20](https://www.amazon.de/UseeLink-Steckdosenleiste-AC-Anschl%C3%BCsse-Schnellladeger%C3%A4t-Sprachsteuerung/dp/B0CT8FXWF5?source=ps-sl-shoppingads-lpcontext&ref_=fplfs&psc=1&smid=A2M76ZDYXTOLGU) which looks nearly the same. But I think mine has the ability to set "child_lock" and at least to control the "power_outage_memory" (or "power_on_state") as mentioned [here](https://github.com/zigpy/zha-device-handlers/issues/1632#issuecomment-1213118577). Initially I integrated this power strip with Z2M on HAOS although not listed as a supported device. Pairing went well but surprisingly it showed up as [SM-0306E-2W](https://www.zigbee2mqtt.io/devices/SM-0306E-2W.html) which caused the wrong picture of the device on the Z2M UI. I was able to toggle the outlets and USB ports like in ZHA and additionally could set the "power_outage_memory" to it's expected values "on", "off", "restore" from the Z2M UI and it worked as expected although I got an error message for an "UNSUPPORTED_ATTRIBUTE". The same goes for the "child_lock". Setting from the UI to both values worked as expected despite an "UNSUPPORTED_ATTRIBUTE" message. I think the behaviour matches the [comment](https://github.com/zigpy/zha-device-handlers/issues/1632#issuecomment-1213380426) from @javicalle as the "power_outage_memory" was only visible  for endpoint 1 (and as general attribute which may be forced the error message).

To figure out what was the problem I tried the other way with ZHA, which I haven't used until then.
I got it working with your quirk, thanks for that. Toggling the 4 AC outlets and the USB ports as a group from the ZHA UI and / or from the developer tools works without issues.
image
Further configuration for "power_on_state", "power_outage_memory" or "start_up_on_off", or whatever you would call it, doesn't work so I extended your quirk with "quirk_id = TUYA_PLUG_ONOFF" in the hope of getting "child_lock" and "power_outage_memory" / "power_on_state" working. I got additional attributes in the "TuyaZBOnOffAttributeCluster"
image
but haven't been able to get it working. I couldn't set the value to "on", "off" or "restore". Initially it has the value "None".
The same is for
image
Any idea how to get it working, for at least endpoint 1?

hangy:
> Hi!

🙋‍♂️

> Based on the description "UseeLink 302EU-Zigbee" I guess [this](https://www.amazon.de/Steckdosenleiste-UseeLink-Mehrfachsteckdose-Unabh%C3%A4ngig-Anschl%C3%BCsse-302EU-Zigbee/dp/B0CTMXYQLZ?th=1) is the device in question. 

That's the one.

> Wouldn't it be better to name the class "Plug_4AC_**4**USB_TZ3000_cfnprab5" to make it clear, that it has 4 USB (3 x USB-A, 1 x USB-C) ports?

Fair point! Maybe just `_USB_` or `_1USB_`, as all USB ports can _only_ be switched on or off as one group? I haven't looked at that many quirks, yet, so I'm not aware of a naming convention for a case like this.

> I got this [SM-302Ez-PA20](https://www.amazon.de/UseeLink-Steckdosenleiste-AC-Anschl%C3%BCsse-Schnellladeger%C3%A4t-Sprachsteuerung/dp/B0CT8FXWF5?source=ps-sl-shoppingads-lpcontext&ref_=fplfs&psc=1&smid=A2M76ZDYXTOLGU) which looks nearly the same. But I think mine has the ability to set "child_lock" and at least to control the "power_outage_memory" (or "power_on_state") as mentioned [here](https://github.com/zigpy/zha-device-handlers/issues/1632#issuecomment-1213118577). Initially I integrated this power strip with Z2M on HAOS although not listed as a supported device. Pairing went well but surprisingly it showed up as [SM-0306E-2W](https://www.zigbee2mqtt.io/devices/SM-0306E-2W.html) which caused the wrong picture of the device on the Z2M UI. I was able to toggle the outlets and USB ports like in ZHA and additionally could set the "power_outage_memory" to it's expected values "on", "off", "restore" from the Z2M UI and it worked as expected although I got an error message for an "UNSUPPORTED_ATTRIBUTE". The same goes for the "child_lock". Setting from the UI to both values worked as expected despite an "UNSUPPORTED_ATTRIBUTE" message. I think the behaviour matches the [comment](https://github.com/zigpy/zha-device-handlers/issues/1632#issuecomment-1213380426) from @javicalle as the "power_outage_memory" was only visible for endpoint 1 (and as general attribute which may be forced the error message).
> 
> To figure out what was the problem I tried the other way with ZHA, which I haven't used until then. I got it working with your quirk, thanks for that. Toggling the 4 AC outlets and the USB ports as a group from the ZHA UI and / or from the developer tools works without issues. image Further configuration for "power_on_state", "power_outage_memory" or "start_up_on_off", or whatever you would call it, doesn't work so I extended your quirk with "quirk_id = TUYA_PLUG_ONOFF" in the hope of getting "child_lock" and "power_outage_memory" / "power_on_state" working. I got additional attributes in the "TuyaZBOnOffAttributeCluster" image but haven't been able to get it working. I couldn't set the value to "on", "off" or "restore". Initially it has the value "None". The same is for image Any idea how to get it working, for at least endpoint 1?

I'm not too experience with zigpy or ZHA, so I guess we're both figuring it out in parallel. 😉 
For my device, the "Startverhalten" also seems to accept the `start_up_on_off` value: 
ZHA screen in Home Assistant showing the different switches of the board with different power-on states
THA screen of Home Assistant showing the 0x4003 property of the 0x0006 cluster endpoint 2
Unfortunately, I cannot test whether this actually works right now, because the plug is neatly hidden behind a cabinet. It does look like the values are mapped and accepted by the device, though.

As for the `child_lock`, my power strip doesn't accept setting them at all, and only returns `None`. To be honest, that sounds about right, though. The `child_lock` property seems to be displayed for each switch, but if it was actually supported/worked, I'd expect there to be one `child_lock` setting for the main power button, and the main button as a separate cluster. Having different child lock settings for each socket doesn't make too much sense

PR #3501, opened 2024-11-10
Add support for Sinopé HP6000ZB-xx thermostats

claudegel:
## Proposed change

Add support for new Sinopé HP6000ZB-xx thermostats

## Checklist

### DIFF
--- a/zhaquirks/sinope/thermostat.py
+++ b/zhaquirks/sinope/thermostat.py
@@ -1,4 +1,4 @@
-"""Module to handle quirks of the  Sinopé Technologies thermostat.
+"""Module to handle quirks of the Sinopé Technologies thermostats.
 
 Manufacturer specific cluster implements attributes to control displaying
 of outdoor temperature, setting occupancy on/off and setting device time.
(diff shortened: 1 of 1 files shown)


zhaquirks/sinope/thermostat.py
```diff
@@ -645,3 +646,93 @@ class SinopeG2Thermostats(SinopeTechnologiesThermostat):
             }
         }
     }
+
+
+class SinopeHPThermostats(SinopeTechnologiesThermostat):
```
TheJulianJES, line 651: I don't see any real reason this needs to extend/inherit from `SinopeTechnologiesThermostat`, instead of just `CustomDevice`. We should probably change it on all Sinope thermostat quirks in a future PR.

For now, I guess it's fine, as the others already do it.

TheJulianJES (approved):
Thanks!

zhaquirks/sinope/thermostat.py
claudegel: Your wright, I gonna change all of them and make a new PR. I guess it'll be the same for light, switch and sensor.

PR #3313, opened 2024-08-23
Convert Sinope light to new style AttributeDefs, add ZHA events

ckm2k1:
## Proposed change
Add ability for Sinope light switches to handle routing attribute reports on the manufacturer cluster attribute `action_report` as ZHA events.

## Additional information
Related PR zigpy/zha/pull/153 made it so that `action_report` doesn't need to be manually configured for reporting (as shown by @claudegel here: https://github.com/claudegel/sinope-zha?tab=readme-ov-file#a-create-the-reporting-).

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [x] Tests have been added to verify that the new code works

PR #3504, opened 2024-11-10
Fix Sinope device triggers

ckm2k1:
## Proposed change
Fixes device automation triggers for the Sinope light quirk. The current definition of the automation triggers
uses the `ButtonAction` enum in the `value` field, which throws an error from Voluptuous when passed into HA's `async_attach_trigger`.

## Additional information
Not a breaking change and doesn't require anything else to be merged into HA Core.

## Checklist

### DIFF
--- a/tests/test_sinope.py
+++ b/tests/test_sinope.py
@@ -247,7 +247,6 @@ async def test_sinope_light_device_triggers_def(zigpy_device_from_quirk, quirk):
     device: Device = zigpy_device_from_quirk(quirk)
 
     for config in device.device_automation_triggers.values():
-        # from pudb import set_trace; set_trace();
         val = config.get("args", {}).get("value")
         if val is not None:
             assert type(val) is int, type(val)
--- a/tests/test_sinope.py
+++ b/tests/test_sinope.py
@@ -239,3 +239,15 @@ async def test_sinope_light_switch_reporting(zigpy_device_from_quirk, quirk):
 
         assert len(request_mock.mock_calls) == 1
         assert len(bind_mock.mock_calls) == 1
+
+
+@pytest.mark.parametrize("quirk", (SinopeTechnologieslight,))
+async def test_sinope_light_device_triggers_def(zigpy_device_from_quirk, quirk):
+    """Test that configuring reporting for action_report works."""
+    device: Device = zigpy_device_from_quirk(quirk)
+
+    for config in device.device_automation_triggers.values():
+        # from pudb import set_trace; set_trace();
+        val = config.get("args", {}).get("value")
+        if val is not None:
+            assert type(val) is int, type(val)
--- a/tests/test_sinope.py
+++ b/tests/test_sinope.py
@@ -243,7 +243,11 @@ async def test_sinope_light_switch_reporting(zigpy_device_from_quirk, quirk):
 
 @pytest.mark.parametrize("quirk", (SinopeTechnologieslight,))
 async def test_sinope_light_device_triggers_def(zigpy_device_from_quirk, quirk):
-    """Test that configuring reporting for action_report works."""
+    """Test device automation triggers.
+
+    Make sure that values are actual ints and not instances of an enum class.
+    """
+
     device: Device = zigpy_device_from_quirk(quirk)
 
     for config in device.device_automation_triggers.values():
--- a/zhaquirks/sinope/__init__.py
+++ b/zhaquirks/sinope/__init__.py
@@ -52,7 +52,7 @@ LIGHT_DEVICE_TRIGGERS = {
             ATTRIBUTE_ID: 84,
             ATTRIBUTE_NAME: ATTRIBUTE_ACTION,
             BUTTON: TURN_ON,
-            VALUE: ButtonAction.Pressed_on,
+            VALUE: int(ButtonAction.Pressed_on),
         },
     },
     (SHORT_PRESS, TURN_OFF): {
@@ -63,7 +63,7 @@ LIGHT_DEVICE_TRIGGERS = {
             ATTRIBUTE_ID: 84,
             ATTRIBUTE_NAME: ATTRIBUTE_ACTION,
             BUTTON: TURN_OFF,
-            VALUE: ButtonAction.Pressed_off,
+            VALUE: int(ButtonAction.Pressed_off),
         },
     },
     (SHORT_RELEASE, TURN_ON): {
@@ -74,7 +74,7 @@ LIGHT_DEVICE_TRIGGERS = {
             ATTRIBUTE_ID: 84,
             ATTRIBUTE_NAME: ATTRIBUTE_ACTION,
             BUTTON: TURN_ON,
-            VALUE: ButtonAction.Released_on,
+            VALUE: int(ButtonAction.Released_on),
         },
     },
     (SHORT_RELEASE, TURN_OFF): {
@@ -85,7 +85,7 @@ LIGHT_DEVICE_TRIGGERS = {
             ATTRIBUTE_ID: 84,
             ATTRIBUTE_NAME: ATTRIBUTE_ACTION,
             BUTTON: TURN_OFF,
-            VALUE: ButtonAction.Released_off,
+            VALUE: int(ButtonAction.Released_off),
         },
     },
     (DOUBLE_PRESS, TURN_ON): {
@@ -96,7 +96,7 @@ LIGHT_DEVICE_TRIGGERS = {
             ATTRIBUTE_ID: 84,
             ATTRIBUTE_NAME: ATTRIBUTE_ACTION,
             BUTTON: TURN_ON,
-            VALUE: ButtonAction.Double_on,
+            VALUE: int(ButtonAction.Double_on),
         },
     },
     (DOUBLE_PRESS, TURN_OFF): {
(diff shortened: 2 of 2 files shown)


ckm2k1:
I noticed an issue while trying to debug this problem.

To validate my fix, I added a modified copy of the `sinope` quirk into the custom quirks directory in my live HA instance.
Although ZHA loads the modified quirk just fine, since the fix itself happens at definition time, my modifications are loaded too late in the process to fix the call to `async_attach_trigger` in `/zha/core/homeassistant/components/zha/device_trigger.py`. 

Instead, I had to ssh into HA OS as root and modify the `zhaquirks` code manually in `/usr/local/lib/python3.12/site-packages/zhaquirks/sinope/__init__.py`. This worked, but is pretty tedious if you're trying to iterate and check your changes. 

I don't know if it's possible, but it would be nice if the loading order was changed such that if a custom quirk is present and it overrides an existing quirk, it would be loaded _instead_ of the builtin one, and not after it.

----
Separately, I also noticed that there's a helper `schema_type_to_vol`, which is called on cluster command arguments, likely to prevent the exact problem I'm trying to fix in this PR. 

If all fields in device automation triggers were either validated at dev time, or better yet, automatically coerced into voluptuous supported types, quirk authors wouldn't be forced to use scalar types instead of Enum values.

ckm2k1:
@bassdr a couple questions for you: 

* Does the above comment sound like what you ran into here: https://github.com/zigpy/zha-device-handlers/pull/3438#discussion_r1815884198? If so, how did you work around it?
* Did all your button related automations also break with HA 2024.11.0?

bassdr:
> @ckm2k1 
> https://github.com/zigpy/zha-device-handlers/pull/3504#issuecomment-2466749623
> I don't know if it's possible, but it would be nice if the loading order was changed such that if a custom quirk is present and it overrides an existing quirk, it would be loaded instead of the builtin one, and not after it.


> @bassdr a couple questions for you:
> - Does the above comment sound like what you ran into here: https://github.com/zigpy/zha-device-handlers/pull/3438#discussion_r1815884198? If so, how did you work around it?

Yes, this problem is exactly why I used relative import paths. This way, you can add the `__init__.py` file to your quirk and it works just fine. See https://github.com/zigpy/zha-device-handlers/pull/3438/commits/f922d1439e2665d0a23a290a3e75017ff0544d1a, actually, you want to revert this one.
So in short, in your config folder create a folder called zhaquirks, inside you create a folder named sinope, and you setup your configuration.yaml to point to the zharquirks folder (not the sinope folder). The sinope folder must include the `__init__.py` and all files that will be affected by the changes (take the whole folder just to be safe). You then have to change everywhere you `from zhaquirks.sinope import` to `from . import`. It just works. I did not understand the point of having absolute paths but, not the end of the world either...

> - Did all your button related automations also break with HA 2024.11.0?

I haven't tried 2024.11.0 yet, I keep my non-dev HA setup on the last version of last month... But I'm surprised to see this PR indeed, I'll upgrade to check when I get the chance. Do you still have the exact voluptuous error it gave you? Or maybe a snippet of you automation that is giving the error? I'm personally listening to the zha event directly like this:

``` yaml
  - trigger: event
    event_type: zha_event
    event_data:
      endpoint_id: 1
      cluster_id: 65281
      args:
        attribute_name: action_report
```

To test it's the right device:
```yaml
condition: template
value_template: |-
  {{
    device_id('light.upstairslightswitch_lumiere') == trigger.event.data.device_id
  }}
```

and eventually:
``` yaml
choose:
  - conditions:
      - condition: template
        value_template: "{{ trigger.event.data.command == 'initial_press' }}"
    sequence:
      - action: zha.issue_zigbee_cluster_command
        data:
          ieee: |-
            {{
              device_attr('light.upstairslightswitch_lumiere', 'identifiers')
                | selectattr(0,'eq','zha')
                | map(attribute=1)
                | first
            }}
          endpoint_id: 1
          cluster_id: 8
          cluster_type: in
          command: 5
          command_type: server
          params:
            move_mode: |-
              {{
                iif (trigger.event.data.args.button == 'turn_on', 'Up', 'Down')
              }}
            rate: 50
    alias: Initial Press
```

bassdr:
I can't find the schema of `device_automation_triggers`, TBH that might be the part I missed in the original PR, but I assumed this was not fixed anywhere.

EDIT: I think I figured my original mistake: the `device_automation_triggers` should either listen to `description` field or be casted to `int` in the `value` field like you are doing in this PR, as emitted here:
```python
        event_args = {
            ATTRIBUTE_ID: 84,
            ATTRIBUTE_NAME: ATTRIBUTE_ACTION,
            BUTTON: button,
            DESCRIPTION: action.name,
            VALUE: action.value,
        }
```
`action.value` here is equivalent to `int(action)`

zhaquirks/sinope/__init__.py
```diff
@@ -52,7 +52,7 @@ class ButtonAction(t.enum8):
             ATTRIBUTE_ID: 84,
             ATTRIBUTE_NAME: ATTRIBUTE_ACTION,
             BUTTON: TURN_ON,
-            VALUE: ButtonAction.Pressed_on,
+            VALUE: int(ButtonAction.Pressed_on),
```
bassdr, line 55: Good catch, sorry about it.

I personally prefer `VALUE: ButtonAction.Pressed_on.value` to match the syntax when we emit.

Another fix would be to change to `DESCRIPTION: ButtonAction.Pressed_on.name`, but I don't have a strong preference.

tests/test_sinope.py
```diff
@@ -239,3 +239,18 @@ async def test_sinope_light_switch_reporting(zigpy_device_from_quirk, quirk):
...
+
+    Make sure that values are actual ints and not instances of an enum class.
+    """
+
+    device: Device = zigpy_device_from_quirk(quirk)
+
+    for config in device.device_automation_triggers.values():
+        val = config.get("args", {}).get("value")
+        if val is not None:
+            assert type(val) is int, type(val)
```
bassdr, line 256: I'm OK with this test, but like I said in the other comment, I don't think it's forced anywhere that the value field is an int. I personally think zha should fix rules like this tho (so we don't have to do it in the sinope test, but all fields would be tested at once)

zhaquirks/sinope/__init__.py
ckm2k1: So do I :)

Problem is that when I tested using `.value`, it turned out that it's of type `<class 'zigpy.types.basic.uint8_t'>` and not an `int`! This still made Voluptuous raise an invalid type exception, so I opted for the next simplest solution, coercing to int.

tests/test_sinope.py
ckm2k1: I agree in principal, and would happily participate in a discussion with the ZHA maintainers, but since there are no guarantees, I'd prefer to patch things in a simple way and have our automations work like before.

ckm2k1: > I don't think it's forced anywhere that the value field is an int

This is true, but it's indirectly forced by the fact that Voluptuous fails to handle user types. I did find the root of the problem, described in my comment here: https://github.com/zigpy/zha-device-handlers/pull/3504#issuecomment-2466749623

ckm2k1:
@bassdr 
> I haven't tried 2024.11.0 yet, I keep my non-dev HA setup on the last version of last month... But I'm surprised to see this PR indeed, I'll upgrade to check when I get the chance. Do you still have the exact voluptuous error it gave you? Or maybe a snippet of you automation that is giving the error? I'm personally listening to the zha event directly like this:
> 
> ```yaml
>   - trigger: event
>     event_type: zha_event
>     event_data:
>       endpoint_id: 1
>       cluster_id: 65281
>       args:
>         attribute_name: action_report
> ```
> ...

I'm not sure the above is needed anymore after these PRs: https://github.com/zigpy/zha-device-handlers/pull/3313 and https://github.com/zigpy/zha/pull/153. 

I got tired of writing YAML for these things (I have a whole bunch of Sinope devices, particularly light switches), so I submitted the PRs above to patch things and make it work through the UI. Here's a partial example of an automation I'm running:

image

TheJulianJES:
> I don't know if it's possible, but it would be nice if the loading order was changed such that if a custom quirk is present and it overrides an existing quirk, it would be loaded instead of the builtin one, and not after it.

> If all fields in device automation triggers were either validated at dev time, or better yet, automatically coerced into voluptuous supported types, quirk authors wouldn't be forced to use scalar types instead of Enum values.

Can you create an issue for each of these please? You can just copy paste your comment in those. For the first one, maybe in the ZHA repo? For the second one, we should add probably add some basic tests in the quirks repo to validate device triggers IMO.

ckm2k1:
> > I don't know if it's possible, but it would be nice if the loading order was changed such that if a custom quirk is present and it overrides an existing quirk, it would be loaded instead of the builtin one, and not after it.
> 
> > If all fields in device automation triggers were either validated at dev time, or better yet, automatically coerced into voluptuous supported types, quirk authors wouldn't be forced to use scalar types instead of Enum values.
> 
> Can you create an issue for each of these please? You can just copy paste your comment in those. For the first one, maybe in the ZHA repo? For the second one, we should add probably add some basic tests in the quirks repo to validate device triggers IMO.

@TheJulianJES - no problem, will do!

TheJulianJES (approved):
Thanks!

PR #3509, opened 2024-11-13
Update IKEA Styrbar Remote to support long press releases for left and right buttons

0lento:
## Proposed change
Add "released after long press" triggers for both Left and Right buttons on the IKEA Styrbar Remote quirk.

## Additional information
This change was trivial to make so this makes me wonder if there's some reason why this wasn't done on the earlier version of the quirk? This is new code base for me so maybe I've missed some reasoning here. However I've verified the changes to work on the Styrbar remote I have and I use them on my own automations.

## Checklist

### DIFF
--- a/zhaquirks/ikea/fourbtnremote.py
+++ b/zhaquirks/ikea/fourbtnremote.py
@@ -12,6 +12,7 @@ from zhaquirks.const import (
     COMMAND_OFF,
     COMMAND_ON,
     COMMAND_PRESS,
+    COMMAND_RELEASE,
     COMMAND_STOP,
     COMMAND_STOP_ON_OFF,
     DIM_DOWN,
@@ -85,6 +86,11 @@ from zhaquirks.ikea import IKEA, DoublingPowerConfig2AAACluster, ScenesCluster
                     "param2": 0,
                 },
             },
+            (LONG_RELEASE, LEFT): {
+                COMMAND: COMMAND_RELEASE,
+                CLUSTER_ID: 5,
+                ENDPOINT_ID: 1,
+            },
             (SHORT_PRESS, RIGHT): {
                 COMMAND: COMMAND_PRESS,
                 CLUSTER_ID: 5,
@@ -103,6 +109,11 @@ from zhaquirks.ikea import IKEA, DoublingPowerConfig2AAACluster, ScenesCluster
                     "param1": 3328,
                     "param2": 0,
                 },
+            },
+            (LONG_RELEASE, RIGHT): {
+                COMMAND: COMMAND_RELEASE,
+                CLUSTER_ID: 5,
+                ENDPOINT_ID: 1,
             },
         }
     )


MattWestb:
The reason is that the buttons long pressed is being used for resetting / sync the lights the controller is controlling. Its sending one sequence of commands and is being on large mess getting it working "clean" and not triggering other device automatons.
Also its build of the old 5 buttons remote that is doing the same reset sequence.

Was sniffing the initial released firmware and some more later ones but its little messy then  we dont knowing what magic IKEA is using (attributes and commands from there custom cluster).

Keep in mind its one light controller that have 3 different working mode (Direct with lights ala ZLL, with ZBB ala ZLL / ZB3 and real Zigbee 3.01 with bonding of devices / groups) with some custom things added but its very Zigbee 3 (perhaps the most compatible on the market).

If you is getting it working well then its great and i think many users like have that function in there system.

0lento:
> The reason is that the buttons long pressed is being used for resetting / sync the lights the controller is controlling. Its sending one sequence of commands and is being on large mess getting it working "clean" and not triggering other device automatons. Also its build of the old 5 buttons remote that is doing the same reset sequence.

This is a good point. It actually does that and I forgot that I've worked around it on my automations by adding if condition to relevant places.

I use left/right to control amp volume, so I have separate automation that sets boolean when you hold the left/right and other automation that turns this bool off when you release the left/right buttons. Now on automations that use the "turn on" short press, I have condition check that this left/right hold boolean is off, it would indeed otherwise fire those automations.

But this automation is bit hacky, I have set the boolean also on from "left/right long press released" triggers (in addition to long press), before setting it again off and have delayed check for it on automation that use "Turn on" trigger. I needed the "hold" input boolean for my amp's volume up/down repeat until condition anyway, so it made sense to just use it but this may not be situation for others.

> Keep in mind its one light controller that have 3 different working mode (Direct with lights ala ZLL, with ZBB ala ZLL / ZB3 and real Zigbee 3.01 with bonding of devices / groups) with some custom things added but its very Zigbee 3 (perhaps the most compatible on the market).

Can you use this with ZHA on other than ZB3 mode so this needs to be taken into account?

> If you is getting it working well then its great and i think many users like have that function in there system.

The obvious downside is that mentioned bundling with the other buttons. I don't know this code base well enough if we could make some conditional setup for the quirk to always work right without extra input boolean on automations.

Extra boolean filtering can be done manually with some effort on your automation and lets you use this as intended, but I don't know if this should just work out of the box proper to be exposed/if including this as is would just cause more confusion among users.

0lento:
Example of the log events when you hold Left (or Right) button:

IKEA Remote N2 Remote Button Long Release - Left event was fired with parameters: {'param1': 0}
12:46:21 - 4 seconds ago
IKEA Remote N2 Remote Button Short Press - Turn On event was fired
12:46:22 - 3 seconds ago
IKEA Remote N2 Recall event was fired with parameters: {'group_id': 0x0000, 'scene_id': 0, 'transition_time': 0}
12:46:22 - 3 seconds ago
IKEA Remote N2 Remote Button Long Press - Left event was fired with parameters: {'param1': 3329, 'param2': 0}
12:46:23 - 2 seconds ago
IKEA Remote N2 Remote Button Long Release - Left event was fired with parameters: {'param1': 2630}
12:46:26 - In 1 second

You start with Left Long Release, then you get Turn on short press, then recall, then actual Left Long Press hold event and finally when you release, again Left Long Release.

I don't know if that recall can mess something up but at least that Turn on event is the thing we need to filter out.

0lento:
Initial Long Release - Left event can also be trouble if you rely on this alone. So to make this more robust, one would need to track if the long press was even active before release, and in addition filter the Short Press - Turn On event for brief period after the release event.

MattWestb:
> > Keep in mind its one light controller that have 3 different working mode (Direct with lights ala ZLL, with ZBB ala ZLL / ZB3 and real Zigbee 3.01 with bonding of devices / groups) with some custom things added but its very Zigbee 3 (perhaps the most compatible on the market).
> 
> Can you use this with ZHA on other than ZB3 mode so this needs to be taken into account?
> 

The controller is in "normal Zigbee 3 mode" in ZHA and the other modes is only used then direct pairing lights or with the old TF hub.
Its pure Zigbee commands but if mixing DA with bonded devices it and being very messy and is not  recommended.
The great thing is then binding lights to it its still working then the host system is offline and automatons made in HA is not working at all.

I prefer having lights bonded and ding other automatons that is not critical in HA (i dont like coming home and late and all is black and the light is not working but having one motion sensor plus remote bonded it i have 2 extra levels if ZHA is not working OK).

I knowing how much commands its sending then long pressing and getting the logic for getting it  clean its not easy but do you best and no one can complaining. If users like have more functions then use shortcut buttons or tuya scene switches but then  you cant  controlling lights (its 2 different device types lights /  HA controllers).

0lento:
Realized that COMMAND_RELEASE doesn't actually differentiate left and right buttons, while this works for my use case (I use same input boolean for testing if either left or right is held or released) this PR wouldn't work as general purpose solution, there would need to be more sophisticated preprocessing for these so closing this PR for now.

One could figure out the correct released button from param1 as it seems to be positive value for left button release and negative for right button release but like mentioned, this would need more logic to cover this.

MattWestb:
Look on the Opne/Close remote its have the same but they is saving the last direction and using it for tagging the release events and shall being possible doing the same with  right left long press for you (i cant coding).
https://github.com/zigpy/zha-device-handlers/blob/3bd8498c0849ccff78f8b4d08e3678847a40cd1f/zhaquirks/ikea/opencloseremote.py#L55C1-L76C60

PR #3512, opened 2024-11-13
Add Tuya contact and illuminance sensor `_TZE200_pay2byax`

tr4nt0r:
## Proposed change

This is a custom V2 quirk that I cobbled together and wanted to share. It works in HA, except for battery sensor, which seems to be dysfunctional i think. The dp_id for the battery is 2 btw. 

Maybe someone wants to take this over and implement it correctly. 

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_sensor.py
+++ b/zhaquirks/tuya/ts0601_sensor.py
@@ -1,5 +1,10 @@
 """Tuya temp and humidity sensors."""
 
+from zigpy.quirks.v2.homeassistant import EntityType
+from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
+from zigpy.quirks.v2.homeassistant.sensor import SensorDeviceClass, SensorStateClass
+import zigpy.types as t
+
 from zhaquirks.tuya.builder import TuyaPowerConfigurationCluster2AAA, TuyaQuirkBuilder
 
 (
@@ -68,6 +73,28 @@ from zhaquirks.tuya.builder import TuyaPowerConfigurationCluster2AAA, TuyaQuirkB
     .tuya_temperature(dp_id=5)
     .tuya_battery(dp_id=15)
     .tuya_soil_moisture(dp_id=3)
+    .skip_configuration()
+    .add_to_registry()
+)
+
+(
+    TuyaQuirkBuilder("_TZE200_pay2byax", "TS0601") # Cusam ZG-102ZL
+    .applies_to("_TZE200_n8dljorx",  "TS0601")
+    .tuya_sensor(
+        dp_id=101,
+        attribute_name="measured_value",
+        type=t.uint16_t,
+        fallback_name="Illuminance",
+        device_class=SensorDeviceClass.ILLUMINANCE,
+        state_class=SensorStateClass.MEASUREMENT
+    )
+    .tuya_binary_sensor(
+        dp_id=1,
+        attribute_name="zone_status",
+        device_class=BinarySensorDeviceClass.OPENING,
+        fallback_name="Opening",
+        entity_type=EntityType.STANDARD,
+    )
     .skip_configuration()
     .add_to_registry()
 )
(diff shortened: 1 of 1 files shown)


prairiesnpr:
What are you seeing when you add the battery line? Sometimes, on battery powered devices, it can take awhile to report the value.

tr4nt0r:
@prairiesnpr it always shows 200%. But the value doesn't change. There are some older discussions were people tried getting it working but all had the same problem, even after weeks, the value stayed the same.

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -71,3 +76,25 @@
...
+(
+    TuyaQuirkBuilder("_TZE200_pay2byax", "TS0601")  # Cusam ZG-102ZL
+    .applies_to("_TZE200_n8dljorx", "TS0601")
+    .tuya_sensor(
+        dp_id=101,
+        attribute_name="measured_value",
+        type=t.uint16_t,
+        fallback_name="Illuminance",
+        device_class=SensorDeviceClass.ILLUMINANCE,
+        state_class=SensorStateClass.MEASUREMENT,
```
prairiesnpr, line 89: Need `translation_key`

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -71,3 +76,25 @@
...
+    )
+    .tuya_binary_sensor(
+        dp_id=1,
+        attribute_name="zone_status",
+        device_class=BinarySensorDeviceClass.OPENING,
+        fallback_name="Opening",
+        entity_type=EntityType.STANDARD,
```
prairiesnpr, line 96: Also need `translation_key` here

zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -71,3 +76,25 @@
     .skip_configuration()
     .add_to_registry()
 )
+
+(
+    TuyaQuirkBuilder("_TZE200_pay2byax", "TS0601")  # Cusam ZG-102ZL
+    .applies_to("_TZE200_n8dljorx", "TS0601")
```
prairiesnpr, line 82: I'd lean towards leaving `tuya_battery` in, from the incident, the device does report the value on DP 2 and that's what z2m is doing also.

zhaquirks/tuya/ts0601_sensor.py
tr4nt0r: Translation key is optional if a device class is set.  HA already has translations and icons for illiminance and it is working perfectly fine.
image

zhaquirks/tuya/ts0601_sensor.py
TheJulianJES: Not required here due to the same reason as [here](https://github.com/zigpy/zha-device-handlers/pull/3512/files#r1842444759). Zigpy also validates the `translation_key`, but as there's a `device_class` provided here, HA will use that.

TheJulianJES (approved):
Thanks!

TheJulianJES:
Does the device expose the default illuminance light sensor? So, does the device have the ZCL Illuminance Measurement cluster?


Regarding:
- https://github.com/home-assistant/core/issues/132669#issue-2726376918

tr4nt0r:
Nope, there is no illuminance cluster. I would try to fix the issues but I'm not able to override the quirk with a custom quirk.

TheJulianJES:
We're tracking the custom quirks issue in https://github.com/zigpy/zigpy/issues/1508
There's also a workaround in that issue that you can paste in your custom quirk. It'll then remove the built-in one first, so the custom quirk doesn't conflict.

tr4nt0r:
The reason the illuminance sensor gets duplicated may be from previously using a custom quirk. I was using a custom V1 quirk too. But the sensor is no longer provided by zha and after a reboot it was possible to delete the old illuminance sensor.

For the contact sensor, it is indeed available twice. This may come from the IasZone cluster that the device exposes but that sensor doesn't work, it will always show closed. The binary sensor I added in the quirk should have replaced it. 

The illuminance sensor does not react to lights because it sends it's state only every 30min or if the state of the contact sensor changes the illuminance also gets updated. I'm not sure if this is something that can be improved, but it was also like that with the old v1 quirk.

TheJulianJES:
We can "remove" the entire `IasZone` cluster if it doesn't really work. ZHA won't create the sensor then:
```python
.removes(IasZone.cluster_id)
```

tr4nt0r:
I found out how to extend the IasZone :)

- https://github.com/zigpy/zha-device-handlers/pull/3605

PR #3519, opened 2024-11-17
Use SPDX license identifier

cdce8p:
## Proposed change
Use official SPDX license identifier as recommended by [PEP 639](https://peps.python.org/pep-0639/).
https://spdx.org/licenses/Apache-2.0.html


TheJulianJES (approved):
Thanks!

PR #3521, opened 2024-11-18
Add Aqara FP1E presence sensor v2 quirk

### LINKED ISSUE #3294: [Device Support Request] Aqara FP1e (lumi.sensor_occupy.agl1)
### Problem description

Everything on this device is implemented on the apple cluster 0xfcc0

### Solution description

Implement the proper quirk providing support for the device

### Device signature

```json
{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 140,
    "manufacturer_code": 4660,
    "maximum_buffer_size": 108,
    "maximum_incoming_transfer_size": 127,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 127,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0xfff0",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0xfcc0"
      ],
      "output_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "aqara",
  "model": "lumi.sensor_occupy.agl1",
}
```

</details>

jeverley:
## Proposed change

This quirk adds support for the Aqara FP1E presence sensor, exposing its motion and occupancy sensors in addition to its  configuration parameters.

Fixes https://github.com/zigpy/zha-device-handlers/issues/3294 and is written using the quirks v2 format.

image

## Additional information

There are two issues that we'd ideally want to fix before merging this PR.

- https://github.com/zigpy/zha/issues/290
   - _Currently unable to specify the unit for the distance sensors_
   - **PR raised:** https://github.com/zigpy/zha/pull/307
- https://github.com/zigpy/zha/issues/291
   - _I'd like to make the _Restart device_ button a diagnostic rather than config function_
   - **PR raised:** https://github.com/zigpy/zha/pull/308

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/motion_agl1.py
+++ b/zhaquirks/xiaomi/aqara/motion_agl1.py
@@ -15,7 +15,7 @@ from zigpy.quirks.v2 import (
 from zigpy.zcl.clusters.general import DeviceTemperature
 from zigpy.zcl.clusters.measurement import OccupancySensing
 from zigpy.zcl.clusters.security import IasZone
-from zigpy.zcl.foundation import ZCLAttributeDef
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 from zhaquirks.xiaomi import XiaomiAqaraE1Cluster
 
@@ -54,8 +54,8 @@ class AqaraMotion(types.enum8):
 class OppleCluster(XiaomiAqaraE1Cluster):
     """Aqara manufacturer cluster for the FP1E presence sensor."""
 
-    class AttributeDefs(XiaomiAqaraE1Cluster.AttributeDefs):
-        """Aqara occupancy sensor manufacturer specific attributes."""
+    class AttributeDefs(BaseAttributeDefs):
+        """Manufacturer specific attributes."""
 
         approach_distance = ZCLAttributeDef(
             id=APPROACH_DISTANCE_ATTR_ID,
--- a/zhaquirks/xiaomi/aqara/motion_agl1.py
+++ b/zhaquirks/xiaomi/aqara/motion_agl1.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import logging
 from typing import Any
 
 from zigpy import types
@@ -29,8 +28,6 @@ RESTART_DEVICE_ATTR_ID = 0x00E8  # BOOL Trigger device restart
 
 RESET_NO_PRESENCE_STATUS_WRITE_VALUE = 1
 RESTART_DEVICE_WRITE_VALUE = 0
-
-_LOGGER = logging.getLogger(__name__)
 
 
 class AqaraMotionSensitivity(types.enum8):
--- a/zhaquirks/xiaomi/aqara/motion_agl1.py
+++ b/zhaquirks/xiaomi/aqara/motion_agl1.py
@@ -5,14 +5,13 @@ from __future__ import annotations
 import logging
 from typing import Any
 
+from zigpy import types
 from zigpy.quirks.v2 import (
-    QuirkBuilder,
     NumberDeviceClass,
+    QuirkBuilder,
     SensorDeviceClass,
     SensorStateClass,
 )
-from zigpy.quirks.v2.homeassistant import EntityType, UnitOfLength
-import zigpy.types as types
 from zigpy.zcl.clusters.general import DeviceTemperature
 from zigpy.zcl.clusters.measurement import OccupancySensing
 from zigpy.zcl.clusters.security import IasZone
--- a/zhaquirks/xiaomi/aqara/motion_agl1.py
+++ b/zhaquirks/xiaomi/aqara/motion_agl1.py
@@ -20,35 +20,32 @@ from zhaquirks import LocalDataCluster
 from zhaquirks.xiaomi import XiaomiAqaraE1Cluster
 
 
-class AqaraMotionSensitivity(types.enum8):
-    """Aqara motion sensitivity."""
-
-    Low = 0x01
-    Medium = 0x02
-    High = 0x03
-
-
 class AqaraMotion(types.enum8):
-    """Aqara motion."""
-
-    Unknown_0 = 0x00
-    Unknown_1 = 0x01
     Idle = 0x02
     Moving = 0x03
     Still = 0x04
 
 
+class AqaraMotionSensitivity(types.enum8):
+    Low = 0x01
+    Medium = 0x02
+    High = 0x03
+
+
+class AqaraOccupancy(types.enum8):
+    Unoccupied = 0x00
+    Occupied = 0x01
+
+
 class IasZoneLocal(LocalDataCluster, IasZone):
     """Virtual cluster for IasZone."""
 
-    # required to make sure ZHA creates sensors when initially pairing
     _VALID_ATTRIBUTES = {IasZone.AttributeDefs.zone_status.id}
 
 
 class OccupancySensingLocal(LocalDataCluster, OccupancySensing):
     """Virtual cluster for OccupancySensing."""
 
-    # required to make sure ZHA creates sensors when initially pairing
     _VALID_ATTRIBUTES = {OccupancySensing.AttributeDefs.occupancy.id}
 
 
@@ -66,10 +63,10 @@ class OppleCluster(XiaomiAqaraE1Cluster):
             is_manufacturer_specific=True,
         )
 
-        # Detected motion (0x02 = no movement, 0x03 = large movement, 0x04 = small movement)
+        # Detected motion
         motion = ZCLAttributeDef(
             id=0x0160,
-            type=types.uint8_t,
+            type=AqaraMotion,
             access="rp",
             is_manufacturer_specific=True,
         )
@@ -82,18 +79,18 @@ class OppleCluster(XiaomiAqaraE1Cluster):
             is_manufacturer_specific=True,
         )
 
-        # The configurable detection sensitivity (0x01 = low, 0x02 = medium, 0x03 = high, default 0x03 = high)
+        # The configurable detection sensitivity
         motion_sensitivity = ZCLAttributeDef(
             id=0x010C,
-            type=types.uint8_t,
+            type=AqaraMotionSensitivity,
             access="rw",
             is_manufacturer_specific=True,
         )
 
-        # Occupancy detected (0x0 = absence, 0x01 = presence)
+        # Occupancy detected
         occupancy = ZCLAttributeDef(
             id=0x0142,
-            type=types.uint8_t,
+            type=AqaraOccupancy,
             access="rp",
             is_manufacturer_specific=True,
         )
@@ -119,14 +116,14 @@ class OppleCluster(XiaomiAqaraE1Cluster):
         if attrid == self.AttributeDefs.occupancy.id:
             self.endpoint.occupancy.update_attribute(
                 OccupancySensing.AttributeDefs.occupancy.id,
-                OccupancySensing.Occupancy(value),
+                OccupancySensing.Occupancy.Occupied
+                if value == AqaraOccupancy.Occupied
+                else OccupancySensing.Occupancy.Unoccupied,
             )
         elif attrid == self.AttributeDefs.motion.id:
             self.endpoint.ias_zone.update_attribute(
                 IasZone.AttributeDefs.zone_status.id,
-                IasZone.ZoneStatus(
-                    IasZone.ZoneStatus.Alarm_1 if value == AqaraMotion.Moving else 0
-                ),
+                IasZone.ZoneStatus.Alarm_1 if value == AqaraMotion.Moving else 0,
             )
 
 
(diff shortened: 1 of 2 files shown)


jeverley:
Hi @dmulcahey, do you have any feedback on the current implementation with regards to quirk v2 conventions?

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,179 @@
...
+        restart_device = ZCLAttributeDef(
+            id=RESTART_DEVICE_ATTR_ID,
+            type=types.Bool,
+            access="w",
+            is_manufacturer_specific=True,
+        )
+
+    def _update_attribute(self, attrid: int, value: Any) -> None:
+        super()._update_attribute(attrid, value)
+        if attrid == OCCUPANCY_ATTR_ID:
```
TheJulianJES: This would be more consistent with our other (new) quirks:
```suggestion
        if attrid == self.AttributeDefs.occupancy.id:
```
(or: `OppleCluster.AttributeDefs.occupancy.id`)

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,179 @@
...
+            self.endpoint.occupancy.update_attribute(
+                OccupancySensing.AttributeDefs.occupancy.id,
+                OccupancySensing.Occupancy(value),
+            )
+        elif attrid == MOTION_ATTR_ID:
```
TheJulianJES: and same here:
```suggestion
        elif attrid == self.AttributeDefs.motion.id:
```

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,179 @@
...
+    QuirkBuilder("aqara", "lumi.sensor_occupy.agl1")
+    .friendly_name(model="Presence Sensor FP1E", manufacturer="Aqara")
+    .adds(DeviceTemperature)
+    .adds(OccupancySensing)
+    .adds(
+        IasZone,
+        constant_attributes={
+            IasZone.AttributeDefs.zone_type: IasZone.ZoneType.Motion_Sensor
+        },
+    )
```
TheJulianJES: If these clusters are completely virtual/fake (not present on the device), we'd probably want them to be a `LocalDataCluster`. E.g.:

```python
class IasZoneLocal(LocalDataCluster, IasZone):
    """Virtual cluster for IasZone."""

    # required to make sure ZHA creates sensors when initially pairing
    _VALID_ATTRIBUTES = {IasZone.AttributeDefs.zone_status.id}
```

`_VALID_ATTRIBUTES` is probably required to make sure that ZHA creates the sensors when the device is initially paired (so completely removed from ZHA before pairing again), even when there's no initial data in the zigpy DB (which there is not, as the attribute cannot be read on pairing).

TheJulianJES: It might be a good idea for us to provide a shorthand v2 quirk method for this. But that does't exist yet..

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,179 @@
...
+
+from zhaquirks.xiaomi import XiaomiAqaraE1Cluster
+
+APPROACH_DISTANCE_ATTR_ID = 0x015B  # UINT32 The configurable maximum detection distance in millimeters (default 600 = 6 meters).
+MOTION_ATTR_ID = 0x0160  # UINT8 Detected motion (0x02 = no movement, 0x03 = large movement, 0x04 = small movement)
+MOTION_DISTANCE_ATTR_ID = 0x015F  # UINT32 Distance to the detected motion (mm)
+MOTION_SENSITIVITY_ATTR_ID = 0x010C  # UINT8 The configurable detection sensitivity (0x01 = low, 0x02 = medium, 0x03 = high, default 0x03 = High)
+OCCUPANCY_ATTR_ID = 0x0142  # UINT8 Occupancy detected 0x0 = absence, 0x01 = presence
+RESET_NO_PRESENCE_STATUS_ATTR_ID = 0x0157  # UINT8 Trigger AI spatial learning
+RESTART_DEVICE_ATTR_ID = 0x00E8  # BOOL Trigger device restart
```
TheJulianJES: We can probably remove these constants (define them directly in the `ZCLAttributeDef`) and move the comments above the respective `ZCLAttributeDef`.

jeverley: Agreed that makes sense, the ZCLAttributeDefs allow a much more consistent approach.
I've addressed this in https://github.com/zigpy/zha-device-handlers/pull/3521/commits/0636e87c22e7c247ad57ac9c690953e511745f72.

zhaquirks/xiaomi/aqara/motion_agl1.py
jeverley: Thanks for the feedback @TheJulianJES , updated with https://github.com/zigpy/zha-device-handlers/pull/3521/commits/0636e87c22e7c247ad57ac9c690953e511745f72 to follow this recommendation.

zhaquirks/xiaomi/aqara/motion_agl1.py
jeverley: > It might be a good idea for us to provide a shorthand v2 quirk method for this. But that does't exist yet..

Yes I think that would be a great addition, in the meantime I've updated the quirk to use Local clusters for motion and occupancy.

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,194 @@
...
+(
+    QuirkBuilder("aqara", "lumi.sensor_occupy.agl1")
+    .friendly_name(model="Presence Sensor FP1E", manufacturer="Aqara")
+    .adds(DeviceTemperature)
+    .adds(OccupancySensingLocal)
+    .adds(
+        IasZoneLocal,
+        constant_attributes={
+            IasZone.AttributeDefs.zone_type: IasZone.ZoneType.Motion_Sensor
+        },
```
TheJulianJES: I think I remember someone complaining about `constant_attributes` defined via quirks v2 not always working as expected. And I might see why: zigpy adds the attribute name as the key, not the attribute id. This might work in some places, but is unexpected in others. Zigpy code for reference: [zigpy/quirks/v2/__init__.py#L178-L182](https://github.com/zigpy/zigpy/blob/8e9f5163ef93ff71fa0ce4992076d8bc248eac44/zigpy/quirks/v2/__init__.py#L178-L182)

Until this is fixed in zigpy, let's remove this here and add it in the cluster class above via the attribute id in `_CONSTANT_ATTRIBUTES`. We're defining `_VALID_ATTRIBUTES` there anyway, so they're also together then.
We'll want to update this again when we add valid attributes as a parameters here for quirks v2 (and fix the constant attributes issue).
```suggestion
```

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,194 @@
...
+class AqaraOccupancy(types.enum8):
+    """Aqara occupancy attribute values."""
+
+    Unoccupied = 0x00
+    Occupied = 0x01
+
+
+class IasZoneLocal(LocalDataCluster, IasZone):
+    """Virtual cluster for IasZone."""
+
```
TheJulianJES, line 49: So let's add the zone type as a constant attribute here, per the previous comment.
```suggestion

    _CONSTANT_ATTRIBUTES = {IasZone.AttributeDefs.zone_type.id: IasZone.ZoneType.Motion_Sensor}
```

TheJulianJES (commented):
This looks very good.
I'll make a small change regarding the `constant_attributes`. There's [more info ](https://github.com/zigpy/zha-device-handlers/pull/3521#discussion_r1857687985)in the two comments. Otherwise, this should be good to go.

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,194 @@
...
+
+(
+    QuirkBuilder("aqara", "lumi.sensor_occupy.agl1")
+    .friendly_name(model="Presence Sensor FP1E", manufacturer="Aqara")
+    .adds(DeviceTemperature)
+    .adds(OccupancySensingLocal)
+    .adds(
+        IasZoneLocal,
+    )
```
TheJulianJES, line 147:
```suggestion
    .adds(IasZoneLocal)
```

TheJulianJES:
I've also added a small test to cover the `_update_attribute` method: https://github.com/zigpy/zha-device-handlers/pull/3521/commits/f0d075478de26c6200023ec4a1971a3e9e1b984c

TheJulianJES (approved):
This should be good to go IMO. Thanks, @jeverley!

zhaquirks/xiaomi/aqara/motion_agl1.py
jeverley: Good catch, thanks for making that change.

jeverley:
> I've also added a small test to cover the `_update_attribute` method: [f0d0754](https://github.com/zigpy/zha-device-handlers/commit/f0d075478de26c6200023ec4a1971a3e9e1b984c)

Thanks for adding these, I was still wrapping my head around how best to introduce a v2 quirk amongst the other xiaomi motion devices.
I think we need to make a small tweak to the expected statuses for motion in here, possible values being:

```
class AqaraMotion(types.enum8):
    """Aqara motion attribute values."""

    Idle = 0x02 # no presence
    Moving = 0x03 # present and moving
    Still = 0x04 # present and not moving
```
Therefore the parameters should be (if I'm understanding the test syntax correctly):

```
@pytest.mark.parametrize(
    "occupancy_value, expected_occ_status, motion_value, expected_motion_status",
    [
        (0, OccupancySensing.Occupancy.Unoccupied, 2, 0),
        (1, OccupancySensing.Occupancy.Occupied, 3, IasZone.ZoneStatus.Alarm_1),
        (1, OccupancySensing.Occupancy.Occupied, 4, 0),
    ],
)

```
Will add a commit reflecting this shortly.

Edit: I see you already corrected the flipped occupied/unoccupied values in https://github.com/zigpy/zha-device-handlers/pull/3521/commits/45c49fc50091eabba0981387ca104624cb8e9be5, still good to capture the 3rd scenario of occupied with no movement.

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,192 @@
...
+        elif attrid == self.AttributeDefs.motion.id:
+            self.endpoint.ias_zone.update_attribute(
+                IasZone.AttributeDefs.zone_status.id,
+                IasZone.ZoneStatus.Alarm_1 if value == AqaraMotion.Moving else 0,
+            )
+
+
+(
+    QuirkBuilder("aqara", "lumi.sensor_occupy.agl1")
+    .friendly_name(model="Presence Sensor FP1E", manufacturer="Aqara")
```
TheJulianJES: I know it's `model, manufacturer` in zigpy, but let's do it the other way around to match the order of `QuirkBuilder`.
Since they are keyword arguments here, the order technically doesn't matter. We should still consider swapping it around in zigpy to align with the typical `manufacturer, model` order though.
```suggestion
    .friendly_name(manufacturer="Aqara", model="Presence Sensor FP1E")
```

zhaquirks/xiaomi/aqara/motion_agl1.py
```diff
@@ -0,0 +1,192 @@
...
+    Occupied = 0x01
+
+
+class IasZoneLocal(LocalDataCluster, IasZone):
+    """Virtual cluster for IasZone."""
+
+    _CONSTANT_ATTRIBUTES = {
+        IasZone.AttributeDefs.zone_type.id: IasZone.ZoneType.Motion_Sensor
+    }
+    _VALID_ATTRIBUTES = {IasZone.AttributeDefs.zone_status.id}
```
TheJulianJES, line 53: Note for other reviewers: We need a virtual cluster for `IasZone` here, as we need a virtual attribute for "motion". The Xiaomi `motion` attribute needs to be parsed in a way that we can't yet do with quirks v2.

Also, for motion/occupancy, it makes sense to use the integrated entities in ZHA IMO. We'd have to duplicate "motion" and "occupancy" quirk entities with all properties everywhere otherwise.

We already use virtual clusters in a lot of quirks and they'll always be needed for some (especially for Tuya and some Xiaomi devices), unless we have some kind of other structure we can map attributes to (like Z2M does).

TheJulianJES (approved):

TheJulianJES:
Looks like Codecov is broken across all repos again. This PR is fully covered.
Also noting that this quirk will require the latest ZHA version. You can't install it on a HA Core 2024.11.x system (or earlier) anymore.

PR #3529, opened 2024-11-20
Migrate Schneider dimmer quirks to v2, add missing models

uvNikita:
## Proposed change
I migrated quirks to v2 and added models other models from the spec.
Looking through the spec docs, they all seem to have equivalent functionality.
The original quirk was missing SEOnOff cluster, so I added it too.

## Additional information
Specs:
[ZB Spec - Push-button Dimmer - 110422.pdf](https://github.com/user-attachments/files/17831848/ZB.Spec.-.Push-button.Dimmer.-.110422.pdf)
[ZB Spec - Rotary Dimmer - 110422.pdf](https://github.com/user-attachments/files/17831849/ZB.Spec.-.Rotary.Dimmer.-.110422.pdf)
[ZB Spec - Relay Switch 10A - 110422.pdf](https://github.com/user-attachments/files/17832116/ZB.Spec.-.Relay.Switch.10A.-.110422.pdf)

## Checklist

### DIFF
--- a/zhaquirks/schneiderelectric/dimmers.py
+++ b/zhaquirks/schneiderelectric/dimmers.py
@@ -19,7 +19,7 @@ from zhaquirks.schneiderelectric import (
     .replaces(SEBallast, endpoint_id=3)
     .replaces(SEOnOff, endpoint_id=3)
     .replaces(SEBasic, endpoint_id=21)
-    .adds(SESpecific, endpoint_id=21)
+    .replaces(SESpecific, endpoint_id=21)
     .add_to_registry()
 )
 
@@ -28,6 +28,6 @@ from zhaquirks.schneiderelectric import (
     .replaces(SEBasic)
     .replaces(SEOnOff)
     .replaces(SEBasic, endpoint_id=21)
-    .adds(SESpecific, endpoint_id=21)
+    .replaces(SESpecific, endpoint_id=21)
     .add_to_registry()
 )
--- a/zhaquirks/schneiderelectric/dimmers.py
+++ b/zhaquirks/schneiderelectric/dimmers.py
@@ -1,129 +1,33 @@
 """Schneider Electric dimmers and switches quirks."""
 
-from zigpy.profiles import zgp, zha
-from zigpy.quirks import CustomDevice
-from zigpy.zcl.clusters.closures import WindowCovering
-from zigpy.zcl.clusters.general import (
-    Basic,
-    GreenPowerProxy,
-    Groups,
-    Identify,
-    LevelControl,
-    OnOff,
-    Ota,
-    Scenes,
+from zigpy.quirks.v2 import QuirkBuilder
+
+from zhaquirks.schneiderelectric import (
+    SE_MANUF_NAME,
+    SEBallast,
+    SEBasic,
+    SEOnOff,
+    SESpecific,
 )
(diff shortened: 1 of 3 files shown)


uvNikita:
Adding the original PRs for the reference #1705 #3111

zhaquirks/schneiderelectric/dimmers.py
```diff
@@ -1,129 +1,33 @@
...
+(
+    QuirkBuilder(SE_MANUF_NAME, "NHROTARY/DIMMER/1")
+    .applies_to(SE_MANUF_NAME, "NHROTARY/UNIDIM/1")
+    .applies_to(SE_MANUF_NAME, "NHPB/DIMMER/1")
+    .applies_to(SE_MANUF_NAME, "NHPB/UNIDIM/1")
+    .replaces(SEBasic, endpoint_id=3)
+    .replaces(SEBallast, endpoint_id=3)
+    .replaces(SEOnOff, endpoint_id=3)
+    .replaces(SEBasic, endpoint_id=21)
+    .adds(SESpecific, endpoint_id=21)
```
TheJulianJES: Shouldn't this also be `.replaces`, as the `SESpecific.cluster_id` was in the signature before?

zhaquirks/schneiderelectric/dimmers.py
```diff
@@ -1,129 +1,33 @@
...
-                ],
-            },
-        }
-    }
+(
+    QuirkBuilder(SE_MANUF_NAME, "NHPB/SWITCH/1")
+    .replaces(SEBasic)
+    .replaces(SEOnOff)
+    .replaces(SEBasic, endpoint_id=21)
+    .adds(SESpecific, endpoint_id=21)
```
TheJulianJES: And same here? Also `.replaces`?

zhaquirks/schneiderelectric/dimmers.py
TheJulianJES: Tbh, I'm not entirely sure why we even have `.adds(cluster)`. I think you always want to remove a cluster with the same `cluster_id` when adding/replacing. If it already exists, `.adds(cluster)` won't do anything IIRC.

TheJulianJES:
```suggestion
    .replaces(SESpecific, endpoint_id=21)
```

zhaquirks/schneiderelectric/dimmers.py
TheJulianJES:
```suggestion
    .replaces(SESpecific, endpoint_id=21)
```

TheJulianJES (approved):
Thanks! I've replaced `adds(SESpecific)` with `replaces(SESpecific)`, as we want to ensure the original cluster is removed first (and replaced with our custom implementation).
There's some more info in the review comments above.

uvNikita:
@TheJulianJES I was just assuming `adds` should be used on Manufacturer Specific clusters and `replaces` on the standard ones.

Will keep that in mind for the future, thanks a lot for the review!

TheJulianJES:
Yeah, we need to get some docs together at some point.. I'm not sure if there's ever any reason to use `adds` instead of `replaces`. We should also take a look at that and remove `adds` if needed 😄

PR #3533, opened 2024-11-21
Add Tuya TRV Moes HY368 variation

hlfbt:
## Proposed change
Adds new variation `_TZE200_znlqjmih` of the Moes HY368 TRV.
The clusters seem to be the exact same as the other/older variations of the same TRV.

## Additional information
Signature as reported in ZHA
```json
{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4098,
    "maximum_buffer_size": 82,
    "server_mask": 11264,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_znlqjmih",
  "model": "TS0601",
}
```

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_trv.py
+++ b/zhaquirks/tuya/ts0601_trv.py
@@ -1534,6 +1534,7 @@ class MoesHY368_Type1(TuyaThermostat):
             ("_TZE200_8whxpsiw", "TS0601"),
             ("_TZE200_8thwkzxl", "TS0601"),  # Tervix Pro Line EVA 2
             ("_TZE200_xby0s3ta", "TS0601"),  # Sandy Beach HY367
+            ("_TZE200_znlqjmih", "TS0601"),
         ],
         ENDPOINTS: {
             1: {


hlfbt:
I'm not sure if the previous quirk works with this new variant actually.
When I copy paste the python script as a custom quirk and add the new signature, ZHA does correctly identify it as a TS0601 device and exposes a Thermostat.
The other exposes sensors and switches seem wrong though. I've never used these radiator valves with ZHA before, so I'm not certain this is the expected behaviour.

This is what the device looks like with the custom quirk:
image
The switch has no label to show its purpose, Pi-Wärmebedarf (Pi heat requirement??) seems out of place, I have no idea what HLK Action means.
Also, the set temperature is initially read as >160°C. The set temperature changes to the correct (?) value after a couple minutes though.

In contrast, this is the exact same valve when paired with Z2M:
image
So Z2M seems to be doing something different.

Michuu:
@hlfbt I can confirm that I'm getting similar effects. I have added the _TZE200_znlqjmih as a custom quirk like in the original post. I cannot set the thermostat to any other mode than "Heat" - the "auto" mode is not available, and the "Off" mode does not trigger properly. The temperature readout seems fine, but we're clearly missing several features here compared with Z2M.

PR #3542, opened 2024-11-24
Let Sinopé thermostat and light quirks inherit from CustomDevice

claudegel:
## Proposed change

- All devices class should inherit from CustomDevice.
- Small fix for SinopeHPThermostats for devices list

## Checklist


TheJulianJES (approved):
Thanks!

PR #3550, opened 2024-11-26
Clean up Tuya Power Config

prairiesnpr:
## Proposed change

Currently `TuyaPowerConfigurationCluster2AA` and `TuyaPowerConfigurationCluster3AA` clusters inherit from `TuyaPowerConfigurationCluster` which creates a battery bus. The other TuyaPowerConfiguration clusters are built off of `TuyaLocalCluster`. This causes confusion when building a quirk via TuyaQuirkBuilder since it will fail if either of the two legacy power clusters are used.

This PR moves the power config clusters to tuya.power_config, adds 2AA and 3AA, and marks the two legacy 2AA and 3AA as Legacy.

A follow-on PR should convert the ts0601_trv quirk to use the newer style, remove the legacy clusters, and remove the import and re-export of the 2AAA and 4AA clusters from tuya/__init__.py.

## Checklist

### DIFF
--- a/tests/test_tuya_builder.py
+++ b/tests/test_tuya_builder.py
@@ -13,7 +13,6 @@ from zigpy.zcl.clusters.general import Basic
 from tests.common import ClusterListener, wait_for_zigpy_tasks
 import zhaquirks
 from zhaquirks.tuya.builder import (
-    TuyaPowerConfigurationCluster2AAA,
     TuyaQuirkBuilder,
     TuyaRelativeHumidity,
     TuyaSoilMoisture,
@@ -21,6 +20,7 @@ from zhaquirks.tuya.builder import (
     TuyaValveWaterConsumed,
 )
 from zhaquirks.tuya.mcu import TuyaMCUCluster, TuyaOnOffNM
+from zhaquirks.tuya.power_config import TuyaPowerConfigurationCluster2AAA
 
 from .async_mock import sentinel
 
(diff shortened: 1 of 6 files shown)


TheJulianJES (commented):
For the `TuyaPowerConfigurationCluster`, we could just check if the `battery_bus` is available on the device. It's created by the quirk/`CustomDevice` class for the few devices that still use it.

E.g.
```python
class TuyaPowerConfigurationCluster(PowerConfiguration, TuyaLocalCluster):
    """PowerConfiguration cluster for battery-operated thermostats."""

        """Init."""
        if self.endpoint.device.battery_bus:
            self.endpoint.device.battery_bus.add_listener(self)
```

TheJulianJES (commented):
I think I'd also prefer to not have both new and legacy clusters. We should remove the `battery_bus` at some point, but I bet some custom Tuya TRV quirks rely on it.
Only modifying the code like above might be enough for now..

Moving everything and needing backwards-compatible imports isn't great if we can avoid it. If we move or rename it again in the future, it would become a mess.
Personally, I'd just prefer if we can modify the `TuyaPowerConfigurationCluster` cluster like above, so it works with new quirks not defining the `battery_bus` and old quirks that do.

The other power configuration clusters can stay `PowerConfiguration, TuyaLocalCluster` for now. We don't need the legacy `battery_bus` implementation for those.

prairiesnpr:
Per above discussion, closing this.

PR #3563, opened 2024-11-29
Fix missing unit in _TZE200_pay2byax quirk

tr4nt0r:
## Proposed change

Adds unit lx to _TZE200_pay2byax quirk

Requires:
- https://github.com/zigpy/zha/issues/326 

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_sensor.py
+++ b/zhaquirks/tuya/ts0601_sensor.py
@@ -1,6 +1,6 @@
 """Tuya temp and humidity sensors."""
 
-from zigpy.quirks.v2.homeassistant import EntityType
+from zigpy.quirks.v2.homeassistant import LIGHT_LUX, EntityType
 from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
 from zigpy.quirks.v2.homeassistant.sensor import SensorDeviceClass, SensorStateClass
 import zigpy.types as t
@@ -89,6 +89,7 @@ from zhaquirks.tuya.builder import TuyaPowerConfigurationCluster2AAA, TuyaQuirkB
         fallback_name="Illuminance",
         device_class=SensorDeviceClass.ILLUMINANCE,
         state_class=SensorStateClass.MEASUREMENT,
+        unit=LIGHT_LUX
     )
     .tuya_binary_sensor(
         dp_id=1,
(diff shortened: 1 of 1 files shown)


TheJulianJES:
Whilst the underlying issue with some units being invalid according to ZHA isn't fixed, we replaced this custom sensor with one mapping to the ZCL attribute, right? In that case, we can close this?

tr4nt0r:
@TheJulianJES wasn't that the contact sensor?

TheJulianJES:
Oh, yeah 👍

tr4nt0r:
@TheJulianJES I tried doing the same with illluminance, it adds a `TuyaIlluminance` cluster, the `measured_value` attribute shows correctly the reading from the device. It also adds the sensor entity correctly with lx as unit, but the sensor value always shows 1, don't know why. Can you see what I'm doing wrong? 

```
class TuyaIlluminance(IlluminanceMeasurement, TuyaLocalCluster):
    """Tuya local illuminance cluster."""

    _CONSTANT_ATTRIBUTES = {
        IlluminanceMeasurement.AttributeDefs.light_sensor_type.id: IlluminanceMeasurement.LightSensorType.Photodiode
    }
```
in the quirk builder:
```

    def tuya_illuminance(
        self,
        dp_id: int,
        illuminance_cfg: TuyaLocalCluster = TuyaIlluminance,
        converter: Optional[Callable[[Any], Any]] = None,
    ) -> QuirkBuilder:
        """Add a Tuya Illuminance Configuration."""
        self.tuya_dp(
            dp_id,
            illuminance_cfg.ep_attribute,
            IlluminanceMeasurement.AttributeDefs.measured_value.name,
            converter=converter,
        )
        self.adds(illuminance_cfg)
        return self

```

the device
```
(
    CustomTuyaQuirkBuilder("_TZE200_pay2byax", "TS0601")
    .applies_to("_TZE200_n8dljorx",  "TS0601")
    .tuya_illuminance(dp_id=101)
    .tuya_contact(dp_id=1)
    .tuya_battery(dp_id=2)
    .skip_configuration()
    .add_to_registry()
)
```

TheJulianJES:
We use this for the Tuya motion sensors: https://github.com/zigpy/zha-device-handlers/blob/e61ecbb4f64291c663fd19d23f95f7db5d33e1f0/zhaquirks/tuya/ts0601_motion.py#L201-L206

The `converter` is needed because ZHA also converts the value back from the ZCL spec to the lx value used in HA: https://github.com/zigpy/zha/blob/a833f43c0c3507c62b4331ae102e924e86bbb96c/zha/application/platforms/sensor/__init__.py#L831-L837

tr4nt0r:
@TheJulianJES that wasn't it 😕

But i figured out that it is showing a binary value, 1 light, 0 no light. Any idea were that comes from?

TheJulianJES:
Hmm, but it works correctly with the custom quirks v2 sensor? What are some values of that entity (so the ones that are published to the Tuya datapoint)?

tr4nt0r:
I got it working, I will open a new PR

PR #3564, opened 2024-11-29
Refactor Aqara remote switch H1 quirk to v2 format to support all variants

jeverley:
## Proposed change
This PR refactors the existing Aqara H1 quirk into the new v2 structure (originally undertaken as I'd encountered a new device variant for the double switch):

```
"signature": {
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4447,
    "maximum_buffer_size": 127,
    "maximum_incoming_transfer_size": 100,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 100,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0103",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003"
      ],
      "output_clusters": [
        "0x0003",
        "0x0006"
      ]
    },
    "2": {
      "profile_id": "0x0104",
      "device_type": "0x0103",
      "input_clusters": [
        "0x0003"
      ],
      "output_clusters": [
        "0x0003",
        "0x0006"
      ]
    },
    "3": {
      "profile_id": "0x0104",
      "device_type": "0x0103",
      "input_clusters": [
        "0x0003"
      ],
      "output_clusters": [
        "0x0006"
      ]
    },
    "4": {
      "profile_id": "",
      "device_type": "",
      "input_clusters": [],
      "output_clusters": []
    },
    "5": {
      "profile_id": "",
      "device_type": "",
      "input_clusters": [],
      "output_clusters": []
    },
    "6": {
      "profile_id": "",
      "device_type": "",
      "input_clusters": [],
      "output_clusters": []
    }
  },
  "manufacturer": "LUMI",
  "model": "lumi.remote.b28ac1"
},
```

## Additional information

I've two switches, one of which seems to be intermittently affected by the issue https://github.com/home-assistant/core/issues/127874 whereby it's recognised and works correctly after initial pairing, but following a restart of the ZHA integration or Home Assistant it's recognised as unk_manufacturer, unk_model and no longer functions.

Update:
I haven't encountered the issue above in the last week with the new v2 quirk (following multiple reboots and zha reloads).

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/remote_h1.py
+++ b/zhaquirks/xiaomi/aqara/remote_h1.py
@@ -1,7 +1,7 @@
 """Aqara H1-series wireless remote."""
 
+from zigpy import types
 from zigpy.quirks.v2 import ClusterType, QuirkBuilder
-import zigpy.types as t
 from zigpy.zcl.clusters.general import Identify, OnOff
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
@@ -42,14 +42,14 @@ from zhaquirks.xiaomi.aqara.opple_remote import (
 BOTH_BUTTONS = "both_buttons"
 
 
-class AqaraSwitchClickMode(t.enum8):
+class AqaraSwitchClickMode(types.enum8):
     """Aqara switch click mode attribute values."""
 
     Single = 1  # Low latency (50ms) but only sends single click events.
     Multiple = 2  # (default) Slightly higher latency but supports single/double/triple click and long press.
 
 
-class AqaraSwitchOperationMode(t.enum8):
+class AqaraSwitchOperationMode(types.enum8):
     """Aqara switch operation mode attribute values."""
 
     Command = 0
@@ -66,14 +66,14 @@ class AqaraRemoteManuSpecificCluster(XiaomiAqaraE1Cluster):
 
         operation_mode = ZCLAttributeDef(
             id=0x0009,
-            type=t.uint8_t,
+            type=types.uint8_t,
             access="rw",
             is_manufacturer_specific=True,
         )
 
         click_mode = ZCLAttributeDef(
             id=0x0125,
-            type=t.uint8_t,
+            type=types.uint8_t,
             access="rw",
             is_manufacturer_specific=True,
         )
@@ -141,13 +141,18 @@ class PowerConfigurationClusterH1Remote(PowerConfigurationCluster):
         manufacturer="Aqara", model="Wireless Remote Switch H1 (Double Rocker)"
     )
     .adds(AqaraRemoteManuSpecificCluster)
+    .adds(Identify)
+    .adds(Identify, cluster_type=ClusterType.Client)
     .adds(Identify, endpoint_id=2)
     .adds(Identify, endpoint_id=2, cluster_type=ClusterType.Client)
+    .adds(Identify, endpoint_id=3)
     .adds(Identify, endpoint_id=3, cluster_type=ClusterType.Client)
     .adds(MultistateInputCluster)
     .adds(MultistateInputCluster, endpoint_id=2)
     .adds(MultistateInputCluster, endpoint_id=3)
+    .adds(OnOff, cluster_type=ClusterType.Client)
     .adds(OnOff, endpoint_id=2, cluster_type=ClusterType.Client)
+    .adds(OnOff, endpoint_id=3, cluster_type=ClusterType.Client)
     .replaces(PowerConfigurationClusterH1Remote)
     .enum(
         AqaraRemoteManuSpecificCluster.AttributeDefs.click_mode.name,
--- a/zhaquirks/xiaomi/aqara/remote_h1.py
+++ b/zhaquirks/xiaomi/aqara/remote_h1.py
@@ -90,11 +90,11 @@ class PowerConfigurationClusterH1Remote(PowerConfigurationCluster):
 
 (
     QuirkBuilder(LUMI, "lumi.remote.b18ac1")
-    .friendly_name(
-        manufacturer="Aqara", model="Wireless Remote Switch H1 (Single Rocker)"
-    )
-    .adds(AqaraRemoteManuSpecificCluster)
-    .adds(MultistateInputCluster)
+    #    .friendly_name(
+    #        manufacturer="Aqara", model="Wireless Remote Switch H1 (Single Rocker)"
+    #    )
+    .replaces(AqaraRemoteManuSpecificCluster)
+    .replaces(MultistateInputCluster)
     .replaces(PowerConfigurationClusterH1Remote)
     .enum(
         AqaraRemoteManuSpecificCluster.AttributeDefs.click_mode.name,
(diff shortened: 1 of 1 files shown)


jeverley:
I'm re-implementing this as a v2 quirk to expose the configuration options as well as exploring whether this resolves the unknown device bug https://github.com/home-assistant/core/issues/127874 - we shall see.

zhaquirks/xiaomi/aqara/remote_h1.py
```diff
@@ -60,336 +42,157 @@
...
+    .friendly_name(
+        manufacturer="Aqara", model="Wireless Remote Switch H1 (Double Rocker)"
+    )
+    .adds(AqaraRemoteManuSpecificCluster)
+    .adds(Identify)
+    .adds(Identify, cluster_type=ClusterType.Client)
+    .adds(Identify, endpoint_id=2)
+    .adds(Identify, endpoint_id=2, cluster_type=ClusterType.Client)
+    .adds(Identify, endpoint_id=3)
+    .adds(Identify, endpoint_id=3, cluster_type=ClusterType.Client)
```
TheJulianJES: Are we adding all these Identify clusters because the signature reported by the device is sometimes incomplete?
Is there any real function for these on endpoint 2 and 3 (or the client clusters)? If not, I wonder if we can just not add them?

zhaquirks/xiaomi/aqara/remote_h1.py
```diff
@@ -60,336 +42,157 @@
...
-                    Identify.cluster_id,
-                    MultistateInput.cluster_id,
-                ],
-                OUTPUT_CLUSTERS: [OnOff.cluster_id],
+(
+    QuirkBuilder(LUMI, "lumi.remote.b28ac1")
+    .friendly_name(
+        manufacturer="Aqara", model="Wireless Remote Switch H1 (Double Rocker)"
+    )
+    .adds(AqaraRemoteManuSpecificCluster)
```
TheJulianJES: IIRC, if a cluster is ever reported by the device, we should use `.replaces`. It also removes the existing cluster with the same cluster id, should there be one already. It should *not* error out if there is no cluster with that ID.
Then, `replaces` also adds the custom cluster.

I don't see it in the `signature` from the v1 quirks, but I think we should still consider using `replaces` here, as it's a real cluster present on the device, just not exposed in the signature (by the device).
I guess that could theoretically change. With `adds`, we only add the custom cluster if one doesn't exist with the same cluster ID already (again, IIRC). In that case, we would not add the attribute definitions.

I think we should just use `replaces` for this one.

zhaquirks/xiaomi/aqara/remote_h1.py
```diff
@@ -60,336 +42,157 @@
...
+    .adds(Identify)
+    .adds(Identify, cluster_type=ClusterType.Client)
+    .adds(Identify, endpoint_id=2)
+    .adds(Identify, endpoint_id=2, cluster_type=ClusterType.Client)
+    .adds(Identify, endpoint_id=3)
+    .adds(Identify, endpoint_id=3, cluster_type=ClusterType.Client)
+    .adds(MultistateInputCluster)
+    .adds(MultistateInputCluster, endpoint_id=2)
+    .adds(MultistateInputCluster, endpoint_id=3)
```
TheJulianJES: Hmm, I guess the same might apply here. These are real clusters, just not exposed in the signature by the current device firmware.
I'd also consider using `replaces` for these.

zhaquirks/xiaomi/aqara/remote_h1.py
```diff
@@ -60,336 +42,157 @@
...
+    .adds(OnOff, cluster_type=ClusterType.Client)
+    .adds(OnOff, endpoint_id=2, cluster_type=ClusterType.Client)
+    .adds(OnOff, endpoint_id=3, cluster_type=ClusterType.Client)
```
TheJulianJES, line 154: These are probably fine to leave as-is. If they're not exposed by the device signature, we add them. If they are, we don't add them, but it's still the same cluster, as it's the default ZCL `OnOff` one.
I believe these clusters are also used if bound to another device when the remote is in that mode.

zhaquirks/xiaomi/aqara/remote_h1.py
```diff
@@ -60,336 +42,157 @@
...
-            },
-            3: {
-                PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT_SWITCH,
-                INPUT_CLUSTERS: [Identify.cluster_id],
-                OUTPUT_CLUSTERS: [OnOff.cluster_id],
+(
+    QuirkBuilder(LUMI, "lumi.remote.b18ac1")
+    .friendly_name(
+        manufacturer="Aqara", model="Wireless Remote Switch H1 (Single Rocker)"
```
TheJulianJES: Due to the current implementation we have for `friendly_name`, this might break existing blueprints.
I wonder if we should comment it out for now (until the ZHA implementation is changed?).

TheJulianJES (commented):
Nice work!

zhaquirks/xiaomi/aqara/remote_h1.py
jeverley: I've simplified this in https://github.com/zigpy/zha-device-handlers/pull/3564/commits/76158d1927a1bcad8fbe451a2dbeebbd423f1544 to just add the first on endpoint 1 (it seems to be the only one that actually responds).

zhaquirks/xiaomi/aqara/remote_h1.py
jeverley: Agreed, using replace should shield us from future variants which do expose the cluster.
Amended in https://github.com/zigpy/zha-device-handlers/pull/3564/commits/76158d1927a1bcad8fbe451a2dbeebbd423f1544

zhaquirks/xiaomi/aqara/remote_h1.py
jeverley: Some variants to expose the cluster in their signature, definitely correct to use replace.
Amended in https://github.com/zigpy/zha-device-handlers/pull/3564/commits/76158d1927a1bcad8fbe451a2dbeebbd423f1544

zhaquirks/xiaomi/aqara/remote_h1.py
jeverley: Have commented these out for now in https://github.com/zigpy/zha-device-handlers/pull/3564/commits/76158d1927a1bcad8fbe451a2dbeebbd423f1544

zhaquirks/xiaomi/aqara/remote_h1.py
```diff
@@ -60,336 +42,154 @@
...
-                INPUT_CLUSTERS: [
-                    Identify.cluster_id,
-                    MultistateInput.cluster_id,
-                ],
-                OUTPUT_CLUSTERS: [OnOff.cluster_id],
+(
+    QuirkBuilder(LUMI, "lumi.remote.b28ac1")
+    #    .friendly_name(
+    #        manufacturer="Aqara", model="Wireless Remote Switch H1 (Double Rocker)"
+    #    )
```
TheJulianJES: Let's just add a small comment here about why this is commented out. I hope to double check if we can actually add this soon.
```suggestion
    # temporarily commented out due to potentially breaking existing blueprints
    # .friendly_name(
    #     manufacturer="Aqara", model="Wireless Remote Switch H1 (Double Rocker)"
    # )
```

zhaquirks/xiaomi/aqara/remote_h1.py
```diff
@@ -60,336 +42,154 @@
...
-            3: {
-                PROFILE_ID: zha.PROFILE_ID,
-                DEVICE_TYPE: zha.DeviceType.ON_OFF_LIGHT_SWITCH,
-                INPUT_CLUSTERS: [Identify.cluster_id],
-                OUTPUT_CLUSTERS: [OnOff.cluster_id],
+(
+    QuirkBuilder(LUMI, "lumi.remote.b18ac1")
+    #    .friendly_name(
+    #        manufacturer="Aqara", model="Wireless Remote Switch H1 (Single Rocker)"
+    #    )
```
TheJulianJES: Let's just add a small comment here about why this is commented out. I hope to double check if we can actually add this soon.
```suggestion
    # temporarily commented out due to potentially breaking existing blueprints
    # .friendly_name(
    #     manufacturer="Aqara", model="Wireless Remote Switch H1 (Single Rocker)"
    # )
```

TheJulianJES (approved):
Awesome! Let's try this.

PR #3568, opened 2024-11-29
Add NodOn switch options for SIN-4-1-20, SIN-4-1-21, SIN-4-2-20

ikruglov:
## Proposed change
This PR adds the following configuration options for NodOn switches:

SIN-4-1-20, SIN-4-1-21, SIN-4-1-20_PRO
- switch_type (Bistable, Monostable, AutoDetect) 
   This setting is only available with firmware 3.4.0 and older. It allows one to manually configure SwitchType. In principle, NodOn switches auto-detect Toggle vs. Momentary mode. However, the auto-detection mechanism is not always reliable, and this option allows one to override it.

- impulse_mode_duration
  These switches have pulse mode, i.e., they turn off after a configured number of milliseconds. This setting allows one to configure it. 0 (the default value) means deactivation of the impulse mode.

SIN-4-2-20
- switch type (Bistable, Monostable, AutoDetect) 
  Same as above. For both endpoints.

## Additional information
The logic is a port of https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/nodon.ts#L100

This is how it looks in UI:
Screenshot 2024-11-29 at 23 22 38

All UI entities are initially disabled.

## Checklist

### DIFF
--- a/zhaquirks/nodon/switch.py
+++ b/zhaquirks/nodon/switch.py
@@ -6,15 +6,33 @@ from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
 import zigpy.types as t
 from zigpy.zcl.clusters.general import LevelControl
 from zigpy.zcl.clusters.general import OnOff
-from zigpy.zcl.foundation import ZCLAttributeDef
+from zigpy.zcl.foundation import ZCLAttributeDef, DataTypeId
 from zha.units import UnitOfTime
 
 NODON = "NodOn"
 
-class NodOnOnOff(OnOff, CustomCluster):
-    """NodOn custom OnOff cluster"""
+class NodOnSwitchType(t.enum8):
+    """NodOn switch type."""
+
+    Bistable = 0x00
+    Monostable = 0x01
+    AutoDetect = 0x02
+
+
+class OnOffSIN4_1_20(OnOff, CustomCluster):
+    """NodOn custom OnOff cluster for SIN-4-1-20 and alike"""
 
     class AttributeDefs(OnOff.AttributeDefs):
+        """Attribute definitions."""
+
+        """Select the switch type wire to the device. Available from version > V3.4.0"""
+        switch_type = ZCLAttributeDef(
+            id=0x1001,
+            type=NodOnSwitchType,
+            zcl_type=DataTypeId.enum8, # need to explicitly set ZCL type
+            is_manufacturer_specific=True,
+        )
+
         """ Set the impulse duration in milliseconds (set value to 0 to deactivate the impulse mode). """
         impulse_mode_duration = ZCLAttributeDef(
             id=0x0001,
@@ -22,13 +40,39 @@ class NodOnOnOff(OnOff, CustomCluster):
             is_manufacturer_specific=True,
         )
 
+class OnOffSIN4_2_20(OnOff, CustomCluster):
+    """NodOn custom OnOff cluster for SIN-4-2-20 and alike"""
+
+    class AttributeDefs(OnOff.AttributeDefs):
+        """Attribute definitions."""
+
+        """Select the switch type wire to the device. Available from version > V3.4.0"""
+        switch_type = ZCLAttributeDef(
+            id=0x1001,
+            type=NodOnSwitchType,
+            zcl_type=DataTypeId.enum8, # need to explicitly set ZCL type
+            is_manufacturer_specific=True,
+        )
+
 (
     # quirk is similar to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/nodon.ts#L100
     QuirkBuilder(NODON, "SIN-4-1-20")
-    .replaces(NodOnOnOff)
+    .applies_to(NODON, "SIN-4-1-21")
+    .applies_to(NODON, "SIN-4-1-20_PRO")
+    # .filter() TODO this should be applied to firmware 3.5.0+, but I didn't find a way to reliably detect it .
+    #           Please let me know if you know how
+    .replaces(OnOffSIN4_1_20)
+    .enum(
+        attribute_name=OnOffSIN4_1_20.AttributeDefs.switch_type.name,
+        enum_class=NodOnSwitchType,
+        cluster_id=OnOffSIN4_1_20.cluster_id,
+        initially_disabled=True,
+        translation_key="switch_type",
+        fallback_name="Switch type",
+    )
     .number(
-        attribute_name=NodOnOnOff.AttributeDefs.impulse_mode_duration.name,
-        cluster_id=NodOnOnOff.cluster_id,
+        attribute_name=OnOffSIN4_1_20.AttributeDefs.impulse_mode_duration.name,
+        cluster_id=OnOffSIN4_1_20.cluster_id,
         min_value=0,
         max_value=10000,
         step=1,
@@ -44,7 +88,30 @@ class NodOnOnOff(OnOff, CustomCluster):
 (
     # this quirk is a v2 version of 7397b6a
     QuirkBuilder(NODON, "SIN-4-2-20")
+    .applies_to(NODON, "SIN-4-2-20_PRO")
     .removes(cluster_id=LevelControl.cluster_id, endpoint_id=1)
     .removes(cluster_id=LevelControl.cluster_id, endpoint_id=2)
+    # .filter() TODO this should be applied to firmware 3.5.0+, but I didn't find a way to reliably detect it .
+    #           Please let me know if you know how
+    .replaces(OnOffSIN4_2_20, endpoint_id=1)
+    .replaces(OnOffSIN4_2_20, endpoint_id=2)
+    .enum(
+        attribute_name=OnOffSIN4_1_20.AttributeDefs.switch_type.name,
+        enum_class=NodOnSwitchType,
+        cluster_id=OnOffSIN4_1_20.cluster_id,
+        endpoint_id=1,
+        initially_disabled=True,
+        translation_key="switch_type",
+        fallback_name="Switch type",
+    )
+    .enum(
+        attribute_name=OnOffSIN4_1_20.AttributeDefs.switch_type.name,
+        enum_class=NodOnSwitchType,
+        cluster_id=OnOffSIN4_1_20.cluster_id,
+        endpoint_id=2,
+        initially_disabled=True,
+        translation_key="switch_type",
+        fallback_name="Switch type",
+    )
     .add_to_registry()
 )
--- a/zhaquirks/nodon/switch.py
+++ b/zhaquirks/nodon/switch.py
@@ -61,7 +61,7 @@ class OnOffSIN4_2_20(OnOff, CustomCluster):
     QuirkBuilder(NODON, "SIN-4-1-20")
     .applies_to(NODON, "SIN-4-1-21")
     .applies_to(NODON, "SIN-4-1-20_PRO")
-    # .filter() TODO this should be applied to firmware 3.5.0+, but I didn't find a way to reliably detect it .
+    # .filter() TODO this should be applied to firmware 3.4.0+, but I didn't find a way to reliably detect it .
     #           Please let me know if you know how
     .replaces(OnOffSIN4_1_20)
     .enum(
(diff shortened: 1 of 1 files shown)


ikruglov:
One TODO for this PR: `switch_type` is available only after 3.4.0 firmware. I was thinking about using `filter()` from the QuirkBuilder. However, I needed help finding a reliable way to detect the current firmware version. The one that zha/zigpy reports doesn't match the real firmware version. If you know how to address this, please let me know.

TheJulianJES:
ZHA should try and read the attribute during pairing. If it's returned as unsupported, it shouldn't create the select entity.
Do you have a device with the older firmware to verify?

ikruglov:
> ZHA should try and read the attribute during pairing. If it's returned as unsupported, it shouldn't create the select entity. Do you have a device with the older firmware to verify?


Yes, I still have a few. The entity can be updated without any extra code, but an error occurs during the update: Status.UNSUPPORTED_ATTRIBUTE: 134. This is not good behavior cause it will cause questions.

Screenshot 2024-11-30 at 00 00 32

TheJulianJES:
Hmm, when ZHA reads the attribute during pairing, we should get "unsupported attribute" back. It will cause that attribute to be added to the "unsupported attributes" list for that device. ZHA should not create the entity then:
https://github.com/zigpy/zha/blob/6737b3628f1166af1ac90dc39019a402024dc331/zha/application/platforms/select.py#L190

Can you try and restart HA to see if the entity disappears then? Also, when you download the diagnostics file for that device (from the device page -> three dots), is `switch_type` listed in the "unsupported attributes" section in the JSON?

ikruglov:
@TheJulianJES, it's weird. `switch_type` is present is both `attributes` and `unsupported attributes`

```
          "0x0006": {
            "attributes": {
              "0xfffd": {
                "attribute": "ZCLAttributeDef(id=0xFFFD, name='cluster_revision', type=<class 'zigpy.types.basic.uint16_t'>, zcl_type=<DataTypeId.uint16: 33>, access=<ZCLAttributeAccess.Read: 1>, mandatory=True, is_manufacturer_specific=False)",
                "value": null
              },
              "0x4000": {
                "attribute": "ZCLAttributeDef(id=0x4000, name='global_scene_control', type=<enum 'Bool'>, zcl_type=<DataTypeId.bool_: 16>, access=<ZCLAttributeAccess.Read: 1>, mandatory=False, is_manufacturer_specific=False)",
                "value": null
              },
              "0x4002": {
                "attribute": "ZCLAttributeDef(id=0x4002, name='off_wait_time', type=<class 'zigpy.types.basic.uint16_t'>, zcl_type=<DataTypeId.uint16: 33>, access=<ZCLAttributeAccess.Read|Write: 3>, mandatory=False, is_manufacturer_specific=False)",
                "value": null
              },
              "0x0000": {
                "attribute": "ZCLAttributeDef(id=0x0000, name='on_off', type=<enum 'Bool'>, zcl_type=<DataTypeId.bool_: 16>, access=<ZCLAttributeAccess.Read|Report|Scene: 25>, mandatory=True, is_manufacturer_specific=False)",
                "value": 0
              },
              "0x4001": {
                "attribute": "ZCLAttributeDef(id=0x4001, name='on_time', type=<class 'zigpy.types.basic.uint16_t'>, zcl_type=<DataTypeId.uint16: 33>, access=<ZCLAttributeAccess.Read|Write: 3>, mandatory=False, is_manufacturer_specific=False)",
                "value": null
              },
              "0xfffe": {
                "attribute": "ZCLAttributeDef(id=0xFFFE, name='reporting_status', type=<enum 'AttributeReportingStatus'>, zcl_type=<DataTypeId.enum8: 48>, access=<ZCLAttributeAccess.Read: 1>, mandatory=False, is_manufacturer_specific=False)",
                "value": null
              },
              "0x4003": {
                "attribute": "ZCLAttributeDef(id=0x4003, name='start_up_on_off', type=<enum 'StartUpOnOff'>, zcl_type=<DataTypeId.enum8: 48>, access=<ZCLAttributeAccess.Read|Write: 3>, mandatory=False, is_manufacturer_specific=False)",
                "value": 255
              },
              "0x1001": {
                "attribute": "ZCLAttributeDef(id=0x1001, name='switch_type', type=<enum 'NodOnSwitchType'>, zcl_type=<DataTypeId.enum8: 48>, access=<ZCLAttributeAccess.Read|Write|Report: 11>, mandatory=False, is_manufacturer_specific=True)",
                "value": null
              }
            },
            "unsupported_attributes": [
              4097,
              "switch_type"
            ]
```

ikruglov:
~@TheJulianJES I've done a bit of debugging. This code  never executes for `switch_type` attribute:~ https://github.com/zigpy/zha/blob/6737b3628f1166af1ac90dc39019a402024dc331/zha/application/platforms/select.py#L190

No wait. I think I missed something.

ikruglov:
okay, I got why the code is not executed for `switch type`. 

the if condition is:
```
        if ENTITY_METADATA not in kwargs and (
            cls._attribute_name in cluster_handler.cluster.unsupported_attributes
            or cls._attribute_name not in cluster_handler.cluster.attributes_by_name
            or cluster_handler.cluster.get(cls._attribute_name) is None
        ):
```

`ENTITY_METADATA not in kwargs` part is not true for `switch_type` because:

```
kwargs={'entity_metadata': ZCLEnumMetadata(entity_platform=<EntityPlatform.SELECT: 'select'>, entity_type=<EntityType.CONF
IG: 'config'>, cluster_id=6, endpoint_id=1, cluster_type=<ClusterType.Server: 0>, initially_disabled=True, attribute_initialized_from_cache=True, translation_key='switch_type', fallback_name='Switch type', en
um=<enum 'NodOnSwitchType'>, attribute_name='switch_type', reporting_config=None)}
```

Looks like either a feature of ZHA, or a bug :-)

ikruglov:
This small patch makes it work as expected. I'm not sure if this is the right approach, though

```diff
$ git diff
diff --git a/zha/application/platforms/select.py b/zha/application/platforms/select.py
index 1012966..4f42dd2 100644
--- a/zha/application/platforms/select.py
+++ b/zha/application/platforms/select.py
@@ -185,15 +185,22 @@ class ZCLEnumSelectEntity(PlatformEntity):

         Return entity if it is a supported configuration, otherwise return None
         """
+
+        if ENTITY_METADATA in kwargs:
+            entity_metadata = kwargs.get(ENTITY_METADATA)
+            attr_name = entity_metadata.attribute_name if hasattr(entity_metadata, 'attribute_name') else None
+        else:
+            attr_name = cls._attribute_name
+
         cluster_handler = cluster_handlers[0]
-        if ENTITY_METADATA not in kwargs and (
-            cls._attribute_name in cluster_handler.cluster.unsupported_attributes
-            or cls._attribute_name not in cluster_handler.cluster.attributes_by_name
-            or cluster_handler.cluster.get(cls._attribute_name) is None
+        if attr_name is not None and (
+            attr_name in cluster_handler.cluster.unsupported_attributes
+            or attr_name not in cluster_handler.cluster.attributes_by_name
+            or cluster_handler.cluster.get(attr_name) is None
         ):
             _LOGGER.debug(
                 "%s is not supported - skipping %s entity creation",
-                cls._attribute_name,
+                attr_name,
                 cls.__name__,
             )
             return None
```

ikruglov:
FTR: Firmware images I used are from https://github.com/Koenkk/zigbee-OTA/pull/599. NodOn's CTO submits them.

TheJulianJES:
> `ENTITY_METADATA not in kwargs` part is not true for `switch_type`

Ah, of course! So we always "force create" all entities defined via quirks v2. I think the original intention was to skip this part:
`or cluster_handler.cluster.get(cls._attribute_name) is None` (and/or because the implementation/patch is a bit awkward otherwise because of the `ENTITY_METADATA`).
There, we check that there's at least a `not None` value in the attribute cache. That might make sense, but if the attributes are marked as "unsupported" or do not even have an attribute definition, I think we should also skip these.

Yeah, well, I guess we should change this for all platforms.. I'll try to have another look at this in the next few days. We might need to adopt something similar to your patch. I think we could compact it a bit though.

ikruglov:
@TheJulianJES I'd be happy to prepare a PR for all platforms. As long as the direction I initially took is the right one. Please let me know.

I've also created https://github.com/zigpy/zha/issues/373 to track the problem

Nbird67:
Could you please add Lexman modules too ? SIN-4-1-20_LEX that are Nodon rebranded module (Leroy Merlin)
"manufacturer_code": 4647,

Thank you.

ikruglov:
@TheJulianJES, would you be open to merging this PR as is? Later, I can follow up with changes to remove entities for unsupported firmware.

TheJulianJES:
Personally, I think it might be fine to merge as is, but I'd need to discuss that. We're currently still focused on HA 2025.2 to fix some issues there though.

https://github.com/zigpy/zha/pull/365 redoes discovery quite a bit. I think/hope to get that in time for HA 2025.3.
We might be able to add `_skip_creation_if_no_attr_cache` to other platforms (including select) and/or expose that via quirks v2.
So, it would be ideal if we can optionally add the same limitation of only creating v2 quirks if an initial value was read once and/or if they're not marked as unsupported, similar to normal ZHA entities.

Force creating v2 quirk entities by default makes sense IMO (ref: https://github.com/zigpy/zha/issues/373#issuecomment-2628420496), but it would still be nice to have the option to disable this. Hopefully, we can get that done for 2025.3.
That would also avoid the need for the (slightly more complicated) quirk matching by firmware.

I'll come back to this as soon as possible.

zhaquirks/nodon/switch.py
```diff
@@ -1,14 +1,122 @@
...
+    .applies_to(NODON, "SIN-4-1-21")
+    .applies_to(NODON, "SIN-4-1-20_PRO")
+    # .filter() TODO this should be applied to firmware 3.4.0+, but I didn't find a way to reliably detect it .
+    #           Please let me know if you know how
+    .replaces(OnOffSIN4_1_20)
+    .enum(
+        attribute_name=OnOffSIN4_1_20.AttributeDefs.switch_type.name,
+        enum_class=NodOnSwitchType,
+        cluster_id=OnOffSIN4_1_20.cluster_id,
+        initially_disabled=True,
```
TheJulianJES, line 72: I think we should just remove `initially_disabled=True` for all these entities as well. We don't really use that in other quirks and if users don't want to see the entities, they can disable them themselves.

Same for the others.

zhaquirks/nodon/switch.py
```diff
@@ -1,14 +1,122 @@
...
+        )
+
+
+(
+    # quirk is similar to https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/nodon.ts#L100
+    QuirkBuilder(NODON, "SIN-4-1-20")
+    .applies_to(NODON, "SIN-4-1-21")
+    .applies_to(NODON, "SIN-4-1-20_PRO")
+    # .filter() TODO this should be applied to firmware 3.4.0+, but I didn't find a way to reliably detect it .
+    #           Please let me know if you know how
```
TheJulianJES: Let's remove this `filter()` command and just put a comment above the `(` from the quirk that the entities should only be created with firmware 3.4.0+.

Same below.

zhaquirks/nodon/switch.py
```diff
@@ -1,14 +1,122 @@
...
+        """Select the switch type wire to the device. Available from version > V3.4.0."""
+        switch_type = ZCLAttributeDef(
+            id=0x1001,
+            type=NodOnSwitchType,
+            zcl_type=DataTypeId.enum8,  # need to explicitly set ZCL type
+            access="rw",
+            is_manufacturer_specific=True,
+        )
+
+        """ Set the impulse duration in milliseconds (set value to 0 to deactivate the impulse mode). """
```
TheJulianJES: This one space here should be removed:
```suggestion
        """Set the impulse duration in milliseconds (set value to 0 to deactivate the impulse mode). """
```

zhaquirks/nodon/switch.py
ikruglov: Same logic as in my other PR. These settings are to override results of internal calibration. So, not useful in many cases, I think (though was useful in mine). Hence, I think it make sense to have them disabled by default. Not a strong opinion though.

ikruglov: Same logic as in my other PR. These settings are to override results of internal calibration. So, not useful in many cases, I think (though was useful in mine). Hence, I think it make sense to have them disabled by default. Not a strong opinion though.

TheJulianJES: Ok, yeah. Seems fine. Also doesn't matter as much then with the firmware versions.
Still, we should address that in ZHA in the future to allow these entities to only be created if an attribute is cached for them (or somehow allow depending on the firmware version).

TheJulianJES (approved):
Thanks! Seems fine.

Issue #3597, opened 2024-12-08
[BUG] INVALID_DATA_TYPE error when selecting a new value for the Aqara FP1E motion sensitivity attribute


jeverley:
### Bug description

Using enum types to document attribute values in the new quirk I wrote has inadvertently resulted in the incorrect zigbee attribute type being sent to the device in write_attributes requests for the motion_sensitivity attribute.

I had not appreciated at the time that whilst it's derived from uint8_t, the code in https://github.com/zigpy/zigpy/blob/dev/zigpy/zcl/foundation.py maps the two types to separate zigbee type values (0x20 and 0x30).

This should be corrected by using the device's native uint8 type (raising a PR shortly).

### Steps to reproduce

Attempt to change the device Motion Sensitivity setting through the select entity/manage device attribute write.

### Expected behavior

The new value writes successfully to the device.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

![image](https://github.com/user-attachments/assets/99f4f3e0-8f13-48b2-97cc-07c67adf4a20)

</details>

### Logs

<details><summary>Logs</summary>

```python

2024-12-08 11:25:34.970 DEBUG (MainThread) [zigpy.zcl] [0xBBCE:1:0xfcc0] Sending request header: ZCLHeader(frame_control=FrameControl<0x04>(frame_type=<FrameType.GLOBAL_COMMAND: 0>, is_manufacturer_specific=True, direction=<Direction.Client_to_Server: 0>, disable_default_response=0, reserved=0, *is_cluster=False, *is_general=True), manufacturer=4660, tsn=3, command_id=<GeneralCommand.Write_Attributes: 2>, *direction=<Direction.Client_to_Server: 0>)
2024-12-08 11:25:34.972 DEBUG (MainThread) [zigpy.zcl] [0xBBCE:1:0xfcc0] Sending request: Write_Attributes(attributes=[Attribute(attrid=0x010C, value=TypeValue(type=AqaraMotionSensitivity, value=<AqaraMotionSensitivity.High: 3>))])
2024-12-08 11:25:34.977 DEBUG (MainThread) [zigpy.device] [0xbbce] Extending timeout for 0x03 request
[...]

PR #3598, opened 2024-12-08
Add `zcl_type` for Aqara FP1E sensor attributes

jeverley:
## Proposed change

Addresses the bug raised https://github.com/zigpy/zha-device-handlers/issues/3597

## Additional information

Following investigation of the errors documented in https://github.com/zigpy/zha-device-handlers/issues/3597 and feedback from @TheJulianJES in https://github.com/zigpy/zha-device-handlers/issues/3597#issuecomment-2526306000 I've explicitly defined zcl_type for attributes documented using enums and validated that writes to the attribute now complete successfully without introducing further issues.

## Checklist


TheJulianJES (approved):
Thanks!

PR #3605, opened 2024-12-12
Fix duplicate opening sensor for `_TZE200_pay2byax`

tr4nt0r:
## Proposed change

image

This should fix the issue with duplicate opening sensor for Tuya _TZE200_pay2byax.  

Tested only as a custom quirk, not the end result as only my production machine has a zigbee radio.

## Additional information

```python
"""Tuya temp and humidity sensors."""
from zigpy.quirks import DEVICE_REGISTRY
from zigpy.quirks.v2 import QuirkBuilder
from zigpy.quirks.v2.homeassistant.sensor import SensorDeviceClass, SensorStateClass
import zigpy.types as t
from zigpy.zcl.clusters.security import IasZone, ZoneType

from zhaquirks.tuya import TuyaLocalCluster
from zhaquirks.tuya.builder import TuyaQuirkBuilder

del DEVICE_REGISTRY._registry_v2[("_TZE200_pay2byax", "TS0601")]
del DEVICE_REGISTRY._registry_v2[("_TZE200_n8dljorx",  "TS0601")]

class TuyaIASContact(TuyaLocalCluster, IasZone):
    """Custom IasZone cluster that represents the Open/Closed sensor."""

    _CONSTANT_ATTRIBUTES = {
        IasZone.AttributeDefs.zone_type.id: ZoneType.Contact_Switch,
    }

class CustomTuyaQuirkBuilder(TuyaQuirkBuilder):
        
    def tuya_contact(
        self,
        dp_id: int,
        ias_cfg:  TuyaLocalCluster = TuyaIASContact,
    ) -> QuirkBuilder:
        """Add a Tuya contact switch configuration."""
        self.tuya_dp(
            dp_id,
            ias_cfg.ep_attribute,
            IasZone.AttributeDefs.zone_status.name,
        )
        self.adds(ias_cfg)
        return self

(
    CustomTuyaQuirkBuilder("_TZE200_pay2byax", "TS0601")
    .applies_to("_TZE200_n8dljorx",  "TS0601")
    .tuya_sensor(
        dp_id=101,
        attribute_name="measured_value",
        type=t.uint16_t,
        fallback_name="Illuminance",
        device_class=SensorDeviceClass.ILLUMINANCE,
        state_class=SensorStateClass.MEASUREMENT,
    )
    .tuya_contact(dp_id=1)
    .tuya_battery(dp_id=2)
    .skip_configuration()
    .add_to_registry()
)

```

Fixes: https://github.com/home-assistant/core/issues/132669

## Checklist

[...]

PR #3608, opened 2024-12-12
Split Tuya IAS function into contact & smoke

prairiesnpr:
## Proposed change

Originally, we had tuya_ias with the intention of having one method to handle any IAS requirements. Looking at https://github.com/zigpy/zha-device-handlers/pull/3605, I think having separate convenience methods does make sense. Unfortunately, adding `tuya_contact` for contact sensors while retaining `tuya_ias` for Fire/smoke sensors would be confusing. This PR add `tuya_contact` and `tuya_smoke` and then modifies `tuya_ias` to accept a converter.

## Checklist

### DIFF
--- a/tests/test_tuya_builder.py
+++ b/tests/test_tuya_builder.py
@@ -13,6 +13,8 @@ from zigpy.zcl.clusters.general import Basic
 from tests.common import ClusterListener, wait_for_zigpy_tasks
 import zhaquirks
 from zhaquirks.tuya.builder import (
+    TuyaIASContact,
+    TuyaIASFire,
     TuyaPowerConfigurationCluster2AAA,
     TuyaQuirkBuilder,
     TuyaRelativeHumidity,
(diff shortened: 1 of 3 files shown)


zhaquirks/tuya/builder/__init__.py
```diff
@@ -99,35 +107,54 @@ def tuya_battery(
...
-        self.tuya_dp(
-            dp_id,
-            metering_cfg.ep_attribute,
-            "current_summ_delivered",
+    def tuya_contact(self, dp_id: int):
+        """Add a Tuya IAS contact sensor."""
+        self.tuya_ias(
+            dp_id=dp_id,
+            ias_cfg=TuyaIASContact,
+            converter=lambda x: IasZone.ZoneStatus.Alarm_1 if x == 1 else 0,
```
TheJulianJES: I assume most contact sensors will only push `0` and `1` as values, but in case some sensor uses `2` (e.g. if two alarms are triggered), we might wanna catch that too and still trigger the alarm.

So, we could have any non-0 value activate the alarm.
It would be similar to the binary sensor parsing for the ZCL sensors defined in ZHA then, as they just use `bool(value)`.

```suggestion
            converter=lambda x: IasZone.ZoneStatus.Alarm_1 if x != 0 else 0,
```

tests/test_tuya_builder.py
```diff
@@ -46,6 +48,41 @@ def real_device(MockAppController):
...
+async def test_convenience_methods(device_mock, method_name, attr_name, exp_class):
+    """Test TuyaQuirkBuilder convenience methods."""
+
+    registry = DeviceRegistry()
+
+    entry = TuyaQuirkBuilder(
+        device_mock.manufacturer, device_mock.model, registry=registry
+    )
+    entry = getattr(entry, method_name)(dp_id=1)
+    entry.skip_configuration().skip_configuration().add_to_registry()
```
TheJulianJES: `skip_configuration()` is set twice right now, I guess one time is enough 😄 
```suggestion
    entry.skip_configuration().add_to_registry()
```

zhaquirks/tuya/builder/__init__.py
```diff
@@ -30,8 +30,16 @@
 from zhaquirks.tuya.mcu import DPToAttributeMapping, TuyaMCUCluster, TuyaOnOffNM
 
 
+class TuyaIASContact(IasZone, TuyaLocalCluster):
```
TheJulianJES: nit: For zigpy, we use `IasZone`. Here, we use `IAS` in upper-case. Do we want to change that to `Ias` to align with zigpy or do you prefer to leave it as-is?

TheJulianJES (approved):
Thanks!

PR #3612, opened 2024-12-14
Convert Tuya motion quirks to v2, add `_TZE200_ya4ft0w4`

prairiesnpr:
## Proposed change

Adds _TZE200_ya4ft0w4, converts rest of motion to v2 quirks

## Additional information

Closes: https://github.com/zigpy/zha-device-handlers/issues/3584
Closes: https://github.com/zigpy/zha-device-handlers/issues/3359

## Checklist

- [ ] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [x] Tests have been added to verify that the new code works

PR #3629, opened 2024-12-23
Fix Tuya BlitzWolf motion sensor

TheJulianJES:
## Proposed change
This reverts the BlitzWolf motion sensors to use the `IasZone` cluster instead of using the `OccupancySensing` cluster.
This change was made to keep existing entities working. The missing auto-reset delay was also reintroduced.

cc @prairiesnpr

## Additional information
Small follow-up to https://github.com/zigpy/zha-device-handlers/pull/3612

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_motion.py
+++ b/zhaquirks/tuya/ts0601_motion.py
@@ -1,12 +1,15 @@
 """BlitzWolf IS-3/Tuya motion rechargeable occupancy sensor."""
 
+import asyncio
 import math
+from typing import Any
 
 from zigpy.quirks.v2 import EntityType
 from zigpy.quirks.v2.homeassistant import UnitOfLength, UnitOfTime
 from zigpy.quirks.v2.homeassistant.sensor import SensorDeviceClass, SensorStateClass
 import zigpy.types as t
 from zigpy.zcl.clusters.measurement import IlluminanceMeasurement, OccupancySensing
+from zigpy.zcl.clusters.security import IasZone
 
 from zhaquirks.tuya import TuyaLocalCluster
 from zhaquirks.tuya.builder import TuyaQuirkBuilder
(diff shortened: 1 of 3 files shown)


TheJulianJES:
Hmm, looking at the code [we had for this before](https://github.com/zigpy/zha-device-handlers/blob/6d029efba11949230b928f09616d7ebcadf3f1b1/zhaquirks/tuya/ts0601_motion.py#L260-L277), I'm now also wondering if this sensor correctly sends a "reset" message.
Previously, we just reset it after 15 seconds of no messages.

I think that code was written before we had a proper Tuya datapoint conversion though.

TheJulianJES:
Ok, got the sensor now and we need a timeout. The current zhaquirks code for `MotionWithReset` and so on still uses `Bus`, so that doesn't play well with quirks v2. I'll take another stab or revert that sensor part otherwise.

TheJulianJES:
Changes work as expected now.
The timeout code is somewhat copied from `MotionWithReset` and related clusters from the base `zhaquirks` module.
However, that code is outdated, still uses `Bus` somewhat, and listens to the IAS zone status change commands.
We generally listen to IAS zone status attribute updates now. This is also needed for the Tuya DP conversion.

zhaquirks/tuya/ts0601_motion.py
```diff
@@ -20,6 +23,40 @@ class TuyaOccupancySensing(OccupancySensing, TuyaLocalCluster):
...
+
+    def _update_attribute(self, attrid: int | t.uint16_t, value: Any) -> None:
+        """Catch zone status updates and potentially schedule reset."""
+        if (
+            attrid == IasZone.AttributeDefs.zone_status.id
+            and value == IasZone.ZoneStatus.Alarm_1
+        ):
+            self.debug("%s - Received Tuya motion event", self.endpoint.device.ieee)
+            if self._timer_handle:
+                self._timer_handle.cancel()
```
TheJulianJES, line 54: Ah, missed to cover this one line. During a long-ish `reset_s` time, we should just re-send the motion message once to also cover this and make sure the delay restarts in tests, although the latter is a bit awkward to do, as we currently modify `reset_s` to `0`. We may want to consider using the `looptime` lib here(?)

TheJulianJES:
Hmm, `reset_s` was 15 before the changes in https://github.com/zigpy/zha-device-handlers/pull/3612 and is again with this PR, but my sensor seems to need a longer delay.
I'll leave this as-is for now to keep behavior "unchanged", but I wonder why nobody complained before..? 😄

PR #3633, opened 2024-12-24
Add Third Reality custom attributes and entities

hwzolin:
## Checklist


hwzolin:
@TheJulianJES 

Hello, the last two revisions involve the following aspects:
Add temperature and humidity sensor private clusters: temperature correction value setting item, humidity correction value setting item.
Add private clusters for soil moisture sensor: temperature correction value setting, humidity correction value setting.
Add switch private cluster: countdown time.
Add plug private cluster: countdown time.
Add watering kit private clusters: water duration, water interval.
Add vibration sensor private cluster: detection interval.
Add radar sensor private cluster: detection interval.
Add door sensor private cluster: open delay time.
Add curtain private cluster: enable disable pir remote.
Modify the min_value of motion sensor private cluster: detection_interval to 0.

Please review it, thank you.

hwzolin:
@TheJulianJES Hello, what does "needs reviewer answer" mean? Is this what I need to do? If there is no problem with my code, can you please merge it as soon as possible?

zhaquirks/thirdreality/door_sensor.py
```diff
@@ -0,0 +1,40 @@
...
+    QuirkBuilder("Third Reality, Inc", "3RDS17BZ")
+    .replaces(ThirdRealityDoorCluster)
+    .number(
+        attribute_name=ThirdRealityDoorCluster.AttributeDefs.open_delay_time.name,
+        min_value=0,
+        max_value=3600,
+        unit=UnitOfTime.SECONDS,
+        cluster_id=ThirdRealityDoorCluster.cluster_id,
+        translation_key="open_delay_time",
+        fallback_name="Open Delay Time",
```
TheJulianJES, line 37: According to HA specs, only the first letter should be capitalized, everything else should be lowercase (except for e.g. abbreviations). So, this should be "Open delay time" for the `fallback_name`, as that's what we'll use for the HA `strings.json` by default. 
This also needs to be changed for other entities.

TheJulianJES (commented):
PRs are way easier and faster to review when only a single device (type) is added in each one. You can create multiple PRs (with multiple different branches) at the same time though.
You can leave this one as is now, but maybe something to keep in the mind for next time. And some PRs can already be merged, even if others still need changes.

Also, the branch contains a lot of unnecessary commits from past PRs (that were squashed when merged into `dev`). In the future, please rebase your branch / always make changes coming from `dev`.
The changes are squashed at the end, so you can leave it like this for now too, but it makes reviews harder.

zhaquirks/thirdreality/door_sensor.py
hwzolin: OK

hwzolin:
> PRs are way easier and faster to review when only a single device (type) is added in each one. You can create multiple PRs (with multiple different branches) at the same time though. You can leave this one as is now, but maybe something to keep in the mind for next time. And some PRs can already be merged, even if others still need changes.
> 
> Also, the branch contains a lot of unnecessary commits from past PRs (that were squashed when merged into `dev`). In the future, please rebase your branch / always make changes coming from `dev`. The changes are squashed at the end, so you can leave it like this for now too, but it makes reviews harder.

@TheJulianJES 
Thank you for your patient reply.
Based on your reply, can I understand it this way: if I submit code in the future, I should create a PR after implementing the function of a device, which is more convenient for review. Instead of creating a PR after implementing the functions of all devices at once like now? This will cause trouble for the review.
Is it right? If my understanding is correct, I will pay attention to this point when submitting my code in the future.

PR #3643, opened 2024-12-25
Add support for Tuya DPs mapped to multiple cluster attributes

abmantis:
## Proposed change

Some Tuya devices expose multiple atributes (like current+voltage+power) in a single DP. This PR adds support for easilly mapping those, including Builder support.

## Checklist


zhaquirks/tuya/__init__.py
```diff
@@ -1532,7 +1532,7 @@ class AttributeDefs(BaseAttributeDefs):
         ),
     }
 
-    dp_to_attribute: dict[int, DPToAttributeMapping] = {}
+    dp_to_attribute: dict[int, list[DPToAttributeMapping]] = {}
```
TheJulianJES: Hmm, if we force `dp_to_attribute` to now have a list of `DPToAttributeMapping`, we will break basically all custom Tuya quirks that are floating around. That's still quite a few, since we need to integrate a lot of TRV related quirks (and others) still.

What about allowing `list[DPToAttributeMapping] | DPToAttributeMapping` as a dict value to also stay backwards compatible? Then, we don't need to change individual/existing quirks.

We'd need to handle that where `dp_to_attribute` is used OR we could have another attribute of `dict[int, list[DPToAttributeMapping]]` (e.g. `_dp_to_attributes`) and then just convert the `dp_to_attribute` to fit the type of that new attribute (so wrap single `DPToAttributeMapping` in a list). Then use that new attribute in all cases where we process/use `dp_to_attribute` currently.

abmantis: Good point! I forgot that we support custom quirks! I've followed your 2nd approach to make it backwards compatible.

zhaquirks/tuya/__init__.py
```diff
@@ -1627,30 +1630,40 @@ def handle_set_time_request(self, payload: t.uint16_t) -> foundation.Status:
         """Handle Time set request."""
         return foundation.Status.SUCCESS
 
+    @property
+    def _dp_to_attributes(self) -> dict[int, list[DPToAttributeMapping]]:
+        """Convert from legacy single attribute dp mappings."""
+        return {
+            dp: attr if isinstance(attr, list) else [attr]
+            for dp, attr in self.dp_to_attribute.items()
+        }
```
TheJulianJES: Hmm, this will be "regenerated" every time it's accessed and that'll be quite often.
We might be able to make this a `cached_property`, but I'd prefer if we can convert this in when initializing (in `__init__` of this cluster?).
I'm not sure if we have quirks that change `dp_to_attribute` at runtime though. Probably unlikely, but we'll also have to check that we don't break the Tuya quirk builder.

I'll get back to this.

zhaquirks/tuya/builder/__init__.py
```diff
@@ -443,18 +443,31 @@ def tuya_dp(
...
+    def tuya_dp_multi(
+        self,
+        dp_id: int,
+        attribute_mapping: list[DPToAttributeMapping],
+        dp_handler: str = "_dp_2_attr_update",
+    ) -> QuirkBuilder:  # fmt: skip
+        """Add Tuya DP Converter that maps to multiple attributes."""
+        self.tuya_dp_to_attribute.update({dp_id: attribute_mapping})
+        self.tuya_data_point_handlers.update({dp_id: dp_handler})
+        return self
```
TheJulianJES: Hmm, is there any benefit to also having this method instead of just calling `tuya_dp` twice for the same endpoint?

If so, we may want to add a test for this, though I'd need to check if `tuya_dp` is tested explicitly (or just by one of the other Tuya methods).

zhaquirks/tuya/__init__.py
```diff
@@ -1532,35 +1532,38 @@ class AttributeDefs(BaseAttributeDefs):
...
-                continue
-
-            # mark mapped to attribute as valid if existing and if on a LocalDataCluster
-            attr = cluster.attributes_by_name.get(dp_map.attribute_name)
-            if attr and isinstance(cluster, LocalDataCluster):
-                # _VALID_ATTRIBUTES is only a class variable, but as want to modify it
-                # per instance here, we need to create an instance variable first
-                if "_VALID_ATTRIBUTES" not in cluster.__dict__:
-                    cluster._VALID_ATTRIBUTES = set()
-                cluster._VALID_ATTRIBUTES.add(attr.id)
```
TheJulianJES, line 1564: Well, this code also never expected a tuple, so the "old multi DP code".
Needs updating in a future PR if we keep that code around.

TheJulianJES (commented):
Just went through the existing DP mapping code and noticed this was merged at some point: https://github.com/zigpy/zha-device-handlers/pull/1989
The example at the bottom is interesting. It seems like we already support mapping one datapoint to multiple attributes if they're on the same cluster.

I generally prefer this implementation though, as it's more explicit/clear than the older one. In a follow-up PR, we can probably convert old quirks and remove the old parsing code, unless we find that a bunch of custom quirks use it..

---

Overall, I think this is fine. We should see if we can optimize how often `_dp_to_attributes` is generated though. It might be enough to just do that once in the cluster `__init__`.

zhaquirks/tuya/builder/__init__.py
abmantis: Part of the reason was that calling `tuya_dp` multiple times would allow passing a different `dp_handler`, which will not work properly. It would still handle multiple `DPToAttributeMapping`, but just overwrite the `dp_handler`.
Having a separate method prevents that and makes it clearer to who uses it. I can drop it if you think it would be better that way.

jeverley:
I was under the impression that this is already supported in v1 quirks, see the DPToAttributeMapping class:

```
@dataclasses.dataclass
class DPToAttributeMapping:
    """Container for datapoint to cluster attribute update mapping."""

    ep_attribute: str
    attribute_name: Union[str, tuple]
    converter: Optional[
        Callable[
            [
                Any,
            ],
            Any,
        ]
    ] = None
    endpoint_id: Optional[int] = None
```

`attribute_name` can be provided as a tuple containing multiple attribute names, this is then used in conjunction with a converter that returns a value for each attribute.

Below is an example that I used previously that calls a function in the TuyaPowerPhase class to extract voltage, current and power:

```
        TUYA_DP_POWER_PHASE: DPToAttributeMapping(
            TuyaElectricalMeasurement.ep_attribute,
            (
                "rms_voltage",
                "rms_current",
                "active_power",
            ),
            converter=lambda x: TuyaPowerPhase.variant_3(x),
        ),
    }

class TuyaPowerPhase:
    """Extract values from a Tuya power phase datapoint."""

    @staticmethod
    def variant_3(value) -> Tuple[t.uint_t, t.uint_t, int]:
        voltage = (value[0] << 8) | value[1]
        current = (value[2] << 16) | (value[3] << 8) | value[4]
        power = (value[5] << 16) | (value[6] << 8) | value[7]
        return voltage, current, power * 10
```

This would look like this in a v2 quirk:


```
    .tuya_dp(
        dp_id=6,
        ep_attribute=TuyaElectricalMeasurement.ep_attribute,
        attribute_name=(
            TuyaElectricalMeasurement.AttributeDefs.rms_voltage.name,
            TuyaElectricalMeasurement.AttributeDefs.rms_current.name,
            TuyaElectricalMeasurement.AttributeDefs.active_power.name,
        ),
        converter=lambda x: TuyaPowerPhase.variant_3(x),
    )
```

Before seeing this PR I as about to raise my own (linked https://github.com/zigpy/zha-device-handlers/pull/3811) to amend the v2 tuya builder methods `tuya_dp` and `tuya_dp_attribute`  to support `str | tuple` for `attribute_name` to match the underlying DPToAttributeMapping class (the functions currently only use `str` as the type).

Are we overcomplicating things here?

abmantis:
@jeverley TheJulianJES pointed that out too here: https://github.com/zigpy/zha-device-handlers/pull/3643#pullrequestreview-2574122550

I have missed that when I opened this. Still, this PR makes it a bit more explicit, and allows having a single DP mapped to different clusters, which is not possible with the old DPToAttributeMapping tuple approach.

Also, not all code is currently properly handling the tuple. See https://github.com/zigpy/zha-device-handlers/pull/3643#discussion_r1929645872

The plan is to check if we can drop the old tuple after this is merge (and migrate any quirks using it).

zhaquirks/tuya/__init__.py
abmantis: moved to `__init__`. The Tuya builder sets `dp_to_attribute` when defining the class, before initialization, so it should be ok.

zhaquirks/tuya/__init__.py
abmantis: Personally I would prefer to drop the tuple from `DPToAttributeMapping`. Even if a few custom quirks break, I think it is worth it to have the code a bit clean. Except if there are a lot of custom quirks using it.

zhaquirks/tuya/builder/__init__.py
abmantis: Oh, I just noticed that I missed the tests when opening this PR from the original one: https://github.com/zigpy/zha-device-handlers/pull/3644/files#diff-d01eb6338c5bfa4635fb433e6a8fc5b9cf3de58e784218f8930c5ccaedd927f5
:man_facepalming: 

Copied them to this PR now.

zhaquirks/tuya/__init__.py
TheJulianJES: Ok, we probably should remove it soon then.
I guess we could also convert the old tuple-style to the new `_dp_to_attributes` format with the list in `__init__` if we really wanted to. Then, we wouldn't need to care about handling it again in `_dp_2_attr_update` or for the "mark attributes as valid". I think that could be nice. Otherwise, we can just remove it soon.

abmantis: I am fine with both removing or converting it to `_dp_to_attributes`. Personally, I prefer to remove it just to keep the code simpler and easier to maintain.

But let me know which approach you prefer and I'll do that. Is it fine if we leave that to a follow-up PR? This one is not breaking it anyway.

TheJulianJES: yeah, that's fine.

zhaquirks/tuya/builder/__init__.py
TheJulianJES: Ok, one more thing regarding this. Can we either block subsequent calls to `tuya_dp` if the `dp_id` is already in `tuya_dp_to_attribute` or just have it update the existing list of `DPToAttributeMapping` per id?
For the latter, both `tuya_dp_multi` and subsequent calls to `tuya_dp` would work as expected then.

abmantis: Added that in a new PR https://github.com/zigpy/zha-device-handlers/pull/3827 since it required changing 2 quirks that were invalid.

abmantis: Updated the `tuya_dp` method to call `tuya_dp_multi` since the code is mostly the same internally.

TheJulianJES (approved):
Should be good for now. There was another PR that wanted to use multiple `tuya_sensor` entities with the same datapoint, but different converters.
As these methods always call `tuya_dp`, it can't be done in a "nice way". So, in the future, we may want to consider supporting `tuya_dp` to be called multiple times with the same DP.

TheJulianJES:
And we should still remove the old "multiple DP" way in another PR.

PR #3644, opened 2024-12-25
Add Tuya PC321-Z-TY 3 phase power meter

abmantis:
## Proposed change

Device in Z2M: https://www.zigbee2mqtt.io/devices/TS0601_3_phase_clamp_meter.html

## Additional information

Requires: #3643

## Checklist


jeverley:
@abmantis, @prairiesnpr I've been working on a v2 quirk specifically for supporting multiple Tuya energy meter devices (https://github.com/zigpy/zha-device-handlers/pull/3824) - it already has mappings for 4 devices.

It would be pretty straightforward to support this device in that quirk by adding the relevant datapoints, I'll add those in this evening if you're happy to go that route (I'm conscious we want to avoid duplication of effort & code).

Tagging https://github.com/zigpy/zha-device-handlers/pull/3764 too for visibility.

abmantis:
@jeverley Thanks for the suggestion, but I don't think this device would benefit from that. 

As far as I know, this device only measures current in one direction, so no need for direction calculation/mitigation.
Also, I am not sure we should have virtual channels, as those can be added by the users on HA side, if they want to. That would add extra complexity to the quirks, making it harder for people to contribute to them.

Personally, given the added complexity, I would prefer to keep it simple unless there is something it is really missing.

jeverley:
> @jeverley Thanks for the suggestion, but I don't think this device would benefit from that.
> 
> As far as I know, this device only measures current in one direction, so no need for direction calculation/mitigation. Also, I am not sure we should have virtual channels, as those can be added by the users on HA side, if they want to. That would add extra complexity to the quirks, making it harder for people to contribute to them.
> 
> Personally, given the added complexity, I would prefer to keep it simple unless there is something it is really missing.

I was aiming to create a single file in which we could maintain Tuya energy meter devices consistently (utilising native zigbee clusters rather than custom sensors - this comes with the benefit of translations being available out of the box for the entities).
The intent being it would keep things easier to maintain than having per device implementations scattered accross multiple files.

The POC https://github.com/zigpy/zha-device-handlers/pull/3824/commits/b56613953b53f8aa8659f11613590c24368e474a commit excludes this device from the virtual channel + directionality logic (this is optionally be used for a device if required).

I suppose an alternative approach might be to split my PR into two files, one with 'simple' energy meter devices which don't require the energy direction logic to sign values, and those that do.

abmantis:
> I was aiming to create a single file in which we could maintain Tuya energy meter devices consistently (utilising native zigbee clusters rather than custom sensors - this comes with the benefit of translations being available out of the box for the entities). The intent being it would keep things easier to maintain than having per device implementations scattered accross multiple files.
> 
> The POC [b566139](https://github.com/zigpy/zha-device-handlers/commit/b56613953b53f8aa8659f11613590c24368e474a) commit excludes this device from the virtual channel + directionality logic (this is optionally be used for a device if required).
> 
> I suppose an alternative approach might be to split my PR into two files, one with 'simple' energy meter devices which don't require the energy direction logic to sign values, and those that do.

I just updated this PR because it had a lot of outdated stuff from the multi_dp branch. You can see from https://github.com/zigpy/zha-device-handlers/pull/3644/files#diff-5da7f6ba0921c69699fbf6b68fb97058fef0d107cbb4193173309e0894833f05 that it doesn't have much that could be reused by other quirks.

Regarding using standard zigbee attributes, I tried to do it for most attributes, except for energy since there is no 3 phase energy stuff in the standard. I tried using different endpoints, but ended up going for custom tuya attributes (I don't remember why I decided on that, maybe because of the naming?). I'll check that again before moving this to ready just to be sure.

Personally, I prefer to keep the PRs as small as possible instead of bundling multiple changes in one big PR. So I would add this one as a simple quirk, and then add shared stuff later, as a different PR. 
But I'll leave that for the maintainers to decide, as I am also ok with doing it all in one big PR if that is preferred.

rhakbari:
Support for Tuya Zemismart 3 Phase Power Meter _TZE200_v9hkz2yn variant needs to be added as I tested your current quirk and  it is not working with the Zemismart 3 phase device. Kindly check this PR [https://github.com/zigpy/zha-device-handlers/pull/3764] in which I have worked on. This custom quirk is working but it is on v1 quirk and the PR wasn't merged due to the versioning.

image

abmantis:
> Support for Tuya Zemismart 3 Phase Power Meter _TZE200_v9hkz2yn variant needs to be added as I tested your current quirk and  it is not working with the Zemismart 3 phase device.

That does not seem to be the same device as the one in this PR. Are the data points the same?

rhakbari:
> > Support for Tuya Zemismart 3 Phase Power Meter _TZE200_v9hkz2yn variant needs to be added as I tested your current quirk and  it is not working with the Zemismart 3 phase device.
> 
> That does not seem to be the same device as the one in this PR. Are the data points the same?

Yeah, it's not the same device as in your PR, which is why we would need to add it to the PR with v2 quirks. My PR was created using v1 quirks since I wasn't aware of v2 at the time.

abmantis:
> > > Support for Tuya Zemismart 3 Phase Power Meter _TZE200_v9hkz2yn variant needs to be added as I tested your current quirk and  it is not working with the Zemismart 3 phase device.
> > 
> > 
> > That does not seem to be the same device as the one in this PR. Are the data points the same?
> 
> Yeah, it's not the same device as in your PR, which is why we would need to add it to the PR with v2 quirks. My PR was created using v1 quirks since I wasn't aware of v2 at the time.

In that case it should be on a new separate PR, not on this one (except it shares the same endpoints and has just a different model id)

zhaquirks/tuya/ts0601_power.py
```diff
@@ -0,0 +1,218 @@
...
+)
+import zigpy.types as t
+from zigpy.zcl.clusters.general import LevelControl, OnOff
+from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
+
+from zhaquirks.tuya import DPToAttributeMapping, TuyaLocalCluster
+from zhaquirks.tuya.builder import TuyaQuirkBuilder
+
+
+def dpToPower(data: ByteString) -> int:
```
TheJulianJES: Could we use snake case for all these functions?

zhaquirks/tuya/ts0601_power.py
```diff
@@ -0,0 +1,218 @@
...
+    """Convert DP data to power value."""
+    # From https://github.com/Koenkk/zigbee2mqtt/issues/18603#issuecomment-2277697295
+    power = int(data)
+    if power > 0x0FFFFFFF:
+        power = (0x1999999C - power) * -1
+    return power
```
TheJulianJES, line 25: Could we add coverage for them as well?
Ideally, we'd send the Tuya DP update in the tests (or get real data from the device) and check that the local ZCL attributes are updated with the expected data.

Just calling these directly from the tests would not be as nice, but better than nothing.

zhaquirks/tuya/ts0601_power.py
```diff
@@ -0,0 +1,218 @@
...
+        ElectricalMeasurement.AttributeDefs.ac_voltage_multiplier: 1,
+        ElectricalMeasurement.AttributeDefs.ac_voltage_divisor.id: 10,
+    }
+
+
+(
+    TuyaQuirkBuilder("_TZE200_nslr42tt", "TS0601")
+    .tuya_temperature(dp_id=0x85, scale=10)
+    .tuya_sensor(
+        dp_id=0x86,
```
TheJulianJES: Can we use the decimal formatting for `dp_id` everywhere? (e.g. 133 and 134 here) 
We use that for all other Tuya v2 quirks at the moment, so might make sense to do the same here: https://github.com/search?q=repo%3Azigpy%2Fzha-device-handlers+dp_id%3D&type=code

TheJulianJES (commented):
Ok, one small (hopefully) last thing. Looks good otherwise.

zhaquirks/tuya/ts0601_power.py
abmantis: I wonder if we should switch to hex by default instead (and eventually migrate the existing ones), for two reasons:

- easier when porting over Z2M 
- our loggers log the dp ids as hex, so it makes things consistent and easier when adding/debugging quirks.


I am fine in changing it to decimal, though if that is still preferred.

TheJulianJES: The modern [Z2M Tuya definitions with `tuyaDatapoints`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/db9f408abf87d6d6ccd45115ae056458db08c135/src/devices/tuya.ts#L10056-L10071) all seem to to use decimal, as far as I can see.
I think this is why we also went with decimal for all existing v2 Tuya quirks.

We may want to change the logging for dp_ids to decimal then as well?

abmantis: > We may want to change the logging for dp_ids to decimal then as well?

Yeah, I think so. I'll open a separate PR for that.

zhaquirks/tuya/ts0601_power.py
```diff
@@ -0,0 +1,218 @@
...
+        attribute_name="power_factor_ph_b",
+    )
+    .tuya_dp(
+        dp_id=122,
+        ep_attribute=Tuya3PhaseElectricalMeasurement.ep_attribute,
+        attribute_name="power_factor_ph_c",
+    )
+    .adds(Tuya3PhaseElectricalMeasurement)
+    .removes(LevelControl.cluster_id)
+    .removes(OnOff.cluster_id)
```
TheJulianJES, line 215: Not sure if these clusters have any use at all, but we can now use `prevent_default_entity_creation(endpoint_id=1, cluster_id=OnOff.cluster_id)` as well (instead of completely "removing" the clusters).

zhaquirks/tuya/ts0601_power.py
```diff
@@ -0,0 +1,218 @@
...
+(
+    TuyaQuirkBuilder("_TZE200_nslr42tt", "TS0601")
+    .tuya_temperature(dp_id=133, scale=10)
+    .tuya_sensor(
+        dp_id=134,
+        attribute_name="device_status",
+        type=t.int32s,
+        fallback_name="Device status",
+        translation_key="device_status",
+    )
```
TheJulianJES, line 68: What does this sensor show? Since it's only numerical..

TheJulianJES (approved):
Thanks!

zhaquirks/tuya/ts0601_power.py
abmantis: They did not seem to do anything at all. Should `prevent_default_entity_creation` be used instead for these cases?

zhaquirks/tuya/ts0601_power.py
abmantis: I have no ideia, I just added it since it is also on Z2M :/ On mine it is always 0.
Should I remove it?

Issue #3626, opened 2024-12-20
[Device Support Request] TZE204_dapwryy7 - TS0601


dexterix88:
### Problem description

No Sensor present (Occupancy) after adding this device. Initial device onboarding log.

### Solution description

I already tested adding the device identifier to existing quirk [ts0601_motion.py](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts0601_motion.py) my only customisation is on line 425: ("_TZE204_dapwryy7", "TS0601"), And at least Occupancy sensor works - see attached screen from my HA.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

![image](https://github.com/user-attachments/assets/d828aed4-56a2-44b1-b14d-536c193dd7ea)

</details>

### Device signature

<details><summary>Device signature</summary>

```json
[Paste the device signature here]
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
[Paste the diagnostic information here]
```

</details>

### Logs

<details><summary>Logs</summary>

```python
[...]

PR #3648, opened 2024-12-28
add support for _TZE204_dapwryy7

bangert:
## Proposed change
Add _TZE204_dapwryy7 to the list of known devices. For #3626 

## Additional information
I dont have the device myself. this is purely based on the information available in the linked issue.

## Checklist


prairiesnpr:
For reference the z2m converter is here https://github.com/Koenkk/zigbee-herdsman-converters/blob/9ede78a950d703dac4e60271e08dabf7808661e6/src/devices/tuya.ts#L10736

I think this will require a separate quirk, the DPs don't seem to match the existing one.

bangert:
Thanks for the feedback - true i didnt check that.

i would not hold this back, though, as it is a step in the right direction. with this change one can make (limited) use of the device - which is more than one can now. but thats not my call, obviously.

bangert:
let's close this for now, i do not intend to continue on this. Thanks for your work on ZHA

Issue #2517, opened 2023-08-09
[Device Support Request] Hue bulbs missing Candle Effect


phbasler:
### Problem description

I have two Hue Go here which have a built in canlde effect.
According to https://github.com/Koenkk/zigbee2mqtt/issues/15699 it is possible to start this by sending certain commands to the cluster `0xfc03`.

This cluster supports the Canlde effect, the fireplace effect and also the color loop.
On my device page for the light I can only see the color loop effect.

If I go to manage zigbee device from the device page, then I don't have this cluster, only `0xfc01` in that range.
Is there a possibility to add these effects to zha so that we can call them from Home Assistant?

### Solution description

Some possibility to trigger the candle effect of the hue go through a script.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

![image](https://github.com/zigpy/zha-device-handlers/assets/28863303/d2c3cf8c-40ac-43c3-931a-9f5c70e2f5f0)

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.FullFunctionDevice|MainsPowered|RxOnWhenIdle|AllocateAddress: 142>, manufacturer_code=4107, maximum_buffer_size=82, maximum_incoming_transfer_size=128, server_mask=11264, maximum_outgoing_transfer_size=128, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "11": {
      "profile_id": "0x0104",
      "device_type": "0x010d",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0300",
        "0x1000",
        "0xfc01"
      ],
      "output_clusters": [
        "0x0019"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
[...]

Issue #3500, opened 2024-11-09
[Device Support Request] Philips Hue Festavia string lights LCX015


ghuser238490:
### Problem description

At the moment only monochrome colors are supported. It would be great if you could use gradient scenes and effects as well.
https://www.philips-hue.com/en-us/p/hue-white-and-color-ambiance-festavia-string-lights/046677585303

### Solution description

In Zigbee2MQTT those additional features are already implemented:
https://www.zigbee2mqtt.io/devices/9290036744.html#gradient-scene-enum

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4107,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 128,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 128,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "11": {
      "profile_id": "0x0104",
      "device_type": "0x010d",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0300",
        "0x1000",
        "0xfc01",
        "0xfc03",
        "0xfc04"
      ],
      "output_clusters": [
        "0x0019"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "Signify Netherlands B.V.",
  "model": "LCX015",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json

  "data": {
    "ieee": "**REDACTED**",
    "nwk": 38455,
    "manufacturer": "Signify Netherlands B.V.",
    "model": "LCX015",
    "name": "Signify Netherlands B.V. LCX015",
    "quirk_applied": false,
    "quirk_class": "zigpy.device.Device",
    "quirk_id": null,
    "manufacturer_code": 4107,
    "available": true,
[...]

Issue #2488, opened 2023-07-25
[Device Support Request] Hue Ambiance Gradient Lightstrip


scottgerring:
### Problem description

The [Hue Ambiance Gradient lightstrip](https://www.philips-hue.com/de-de/p/hue-white---color-ambiance-gradient-ambiance-lightstrip-basis-set-2-meter/8719514339965) currently shows up as a single light, but is configurable to some degree in segments. 

### Solution description

Something like what deconz and [zigbee2mqtt ](https://github.com/Koenkk/zigbee2mqtt/issues/10674)are doing - it would be great to configure the different colours of the strip to get the gradient effect. 

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4107, maximum_buffer_size=82, maximum_incoming_transfer_size=128, server_mask=11264, maximum_outgoing_transfer_size=128, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "11": {
      "profile_id": 260,
      "device_type": "0x010d",
      "in_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0300",
        "0x1000",
        "0xfc01",
        "0xfc03"
      ],
      "out_clusters": [
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "Signify Netherlands B.V.",
  "model": "LCX004",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
 "data": {
    "ieee": "**REDACTED**",
[...]

PR #3664, opened 2025-01-02
Expose the 'multicolor' command on Hue lights

kjagiello:
## Proposed change

I have sourced the Hue device models that support gradients from [z2m](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5be3337374c8ee4120966e8cb30b8a80f624391f/src/devices/philips.ts) and added a quirk for them, exposing Hue's "multicolor" command that can be used to control the gradient, i.e. set a custom gradient or activate a built-in gradient effect. 

Below you will find more details on how this command can be used.

### Static gradients

Here comes an example of how to set a custom gradient using an HA action:

```yaml
action: zha.issue_zigbee_cluster_command
data:
  ieee: <ieee>
  cluster_id: 0xFC03
  endpoint_id: 11
  command: 0x0000
  command_type: server
  params:
      data: '50010400135000000037a7a30edb6337a7a30edb6337a7a32800'
```

To try out some other gradients, you can try one of the [gradient scenes](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5be3337374c8ee4120966e8cb30b8a80f624391f/src/lib/philips.ts#L285-L361).

The payload structure is very cryptic, but, thankfully, the [deconz](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/8ae69a976bca13f22e8002a13ebe798d1e26c086/hue.cpp#L280) and [z2m](https://github.com/Koenkk/zigbee2mqtt/issues/10674#issuecomment-1284527609) communities have done a great job deciphering it. This PR does however not provide any handy interface for constructing the payload, since I was not sure how to do it properly.

If you want to create your own gradients, you can do it with the help of a simple wizard that I have put together: https://kjagiello.github.io/hue-gradient-command-wizard/..

### Built-in effects

Using an HA action, you can activate, for example, the "opal" Hue effect:

```yaml
action: zha.issue_zigbee_cluster_command
data:
  ieee: <ieee>
  cluster_id: 0xFC03
  endpoint_id: 11
  command: 0x0000
  command_type: server
  params:
      data: '2100010b'
```

You can find more effects [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5be3337374c8ee4120966e8cb30b8a80f624391f/src/lib/philips.ts#L363-L372).

## Additional info

This PR should, at least partially, resolve the following issues:
- https://github.com/zigpy/zha-device-handlers/issues/3500
- https://github.com/zigpy/zha-device-handlers/issues/2488

## Checklist

    - Caveat: As mentioned above, I have relied on the device signatures provided by z2m. The only gradient device that I own is LCX005, which I performed testing on.


kjagiello:
For an easier way to generate the command payloads, you can check out https://kjagiello.github.io/hue-gradient-command-wizard/.

BetaRavener:
Fixes https://github.com/zigpy/zha-device-handlers/issues/2517.

jtbandes:
I also have some of [these bulbs](https://www.philips-hue.com/en-us/p/hue-white-ambiance-br30-e26-smart-bulb/046677577995) (Signify LTB003) and they support the candle effect in the Hue app. Should these (and other similar products) be included as well? I'm not entirely sure how to help test the changes but I could try to do so if that's helpful.

zhaquirks/philips/perifo.py
```diff
@@ -0,0 +1,19 @@
+"""Hue Perifo device."""
+
+from zigpy.quirks.v2 import QuirkBuilder
+
+from zhaquirks.philips import PhilipsHueCluster
+
+(
+    QuirkBuilder()
+    .applies_to("Signify Netherlands B.V.", "929003116301")
```
BetaRavener: I missed these few files when using SIGNIFY name constant. Could be used here too for uniformity sake.

zhaquirks/philips/hue_go.py
```diff
@@ -0,0 +1,12 @@
+"""Philips Hue Go device."""
+
+from zigpy.quirks.v2 import QuirkBuilder
+
+from zhaquirks.philips import PHILIPS, PhilipsHueCluster
+
+(
+    QuirkBuilder(PHILIPS, "7602031P7")
+    .also_applies_to(PHILIPS, "7602031U7")
```
BetaRavener: I've spotted a difference from your quirk builders, and it turns our `also_aplies_to` is just compatibility alias, so it could be changed.
```suggestion
    .applies_to(PHILIPS, "7602031U7")
```

source: https://github.com/zigpy/zigpy/blob/0f42cab8f7058cc68044678e4e6e46e59d1747b9/zigpy/quirks/v2/__init__.py#L554

kjagiello: Good catch!

zhaquirks/philips/perifo.py
kjagiello: Fixed it!

BetaRavener:
@jtbandes I have HomeAssistant running in docker. To test the changes, I've logged in to the container running on remote server with VS Code as it's more convenient for text and folder editing, but you can also use classic `ssh <hostname>` and then `docker exec -ti <container_name> sh` and use console text editor. Then you can directly edit files under `/usr/local/lib/python3.13/site-packages/zhaquirks/philips/` and once ready, restart home assistant. Even if you make mistake it will start up and you will find your error in the logs. If everything works, the new cluster will show up in GUI on your device:
image
and then
image

If you installed HA directly without docker, I'm rather positive you can use the same approach, minus `docker exec`, but if you are having troubles, `find / -name "zhaquirks" 2>/dev/null` could help you track down the file location.

TheJulianJES (commented):
I think we should combine most/all quirks into one file, e.g. `gradient.py`, since it's essentially the "same quirk", just adding the `PhilipsHueCluster` (and a friendly name).
Like, just have all `QuirkBuilder`s in one file.

jtbandes:
@BetaRavener cool, thanks for the info! Btw, are you still planning to pursue the UI work mentioned in https://github.com/zigpy/zha-device-handlers/issues/2517#issuecomment-1949528398 to make the effects show up in the light's UI?

I was able to confirm that by adding `.applies_to(SIGNIFY, "LTB003")` I got access to the new command and was able to successfully activate the "candle" effect on my bulbs. I see this bulb is also mentioned in the [z2m converter code](https://github.com/Koenkk/zigbee-herdsman-converters/blob/5be3337374c8ee4120966e8cb30b8a80f624391f/src/devices/philips.ts#L3652), so I wonder if another pass should be done to copy all the supported devices from there?

Actually it looks like in Z2M, if I’m reading this right, **all Hue devices get the `candle` effect [by default](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3d892edcb684dda14467a360cc953d86df629530/src/lib/philips.ts#L64)**. Should the same be done here, rather than just a handful of the devices being included?  I'm not sure what the other 4 default effects (`blink`, `breathe`, `okay`, `channel_change`) are referring to, but they don't seem to be Philips-specific...wonder what they do? (I'm not a Z2M user so I can't easily play with them.)

image


By the way, the Hue app gives me the following options for these bulbs: a few pre-configured candle effects. Curious if anyone knows how the "custom candle" would be accessed via this command's parameters? (Probably not a big deal as I can barely tell the difference between the 3 premade candle effects anyway 😅)

kjagiello:
> I think we should combine most/all quirks into one file, e.g. `gradient.py`, since it's essentially the "same quirk", just adding the `PhilipsHueCluster` (and a friendly name). Like, just have all `QuirkBuilder`s in one file.

Done! Though I have decided to put them under `hue.py` instead, because not all of Hue devices are gradient compatible, e.g. Hue Go or all the light bulbs. Let me know if you disagree! 🙏

BetaRavener:
@jtbandes thanks for testing, we could include your bulb too then. I don't know what's best solution for dealing with all these models and there are apparently lot of them. With this PR, we are only making it possible to send command to the philips cluster. We could be implicit and enable the quirk for every philips device that has 0xFC03 cluster, but then we run risk that later philips might use the cluster for something else or otherwise break compatibility. Then we have to be explicit anyway and list all models. And for that we could either take something like z2m as source of truth, or add devices gradually as requested. In any case the list will have to be updated as Philips keeps creating new lights.

Regarding the effects, I will use the hue go issue thread as not to spam here, but yes I'll be looking into the HA GUI part. To just answer rest - in z2m code you shared it does does indeed seem like all philips lights have the hueEffect enabled by default because of this [line](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3d892edcb684dda14467a360cc953d86df629530/src/lib/philips.ts#L57C13-L57C22). The definition of light would have to contain `hueEffect: false`, which is done only for single light in that codebase. That would mean quirk in this PR could apply to all those lights, but I'm still not persuaded we should just copy the whole list over without verifying.. And lastly, the other effects are not implemented by cluster in this PR but by Identify. Those are actually already exposed in (Z)HA:
image, 

so it's rather strange that z2m throws the "candle" effect into that mix. Maybe it's because any bulb that can change brightness (majority) could support "candle", but I'd still not do it that way.

zhaquirks/philips/hue.py
```diff
@@ -0,0 +1,123 @@
...
+    .applies_to(SIGNIFY, "LCX016")
+    .applies_to(SIGNIFY, "LCX017")
+    .friendly_name(
+        model="Hue Festavia Gradient Light String",
+        manufacturer="Philips",
+    )
+    .replaces(PhilipsHueCluster, endpoint_id=11)
+    .add_to_registry()
+)
+
```
jtbandes:
```suggestion

(
    QuirkBuilder()
    .applies_to(SIGNIFY, "LTB003")
    .friendly_name(
        model="Hue White Ambiance BR30 E26",
        manufacturer="Philips",
    )
    .replaces(PhilipsHueCluster, endpoint_id=11)
    .add_to_registry()
)

```

zhaquirks/philips/__init__.py
```diff
@@ -340,3 +345,21 @@ class PhilipsRwlRemoteCluster(PhilipsRemoteCluster):
         3: Button("down", DIM_DOWN),
         4: Button("off", TURN_OFF),
     }
+
+
+class PhilipsHueCluster(CustomCluster):
```
TheJulianJES: Would renaming this to `PhilipsLightCluster` or `PhilipsHueLightCluster` make sense?

TheJulianJES (commented):
And, do we wanna rename `hue.py` to `light.py` or `hue_light.py`? I think that makes it a bit clearer that those quirks are for Hue lightning.

Otherwise, I think we can merge this. I really hope that `friendly_name` doesn't cause us any issues. I don't think the implementation is ideal and we also don't implement `model_id` in ZHA, but I guess we can still try it. There shouldn't be any blueprints looking for the previous names.

HA naming generally specifies that the names should be written like this: `Hue ambiance spot`.

Looking at the US Hue page, this is also how Hue does it, so we should follow that:
image
"Hue Perifo light tube" should be fine, for example. This is also how Z2M does it: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/philips.ts

kjagiello:
> And, do we wanna rename hue.py to light.py or hue_light.py? I think that makes it a bit clearer that those quirks are for Hue lightning.

Here's some context behind my thought process when choosing the current naming.

I went back and forth on the naming, but [this Wiki article](https://en.wikipedia.org/wiki/Philips_Hue) ultimately guided my choice. It describes Philips Hue as "a line of color-changing LED lamps and white bulbs".

Using just "Hue" felt appropriate, as it directly ties to the product line, whereas "Hue lights" seemed a bit redundant. On the other hand, calling it simply "lights" felt too broad (far fetched scenario perhaps, but Philips could introduce a new line of lighting products).

That said, I'm fully open to revisiting this if you feel a different name would be more appropriate. Let me know what you think!

> HA naming generally specifies that the names should be written like this: Hue ambiance spot.

Makes sense! I've just pushed a change that aligns the naming.

TheJulianJES:
> Using just "Hue" felt appropriate, as it directly ties to the product line, whereas "Hue lights" seemed a bit redundant.

Hmm, there are Philips Hue remotes/dimmers and their Hue Secure line with Zigbee door sensors and other stuff.
Even if it may be a bit redundant to add use `hue_light` and `PhilipsLightCluster`/`PhilipsHueLightCluster`, I think it does make it a bit clearer that this is for the lights part of Philips Hue.

I mean, we can always freely rename this in the future, but let's say you're working on a quirk for a new Hue dimmer and try to use and import the Hue cluster for the remotes, you might stumble upon `PhilipsHueCluster` cluster and try to use that for your remotes. Granted, I doubt this will happen, but I think we could just clarify the lights part a bit.

kjagiello:
Fair enough! Renamed everything now.

@BetaRavener I think you will need to take this rename into account in https://github.com/zigpy/zha/pull/355

TheJulianJES (approved):
LGTM. Thanks!

BetaRavener:
Amazing this landed, I'll updated the other PR accordingly. 👍

ylnest2018:
apologies for being a noob here, how do i get started with this, I am trying to change colours and effects through automations on my Phillips Hue Gradient Strip which is paired to ZHA...currently it is using Quirk: zigpy.quirks.v2.CustomDeviceV2

tomasbedrich:
Great work! 💪🏻  Are you aware of any plans to make it "user-friendly" to set the pre-defined scenes, i.e. using "Effect" light option?

Screenshot 2025-11-29 at 12 22 06

TheJulianJES:
There's an open PR for that: https://github.com/zigpy/zha/pull/355, but it'll require some more changes in ZHA before we can go ahead with it. It's something we want to do eventually though.

jtbandes:
On that note, the open PR you referenced seems to be stalled, is there anything we can do to help with figuring out next steps? (What are the other ZHA changes that would be blocking this?)

PR #3674, opened 2025-01-04
Fix missing endpoints for Aqara H1 wireless remotes

TheJulianJES:
## Proposed change
Fix missing endpoints for Aqara H1 wireless remotes.

The device sometimes doesn't report these endpoints as available, so quirks v2 code that `adds` or `replaces` clusters on those endpoints will cause ZHA to break.
Zigpy now allows us to add these endpoints to the zigpy device, so we can also add clusters to the endpoints in the same quirk without issues.

## Additional information

Addresses https://github.com/home-assistant/core/issues/134686 and https://github.com/home-assistant/core/issues/134691.
Follow-up to https://github.com/zigpy/zha-device-handlers/pull/3564

~~This needs https://github.com/zigpy/zigpy/pull/1524 and a corresponding [zigpy bump](https://github.com/zigpy/zha-device-handlers/pull/3675), then a rebase before tests will pass.~~ (done)

## Checklist

### DIFF
--- a/tests/test_xiaomi.py
+++ b/tests/test_xiaomi.py
@@ -1804,3 +1804,20 @@ async def test_aqara_fp1e_sensor(
     assert len(ias_listener.attribute_updates) == 1
     assert ias_listener.attribute_updates[0][0] == IasZone.AttributeDefs.zone_status.id
     assert ias_listener.attribute_updates[0][1] == expected_motion_status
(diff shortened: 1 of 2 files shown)


phil-lipp:
Is there a way to update zha directly or do we need for HA Core to update?

PR #3681, opened 2025-01-06
Add IKEA Vallhorn quirk

vinzent:
## Proposed change

Allows to configure the "on only when dark" and the "on time"
value. Both "only when dark" (on/off) and "on time" (only 60s or 300s) can also be
configured by buttons on the device itself.
    
There is another attribute 0x0004 in this cluster which I haven't figured out what
it does.

## Additional information

Looks like this in my 2025.01.0 HomeAssistant:

grafik

## Checklist

### DIFF
--- a/zhaquirks/ikea/vallhorn.py
+++ b/zhaquirks/ikea/vallhorn.py
@@ -1,0 +1,52 @@
+"""IKEA Vallhorn quirk."""
+
+from typing import Final
+
+from zigpy import types as t
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class IkeaVallhornManufSpecificConfig(CustomCluster):
+    """Ikea Vallhorn manufacturer specific config cluster."""
+
+    name = "IKEA manufacturer specific config"
+    cluster_id = 0xFC81
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        on_only_when_dark: Final = ZCLAttributeDef(
+            id=0x0000,
+            type=t.Bool,
+        )
+
+        on_time: Final = ZCLAttributeDef(
+            id=0x0002,
+            type=t.uint16_t,
+        )
+
+
+(
+    QuirkBuilder("IKEA of Sweden", "VALLHORN Wireless Motion Sensor")
+    .replaces(IkeaVallhornManufSpecificConfig)
+    .switch(
+        IkeaVallhornManufSpecificConfig.AttributeDefs.on_only_when_dark.name,
+        IkeaVallhornManufSpecificConfig.cluster_id,
+        translation_key="on_only_when_dark",
+        fallback_name="On only when dark",
+    )
+    .number(
+        IkeaVallhornManufSpecificConfig.AttributeDefs.on_time.name,
+        IkeaVallhornManufSpecificConfig.cluster_id,
+        step=1,
+        min_value=10,
+        max_value=65534,
+        unit=UnitOfTime.SECONDS,
+        translation_key="on_time",
+        fallback_name="On time",
+    )
+    .add_to_registry()
+)


vinzent:
zha-toolkit device scan: https://gist.github.com/vinzent/af5956db77a20c94556030e812b2211b#file-ikea-vallhorn-zha-tookit-device-scan-json

vinzent:
I'm unsure if "dark threshold" is really working. Need to check again.

vinzent:
the 0x0004 attribute might be the reporting interval of the illuminace value. defaults to 60s. the illuminance value seems to be reported on every change of value tough - so I dont know what the benefit is.

MattWestb:
Its depends of how you is using the data from the sensor.
I using one in the bedroom window and in the summer its closing the blinds then the sun is high enough for heating it up up,
Fast = good and setting one time in HA automation for some minutes so not getting "jo-jo" effect. 
Too fast = eating batteries and spamming the logs for no use,
Normally it also possible setting one value limit for updating so not sending every litle change,

Was looking in the ZCL and attribute 0x0004 on cluster type 0x0400 is not reporting its LightSensorType.
``` 
7709 4.2.2.2.5 LightSensorType Attribute 
The LightSensorType attribute specifies the electronic type of the light sensor. This attribute shall be set to 
one of the non-reserved values listed in Table 4-6.
Table 4-6. Values of the LightSensorType Attribute
Attribute Value Description
0x00 Photodiode
0x01 CMOS
0x40 – 0xfe Reserved for manufacturer specific light sensor types
0xff Unknown
 ```
If reporting is configuball it shall using the normal Zigbee standard for doing it then its one IKEA device = normally very standard.

**Comment by MattWestb** [comment]
He he eh !!
And the next paragraph was:
```
7717 4.2.2.5 Attribute Reporting
 This cluster shall support attribute reporting using the Report Attributes command and according to the min7719 imum and maximum reporting intervals and reportable change settings described in the ZCL Foundation 
 specification (see 2.4.7). The following attributes shall be reported:
 MeasuredValue
```
So normal reporting setting of the attribute **0x0000 MeasuredValue uint16 0x0000 to 0xffff RP 0x0000 M**

**Comment by MattWestb** [comment]
If you like have the latest ZCL (Zigbee Cluster Library Specification R8) you can getting it here https://csa-iot.org/developer-resource/specifications-download-request/ only need putting in one valid email address.
But for some new IKEA devices it can being they is using upcoming ZCL R9 commands on private cluster but with the matter ZCL (Matter 1.4 Application Cluster Specification) like Symfonisk 2 and sommring is doing. 

## By the way great work done !!!!!!

**Comment by vinzent** [comment]
> Its depends of how you is using the data from the sensor. I using one in the bedroom window and in the summer its closing the blinds then the sun is high enough for heating it up up, Fast = good and setting one time in HA automation for some minutes so not getting "jo-jo" effect. Too fast = eating batteries and spamming the logs for no use, Normally it also possible setting one value limit for updating so not sending every litle change,

Regardless if I set the cluster 0xFC81 attribute 0x0004 to 300 or anything else, changes of illumination values are reported immediatly (e.g. if I turn on the light in the dark (by HA, not motion), the change is reported immediately). It seems tough without change, the value is reported only every 300 seconds if I configure 300. but its not reflected on export of the history if the value doesn't change but if I hover the mouse in the graph i can see that there seems to be a data point. Maybe i'd need to enable ZHA debug to verify this.

> 
> Was looking in the ZCL and attribute 0x0004 on cluster type 0x0400 is not reporting its LightSensorType.

Actually, the IKEA Vallhorn reports this:

grafik

which is defined here: https://github.com/zigpy/zigpy/blob/dev/zigpy/zcl/clusters/measurement.py#L13 and matches the zha-toolkit scan: https://gist.github.com/vinzent/af5956db77a20c94556030e812b2211b#file-ikea-vallhorn-zha-tookit-device-scan-json-L1714-L1724

**Comment by MattWestb** [comment]
Was moving one Vallhorn from Dirigera to test setup and installed your quirk and its looks working very well !!!
If setting the time to 10 and sniffing the network wireschark is showing the frame sent:
```
ZigBee Cluster Library Frame
    Frame Control Field: Cluster-specific (0x01)
    Sequence Number: 40
    Command: On with timed off (0x42)
    Payload
        On/Off Control Mask: 0x00
        On Time: 10.0 seconds
        Off Wait Time: 0.0 seconds
```
So indeed its possible setting how long time lights shall staying on and that is GREAT !!!
If putting in less then 10 the device is complaining but that is OK then its not user lighting one light for under 10 seconds then somthing is moving.
Do you knowing the max time the device is accepting ?

File: zhaquirks/ikea/vallhorn.py [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+        IkeaVallhornManufSpecificConfig.AttributeDefs.on_only_when_dark.name,
+        IkeaVallhornManufSpecificConfig.cluster_id,
+        translation_key="on_only_when_dark",
+        fallback_name="On only when dark",
+    )
+    .number(
+        IkeaVallhornManufSpecificConfig.AttributeDefs.on_time.name,
+        IkeaVallhornManufSpecificConfig.cluster_id,
+        step=1,
+        min_value=1,
```
MattWestb: I think 10 is the device lowest so if its possible limiting it here users is not getting errors that they dont understand.
I have not testing the max but if we is getting it i think it shall also being in ZHAs GUI.

vinzent: I just used min/max of uint16 here because if I don't specify anything, the slider is 1-100. I can adjust it to 10. will also try to test it myself.

probably the on_time should not be lower than the pir_o_to_u_delay which seems to be 20s.

grafik

also in the ZCL specs R8 I found the On/Off cluster provides the on_time attribute, which isn't used by vallhorn:

grafik


I also learned that the On/off cluster attribute on_off is mapped to "Opening" in HA/ZHA, which at least for me, a bit confusing word for this.

MattWestb: Servus Thomas.
Its depends on with endpoint type the On/Off is on how ZHA is using it if its one type light its making one light and if its one type outlet its make on out let of it and so on plus if its one in or out cluster.
I was using 10 seconds and its looks like the PIR can being triggered OK but i dont knowing if its so exact (perhaps sleeping little more or less some time) but i think its OK having it limited to 10 and if its stable at 20 its up to the user then its only plus functionality in the device.

PS: if you having on extra EZSP coordinator you can using it for sniffing the network and see what its being sent to and from the device and its great for finding strange things devices is doing.

PPS: I have putting my 2 Philips HUE motion sensors in the black box of bad Zigbee devices and im very happy with that !!!!!

MattWestb:
Was testing the the max time setting and 65535 is not valid but 65534 is so shall being the max in the code.
I was testing by sniffing then ZHA was setting the attribute the waiting its not option (the max is 18 hours) and the device is reporting susses.
```
ZigBee Network Layer Data, Dst: 0x71cd, Src: 0x0000
ZigBee Application Support Layer Data, Dst Endpt: 1, Src Endpt: 1
ZigBee Cluster Library Frame, Mfr: Ikea of Sweden (0x117c), Command: Write Attributes, Seq: 1
    Frame Control Field: Profile-wide (0x04)
    Manufacturer Code: Ikea of Sweden (0x117c)
    Sequence Number: 1
    Command: Write Attributes (0x02)
    Attribute Field, Uint16: 65534
        Attribute: 0x0002
        Data Type: 16-Bit Unsigned Integer (0x21)
        Uint16: 65534 (0xfffe)


ZigBee Network Layer Data, Dst: 0x0000, Src: 0x71cd
ZigBee Application Support Layer Data, Dst Endpt: 1, Src Endpt: 1
ZigBee Cluster Library Frame, Mfr: Ikea of Sweden (0x117c), Command: Write Attributes Response, Seq: 1
    Frame Control Field: Profile-wide (0x0c)
    Manufacturer Code: Ikea of Sweden (0x117c)
    Sequence Number: 1
    Command: Write Attributes Response (0x04)
    Status Record
        Status: Success (0x00)
```
And the device is rounding the time then using the parameter:
```
IKEA VALLHORN Wireless Motion Sensor On With Timed Off event was fired with parameters: {'on_off_control': <OnOffControl: 0>, 'on_time': 65516, 'off_wait_time': 0}
```

MattWestb:
Was potting the limits in my local quirk and the lower is working well but i have not testing the max but now no error then mixing with the value then it cant going under 10.

MattWestb:
@TheJulianJES Is you having one Vallhorn (8€) ?
If yes can you trying this then its making it to one of the (or more exact **the**) best motion sensor for searing light on the market.
Also adding the proposed min and max for the time ON.
Is shall being very good getting it in the next release of ZHA.

MattWestb:
Also shall we changing the _Opening_ sensor class / type to _Motion_ like the 2 old is using (2und gen is using Occupancy for only movement plus Motion) for make it looks nicer and being more understandabebal  ??

vinzent:
> Also shall we changing the _Opening_ sensor class / type to _Motion_ like the 2 old is using (2und gen is using Occupancy for only movement plus Motion) for make it looks nicer and being more understandabebal ??

I don't know that the correct terms are. but what "Occupancy" shows is for me more "Motion". The "Opening" is also confusing me, but I don't know how I would call it.

The sensors currently:

sensors

Occupancy (Belegung) is "Occupied" until the OccupancySensing pir_o_to_u_delay is done. Default is 20s:
 
zha-occupancy-sensing

The Opening (Öffnung) sensor is open until no motion is recognized for the new IkeaVallhornManufSpecificConfig on_time duration.

MattWestb:
This is little messy !!
Its endpoint can have different device types so the Gen 2 is https://github.com/zigpy/zigpy/blob/07e0451ce5317185e04b38e43aeb9fb168c62f28/zigpy/profiles/zha.py#L70 and looking like this:
IKEA Motion2
as as you see its one OnOff Sensor in the light domain so its sending  OnOff commands to light and / or (Zigbee) group of lights
Vallhorns endpoints is OnOff **and** Occupancy. The latest is:
https://github.com/zigpy/zigpy/blob/07e0451ce5317185e04b38e43aeb9fb168c62f28/zigpy/profiles/zha.py#L36C5-L36C30
Then the _Occupancy_ is one HA / Alarm sensor that cant searing OnOff commands to light and can only  sending  "HA" OnOff / Open/Closed and so on.

If all was straight all OnOff sensors shall being OpenCloce but i think ZHA have one patch for making the IKEA Gen 1 and 2 to motion sensors to motion ones.

MattWestb:
@vinzent If you is having one Parasoll its also have the same manufacture cluster "0xfc81" and is also sending On and Off to lights then open / closing the door/Window but without On With Time. 
If like digging its also one interesting if its also possible getting more functions from it !!

MattWestb (approved):
Looks nice and working well !!!
Great thanks for work done !!

TheJulianJES (approved):
Thanks!

Issue #1397, opened 2022-02-26
[Device Support Request] Insta GmbH Nexentro Blinds Actuator Mini


cremor:
**Is your feature request related to a problem? Please describe.**
I bought a few "Nexentro Blinds Actuator Mini" devices made by "Insta GmbH". Those are mounted behind existing blind switches and enable the blinds to be controlled both by the old manual switches and by Zigbee (and by a config app that uses Bluetooth).

What works:
- cover.close_cover
- cover.open_cover
- cover.stop_cover
- cover.set_cover_position
- Cover position attribute
  Note: Initially only one of my devices reported that correctly. The other ones always reported position as 100 and state as open. This fixed itself after I used the "reconfigure device" command on the device page.

What works unreliably:
- Cover state
  Sometimes the open/close/stop cover services don't change the state of all devices when executed for all at once (e.g. with `entity_id: all`). The actual commands are correctly sent to all devices, it's just the state shown in HA which isn't updated for all.

What doesn't work:
- ~~Nothing related to tilt works.~~
  **Update:** Fixed in HA 2023.11. ZHA didn't support tilt for covers at all before that version.
- Cover state only reports "opening" or "closing" when triggerd by HA commands, never when triggered by the device itself (by pressing the hardware switch buttons).
  But the cover position attribute does change correctly. And I do see zha_events being triggered when the device starts/stops the blinds. See examples below.
  This might be related to home-assistant/core#98933
- It looks like the device is identified as a window. At least it's shown with a window icon.

**Describe the solution you'd like**
All cover services should work. State and all attributes should be reported correctly.

**Device signature**
```
{
[...]

PR #3686, opened 2025-01-08
Refactor Aqara Roller Driver E1 as v2 quirk for more entities

jeverley:
## Proposed change
This PR refactors the existing Aqara Roller Driver E1 quirk into the new v2 structure to expose the configuration and status attributes, below is a summary of changes:

 - Refactor into v2 quirk format
   - Refactor clusters to use AttributeDefs format
 - Expose attributes as entities:
   - **Motor speed** (select)
   - **Charging status** (binary sensor)
   - **Calibration status** (binary sensor)
 - Represent the device as a **_Rollershade_** WindowCoveringType rather than **_Drapery_**
 - Device cluster attribute handling:
   - Reads/reports of `AnalogOutput: present_value` (current position) are replicated into `WindowCovering: current_position_lift_percentage`
   - Writes to `AnalogOutput: present_value` (go to position) do not update the `WindowCovering: current_position_lift_percentage`
   - Reads of `WindowCovering: current_position_lift_percentage` are redirected to `AnalogOutput: present_value`
 - Window Covering command handling:
   - `up_open`, `down_close`, `stop` are handled by `MultistateOutput: present_value`
   - `go_to_lift_percentage` value is written to `AnalogOutput: present_value`
   - The `current_position_lift_percentage` is read prior to returning command response, this ensures the correct state is shown following a change in direction

**Functional enhancements**

- Users can change the Speed using the select entity
- Users can see if the device is Charging
- Users can see if the device is Calibrated
- The cover uses the Rollershade type
- The redundant number entity for current position is disabled by default
- The redundant switch entity is disabled by default

image

## Additional information

Depends on the following PRs for binary_sensor value conversion:
 - https://github.com/zigpy/zigpy/pull/1540
 - https://github.com/zigpy/zha/pull/360

Depends on the following PR for disabling the redundant entities:
 - https://github.com/zigpy/zigpy/pull/1535
 - https://github.com/zigpy/zha/pull/387

Associated PRs:
 - Fixes cover handling of Rollershade device class (for the correct icon) https://github.com/zigpy/zha/pull/376

Whilst I've validated that is works with my own E1 roller shade devices, I'd like to ensure other users do not encounter any regressions with this refactor.
@dmulcahey / @schwickster / @javicalle / @badrpc / @TheJulianJES I can see in the past you've previously contributed for this device, if you still use it would you mind checking if this updated quirk works as expected without introducing regressions?

## Checklist


jeverley:
Have identified an issue with the cover state not reliably changing to open/closed. Investigating this currently.

jeverley:
Edit: https://github.com/zigpy/zha/pull/376 now raised to address the below.



@TheJulianJES, @dmulcahey looking at the code in https://github.com/zigpy/zha/blob/dev/zha/application/platforms/cover/__init__.py
it looks to me like the `_determine_state` function and stored target positions might be the cause of the issue, I'm seeing similar issues reported for other devices which would appear to have the same root cause as below:

- https://github.com/home-assistant/core/issues/93836
- https://github.com/home-assistant/core/issues/98933
- https://github.com/zigpy/zha-device-handlers/issues/1397

There was a PR against the old core ZHA repo to try and address this but that looks abandoned https://github.com/home-assistant/core/pull/105467

I've tested the following scenarios:

**1. _go_to_lift_percentage_ has never been called, _down_close_ is called, then the device reports _position_or_tilt_ of 0**
_target_lift_position == None
position_or_tilt == 0
returns **STATE_CLOSED**

**2. _go_to_lift_percentage_ has never been called, _up_open_ is called, then the device reports _position_or_tilt_ of 100**
_target_lift_position == None
position_or_tilt == 100
returns: **STATE_OPEN**

**3. _go_to_lift_percentage_ is called (go to 0), then the device reports _position_or_tilt_ of 0**
_target_lift_position == 0
position_or_tilt == 0
returns **STATE_CLOSED**

**4. _go_to_lift_percentage_ is called (go to 100), then the device reports _position_or_tilt_ of 100**
_target_lift_position == 100
position_or_tilt == 100
returns **STATE_OPEN**

**5. _go_to_lift_percentage_ is called (go to 50), then the device reports _position_or_tilt_ of 50**
_target_lift_position == 50
position_or_tilt == 50
returns **STATE_OPEN**

**6. _go_to_lift_percentage_ has previously been called (stored value is 50), _down_close_ is called, then the device reports _position_or_tilt_ of 0**
_target_lift_position == 50
position_or_tilt == 0
returns **STATE_CLOSED**

**7. _go_to_lift_percentage_ has previously been called (stored value is 50), _up_open_ is called, then the device reports _position_or_tilt_ of 100**
_target_lift_position == 50
position_or_tilt == 100
returns **None, state remains as STATE_OPENING**

**8. _go_to_lift_percentage_ has previously been called (stored value is 50) and current position is 100, _stop_ is called**
_target_lift_position == 100 (the function async_stop_cover sets it to the current cover position)
position_or_tilt == 100
returns **STATE_OPEN**

I'd suggest we can handle this in one of 3 ways:

1. set the stored _target_lift_position to 100 or 0 when calling the _up_open_ or _down_close_ commands respectively
2. clear the stored _target_lift_position when calling the _up_open_ or _down_close_ commands
3. modify the _determine_state function to return STATE_OPEN when `self.is_opening == True and position_or_tilt == 100` is true


Below is the current function for reference:
```
    def _determine_state(self, position_or_tilt, is_lift_update=True) -> None:
        """Determine the state of the cover.

        In HA None is unknown, 0 is closed, 100 is fully open.
        In ZCL 0 is fully open, 100 is fully closed.
        Keep in mind the values have already been flipped to match HA
        in the WindowCovering cluster handler
        """
        if is_lift_update:
            target = self._target_lift_position
            current = self.current_cover_position
        else:
            target = self._target_tilt_position
            current = self.current_cover_tilt_position

        if position_or_tilt == 0:
            self._state = (
                STATE_CLOSED
                if is_lift_update
                else STATE_OPEN
                if self.current_cover_position is not None
                and self.current_cover_position > 0
                else STATE_CLOSED
            )
            return
        if target is not None and target != current:
            # we are mid transition and shouldn't update the state
            return
        self._state = STATE_OPEN
```

jeverley:
I've been working on changes to the ZHA cover entity hander to address the underlying status issue outlined above (it also affects the existing v1 quirk code for this and other devices).

I'll raise a bug and PR to address these over the weekend, current WIP code here:
https://github.com/jeverley/zha/blob/cover-platform-entity-status/zha/application/platforms/cover/__init__.py

jeverley:
@TheJulianJES if I'm reading https://github.com/zigpy/zigpy/pull/1540 I believe I should be able to use the proposed `attribute_converter` functionality to remap the Aqara values for the charging binary_sensor?

If so I'll remove the custom converter I've implemented in this quirk once your PR is merged.

jeverley:
@TheJulianJES the `AnalogOutput` cluster `present_value` on this device is used to write the intended target position, whereas reads/reports on the attribute report the device current position.

Right now ZHA exposes this as a separate number entity in addition to the cover entity.
In the UI the number entity is redundant since interactions & current position can all be handled through the cover.

When adding custom entities in a v2 quirk we're able to mark them as disabled by default, do you think it would be beneficial to extend this to regular cluster attributes?
This would allow us to present devices more uniformly, without exposing redundant entities arising from manufacturers' non-standard implementations.

Critically this would be different to entirely removing the cluster/marking an attribute as unsupported in the cluster, since we'd still want to remap attribute updates/writes from the cover commands to the cluster.

TheJulianJES:
> When adding custom entities in a v2 quirk we're able to mark them as disabled by default, do you think it would be beneficial to extend this to regular cluster attributes?

With https://github.com/zigpy/zigpy/pull/1535, that should be allowed in the future. Do note that the ZHA part wasn't PR'd yet, so it won't do anything for now.
`attribute_converter` support also isn't merged, yet.

jeverley:
@TheJulianJES I think this should be ready for merging, I'll raise a PR for the zha cover status issue outlined in https://github.com/zigpy/zha-device-handlers/pull/3686#issuecomment-2578907234 separately (it's not a pre-req for this change).

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
```diff
@@ -99,218 +245,104 @@ async def command(
...
+    QuirkBuilder(LUMI, "lumi.curtain.acn002")
+    # temporarily commented out due to potentially breaking existing blueprints
+    #    .friendly_name(
+    #        manufacturer="Aqara", model="Roller Shade Driver E1"
+    #    )
+    .prevent_default_entity_creation(endpoint_id=1, cluster_id=AnalogOutput.cluster_id)
+    .prevent_default_entity_creation(
+        endpoint_id=1, cluster_id=MultistateOutput.cluster_id
+    )
+    .removes(OnOff.cluster_id)
```
puddly: For this use case, why remove the cluster entirely? I'm hoping to replace uses of `removes` with `prevent_default_entity_creation`.

jeverley: As far as I can tell this cluster is non-functional.

Interacting with the switch results in an error, so there would be no benefit exposing it.
If it's an optional entity a user might be tempted to enable it and raise a bug report that the switch doesn't work.

puddly: I think `removes(OnOff.cluster_id)` can be replaced with `prevent_default_entity_creation(cluster_id=OnOff.cluster_id)`. Removing the cluster prevent prevents automatic entity discovery, which is more explicit with `prevent_default_entity_creation` and would allow you to still interact with the `OnOff` cluster via cluster commands if necessary.

jeverley: I'd usually agree, but the device itself is responding with the below error to **any** command to OnOff:

Failed to perform the action zha/issue_zigbee_cluster_command. Failed to issue cluster command with status: <Status.UNSUPPORTED_CLUSTER: 195>

jeverley: @puddly if we're planning to deprecate `removes` then happy to change this to use `prevent_default_entity_creation`.

jeverley:
I've now raised https://github.com/zigpy/zha/pull/376 to address the ZHA logic issues I outlined in https://github.com/zigpy/zha-device-handlers/pull/3686#issuecomment-2578907234.
This isn't a pre-requisite for merging this PR.

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
jeverley: I've updated the PR to use `prevent_default_entity_creation` https://github.com/zigpy/zha-device-handlers/pull/3686/commits/872442a3be297f93968ec0d9519d8cb2b33dc43f

jeverley: Hi @puddly, did you have any other feedback on these changes?

puddly: Not at the moment, sorry. The above method depends on https://github.com/zigpy/zha/pull/365 to be finished and there were a few hiccups getting it merged.

jeverley:
Will update the tests following https://github.com/zigpy/zha-device-handlers/pull/3686/commits/9e5b67ac9a7749c1bfb3f95827462da4cede735c later this evening.

Edit: _fixed and ready with https://github.com/zigpy/zha-device-handlers/pull/3686/commits/d342b64b61c25395e5ae7bb76f9f94ddd3a6bc62_

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
```diff
@@ -94,223 +240,121 @@ async def command(
...
+    )
+    .binary_sensor(
+        XiaomiAqaraRollerE1.AttributeDefs.charging.name,
+        XiaomiAqaraRollerE1.cluster_id,
+        device_class=BinarySensorDeviceClass.BATTERY_CHARGING,
+        translation_key="charging",
+        fallback_name="Charging",
+        attribute_converter=lambda x: True
+        if x == AqaraRollerDriverCharging.true
+        else False,
```
TheJulianJES:
```suggestion
        attribute_converter=lambda x: x == AqaraRollerDriverCharging.true,
```

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
```diff
@@ -2,64 +2,194 @@
...
-    XiaomiCustomDevice,
     XiaomiPowerConfigurationPercent,
 )
 
 
+class AqaraRollerDriverCharging(t.enum8):
+    """Aqara roller driver charging status attribute values."""
+
+    true = 0x01
+    false = 0x02
```
TheJulianJES: This should really be (upper-case) `True` and `False`, but that would conflict.
Maybe something like `Charging` and `Unplugged`? (assuming "not charging" => "unplugged").

Technically, we don't even need this at all, as we could just parse `x == 1` for the charging binary sensor and have no enum.

TheJulianJES (approved):
Not sure if I'm a huge fan of the `is_write`/`WriteAwareUpdateAttribute` stuff and `RedirectAttributes` (we do need a good way of doing this in general), but it seems to look fine. Since the ZHA PR is in, we should probably get this in as well.

I'm assuming that you have tested it quite a bit. Unfortunately, I don't have any cover devices, so unable to test anything. Again, looks fine, so I'll quickly correct the two small things I noted and get it in for the beta tomorrow. If there are any huge issues, we can always fix them during the beta or revert if necessary.

PR #3689, opened 2025-01-09
Add Inovelli switch VZM30-SN

InovelliUSA:
## Proposed change

Inovelli - Adding Support For VZM30-SN Switch

## Checklist


TheJulianJES (approved):
Thanks! I'll merge this for now, so we don't completely loose out of support, but:

We really need to use the new zigpy `AttributeDefs` style to define attributes (and `ServerCommandDefs` for commands) in the future.
We should also consider converting these to quirks v2—I think there are already one or two PRs open for that.

PR #3693, opened 2025-01-09
Add Tuya air sensor `_TZE200_mja3fuja` variant

duhow:
## Proposed change

After adding this change, the device reports all sensors available (VOC, CO2, Formaldehyde, Hum, Temp).
Otherwise it would only report VOC.

## Checklist


TheJulianJES (approved):
Thanks!

PR #3694, opened 2025-01-10
add 3r py

3reality-support:
Remove the poll control because the coordinator sent the poll control, which caused the device to malfunction


3reality-support:
Due to time constraints, I hope this can be merged as soon as possible and a new version can be released to fix the issue of not being able to OTA. Thank you

PR #3701, opened 2025-01-11
Add Niko Connected single/double switch and dimmer

RubenVerborgh:
## Proposed change
Add support for the Niko [Connected single switch](https://www.niko.eu/en/products/niko-home-control/products-for-traditional-wiring/connected-single-switch-10-a-zigbee-productmodel-niko-6f86be91-a2f1-5421-a5ce-234092bfe9c3) (`552-721X1`) and [Connected double switch](https://www.niko.eu/en/products/niko-home-control/products-for-traditional-wiring/connected-double-switch-2-x-10-a-max-16-a-in-total-zigbee-productmodel-niko-c9127021-e751-5640-9021-a6b458788b2e) (`552-721X2`).

### Request and previous work
- This PR implements https://github.com/zigpy/zha-device-handlers/issues/2286, informed by https://github.com/zigpy/zha-device-handlers/pull/2723
- Not all switches were working with previous approaches (https://github.com/Koenkk/zigbee2mqtt/issues/13737), so I [re-performed an analysis of the Niko protocol](https://github.com/Koenkk/zigbee2mqtt/issues/13737#issuecomment-2563955184), which I incorporated into this PR.

### Functionality
- These Niko switches actually consist of three kinds of entities, which are coupled in the default configuration but can be decoupled:
  - 1 or 2 **switches**
  - 2 or 4 **buttons** (each on-device button can have its external extension button)
  - 1 of 2 **status LEDs**
- This PR exposes each of those entities separately, such that they can be controlled individually:

|||
|---|---|
|||

### Events
- The buttons additionally generate events (short and long presses):

### Configuration
- Several configuration options are exposed as entities:

## Checklist

### DIFF
--- a/tests/test_niko.py
+++ b/tests/test_niko.py
@@ -23,12 +23,17 @@ SWITCH_DOUBLE = (
     {"endpoint_ids": [0, 1, 2, 242]},
 )
 
+DIMMER = (
+    ["NIKO NV", "Connectable dimmer,3-200W,2-wire"],
+    {"endpoint_ids": [0, 1, 242]},
+)
+
 
 # pylint: disable=R0903
 class TestNikoSwitch:
     """Tests for Niko Connected Switches (552-721X1 and 552-721X2)."""
 
-    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE])
+    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE, DIMMER])
     class TestClusters:
         """Test whether all clusters are present and complete."""
 
@@ -91,7 +96,7 @@ class TestNikoSwitch:
                 0x0100: foundation.ZCLAttributeAccess.from_str("rw"),
             }
 
-    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE])
+    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE, DIMMER])
     # pylint: disable=R0903
     class TestConfiguration:
         """Test device configuration."""
@@ -237,7 +242,7 @@ class TestNikoSwitch:
                 },
             }
 
-    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE])
+    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE, DIMMER])
     class TestButtonState:
         """Test button state and associated events."""
 
@@ -379,7 +384,7 @@ class TestNikoSwitch:
                         },
                     )
 
-    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE])
+    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE, DIMMER])
     class TestLedState:
         """Test reading and writing status LED state."""
 
@@ -519,7 +524,7 @@ class TestNikoSwitch:
                 attr = config_cluster.get(0x0105)
                 assert (0 if attr is None else attr) == case["result"]
 
-    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE])
+    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE, DIMMER])
     class TestLedSync:
         """Test reading and writing status LED synchronization state."""
 
@@ -597,7 +602,7 @@ class TestNikoSwitch:
                     attr = config_cluster.get(0x0107)
                     assert (0 if attr is None else attr) == expected
 
-    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE])
+    @pytest.mark.parametrize("switch", [SWITCH_SINGLE, SWITCH_DOUBLE, DIMMER])
     class TestLedsAlert:
         """Test reading and writing status LED alert state."""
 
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Feel free to already test and provide feedback; perhaps some points before this PR is merged.

## Improvements

- The `apply_custom_configuration` (https://github.com/zigpy/zha-device-handlers/pull/3701/files#diff-8c13a1ac582c2c486609043c4ca29ce34c03f0a9e07000e97800f953921817d7R38) does not seem to be fired / work as expected. The goal is that a couple of initialization parameters are written, using a pattern I've copied from some other quirks in the repo (grep for `attr_config`).
- Similarly, `bind` (https://github.com/zigpy/zha-device-handlers/pull/3701/files#diff-8c13a1ac582c2c486609043c4ca29ce34c03f0a9e07000e97800f953921817d7R156) does not seem to be called, and the quirk is relying on it for pre-loading values. This may or may not be related to https://github.com/zigpy/zha-device-handlers/pull/3704, but again I've copied the pattern from other quirks.

## Functionality
- Currently, I'm exposing the option to have different colored lights (white / red / blue / purple) as an _alert_ functionality, which seems to be the original intention. However, I could also expose this as a choice for indicator light color, such that, for instance, an outside toilet/bathroom switch could light up in red rather than white to indicate occupation. This is trivial for the one-LED version, and feasible for the two-LED version. But I'm not sure to what extent it is desirable to encode such more advanced behavior in quirks, as opposed to directly translating the manufacturer's intentions.

Any feedback on the above would be helpful!

TheJulianJES (commented):
I'll need to go through this properly when I find some free time, but one thing:
`Bus` shouldn't really be needed anymore. The docs/README really have to be updated, but I just haven't found the time yet.

You can access `update_attribute` and other methods on a different cluster directly (from your current cluster) using something like this:
https://github.com/zigpy/zha-device-handlers/blob/b15a6dc4a63061d9c3d7075002b0fd465f70f020/zhaquirks/xiaomi/aqara/motion_agl1.py#L137-L140
You can also call methods on a different endpoint with `self.endpoint.device.endpoints[2].ias_zone.do_something()` for example (IIRC).

 `ias_zone` is the `ep_attribute` of a cluster. You can also access it with `self.endpoint.in_clusters[IasZone.cluster_id]` e.g.

Doing this might avoid needing some of the tasks? Again, I have to look at this properly some time soon, but it seems a bit weird looking at it now.

Same goes for the inheritance with `NikoCluster`/the other Niko clusters.
And the `NikoQuirkBuilder` also is a bit out of the usual pattern that we wanted to have with "quirks v2", though I see why you're doing it.

Again, this isn't a review. Just some thoughts whilst having a very quick look at this.
Not sure why `apply_custom_configuration` wouldn't be working.. Your HA version is somewhat up-do-date? 
You could also try overriding it in the `CustomDeviceV2` class: https://github.com/zigpy/zigpy/blob/08f8f99df3d235aa0c0154331cb7c5b3dffde3e1/zigpy/quirks/__init__.py#L119-L137

ccantill:
The same quirk also appears to work fine with the NIKO zigbee dimmer:

```
class NikoDimmer(NikoSwitch):
    """Niko Connected Dimmer 552-722X1."""

    model = "Connectable dimmer,3-200W,2-wire"
    model_friendly = "Connected dimmer"
    button_count = 2
```

RubenVerborgh:
Thanks @TheJulianJES for looking at this!

> I'll need to go through this properly when I find some free time, but one thing: `Bus` shouldn't really be needed anymore. 

I see! I rewrote the code without `Bus`.
I still kept a listener-like mechanism to avoid inter-cluster dependencies.

> Same goes for the inheritance with `NikoCluster`/the other Niko clusters.

The main thing this class now does is implementing the listener mechanism.

> And the `NikoQuirkBuilder` also is a bit out of the usual pattern that we wanted to have with "quirks v2", though I see why you're doing it.

Yeah, there's quite some reuse between the switches. While I could have a static `setup_buttons` function somewhere, this would cause us to lose the chaining calling pattern. That's why the `QuirkBuilder` invites subclassing (as also used in `TuyaQuirkBuilder`).

I did get rid of Niko-specific device classes though, by putting all their functionality in the quirk builder.

> Not sure why `apply_custom_configuration` wouldn't be working.. Your HA version is somewhat up-do-date?

Latest versions, yes. Had trouble getting `apply_custom_configuration` and `bind` to run.

----

The code's been updated based on your suggestions. I'll still do some testing myself, and would be happy if others provided feedback as well.

@ccantill That's neat! I don't have a dimmer here to test though. Would it need anything beyond the code you've added? There might be more dimmer-specific properties.

ccantill:
> @ccantill That's neat! I don't have a dimmer here to test though. Would it need anything beyond the code you've added? There might be more dimmer-specific properties.

Nothing additional needed. There might be more functionalities but I have no idea how to go about discovering those.

ccantill:
Any update on when this is going to be merged?

RubenVerborgh:
Ready from my side. Dozens of devices been working without issue for months, if that counts as a test.

@TheJulianJES?

RubenVerborgh:
Been running fine in 2 homes for close to a year now. What can I do to move this on?

luc-sz:
I've been running this for a few month with the "connected double switch" and 2 dimmers.
As ccantill mentionned, they work just fine.

Not sure if it's the correct way to do it, but I've added this to switch.py to make them work:
```
(
    NikoQuirkBuilder("Connectable dimmer,3-200W,2-wire")
    .friendly_name(manufacturer="Niko", model="Connected dimmer")
    .setup_buttons(2)
    .add_to_registry()
)
```

RubenVerborgh:
@luc-sz Thanks! If that works, then I will just add it to the PR.

RubenVerborgh:
Updated after https://github.com/zigpy/zha-device-handlers/pull/4658

PR #3704, opened 2025-01-12
Cluster.bind is not async

RubenVerborgh:
## Proposed change
[`Cluster.bind`](https://github.com/zigpy/zigpy/blob/dev/zigpy/zcl/__init__.py#L682) is not defined as `async`, yet `CustomCluster` defines it as such. During the development of #3701, I found this to lead to unpredictable behavior where `bind` was not consistently called by ZHA. Pylint also calls this out:

```
% pylint zhaquirks/__init__.py
************* Module zhaquirks
zhaquirks/__init__.py:74:4: W0236: Method 'bind' was expected to be 'non-async', found it instead as 'async' (invalid-overridden-method)
```

## Checklist

### DIFF
--- a/tests/test_danfoss.py
+++ b/tests/test_danfoss.py
@@ -8,6 +8,7 @@ from zigpy.zcl import foundation
 from zigpy.zcl.clusters.hvac import Thermostat
 from zigpy.zcl.foundation import WriteAttributesStatusRecord, ZCLAttributeDef
 
+from tests.common import wait_for_zigpy_tasks
 import zhaquirks
 from zhaquirks.danfoss.thermostat import CustomizedStandardCluster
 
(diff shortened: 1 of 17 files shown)


TheJulianJES:
Nice catch! Interesting this wasn't noticed for so long..

I think `bind` in `Cluster` should be marked as `async` in zigpy though, but I'll check this with the others. (cc @puddly)
We await [`bind()` in ZHA](https://github.com/zigpy/zha/blob/fd6cf2f570f1778a9c0933a56ebf476caafe859d/zha/zigbee/cluster_handlers/__init__.py#L280), as it's a normal request.

[`ZDO.bind()`](https://github.com/zigpy/zigpy/blob/07e0451ce5317185e04b38e43aeb9fb168c62f28/zigpy/zdo/__init__.py#L246-L252) will execute [`ZDO.__getattr__`](https://github.com/zigpy/zigpy/blob/07e0451ce5317185e04b38e43aeb9fb168c62f28/zigpy/zdo/__init__.py#L290-L291). That then calls [`ZDO.request`](https://github.com/zigpy/zigpy/blob/07e0451ce5317185e04b38e43aeb9fb168c62f28/zigpy/zdo/__init__.py#L67) which finally calls
[(async) `Device.request`](https://github.com/zigpy/zigpy/blob/07e0451ce5317185e04b38e43aeb9fb168c62f28/zigpy/device.py#L334), so we do have to `await` this in the end.

The ZDO stuff seems to be unchanged since quite a while, while other parts of zigpy were refactored. This code could be cleaner IMO and some methods should just be marked as async I think (e.g. `ZDO.request` and likely most others).

RubenVerborgh:
Thanks @TheJulianJES, I agree with your assessment and opened https://github.com/zigpy/zigpy/pull/1533 instead.

PR #3706, opened 2025-01-12
Add quirk for CentraLite 3156105 thermostat

gmacon:
## Proposed change

This adds a quirk for the CentraLite 3156105 thermostat, which has the semantics of the Running Mode and the heat/cool bits of Running State swapped.

## Additional information

I'm using this as a custom quirk with my thermostat, and it correctly reports Idle, Heating, and Fan states. ~It's freezing here, so I haven't tried cool, but I see no reason that it wouldn't work.~ Edit: It's warmed up, and I was wrong. See below for details.

This fixes the issue I originally reported [on Discourse](https://community.home-assistant.io/t/centralite-3156105-hvac-action-not-changing-to-idle/823732/4).

## Checklist

### DIFF
--- a/zhaquirks/centralite/cl_3156105.py
+++ b/zhaquirks/centralite/cl_3156105.py
@@ -1,0 +1,154 @@
+"""Device handler for CentraLite 3156105."""
+
+from typing import Any
+
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster, CustomDevice
+from zigpy.zcl.clusters.general import (
+    Alarms,
+    Basic,
+    Groups,
+    Identify,
+    Ota,
+    PowerConfiguration,
+    Time,
+)
+from zigpy.zcl.clusters.hvac import (
+    Fan,
+    RunningMode,
+    RunningState,
+    Thermostat,
+    UserInterface,
+)
+
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
+    OUTPUT_CLUSTERS,
+    PROFILE_ID,
+)
+
+
+class CustomThermostatCluster(CustomCluster, Thermostat):
+    """Custom Thermostat cluster.
+
+    With the system type set to "Heat Pump":
+
+    System Mode | Fan Mode | Running? | `running_mode` | `running_state`
+    ----------- | -------- | -------- | -------------- | ---------------
+    Off         | Auto     | No       | Off (0)        | Heat (1)
+    Heat        | Auto     | No       | Off (0)        | Heat (1)
+    Heat        | Auto     | Yes      | Heat (4)       | Heat + Fan (5)
+    Heat        | On       | No       | Off (0)        | Heat + Fan (5)
+    Heat        | On       | Yes      | Heat (4)       | Heat + Fan (5)
+    Cool        | Auto     | No       | Off (0)        | Heat (1)
+    Cool        | Auto     | Yes      | Cool (3)       | Heat + Cool + Fan (7)
+    Cool        | On       | No       | Off (0)        | Heat + Fan (5)
+    Cool        | On       | Yes      | Cool (3)       | Heat + Cool + Fan (7)
+
+    Note: The combination System Mode Off and Fan Mode On isn't supported.
+
+    To fix this up,
+    we correct the `running_state` based on the `running_mode`.
+    The `running_mode` is the most recent value of `system_mode` that fits.
+
+    """
+
+    raw_running_mode: RunningMode
+    corrected_running_mode: RunningMode
+    raw_running_state: RunningState
+
+    def __init__(self, *args, **kwargs):
+        """Set up state."""
+        super().__init__(*args, **kwargs)
+        self.raw_running_mode = RunningMode.Off
+        self.corrected_running_mode = RunningMode.Off
+        self.raw_running_state = RunningState.Idle
+
+    def _corrected_running_state(self):
+        masked_running_state = self.raw_running_state & ~(
+            RunningState.Heat_State_On | RunningState.Cool_State_On
+        )
+        if self.raw_running_mode == RunningMode.Off:
+            corrected_running_state = masked_running_state
+        elif self.raw_running_mode == RunningMode.Cool:
+            corrected_running_state = masked_running_state | RunningState.Cool_State_On
+        elif self.raw_running_mode == RunningMode.Heat:
+            corrected_running_state = masked_running_state | RunningState.Heat_State_On
+        else:
+            raise RuntimeError("Unreachable!")
+
+        return corrected_running_state
+
+    def _update_attribute(self, attrid: int, value: Any) -> None:
+        if attrid == self.AttributeDefs.running_state.id:
+            assert isinstance(value, int)
+            self.raw_running_state = value
+            rs = self._corrected_running_state()
+            super()._update_attribute(self.AttributeDefs.running_state.id, rs)
+        elif attrid == self.AttributeDefs.running_mode.id:
+            assert isinstance(value, int)
+            self.raw_running_mode = value
+            rs = self._corrected_running_state()
+            super()._update_attribute(self.AttributeDefs.running_state.id, rs)
+        elif attrid == self.AttributeDefs.system_mode.id:
+            assert isinstance(value, int)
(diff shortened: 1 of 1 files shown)


gmacon:
OK, here's a slight modification. I noticed that the updates were instantaneous when the fan mode was "Auto", but delayed when it was "On" and realized this was because the incorrect running_state implementation in the thermostat resulted in no change to the running state when the system stopped. This updates both attributes whenever either is updated, which should prevent the delay.

gmacon:
It's gotten warmer here, and I no longer believe this is 100% correct. When the system is running in cooling mode, it's reporting heat. This makes me suspect that something unusual is happening related to my home having a reversible heat pump...

gmacon:
Now I have no idea what the firmware developers were thinking. With my heat pump:

System Mode | Fan Mode | Running? | `running_mode` | `running_state`      
----------- | -------- | -------- | -------------- | ---------------      
Off         | Auto     | No       | Off (0)        | Heat (1)             
Heat        | Auto     | No       | Off (0)        | Heat (1)             
Heat        | Auto     | Yes      | Heat (4)       | Heat + Fan (5)       
Heat        | On       | No       | Off (0)        | Heat + Fan (5)       
Heat        | On       | Yes      | Heat (4)       | Heat + Fan (5)       
Cool        | Auto     | No       | Off (0)        | Heat (1)             
Cool        | Auto     | Yes      | Cool (3)       | Heat + Cool + Fan (7)
Cool        | On       | No       | Off (0)        | Heat + Fan (5)       
Cool        | On       | Yes      | Cool (3)       | Heat + Cool + Fan (7)

Edited to add:

I know @bbenne10 has one of these thermostats and a conventional AC/furnace system. Can you chime in with the corresponding table for your system when you get a chance? Thanks!

bbenne10:
@gmacon: How did you go about grabbing these values? As discussed (out-of-band, for anyone else watching), I'm happy to report back, but I'm not sure I see a way to grab these values through either a dashboard or log graph. Based on your table and reading the diff, some assumptions:

* "system mode" is the explicitly user-set preferred heating mode (off, heat, cool)
* "fan mode" is the explicitly user-set fan state (on, off)
* "running" is...well, whether the unit has decided it should be running, given the user's preferred set-point, the temperature sensor's detected temperature, and the current system mode.
* "running mode" is a bit-value representation of the current system mode.
* "running state" is intended to be the bitmasked combination of system mode and fan mode (and is assumed to be zeroed when the unit is not running)

Is this right? Reading your table, I agree that it looks like the `Heat` bit is maybe being abused to indicate *either* "I am currently in heat mode" or "I am in heat *pump* mode"...

gmacon:
That's not obvious, is it?

With the quirk disabled, on the device page, in the three dots menu on the "Device info" card, select "Manage zigbee device". In the dialog, select the "Thermostat" cluster, then in the attribute menu select `running_mode` or `running_state` as appropriate, and push "Read Attribute". Here are the enum definitions for these two attributes:

https://github.com/zigpy/zigpy/blob/a082f1257c0acebcdb3cfa3a1bff7535c5a9fa12/zigpy/zcl/clusters/hvac.py#L310-L324

bbenne10:
I should be able to get that to you shortly (likely over lunch today).
I'll have to disable the version of the quirk I have but that should be trivial. 

Thanks for the hand-holding :)

bbenne10:
Sorry for the delay. Thought I could do this during lunch, but life had other plans.

I thought this would be simple, but it is getting a bit more interesting.
Without the quirk, my unit refuses to stay in `auto` fan mode while running.
It immediately sets itself to `On` (maybe to report that it is, in fact, running?)
This leads me to question the entries below with fan mode as `auto`, but they do represent what I observed without manually changing the fan mode value to `on`.

Even still: this is about what I would expect given the enum definitions you linked (and a new working understanding of the two variables involved).


System Mode | Fan Mode | Running? | running_mode | running_state
-- | -- | -- | -- | --
Off | Auto | No | Off (0) | Off (0)
Heat | Auto | No | Off (0) | Off (0)
Heat | Auto | Yes | Heat (4) | Heat + Fan (5)
Heat | On | No | Off (0) | Fan (4)
Heat | On | Yes | Heat (4) | Heat + Fan (5)
Cool | Auto | No | Off (0) | Off (0)
Cool | Auto | Yes | Cool (3) |Cool + Fan (6)
Cool | On | No | Off (0) | Fan (4)
Cool | On | Yes | Cool (3) | Cool + Fan (6)

gmacon:
Interesting. It clearly *is* to do with being in heat pump mode. Except for Heat-On-No, your `running_state` looks like it's already correct.

Now that you've pointed it out, I've also noticed that HA displays the fan mode as "on" when the system is running... maybe there's another thing we need to override for that, too...

bbenne10:
I have updated my table - you noticed a mistake (right around then was when I started noticing the fan mode switching in the UI, so I got distracted and didn't check the right table row). I tried again and - as you'd expect - it is actually `Fan (4)`. These state transitions seem to work *fine*. The heatpump is *clearly* doing something weird to this device.

gmacon:
Based on that, I now speculate that I have bad hardware, specifically, the running_state bit 0 is stuck on! Does HA behave normally with the quirk disabled for you? (Other than the fan auto/on thing we previously mentioned, of course.)

bbenne10:
Best I could tell it was working okay, but I didn't futz with it too much other than to generate the table. I'll have a go at it this evening.

gmacon:
I've discovered that I'm not the only person to have observed this, and it *does* appear to be related to the thermostat being in heat pump mode. There's a cluster of issues/PRs in zigbee-herdsman-converters, e.g. https://github.com/Koenkk/zigbee-herdsman-converters/issues/8605 about this.

In theory, we could use the system_type_config attribute to figure out if correction is needed based on whether the system is a heat pump or not, but my thermostat is reporting "None" for that value, which I believe means that the thermostat is not providing that attribute. Sigh.

Devbymitch:
I use zigbee2mqtt not zigpy but I just added a custom option to select if you have the device in heat pump mode or not, which determines whether to use the translation or not provided by this [PR](https://github.com/Koenkk/zigbee-herdsman-converters/pull/4317/files). Not sure if that translation is correct or not, but that might help you here.

Some attributes are not reported by default by this device but I could not find one to request manually that reports what mode it's in. I wasn't able to set mine to heat pump mode on device (assuming because I don't have the wires for it connected). This is also my first foray into zigbee so some attribute might exist I didn't find.

Just a data point that hopefully helps you, good luck, would appreciate a mention or comment over there if you find an attribute dictating if it's in heat pump mode or not.

bbenne10:
I have mentioned this to @gmacon out of band, but with a standard AC/furnace, this device seems working fine without this quirk.

lanrat:
I have this same thermostat and it has been working fine with heat-only mode without this quirk. However the battery level never seems to be reported and is always "unknown". Have you been able to get battery level measurements working?

bbenne10:
Battery measurement is reporting unknown here too. I poked around in the clusters and nothing seems to be reporting anything except MAYBE `battery_voltage` in `Power Configuration` (no idea what this actually means for the device or protocol. I'm just prodding things), which is reporting 33 for me. This seems incorrect...

lanrat:
My `battery_voltage` reports 25, and if I read the voltage across all of the batteries, I read 2.45v. 

image

@bbenne10 can you test if your batteries are at ~3.3v? If so, I it looks like `battery_voltage`/10 is the real voltage.

gmacon:
Spec says (for whatever that's worth):

> The BatteryVoltage attribute specifies the current actual (measured) battery voltage, in units of 100mV.
>
> The value 0xff indicates an invalid or unknown reading.

bbenne10:
33 would make sense if counted in 100mV increments. I am solo dadding right now, but I will try to whip out the multimeter when they go down to sleep later.

nickyivyca:
I have tried the attached quirk, seems to be working well for me with a heat pump setup. My battery voltage reads out at 33 but multimeter says ~3.1V.

bbenne10:
@nickyivyca Does yours work *without* the quirk? I think we basically established that @gmacon's is/was bad?

nickyivyca:
@bbenne10 no, without the quirk it would always show in heating mode (though otherwise it seemed to work fine).

bbenne10:
Oh interesting! It isn't just his. I wonder if it is a specific revision of this family of devices that has this quirk.

PR #3719, opened 2025-01-17
Add Telink devices with custom `ZigbeeTlc` firmware quirk

tr4nt0r:
## Proposed change

Adds quirk for devices that are originally Bluetooth devices but can be modified to use ZigBee with a [custom firmware](https://github.com/pvvx/ZigbeeTLc) 

- MiaMiaoCe MHO-C401N
- MiaMiaoCe MHO-C401
- MiaMiaoCe MHO-C122-z
- Xiaomi LYWSD03MMC
- Qingping CGDK2-z
- Tuya TS0201-z
- Tuya TH03Z-z
- Tuya ZTH01-z
- Tuya ZTH02-z
- Tuya LKTMZL02-z
- Sonoff TH03-z

Edit: Some of these are already Zigbee devices but can be differentiated by the `-z` postfix from the ones with original fw, or in case of the tuya devices because they actually also have the correct model and manufacturer set ^^

## Checklist

### DIFF
--- a/zhaquirks/custom/telink.py
+++ b/zhaquirks/custom/telink.py
@@ -7,7 +7,7 @@ from zigpy.quirks.v2 import QuirkBuilder
 from zigpy.quirks.v2.homeassistant import UnitOfTemperature, UnitOfTime
 import zigpy.types as t
 from zigpy.zcl import ClusterType
-from zigpy.zcl.clusters.hvac import UserInterface
+from zigpy.zcl.clusters.hvac import ScheduleProgrammingVisibility, UserInterface
 from zigpy.zcl.foundation import ZCLAttributeDef
 
 from zhaquirks import CustomCluster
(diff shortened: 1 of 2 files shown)


tr4nt0r:
ZHA seems to be ignoring the `mode=box` setting and defaulting to `auto` instead

zhaquirks/custom/telink.py
```diff
@@ -0,0 +1,208 @@
...
+        multiplier=0.01,
+        mode="box",
+    )
+    .number(
+        CustomUserInterfaceCluster.AttributeDefs.humidity_offset.name,
+        CustomUserInterfaceCluster.cluster_id,
+        min_value=-327.67,
+        max_value=327.67,
+        step=0.01,
+        # unit=PERCENTAGE,
```
TheJulianJES: Support for all units was merged in ZHA with https://github.com/zigpy/zha/pull/352, so we can add this now (and below), as the next HA/ZHA versions will support this.

TheJulianJES (approved):
LGTM. Thanks!

zhaquirks/custom/telink.py
```diff
@@ -0,0 +1,209 @@
...
+        unit=UnitOfTime.SECONDS,
+        translation_key="measurement_interval",
+        fallback_name="Measurement interval",
+        mode="box",
+    )
+    .switch(
+        CustomUserInterfaceCluster.AttributeDefs.display.name,
+        CustomUserInterfaceCluster.cluster_id,
+        off_value=1,
+        on_value=0,
```
TheJulianJES, line 196: Actually, I think we could also use `force_inverted` for this? Shouldn't really make a difference though.

PR #3728, opened 2025-01-19
Enhance TuyaQuirkBuilder battery method

prairiesnpr:
## Proposed change

As we add v2 Tuya quirks, we also keep adding PowerConfiguration clusters. This allows calling `tuya_battery` and specifying the battery type, quantity, and voltage. 

If voltage is omitted, it will be looked up from a dict of known voltages.

Maintains backwards compatibility by allowing a PowerConfiguration cluster to be specified.

```python
.tuya_battery(dp_id=2, battery_type=BatterySize.AA, battery_qty=4)
```

## Checklist

### DIFF
--- a/tests/test_tuya_builder.py
+++ b/tests/test_tuya_builder.py
@@ -78,17 +78,25 @@ async def test_convenience_methods(device_mock, method_name, attr_name, exp_clas
 
 
 @pytest.mark.parametrize(
-    "power_cfg,battery_type,battery_qty,battery_voltage",
+    "power_cfg,battery_type,battery_qty,battery_voltage,"
+    "expected_size,expected_qty,expected_voltage",
     [
-        (TuyaPowerConfigurationCluster2AAA, None, None, None),
-        (None, BatterySize.CR123A, 1, 60),
-        (None, BatterySize.CR123A, 1, None),
-        (None, BatterySize.AA, None, None),
-        (None, None, None, None),
+        (TuyaPowerConfigurationCluster2AAA, None, None, None, BatterySize.AAA, 2, 15),
+        (None, BatterySize.CR123A, 1, 60, BatterySize.CR123A, 1, 60),
+        (None, BatterySize.CR123A, 1, None, BatterySize.CR123A, 1, 30),
+        (None, BatterySize.AA, None, None, BatterySize.AA, None, None),
+        (None, None, None, None, None, None, None),
     ],
 )
 async def test_battery_methods(
-    device_mock, power_cfg, battery_type, battery_qty, battery_voltage
+    device_mock,
+    power_cfg,
+    battery_type,
+    battery_qty,
+    battery_voltage,
+    expected_size,
+    expected_qty,
+    expected_voltage,
 ):
     """Test the battery convenience method."""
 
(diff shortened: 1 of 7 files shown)


tests/test_tuya_builder.py
```diff
@@ -74,6 +77,65 @@ async def test_convenience_methods(device_mock, method_name, attr_name, exp_clas
...
+
+        if battery_qty:
+            assert ep.power.get("battery_quantity") == battery_qty
+        else:
+            assert ep.power.get("battery_quantity") is None
+
+        if battery_voltage:
+            assert ep.power.get("battery_rated_voltage") == battery_voltage
+        elif battery_qty:
+            assert ep.power.get("battery_rated_voltage") == 30
```
TheJulianJES: It might be cleaner to put the expected values into `parametrize` as well.
I'll do that quickly, since I already have it open anyway now.

TheJulianJES: Done with https://github.com/zigpy/zha-device-handlers/pull/3728/commits/32086b58847223e81b926ff5428dba9df0286e50 (hmm, tests did pass locally before I changed the `isinstance` check)

TheJulianJES: We're only checking for instance of `TuyaPowerConfigurationCluster` (always now), not specifically for `TuyaPowerConfigurationCluster2AAA` to avoid an if branch in the test. Should be fine, since we already test the values for that.

TheJulianJES: Ok, I thought `TuyaPowerConfigurationCluster2AAA ` was alsos of `TuyaPowerConfigurationCluster `, but it's obviously not. Fixed that with https://github.com/zigpy/zha-device-handlers/pull/3728/commits/6f9574373781122748980f8f11640738580866b7 and we're now checking for the correct cluster again.

TheJulianJES: Cleaned up slightly more with https://github.com/zigpy/zha-device-handlers/pull/3728/commits/039a47991aed641036ba49d1618d664ed90f5424

zhaquirks/tuya/builder/__init__.py
```diff
@@ -151,10 +167,29 @@ def __init__(
...
+            """PowerConfiguration cluster for Tuya devices."""
+
+            _CONSTANT_ATTRIBUTES = {
+                PowerConfiguration.AttributeDefs.battery_size.id: battery_type,
+                PowerConfiguration.AttributeDefs.battery_rated_voltage.id: battery_voltage,
+                PowerConfiguration.AttributeDefs.battery_quantity.id: battery_qty,
+            }
+
+        if not power_cfg:
+            power_cfg = TuyaPowerConfigurationClusterBattery
```
TheJulianJES: Not that it really matters, but we could avoid creating the class and getting the `battery_voltage` if we're not gonna use it anyway (if a `power_cfg` cluster was explicitly provided).
I'm not completely a fan of how it looks, but yeah:

```python
    def tuya_battery(
        self,
        dp_id: int,
        power_cfg: PowerConfiguration | None = None,
        battery_type: BatterySize | None = BatterySize.AA,
        battery_qty: int | None = 2,
        battery_voltage: int | None = None,
        scale: float = 2,
    ) -> QuirkBuilder:
        """Add a Tuya Battery Power Configuration."""
        if not power_cfg:
            if not battery_voltage and battery_type and battery_qty:
                battery_voltage = battery_voltage or BATTERY_VOLTAGES.get(battery_type)

            class TuyaPowerConfigurationClusterBattery(TuyaPowerConfigurationCluster):
                """PowerConfiguration cluster for Tuya devices."""

                _CONSTANT_ATTRIBUTES = {
                    PowerConfiguration.AttributeDefs.battery_size.id: battery_type,
                    PowerConfiguration.AttributeDefs.battery_rated_voltage.id: battery_voltage,
                    PowerConfiguration.AttributeDefs.battery_quantity.id: battery_qty,
                }
            power_cfg = TuyaPowerConfigurationClusterBattery

        self.tuya_dp(
            dp_id,
            power_cfg.ep_attribute,
            "battery_percentage_remaining",
            converter=lambda x: x * scale,
        )
        self.adds(power_cfg)
        return self
```

prairiesnpr: Thought about this, pushing a change that I think addresses this and maintains readability.

TheJulianJES (approved):
nice!

zhaquirks/tuya/builder/__init__.py
TheJulianJES: yep, that's good!

PR #3743, opened 2025-01-20
Bump ruff to 0.9.4

abmantis:
## Proposed change

SSIA

## Checklist


TheJulianJES (commented):
I'm not sure if we even need this in `requirements_test.txt` at all anymore.
We use the ruff hook for `pre-commit` now ([`.pre-commit-config.yaml`](https://github.com/zigpy/zha-device-handlers/blob/dev/.pre-commit-config.yaml)), so we might be able to drop this completely from `requirements_test.txt`.

(There's an open PR to update the ruff hook from 0.8.3 to 0.9.2 [here](https://github.com/zigpy/zha-device-handlers/pull/3631).)

abmantis:
The downside of not having it on `requirements_test.txt` is that then it will use the system ruff during development which may be a different version than the one we have on `pre-commit`.

prairiesnpr:
> The downside of not having it on `requirements_test.txt` is that then it will use the system ruff during development which may be a different version than the one we have on `pre-commit`.

I'm also in favor of keeping ruff in `requirements_test.txt` and keeping it in sync with pre-commit.

TheJulianJES:
> The downside of not having it on `requirements_test.txt` is that then it will use the system ruff during development

So when using the Ruff extension for VS code for example? (I'm mainly not using VS Code, so not sure about this)

`pre-commit run --all-files` automatically fixes/reformats all files and uses the version from the pre-commit config (that's also what the local Git hook automatically does for the current file and what our CI does).

---

In the zigpy repo, we do not include `ruff` at all: https://github.com/zigpy/zigpy/blob/dev/requirements_test.txt, same with the [`uiprotect`](https://github.com/uilibs/uiprotect/blob/main/pyproject.toml) library for example.
In the ZHA repo, it's still included for some reason, but not pinned at a specific version. The CI/tests do not need it there anymore either, so it should be removed too IMO.

The automatic PRs from the pre-commit CI GitHub app will also only update the pre-commit config, not `requirements_test.txt`. That's why I'd rather drop the requirement completely, unless there's good reason otherwise, as we'll always need to manually update keep both in sync otherwise.

abmantis:
> So when using the Ruff extension for VS code for example? (I'm mainly not using VS Code, so not sure about this)

In my case with the ruff lsp on neovim, but the same should apply to vscode.

> The automatic PRs from the pre-commit CI GitHub app will also only update the pre-commit config, not `requirements_test.txt`. That's why I'd rather drop the requirement completely, unless there's good reason otherwise, as we'll always need to manually update keep both in sync otherwise.

Could removing the installation from `pre-commit` and relying on `requirements_test.txt` for dependency setup be an option? I.e., make the CI do `pip install -r requirements_test.txt` and `pre-commit` will only run the checks (but not install anything)?

I am also fine with just removing it from `requirements_test.txt`, but would prefer to have it as part of the dev environment. In the past I got confused as to why my editor was formatting files (using ruff) differently from the pre-commit, before realizing they were using different versions. But again, fine with dropping it too.

abmantis:
@TheJulianJES I just checked how HA does it, and there we do have the version specified on pre-commit and on the requirements file, but the requirements file is automatically generated. We could do that here too, but I am not sure it is worth the trouble.

I still think it is good to have `ruff` on the requirements file, otherwise it will either be missing during development, so there will be no proper lint/formater in the IDE, or it will use the system one which may not match the one in pre-commit. This can lead to weird errors (like the one that triggered me to open this :smile: )

abmantis:
Also added a comment to both files, to make it easier for contributors to keep them in sync. Tested that both files accept comments.

TheJulianJES (approved):
Ok, yeah. Let's just go with this for now.

We don't update pre-commit hooks that often anyways, as they have to reinstall when working on different PRs/branches with older versions.
I'm not sure if the comment in the pre commit config will conflict with the auto PR though. We'll see.

For `requirements_test.txt`, I guess we could set up Dependabot here at some point. Might be a good idea in general.

abmantis:
Thanks @TheJulianJES !

Issue #3654, opened 2024-12-29
[Device Support Request] TS0601 by _TZE200_ny94onlb


Ankker:
### Problem description

The device from Zemismart, SPM02-D2TZ, is not showing the measured parameters (frecuency, power, current, voltage, ...) with ZHA.

Device: TS0601 by _TZE200_ny94onlb (ZEMISMART)
Zigbee Coordinator: CC2652 by Texas Instruments

Home Assistant
Core   2024.12.5
Supervisor   2024.12.0
Operating System   14.1
Frontend   20241127.8

### Solution description

For devices that do not follow the standard defined in the CSA’s ZCL (Zigbee Cluster Library), the ZHA integration relies on a project called “[ZHA Device Handlers (also known as “zha-quirk”)](https://github.com/zigpy/zha-device-handlers)

If you do not want to develop such a “quirk” Python script yourself, you can submit a “device support request” as a new issue to the [ZHA Device Handlers project repository on GitHub](https://github.com/zigpy/zha-device-handlers/issues)

### Device signature

<details><summary>Device signature</summary>

```json
[Paste the device signature here]
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
[Paste the diagnostic information here]
```

</details>

### Logs

<details><summary>Logs</summary>

```python
[Paste the logs here]
```

</details>

### Custom quirk

<details><summary>Custom quirk</summary>

```python
[Paste your custom quirk here]
```

</details>

thomo66:
I got the device working with this work.
[https://app.codecov.io/gh/zigpy/zha-device-handlers/commit/90dcb4e969c96045ad6c12b09dfdf892de75188b](https://github.com/zigpy/zha-device-handlers/issues/url)
Of course, I still had to enter my model number.
Thank you for the fantastic work.

![Image](https://github.com/user-attachments/assets/5dc01c06-5256-4816-a1cd-ce6ca2c3f2b5)

PR #3747, opened 2025-01-21
Add Tuya SPM02V2.5 3-phase energy meter (_TZE200_ny94onlb)

kali:
## Proposed change

This is quirk for a 3P+N electrical meter from tuya. 
See https://www.amazon.fr/dp/B0CG3CDMM7?ref=ppx_yo2ov_dt_b_fed_asin_title

It is a port from:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/b222aebf2c560ebe18d427497fa65c7bc5ddc288/src/devices/tuya.ts#L8964

## Additional information

It is somewhat working, but I'm a beginner in the zha space, so I think I could use a bit of help to bring this to the finish line.

* I decided against hacking on top the existing `ts0601_din_power.py` because my understanding is that the quirk builder is the preferred way to build new quirks. Additionally, the functional overlap between my device and the one supported by the current file did not feel that big. I hope this was not a mistake, please tell me if it was :)
* I can not find out how to specify using tuya_sensor that some specific measurement are given for a given phase. The device measure a voltage, a current and a power on each line, but I guess I need to match some i18n from HA or ZHA so they appear as such. With the current code, everything appears with generic names, so I have three currents, three voltages and four powers (as there is also a total). I would appreciate some guidance there...
* I am not 100% sure the signature guidelines in the main README apply to quirks built with the quick builder.

Related: #3654
Related: #3799

### DIFF
--- a/zhaquirks/tuya/_TZE200_ny94onlb_din_power.py
+++ b/zhaquirks/tuya/_TZE200_ny94onlb_din_power.py
@@ -1,0 +1,154 @@
+"""Tuya Din Power Meter."""
+
+from zigpy.quirks.v2.homeassistant import (
+    UnitOfElectricCurrent,
+    UnitOfElectricPotential,
+    UnitOfEnergy,
+    UnitOfFrequency,
+    UnitOfPower,
+)
+from zigpy.quirks.v2.homeassistant.sensor import SensorDeviceClass, SensorStateClass
+import zigpy.types as t
+
+from zhaquirks.tuya.builder import TuyaQuirkBuilder
+
+(
+    TuyaQuirkBuilder("_TZE200_ny94onlb", "TS0601")
+    .tuya_sensor(
+        dp_id=1,
+        attribute_name="energy",
+        fallback_name="energy",
+        type=t.uint16_t,
+        state_class=SensorStateClass.TOTAL,
+        device_class=SensorDeviceClass.ENERGY,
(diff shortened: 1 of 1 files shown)


TheJulianJES:
Thanks for the PR! In the mean time, some updated v2 quirks for similar 3 phase power meters are here now: https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts0601_power.py

I think you can adapt yours to be similar to those and also add it to that file instead. They map some Tuya datapoints to local/virtual standardized ZCL attributes for the different phases (for which ZHA creates the sensors).
Others are only mapped to custom virtual attributes with the sensors fully being created by the quirk.

Note that some missing localization in HA will cause you to see the device class name, instead of the `fallback_name` if a device class is assigned.
In the end, so after the PR is merged and quirks/ZHA bumped to HA, if both a translation key and fallback name iares assigned, you'll see the (translation of the) fallback name in HA.
Only during the development, you'll always see the default device class name if there's one device class assigned.

See https://github.com/zigpy/zigpy/issues/1523#issuecomment-2564106284 for more information.

If you really want to only show the device class name in HA at the end (e.g. "Voltage"), just remove the translation key and set the device class. The fallback name always has to be present, but won't be used.
Anyways, I don't think you want to do this for this quirk though.

PR #3752, opened 2025-01-22
Add `BatterySize` enum

### LINKED ISSUE #3713: [Device Support Request] _TZE200_pay2byax

tr4nt0r:
## Proposed change

Sets correct battery size and voltage for `_TZE200_pay2byax TS0601`

Fixes #3713
## Checklist

### DIFF
--- a/tests/test_tuya_builder.py
+++ b/tests/test_tuya_builder.py
@@ -8,10 +8,11 @@ from zigpy.quirks.registry import DeviceRegistry
 from zigpy.quirks.v2 import CustomDeviceV2
 import zigpy.types as t
 from zigpy.zcl import foundation
-from zigpy.zcl.clusters.general import Basic, BatterySize
+from zigpy.zcl.clusters.general import Basic
 
 from tests.common import ClusterListener, MockDatetime, wait_for_zigpy_tasks
 import zhaquirks
+from zhaquirks.const import BatterySize
 from zhaquirks.tuya import (
     TUYA_QUERY_DATA,
     TUYA_SET_TIME,
@@ -93,6 +94,9 @@ async def test_convenience_methods(device_mock, method_name, attr_name, exp_clas
         (TuyaPowerConfigurationCluster2AAA, None, None, None, BatterySize.AAA, 2, 15),
         (None, BatterySize.CR123A, 1, 60, BatterySize.CR123A, 1, 60),
         (None, BatterySize.CR123A, 1, None, BatterySize.CR123A, 1, 30),
+        (None, BatterySize.CR2450, 1, None, BatterySize.CR2450, 1, 30),
+        (None, BatterySize.CR2032, 1, None, BatterySize.CR2032, 1, 30),
+        (None, BatterySize.CR1632, 1, None, BatterySize.CR1632, 1, 30),
         (None, BatterySize.AA, None, None, BatterySize.AA, None, None),
         (None, None, None, None, None, None, None),
     ],
--- a/tests/test_xiaomi.py
+++ b/tests/test_xiaomi.py
@@ -59,6 +59,8 @@ from zhaquirks.xiaomi import (
     XiaomiQuickInitDevice,
     handle_quick_init,
 )
+import zhaquirks.xiaomi.aqara.cube
+import zhaquirks.xiaomi.aqara.cube_aqgl01
 import zhaquirks.xiaomi.aqara.driver_curtain_e1
 from zhaquirks.xiaomi.aqara.feeder_acn001 import (
     FEEDER_ATTR,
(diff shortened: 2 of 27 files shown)


zhaquirks/tuya/ts0601_sensor.py
```diff
@@ -118,7 +118,7 @@
         state_class=SensorStateClass.MEASUREMENT,
     )
     .tuya_contact(dp_id=1)
-    .tuya_battery(dp_id=2)
+    .tuya_battery(dp_id=2, battery_type=10, battery_qty=1, battery_voltage=30)
```
MattWestb: What is the reason setting the device red voltage to static 3.0 V ?
The attribute is sent by the device (if its present) the current voltage the device have red and make no scene putting in somthing hard coded in it. 
Also the battery type is having its default voltage so so IF it was some use of its its redundant and displaying somthing that is not in ZCL is only one mess of it.
The same is made in the [zhaquirks/tuya/tuya_siren.py](https://github.com/zigpy/zha-device-handlers/blob/6c5fc226ec94ff7a30c183888217cb5e0ed27be2/zhaquirks/tuya/tuya_siren.py#L170)

MattWestb (changes requested):

zhaquirks/tuya/ts0601_sensor.py
TheJulianJES: It sets the rated battery voltage, see `_CONSTANT_ATTRIBUTES` here: https://github.com/zigpy/zha-device-handlers/blob/6c5fc226ec94ff7a30c183888217cb5e0ed27be2/zhaquirks/tuya/builder/__init__.py#L191-L215

tr4nt0r: The battery type 10 is not in the ZCL specs and thus there are no default values in the zigpy library, but it is interpreted as `CR2032` in HA. Some other devices, like some from ikea also have battery type 10. Apart from that, this not setting the battery voltage but the battery_rated_voltage.

prairiesnpr: I'd like to see us add an enum for the battery types that are defined in HA, but not the zcl and use that instead of the magic numbers.

tr4nt0r: That would also be a good thing to do but were would I put such an enum?

TheJulianJES: I think we could put it here: https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/const.py
It would also be used in Xiaomi/Aqara quirks.

tr4nt0r: @TheJulianJES I added a BatterySize enum to const.py.  I can proceed and refactor all other quirks to use the enum if that's ok.

MattWestb: > The battery type 10 is not in the ZCL specs and thus there are no default values in the zigpy library, but it is interpreted as `CR2032` in HA. Some other devices, like some from ikea also have battery type 10. Apart from that, this not setting the battery voltage but the battery_rated_voltage.


Wrong its setting the _Attributes of the selected cluster_ `battery_voltage (id: 0x0020)` = the last red device measured voltage in the device. 
From IKEA blinds its reporting `76` and displayed in ZHA as 7.6 V and its one true red value in the device.
Attribute `battery_rated_voltage (id: 0x0034)` is the correct in your case for displaying  the forced rated voltage.
Its being used in the correct way in tuya init https://github.com/zigpy/zha-device-handlers/blob/c82ffed2a1ee72d311396d4d9428cff4a3736a45/zhaquirks/tuya/__init__.py#L919

TheJulianJES: Did you see the comment I posted above? https://github.com/zigpy/zha-device-handlers/pull/3752#discussion_r1926163257 You can scroll down in that code snippet to see the complete `tuya_battery` method and which constant attributes are set.

`tuya_battery` is a method that sets the ZCL `battery_rated_voltage` attribute. It never sets the ZCL `battery_voltage` attribute.

tr4nt0r:
@TheJulianJES so I updated everything now. I think the PR now needs a new name 😂

TheJulianJES (approved):
Nice, thanks!

PR #3753, opened 2025-01-23
Bump zigpy to 0.75.0

TheJulianJES:
## Proposed change
Bump zigpy to [0.75.0](https://github.com/zigpy/zigpy/releases/tag/0.75.0).
Also changes tests to access the quirks registry with the new attribute name(s).

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ readme = "README.md"
 license = {text = "Apache-2.0"}
 requires-python = ">=3.12"
 dependencies = [
-    "zigpy>=0.74.0",
+    "zigpy>=0.75.0",
 ]
 
 [tool.setuptools.packages.find]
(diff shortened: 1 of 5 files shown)


TheJulianJES:
zigpy 0.75.0 accesses the (now) deprecated `registry` attribute internally since https://github.com/zigpy/zigpy/pull/1534.
This should likely be fixed in zigpy (then bumped in quirks again), so our test runs don't all look like this:

<details><summary>Example of a test run with a few warnings</summary>
<p>

image

</p>
</details> 

This should be fixed with https://github.com/zigpy/zigpy/pull/1538 (if bumped in quirks)

TheJulianJES:
This also seems to have broken https://github.com/zigpy/zha-device-handlers/blob/dev/tests/test_quirks_v2.py or v2 quirks in general?

(Also just noticed that the typing for `ALL_QUIRK_V2_CLASSES` in that test is incorrect)

PR #3755, opened 2025-01-23
Only run GitHub stale action once per day

TheJulianJES:
## Proposed change
This makes the GitHub stale action only run once per day. We do not need to run it every hour.

## Additional information
Waiting for a successful run of https://github.com/zigpy/zha-device-handlers/pull/3754 before merging this.
EDIT: v9 seems to work fine. First run: https://github.com/zigpy/zha-device-handlers/actions/runs/12921276628/job/36035002367

## Checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -3,7 +3,7 @@ name: Stale
 # yamllint disable-line rule:truthy
 on:
   schedule:
-    - cron: "0 * * * *"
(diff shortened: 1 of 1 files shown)


TheJulianJES:
We should likely also make the same adjustments (v9 update + only run once per day) to the zigpy repo.
Or just remove the stale action from there entirely, as it has very few issues compared to this repo.

PR #3758, opened 2025-01-23
Test manufacturer/model are unique across v2 quirks

TheJulianJES:
## Proposed change
This adds a test ensuring that the manufacturer/model tuple is unique across all v2 quirks.

## Additional information

Should we use `.filter()` on a v2 quirk in the future, we can just ignore v2 quirks with a filter in the test, either manually or automatically. I'd rather do that when we actually need it, instead of now though.

## Checklist

### DIFF
--- a/tests/test_quirks_v2.py
+++ b/tests/test_quirks_v2.py
@@ -10,8 +10,9 @@ import zhaquirks
 
 zhaquirks.setup()
 
-
-ALL_QUIRK_V2_CLASSES: list[QuirksV2RegistryEntry] = list(
+# zigpy registry v2 contains duplicates (due to being keyed by manufacturer and model),
+# so to avoid duplicates but maintain insertion order, we use a dict instead of a set
+ALL_QUIRK_V2_CLASSES: dict[QuirksV2RegistryEntry, None] = dict.fromkeys(
     itertools.chain.from_iterable(zigpy.quirks.DEVICE_REGISTRY.registry_v2.values())
 )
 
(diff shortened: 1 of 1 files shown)


TheJulianJES:
@prairiesnpr It looks like we already have multiple duplicates. Sometimes, a v2 quirk has the same manufacturer/model added twice, but sometimes, it's also duplicated across multiple quirks (in the Tuya siren quirks).
Due to the test failing as soon as one collision is detected, here's a list of the duplicates:

- https://github.com/zigpy/zha-device-handlers/pull/3759
  - ✅✅ `_TZE204_yjjdcqsq, TS0601` (`tuya_sensor.py:256` - duplicated in same quirk)

- https://github.com/zigpy/zha-device-handlers/pull/3760:
  - ✅ `_TZE200_t1blo2bj, TS0601` (`tuya_siren.py:317`, `tuya_siren.py:362`- in two separate quirks)
  - ✅ `_TZE204_t1blo2bj, TS0601` (`tuya_siren.py:318`, `tuya_siren.py:363`- in two separate quirks)
  - ✅ `_TZE204_q76rtoa9, TS0601` (`tuya_siren.py:319`, `tuya_siren.py:364`- in two separate quirks)

I fixed the first one, but the other two quirks are very similar to another with some differences on the exposed entities, so you might wanna take a look at that.

EDIT: Actually, it seems like the model info for "Tuya ZA03" was copied from the "Neo NAS-AB02B2" siren, but not adjusted. Looked at the Z2M model info and got the correct model now. The entity values also seems to what Z2M has, so this seems like the correct fix. Please double check though 😄

PR #3787, opened 2025-01-26
Only add MCU cluster in `TuyaQuirkBuilder` if attributes are added

prairiesnpr:
## Proposed change

This allows the use of TuyaQuirkBuilder when we don't have an MCU cluster. Allows applying spells for devices that otherwise follow standards. Adds an additional kwarg to force adding the MCU cluster if required.

See: #3781

## Checklist


TheJulianJES (approved):
thanks!

PR #3793, opened 2025-01-27
Ignore Hue dial `RDM002` LevelControl long press events

### LINKED ISSUE #3696: [BUG] Philips RDM002 long press indistinguishable from dial left rotation
### Bug description

Long-pressing buttons will correctly emit long_press/hold events, but they will also emit LevelControl events. Dial rotation will emit only LevelControl events. Due to the args/params attached to the LevelControl events emitted by button long-press, they are indistinguishable from dial left rotation. Dial right rotation uses different arguments, so can be distinguished. See https://github.com/zigpy/zha-device-handlers/pull/2340#issuecomment-2577754688 for more details.

### Steps to reproduce

1. enable debug logging on zha integration, or start listening to `zha_event` in dev tools.
2. perform left rotation on RDM002 dial
3. hold any button for several seconds
4. observe the same step_with_on_off with indistinguishable args/params for both of above actions

### Expected behavior

Ideally holding a button shouldn't emit LevelControl events.

### Screenshots/Video

_No response_

### Device signature

```json
{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4107,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 128,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 128,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0820",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x1000",
        "0xfc00"
      ],
      "output_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0019",
        "0x1000"
      ]
    }
  },
  "manufacturer": "Signify Netherlands B.V.",
  "model": "RDM002",
  "class": "zhaquirks.philips.rdm002.PhilipsRDM002"
}
```
</details>

### Diagnostic information

_No response_

### Logs

_No response_

### Additional information

_No response_

fholzer:
## Proposed change
Override LevelControl's listener_event, and filter out those events that have transition_time == 8, which originate from long-pressing buttons. Rotary events have a transition_time of 4, and won't be muted.

This fixes #3696

## Additional information
Details in #3696

## Checklist

### DIFF
--- a/zhaquirks/philips/rdm002.py
+++ b/zhaquirks/philips/rdm002.py
@@ -1,5 +1,9 @@
 """Signify RDM002 device."""
 
+import logging
+from typing import Any
+
+from zigpy import util
 from zigpy.profiles import zha
 from zigpy.quirks import CustomDevice
 from zigpy.zcl.clusters.general import (
@@ -42,6 +46,8 @@ from zhaquirks.philips import (
     PhilipsRemoteCluster,
 )
 
+_LOGGER = logging.getLogger(__name__)
+
 DIAL_TRIGGERS = {
     (SHORT_PRESS, DIM_UP): {
         COMMAND: COMMAND_STEP_ON_OFF,
@@ -67,6 +73,48 @@ class PhilipsRdm002RemoteCluster(PhilipsRemoteCluster):
         3: Button(BUTTON_3),
         4: Button(BUTTON_4),
     }
+
+
+class PhilipsRdm002LevelControl(LevelControl):
+    """Philips RDM002 LevelControl cluster."""
+
+    def listener_event(self, method_name: str, *args) -> list[Any | None]:
+        """Blackhole requests with transition_time=8, which originate from button long-presses."""
+
+        _LOGGER.debug(
+            "%s - listener_event: method: %s - args: [%s]",
+            self.__class__.__name__,
+            method_name,
+            args,
+        )
+
+        # example args we want to mute:
+        # [(224, 6, step_with_on_off(step_mode=<StepMode.Down: 1>, step_size=255, transition_time=8))]
+        # transition_time=8 when step command is sent for long-press of button 1.
+        # transition_time=4 when dial is rotated.
+        if (
+            method_name == "cluster_command"
+            and len(args) > 2
+            and len(args[2]) > 2
+            and args[2][2] == 8
+        ):
+            _LOGGER.debug(
+                "%s::listener_event - muting level control method: %s - args: [%s]",
+                self.__class__.__name__,
+                method_name,
+                args,
+            )
+
+        else:
+            _LOGGER.debug(
+                "%s::listener_event - forwarding level control method: %s - args: [%s]",
+                self.__class__.__name__,
+                method_name,
+                args,
+            )
+            return util.ListenableMixin.listener_event(self, method_name, *args)
+
+        return []
 
 
 class PhilipsRDM002(CustomDevice):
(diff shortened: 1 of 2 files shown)


fholzer:
Not entirely sure what to do about this failed CustomCluster test...

zhaquirks/philips/rdm002.py
```diff
@@ -69,6 +75,48 @@ class PhilipsRdm002RemoteCluster(PhilipsRemoteCluster):
     }
 
 
+class PhilipsRdm002LevelControl(LevelControl):
```
TheJulianJES:
```suggestion
class PhilipsRdm002LevelControl(CustomCluster, LevelControl):
```
(and import of `CustomCluster` should make it work)

zhaquirks/philips/rdm002.py
```diff
@@ -69,6 +75,48 @@ class PhilipsRdm002RemoteCluster(PhilipsRemoteCluster):
...
+        else:
+            _LOGGER.debug(
+                "%s::listener_event - forwarding level control method: %s - args: [%s]",
+                self.__class__.__name__,
+                method_name,
+                args,
+            )
+            return util.ListenableMixin.listener_event(self, method_name, *args)
+
+        return []
```
TheJulianJES: Can't we just call `super().listener_event(method_name, *args)` (or block that call if it needs to be)?

fholzer: yup, changed to use `super()`

zhaquirks/philips/rdm002.py
```diff
@@ -69,6 +74,48 @@ class PhilipsRdm002RemoteCluster(PhilipsRemoteCluster):
...
+            and len(args) > 2
+            and len(args[2]) > 2
+            and args[2][2] == 8
+        ):
+            _LOGGER.debug(
+                "%s::listener_event - muting level control method: %s - args: [%s]",
+                self.__class__.__name__,
+                method_name,
+                args,
+            )
```
TheJulianJES, line 98: Let's add a `return` after this.

zhaquirks/philips/rdm002.py
```diff
@@ -69,6 +74,48 @@ class PhilipsRdm002RemoteCluster(PhilipsRemoteCluster):
...
+        else:
+            _LOGGER.debug(
+                "%s::listener_event - forwarding level control method: %s - args: [%s]",
+                self.__class__.__name__,
+                method_name,
+                args,
+            )
+            return super().listener_event(method_name, *args)
+
+        return []
```
TheJulianJES: Then we can remove `else` and `return []` and just always `return super()...`.

TheJulianJES (commented):
Also, I think we can safely reduce the log messages a bit. Zigpy probably already logs most of these somewhat, so I'd only log if we block an event from going through.

Also, can we shorten the comments/ZCL frames in the test file a bit? I think the only really important part are the arguments and the command name.

fholzer:
> Also, can we shorten the comments/ZCL frames in the test file a bit? I think the only really important part are the arguments and the command name.

Let me know if that's what you meant wrt. which comments to keep in the test file.

TheJulianJES (approved):
Looks good, thanks!

PR #3808, opened 2025-01-29
Add local clusters to Tuya `_TZ3000_bjawzodf` temp sensor

szogun1987:
Previous commit made this sensor discoverable as no-feature device

## Proposed change
I have device mentioned in the title, it was discoverable but HA couldn't recognize it's features

## Additional information
I've just restored some changes from the past

## Checklist

- [x] The changes are tested and work correctly - tested manually with my device
- [ ] `pre-commit` checks pass / the code has been formatted using Black - edited on smartphone - sorry please format it after merge

### DIFF
--- a/zhaquirks/tuya/ty0201.py
+++ b/zhaquirks/tuya/ty0201.py
@@ -1,7 +1,11 @@
 """Tuya TY0201 temperature and humidity sensor."""
 
 from zhaquirks.tuya import TuyaPowerConfigurationCluster2AA
-from zhaquirks.tuya.builder import TuyaQuirkBuilder, TuyaRelativeHumidity, TuyaTemperatureMeasurement
+from zhaquirks.tuya.builder import (
+    TuyaQuirkBuilder,
+    TuyaRelativeHumidity,
+    TuyaTemperatureMeasurement,
+)
 
 (
     TuyaQuirkBuilder("_TZ3000_bjawzodf", "TY0201")
--- a/zhaquirks/tuya/ty0201.py
+++ b/zhaquirks/tuya/ty0201.py
@@ -1,12 +1,14 @@
 """Tuya TY0201 temperature and humidity sensor."""
 
 from zhaquirks.tuya import TuyaPowerConfigurationCluster2AA
-from zhaquirks.tuya.builder import TuyaQuirkBuilder
+from zhaquirks.tuya.builder import TuyaQuirkBuilder, TuyaRelativeHumidity, TuyaTemperatureMeasurement
 
 (
     TuyaQuirkBuilder("_TZ3000_bjawzodf", "TY0201")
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ty0201.py
```diff
@@ -1,12 +1,18 @@
...
+    TuyaRelativeHumidity,
+    TuyaTemperatureMeasurement,
+)
 
 (
     TuyaQuirkBuilder("_TZ3000_bjawzodf", "TY0201")
     .applies_to("_TZ3000_zl1kmjqx", "TY0201")
     .applies_to("_TZ3000_zl1kmjqx", "")
+    .replaces(TuyaRelativeHumidity)
+    .replaces(TuyaTemperatureMeasurement)
```
TheJulianJES, line 15: These are local clusters that are only required for Tuya devices that exclusively send their data on Tuya (MCU) datapoints (when we convert the Tuya datapoint updates on the fake/local clusters).

We should not need them for this device, as the temperature and humidity clusters on this device seem to (somewhat) work according to ZCL spec.
https://github.com/zigpy/zha-device-handlers/pull/3781 removed these local clusters and added the `tuya_enchantment`, as it seems like the sensor needs to think its connected to the Tuya hub before some data is sent. You might need to re-pair it with those changes in place.

Did you try the changes that were made made in the [linked PR](https://github.com/zigpy/zha-device-handlers/pull/3781) (by installing Home Assistant Core 2025.2.0b0 and re-pairing the sensor)? With them, do the entities not show up?
If you're running Home Assistant Core 2025.1.4 or earlier, the changes from that PR aren't even included.

szogun1987: I use stable release. Device was discovered but without entities on it, so I used this repo as a source of quirk. One from master branch behave like no quirk at all but, as it was shorter than quirks I've seen in the past I started to review file history and found version that works.

TheJulianJES: If you're on HA 2025.1.x, then this quirk v2 version won't do anything at all, as `tuya_enchantment` isn't included in that release yet. You should see warnings in the logs about the quirk not being loaded.

The integrated quirk version in that HA version still contains the local clusters and is probably what causes the issue in the first place. That's why we removed them.

It's likely just the restart of HA made your device work, unrelated to re-adding the local clusters.
When you upgrade to HA 2025.2 eventually, feel free to completely remove the device and then re-pair it (without any custom quirks), so we can see if the issue is indeed fully fixed.

szogun1987: I performed following test:
- renamed quirk to ty0201.__py
- removed device
- restarted HA
- paired device again - no humidity or temperature entities
- renamed quirk back
- removed device
- restarted HA
- paired device one more time - both humidity and temperature available

Issue #3810, opened 2025-01-30
[BUG] IKEA of Sweden RODRET Dimmer quirk not loaded with firmware 1.0.57


jook-network:
### Bug description

I have two IKEA of Sweden RODRET Dimmers. One with firmware 1.0.47 and 1.0.57. For 1.0.47 the quirk zhaquirks.ikea.twobtnremote.IkeaRodretRemote2Btn is applied. For 1.5.7 no quirk is being applied.

When looking at the signatures for both devices. The class differs.

In the ZHA says the following about de device with 1.0.57:

2025-01-29 22:52:50.016 DEBUG (MainThread) [zigpy.quirks.registry] Checking quirks for IKEA of Sweden RODRET Dimmer (d4:48:67:ff:fe:fe:f5:7d)
2025-01-29 22:52:50.017 DEBUG (MainThread) [zigpy.quirks.registry] Considering <class 'zhaquirks.ikea.twobtnremote.IkeaRodretRemote2Btn'>
2025-01-29 22:52:50.017 DEBUG (MainThread) [zigpy.quirks] Fail because input cluster mismatch on at least one endpoint

In the same debug log for device 1.0.47

2025-01-29 22:52:49.990 DEBUG (MainThread) [zigpy.quirks.registry] Checking quirks for IKEA of Sweden RODRET Dimmer (6c:fd:22:ff:fe:66:a8:30)
2025-01-29 22:52:49.990 DEBUG (MainThread) [zigpy.quirks.registry] Considering <class 'zhaquirks.ikea.twobtnremote.IkeaRodretRemote2Btn'>
2025-01-29 22:52:49.990 DEBUG (MainThread) [zigpy.quirks] Device matches filter signature - device ieee[6c:fd:22:ff:fe:66:a8:30]: filter signature[{'models_info': [('IKEA of Sweden', 'RODRET Dimmer')], 'endpoints': {1: {'profile_id': 260, 'device_type': <DeviceType.NON_COLOR_CONTROLLER: 2080>, 'input_clusters': [0, 1, 3, 32, 4096, 64636], 'output_clusters': [3, 4, 6, 8, 25, 4096]}}}]

### Steps to reproduce

1. Open ZHA integration.
2. Find defected device.
3. Examine the info.
4. Notice no quirk is being applied.

### Expected behavior

[...]

PR #3814, opened 2025-01-31
Add IKEA RODRET dimmer v1.0.57 quirk

MattWestb:
## Proposed change
  Adding group on in cluster that is being added i the last firmware update.

## Additional information
Cerdit to @PluueeR

fixing #3810

Tested and quirk is loaded OK and AD is aded on the device card (have not testing them but looks being OK if comparing with the zha_event).

## Checklist

### DIFF
--- a/zhaquirks/ikea/twobtnremote.py
+++ b/zhaquirks/ikea/twobtnremote.py
@@ -263,6 +263,7 @@ class IkeaRodretRemote2Btn(CustomDevice):
 
     device_automation_triggers = IkeaTradfriRemote2Btn.device_automation_triggers.copy()
 
+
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
We may want to consider using a v2 quirk for this in the future.
For now, it's fine. Thanks!

MattWestb:
Thanks.
I dont have the knowledge fixing or doing V2 quirks but i try helping then its possible.

I was finding that Styrbar (4 Button) is having one group in cluster added in later firmware version (one new is in the pipe then its CSA certificated) but is converted to V2 so i cant fixing it like in the V1 quirk (deleting it in replacement so the device cant being added in ZHA groups) that need being fixed for getting the device being "user friendly".

PR #2961, opened 2024-02-01
ts0601_energy_meter quirk for Tuya power meter devices (v1 quirk)

jeverley:
## Proposed change

Adds support for Tuya Energy Meter devices in 1 and 2 clamp configurations.

Uses zigbee2Mqtt https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts as reference for device datapoints (with amendments from community testing in the respective device support issues).

Meter devices providing power direction will report signed values for their supported attributes in accordance with the Zigbee cluster spec (instantaneous demand, active power and reactive power specifically).

**Provides control over metering sensor configuration:**

- The quirk allows the user to control the endpoint 3 power calculation on 2 clamp devices using the "channel_configuration" attribute value (calculated as either _a+b_ or _a-b_).
   - _a+b_ should be used when measuring total consumption from two power sources.
   - _a-b_ should be used to derive a second power value from input power and consumption i.e. clamp A on grid input, clamp B measuring power into the consumer unit.

Some devices report an a+b total natively, however in testing this value is often incorrect due to delay in detecting power direction change on the device.

In the real world when measuring both grid and production an a + b < 0 value would be atypical. 
This is mitigated using the _power_flow_preempt_ filter on device readings, this is optionally is enabled by setting 'power_flow_preempt' to True.

The quirk aims to make adding new devices simpler by utilising device specific dp_to_attribute mappings which feed into common core reporting clusters.

image

**Addresses device support requests:**

2 bidirectional clamps:

- https://github.com/zigpy/zha-device-handlers/issues/2549 (_TZE204_81yrt3lo)
- https://github.com/zigpy/zha-device-handlers/issues/2650 (_TZE200_rks0sgb7)

1 bidirectional clamp:
- https://github.com/zigpy/zha-device-handlers/issues/2420 (_TZE204_ac0fhfiq)

1 clamp:
- https://github.com/zigpy/zha-device-handlers/issues/1973 (_TZE204_cjbofhxw)

@jmuf I've just spotted that in parallel you've also been working on https://github.com/zigpy/zha-device-handlers/pull/2870 for  _TZE200_rks0sgb7, shall we pool ideas?

## Additional information

Associated with merged pre-requisite ZHA metering and electrical measurement changes:

- https://github.com/home-assistant/core/pull/107641
- https://github.com/home-assistant/core/pull/107576
- https://github.com/home-assistant/core/pull/109126

## Checklist

- [ ] The changes are tested and work correctly
[...]

Issue #2650, opened 2023-10-15
[Device Support Request] PC311-Z-TY Single-Phase Power Clamp Zigbee model 'TS0601' and manufacturer name '_TZE200_rks0sgb7


lepalom:
### Problem description

I would like to kindly ask to support this device:

[PC311-Z-TY Single-Phase Power Clamp Zigbee] (https://es.aliexpress.com/item/1005005931365212.html)

This device is the same commented here, in the [zigbee2mqtt project](https://github.com/Koenkk/zigbee2mqtt/issues/18734).

### Solution description

Fully integrated in HA.

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.FullFunctionDevice|MainsPowered|RxOnWhenIdle|AllocateAddress: 142>, manufacturer_code=4098, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00",
        "0xff66"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_rks0sgb7",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
{
  "home_assistant": {
    "installation_type": "Home Assistant Supervised",
    "version": "2023.10.3",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.11.5",
    "docker": true,
    "arch": "armv7l",
    "timezone": "Europe/Andorra",
    "os_name": "Linux",
[...]

Issue #2420, opened 2023-06-01
[Device Support Request] _TZE204_ac0fhfiq TS0601 BiDirectional ZigBee Energy Meter with 150A Clamp Current Sensor


stombi:
### Problem description

Hi I would like to request device support for the _TZE204_ac0fhfiq TS0601 BiDirectional ZigBee Energy Meter with 150A Clamp Current Sensor.

It apears in HA, but without any sensor.

product link : https://www.aliexpress.com/item/1005005466228073.html

### Solution description

Any

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0051",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZE204_ac0fhfiq",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
{
  "home_assistant": {
    "installation_type": "Home Assistant Container",
    "version": "2023.2.5",
    "dev": false,
    "hassio": false,
[...]

Issue #1954, opened 2022-11-22
[Device Support Request] TS0601 PC321-TY 3-phase Clamp Power Meter


Dogmeat88:
Hi I would like to request device support for the TS0601 PC321-TY 3-phase Clamp Power Meter - TZE200_nslr42tt
[https://www.aliexpress.com/item/1005004719888092.html](https://s.click.aliexpress.com/e/_DksL5Lp)

zigbee2mqtt device support thread : [https://github.com/Koenkk/zigbee2mqtt/issues/14790](https://github.com/Koenkk/zigbee2mqtt/issues/14790)

Thanks.

<details>

<summary>Device signature</summary>

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.Router: 1>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress|RxOnWhenIdle|MainsPowered|FullFunctionDevice: 142>, manufacturer_code=4417, maximum_buffer_size=66, maximum_incoming_transfer_size=66, server_mask=10752, maximum_outgoing_transfer_size=66, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=False, *is_full_function_device=True, *is_mains_powered=True, *is_receiver_on_when_idle=True, *is_router=True, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0101",
      "in_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0300",
        "0xef00"
      ],
      "out_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": 41440,
      "device_type": "0x0061",
      "in_clusters": [],
      "out_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZE200_nslr42tt",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

<details>

<summary>Diagnostic information</summary>

```yaml
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2022.11.4",
    "dev": false,
[...]

PR #3824, opened 2025-02-02
ts0601_energy_meter quirk for Tuya power meter devices (v2 quirk)

jeverley:
## Proposed change

Adds support for Tuya Energy Meter devices.

**Reimplementation of my original PR https://github.com/zigpy/zha-device-handlers/pull/2961 using the v2 Tuya quirk builder.**

Uses zigbee2Mqtt https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts as reference for device datapoints (with amendments from community testing in the respective device support issues).

Meter devices providing energy direction will report signed values for their ZCL mapped attributes in accordance with the Zigbee cluster spec (instantaneous demand, active power and reactive power specifically).

To keep the change size manageable this initial PR will not calculate additional power metrics that are not reported by the device (reactive power etc), I can explore re-introducing this if users are keen for the functionality.

**Device configuration entities**

- **Virtual channel:** choose the calculation on 2 clamp devices for calculating the virtual value.
   - Total
   - A minus B
   - B minus A
 - **Calibration:** adjust reported values by a percentage (__TZE204_81yrt3lo_).
 - **Energy direction delay mitigation:** this controls whether the quirk will compensate for delayed direction data point reporting (__TZE204_81yrt3lo_).
   - Automatic (enabled based on device and firmware version)
   - Disabled (values reported immediately)
   - Enabled (value updates are delayed until the next direction is reported).

**Supports 1 clamp devices:**
- **_TZE204_ac0fhfiq**
   - https://github.com/zigpy/zha-device-handlers/issues/2420
   **Prerequisite PR https://github.com/zigpy/zha-device-handlers/pull/3643**
- **_TZE204_cjbofhxw, _TZE284_cjbofhxw**
   - https://github.com/zigpy/zha-device-handlers/issues/1973
   - https://github.com/zigpy/zha-device-handlers/issues/3152

**Supports 2 clamp devices:**
- **_TZE204_81yrt3lo, _TZE284_81yrt3lo**
   - https://github.com/zigpy/zha-device-handlers/issues/2549
- **_TZE200_rks0sgb7**
   - https://github.com/zigpy/zha-device-handlers/issues/2650

**Supports 3 clamp devices:**
 - **_TZE200_nslr42tt**
   - https://github.com/zigpy/zha-device-handlers/issues/1954
 - **_TZE204_v9hkz2yn, _TZE200_v9hkz2yn**

## Additional information

 - Users of the original draft v1 quirk will need to remove and re-add devices to remove redundant entities.
 - Prerequisite pull requests:
   - https://github.com/zigpy/zha-device-handlers/pull/3643
   - https://github.com/zigpy/zha/pull/352
   - https://github.com/zigpy/zigpy/pull/1542
   - https://github.com/zigpy/zigpy/pull/1534
 

## Checklist


jeverley:
This should be ready for testing once https://github.com/zigpy/zha-device-handlers/pull/3643 is merged.

wolsty7:
It might be worth adding support for "_TZE284_81yrt3lo"  https://github.com/zigpy/zha-device-handlers/issues/3658#issuecomment-2643002772

jeverley:
I'm going to split this into two separate PRs and files to make it easier to review and add devices moving forward.
One for devices that require the direction attribute logic (e.g. *_81yrt3lo) and those that don't.

jeverley:
Hi @TheJulianJES,  I'll be looking at progressing support for these devices again next week, but before I do I would much appreciate your opinion on how best to implement these.

So far I've taken the approach of using common Metering/Electrical measurement clusters that handle logic for applying the positive/negative sign to power values (amongst other things) - similar to the draft v1 quirk I'd made last year.

I'm mindful however that with v2 quirks I could instead take the approach of using custom sensor entities - this would allow us to specify a suffix for entity names (i.e. 1/2, or A/B) useful as these devices often have multiple clamps for measuring energy on different circuits.

Also, would you prefer if I broke this out into a PR per device? I don't want to be causing reviewers headaches.

Many thanks, Jack

ouafnico:
Any idea when zha will works with this devices ?

jaromanda:
FYI, there's a **`_TZE284_ac0fhfiq`** which seems to work identically as **`_TZE204_ac0fhfiq`** - support was just added in zigbee2mqtt dev branch, I can confirm that it is functional in Z2M Edge

jaromanda:
Any progress on this?

gillenoarbe:
Hi @jeverley , thanks for the quirk and the long history with this device.

I've been struggling with this device for air conditioning power consumption monitoring to get the data right... With your quirk, main DP seem to be exposed and working (including W and VA power, voltage and frequency.

The problem I am finding is that there is some weird power measurement cap and the deviceisn't capable of measuring (not sure if right or wrong) more than 85W in a >1kW device (measured via external powermeter in the main), idem with 0.5A max of intensity.

Has anyone experienced this? Also the reactive power seems not to work all the time.

Thank you.

Dwouglaz:
Nada ainda para o TZE204_81yrt3lo?

zhaquirks/tuya/ts0601_energy_meter.py
```diff
@@ -0,0 +1,1261 @@
...
+    TuyaZBElectricalMeasurement,
+    TuyaZBMeteringClusterWithUnit,
+)
+from zhaquirks.tuya.builder import TuyaQuirkBuilder
+from zhaquirks.tuya.mcu import TuyaMCUCluster
+
+ENERGY_DIRECTION: Final = "energy_direction"
+
+
+class Channel(t.enum8):
```
epenet, line 29: Why is this an enum?
From my testing, it adds confusion on the unique_id as it ends up with a string-formatted enum rather than a simple integer.

zhaquirks/tuya/ts0601_energy_meter.py
```diff
@@ -0,0 +1,1261 @@
...
+            whole_digits=7, dec_digits=2
+        ),
+        TuyaZBMeteringClusterWithUnit.AttributeDefs.demand_formatting.id: format(
+            whole_digits=7, dec_digits=1
+        ),
+    }
+
+    _EXTENSIVE_ATTRIBUTES: tuple[str] = (
+        TuyaZBMeteringClusterWithUnit.AttributeDefs.instantaneous_demand.name,
+    )
```
epenet, line 536: I think this should be the unsigned attribute:
```suggestion
    _EXTENSIVE_ATTRIBUTES: tuple[str] = (
        TuyaZBMeteringClusterWithUnit.AttributeDefs.instantaneous_demand.name
        + EnergyDirectionHelper.UNSIGNED_ATTR_SUFFIX,
    )
```

epenet (commented):
@jeverley please note that the quirk here did not work for me with a `_TZE284_81yrt3lo`

After a few interations, and looking at attempts by others to adapt your quick in https://github.com/zigpy/zha-device-handlers/issues/3658, I came up with this gist: https://gist.github.com/epenet/39050e3f47f0a8ac4a3aaf526173d5d7 (just for `_TZE284_81yrt3lo`)

jaromanda:
Is this progressing or stalled?

jaromanda:
8 months waiting for a conflict resolution that, as far as I can see, would've been resolved a matter of days after the check failed

I guess this PR is dead in the water

Bizzaro:
how do i use this quirk while i wait for merge? I have 
_TZE200_rks0sgb7

jeverley:
Hi all, an update after a long time - sorry for the delay.

I'm closing this PR and focusing initially on finishing _TZE204_81yrt3lo/_TZE284_81yrt3lo because they require quite specific handling due to firmware issues that would be best separated into their own file (this was creating a lot of overhead in terms of preparing the necessary tests).

I've got a new branch with this simplified approach implemented [here](https://github.com/jeverley/zha-device-handlers/blob/ts0601_ct_power_matseeplus/zhaquirks/tuya/ts0601_power_matseeplus.py), I will be creating a PR with this narrowed scope this weekend once I've finished testing it.

This will allow us to keep the implementation for other devices that don't require the same handling simpler.

PR #3827, opened 2025-02-03
Prevent adding a Tuya DP converter twice for the same DP id

abmantis:
## Proposed change

This prevents calling `tuya_dp` multiple times for the same DP id, which would previously overwrite previous mappings silently.
Two quirks were fixed since they had duplicated mappings. The values for those quirks were updated according to Z2M.

## Checklist

### DIFF
--- a/zhaquirks/tuya/tuya_valve.py
+++ b/zhaquirks/tuya/tuya_valve.py
@@ -577,7 +577,7 @@ class GiexIrrigationStatus(t.enum8):
         fallback_name="Total flow reset switch",
     )
     .tuya_sensor(
-        dp_id=102,
+        dp_id=38,
         attribute_name="irrigation_duration",
         type=t.uint32_t,
         state_class=SensorStateClass.MEASUREMENT,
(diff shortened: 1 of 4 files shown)


zhaquirks/tuya/tuya_valve.py
```diff
@@ -577,7 +577,7 @@ class GiexIrrigationStatus(t.enum8):
         fallback_name="Total flow reset switch",
     )
     .tuya_sensor(
-        dp_id=102,
+        dp_id=38,
         attribute_name="irrigation_duration",
         type=t.uint32_t,
```
TheJulianJES: dp 38 seems to be a `CharacterString`(?) (similar to how we implement dp 106).
https://github.com/Koenkk/zigbee-herdsman-converters/blob/b8008c9fca769668e7da82ace1dabe602dd40b9b/src/devices/neo.ts#L190

prairiesnpr can hopefully take a second look in a few hours.

TheJulianJES (commented):
Other than the `tuya_valve` sensor, this seems good. Thanks!
Targeting this for a HA patch release.

zhaquirks/tuya/tuya_valve.py
TheJulianJES: Fixed by https://github.com/zigpy/zha-device-handlers/pull/3833

TheJulianJES (approved):
Thanks!

PR #3801, opened 2025-01-28
Add IKEA VINDSTYRKA air quality sensor

TheJulianJES:
## Proposed change

This adds support for the IKEA VINDSTYRKA air quality sensor.
The VOC index is exposed and attribute reporting is set up for it.

## Additional information

Supersedes:
- https://github.com/zigpy/zha-device-handlers/pull/2315 - Thanks to @just-oblivious for the initial PR!

And also big thanks to @ikruglov who created and tried a v2 quirk version here: https://github.com/zigpy/zha/pull/254#issuecomment-2507695129

## Checklist

- [ ] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

PR #3832, opened 2025-02-06
Fix missing attribute ids for IKEA Vindstyrka

ikruglov:
## Proposed change
There are reports that the quirk for IKEA Vindstyrka is broken after the upgrade to 2025.2.0. This PR fixes it.

## Additional information
See https://github.com/zigpy/zha-device-handlers/pull/3801

## Checklist

### DIFF
--- a/zhaquirks/ikea/vindstyrka.py
+++ b/zhaquirks/ikea/vindstyrka.py
@@ -26,13 +26,13 @@ class VOCIndex(CustomCluster):
         """Attribute definitions."""
 
         measured_value: Final = ZCLAttributeDef(
-            type=t.Single, access="rp", is_manufacturer_specific=True
+            id=0x0000, type=t.Single, access="rp", is_manufacturer_specific=True
         )
         min_measured_value: Final = ZCLAttributeDef(
-            type=t.Single, access="r", is_manufacturer_specific=True
+            id=0x0001, type=t.Single, access="r", is_manufacturer_specific=True
         )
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
Wow, not sure how this happened.
Thanks for the fix!

ikruglov:
I noted missing IDs when I first checked the PR. I wondered: how is it going to work?! But decided that there is some magic which I don't understand. It turns out there is no black magic :-)

TheJulianJES:
Yeah, I’m not sure when they were lost... Thanks again!
I should add a test to make sure all attributes have an id.

I’ll see if we can make it for a HA patch release with this PRs and some others, but can’t promise it yet.

chevaler27:
Hi all, 
I was Happy to see the COV sensor found After the upgrade 2025.2.1, Thank you for that!
Thing strange, I have COV value at 49 and the value is never updated since I installed HA 2025.2.1 update.
Is there something I should do to fix it?
Or should I wait for a coming update?

Sorry for the question if you feel it's a crasy one, I'm pretty new on HA,

Thanks!

TheJulianJES:
Reconfigure the sensor through the UI once. Or just re-pair it.

chevaler27:
That's where I am a newbie.. I do not understand what I should modify and where in the UI as an example..
Sorry for this, I am learning.😃

chevaler27:
I was able to re pair the device. It worked something like 45 minutes and now not updating anymore.

PR #3839, opened 2025-02-07
Fix Tuya thermostat

delaosa:
## Proposed change

Device tested: Tuya thermostat _TZE284_xnbkhhdr (Avatto ZWT198) 

Home Assistant is not setting the temperature on the device, neither turning it off. However child lock it's working well, same as other settings within TuyaReplacementCluster.

It seems like manufacturer_id_override is not working as expected. Proposed change fix that.

## Additional information

Debugging logs after setting the desired temperature from HA:

` [zigpy.zcl] [0x9054:1:0x0201] writing attributes for LocalDataCluster: attributes={'occupied_heating_setpoint': 1450} manufacturer=None`

In contrast to the previous behavior, this a capture of the same command after 'write_attributes' been overwritten:

` [zigpy.zcl] [0x9054:1:0x0201] writing attributes for LocalDataCluster: attributes={'occupied_heating_setpoint': 1500} manufacturer=-1`

## Checklist


zhaquirks/tuya/tuya_thermostat.py
```diff
@@ -116,6 +116,12 @@ def __init__(self, *args, **kwargs):
         )
         self.add_unsupported_attribute(Thermostat.AttributeDefs.pi_heating_demand.id)
 
+    async def write_attributes(self, attributes, manufacturer=None):
```
prairiesnpr, line 123: I'm hesitant to make this change here, since we know `_TZE204_p3lqqy2r` is tested and working, though `manufacturer_id_override: t.uint16_t = foundation.ZCLHeader.NO_MANUFACTURER_ID` should have prevented the manufacturer id, this risks breaking that thermostat. I'd rather subclass TuyaThermostat, maybe something like this.

```python
class TuyaThermostatNoManuf(TuyaThermostat):
    """Tuya local thermostat cluster, no manufacturer id on attribute write."""

        """Overwrite to force manufacturer code."""
        return await super().write_attributes(
            attributes, manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID
        )
```

Then just update the Avatto thermostats to use this cluster.

TheJulianJES:
Thanks for the PR! It looks like we should never send the manufacturer id on Tuya clusters at all. Some devices don't care, but others seem to not work if it is set. The following PR removes the manufacturer id for the Tuya clusters for all devices (even if the DPs are mapped onto local/fake ZCL clusters):
- https://github.com/zigpy/zha-device-handlers/pull/3860

It should be included in the next HA patch release, so your thermostat should work then.

PR #3840, opened 2025-02-07
Remove Group cluster in IKEA Somrig smart button

MattWestb:
The controller is showing up in ZHA light groups for being added as lights but its one good Zigbee device and is using binding and not adding groups as light (like some tuya controller is doing).

PS i think its one rest from IKEAs direct pairing / binding to devices that is being reused in coordinator paring.

PS2 Styrbar is having the same in the latest firmware but i dont know how to fixing it in V2 quirks.

## Checklist

### DIFF
--- a/zhaquirks/ikea/somrigsmartbtn.py
+++ b/zhaquirks/ikea/somrigsmartbtn.py
@@ -110,7 +110,6 @@ class IkeaSomrigSmartButton(CustomDevice):
                     Basic.cluster_id,
                     PowerConfig1AAACluster,
                     Identify.cluster_id,
-                    Groups.cluster_id,
                     PollControl.cluster_id,
                     LightLink.cluster_id,
                     IKEA_CLUSTER_ID,
@@ -132,7 +131,6 @@ class IkeaSomrigSmartButton(CustomDevice):
                 INPUT_CLUSTERS: [
                     Basic.cluster_id,
                     Identify.cluster_id,
-                    Groups.cluster_id,
                     ShortcutV2Cluster,
                 ],
                 OUTPUT_CLUSTERS: [


MattWestb:
So "only" 10 mounts and still no reaction. . . .

TheJulianJES:
> PS2 Styrbar is having the same in the latest firmware but i dont know how to fixing it in V2 quirks.

You can do `.removes(Group.cluster_id, endpoint_id=X)` in a v2 quirk.

I'm not sure this change is correct though. The devices expose the cluster for a reason, so it's likely working somewhat. Are we sure there aren't maybe old firmware versions (that also match this quirk) still work with `Groups` binding?

TheJulianJES:
Actually, this PR only removed the `Groups` cluster from `INPUT_CLUSTERS`, but not `OUTPUT_CLUSTERS`. It'll still show up as an available cluster to be bound to a group (for this deice).

This also broke CI on `dev`: 😅 
``FAILED tests/test_quirks.py::test_suspicious_cluster_moves[IkeaSomrigSmartButton] - Failed: Quirk <class 'zhaquirks.ikea.somrigsmartbtn.IkeaSomrigSmartButton'> removed input cluster that has output cluster with same ID on EP 1: {4}``

TheJulianJES:
> The controller is showing up in ZHA light groups for being added as lights but its one good Zigbee device and is using binding and not adding groups as light (like some tuya controller is doing).

@MattWestb To make sure I understand your PR description correctly, can you quickly confirm the original issue this PR intended to fix?
1. On the "Manage Zigbee device" dialog for the IKEA Somrig button, the `Groups` cluster showed up for binding, even though binding the cluster to a group does not work. So, in a dialog like this: 
Manage Zigbee device dialog for IKEA Somrig button showing clusters of group to which it can be bound to

---

2. When creating or editing a Zigbee group, the IKEA Somrig button accidentally showed up in the "Add devices" list. So, on a page like this:
Page for adding Zigbee devices to a new Zigbee group

andrei-lazarov:
> Are we sure there aren't maybe old firmware versions (that also match this quirk) still work with `Groups` binding?

Hi, in case this helps: 
- STYRBAR supports binding to groups on all fw versions. 
- On the latest fw, you can bind `Basic` to one group, then you have to unbind it from everything else, including the coordinator
- This is probably true for all old IKEA remotes that were known to 'lose' group binding in firmware updates

I documented it here: 
https://www.zigbee2mqtt.io/devices/E2001_E2002_E2313.html

TheJulianJES:
That is really interesting – huge thanks for sharing that!

MattWestb:
You must have the "IKEA view"  = they implanting function for getting the device working in there 2 scenarios Local with touch link and normal Zigbee coordinator formed networks (or for 3 gen Matter over Thread).
Some functions from ZTT is reused to normal ZB3 and can doing little strange behaving devices but still certificated in Zigbee 3.

Group cluster on in and out is left from the ZLL and is not doing any harm on normal ZB3 systems. But ZHA logic is implementing it that the device is can being added to groups as one light that is not working (tuya remotes is doing it for many remotes that not have binding implemented).
The group cluster in is not used for binding but the in can being added then doing binding group binding but i have not seen any effects of doing it or not (for 13 mouth ago) so shall being OK.

Styrbar need the same fix but i have it on hold until its updated to V2.

I **cant** comforting point 1 then in HA 2026.2.2 the binding panel and signature is **not** longer present in "mange Ziogbee device" so not possible binding remotes to devices / groups but it was working OK in February 2025 then i was doing the PR.

MattWestb:
HA202602A
HA 2026.2.2 missing signature and binding pannel.

TheJulianJES:
You can still click "Bindings" in the dialog you've shared. It being greyed out is just a visual glitch. It's fixed for the next major HA release.

So, from your screenshot, click here to get to the Bindings panel:
image

MattWestb:
Thanks its looks working but not nice done in the frontend (like taking away the color of the top bar).

PR #3859, opened 2025-02-13
Assign default manufacturer code in quirks v2 tests

TheJulianJES:
## Proposed change
This assigns all zigpy devices generated via the `zigpy_device_from_v2_quirk` fixture a node descriptor with a `manufacturer_code`. This is needed to align more closely with actual devices.

We may need to include a full node descriptor instead of only the code?
Also, what code do we want to use? Shouldn't conflict with any real ones ideally.

## Additional information
cc @prairiesnpr 

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -176,9 +176,9 @@ def zigpy_device_from_v2_quirk(MockAppController, ieee_mock):
             ieee = ieee_mock
 
         raw_device = zigpy.device.Device(MockAppController, ieee, nwk)
-        raw_device.node_desc = NodeDescriptor(manufacturer_code=1234)
         raw_device.manufacturer = manufacturer
         raw_device.model = model
+        raw_device.node_desc = NodeDescriptor(manufacturer_code=1234)
 
         for endpoint_id in endpoint_ids:
             ep = raw_device.add_endpoint(endpoint_id)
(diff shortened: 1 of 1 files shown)


TheJulianJES:
These are all the officially defined manufacturer codes: https://github.com/SiliconLabs/gecko_sdk/blob/gsdk_4.4/app/zcl/manufacturers.xml
1234 / 0x04D2 seems to be unused, so that's fine. But we could also use another one.
It's also not really *that* important that they don't conflict, but would be nice. We should also consider making this an argument?

tests/conftest.py
```diff
@@ -175,6 +176,7 @@ def _dev(
             ieee = ieee_mock
 
         raw_device = zigpy.device.Device(MockAppController, ieee, nwk)
+        raw_device.node_desc = NodeDescriptor(manufacturer_code=1234)
         raw_device.manufacturer = manufacturer
         raw_device.model = model
```
TheJulianJES:
```suggestion
        raw_device.manufacturer = manufacturer
        raw_device.model = model
        raw_device.node_desc = NodeDescriptor(manufacturer_code=1234)
```

TheJulianJES:
I also want to do this for v1 quirks in the future, but the old(!) Tuya manufacturer clusters don't yet override with "no manufacturer id", so some tests would fail after that, since the device's manufacturer code would be used.

I think we should also override it with something similar to https://github.com/zigpy/zha-device-handlers/pull/3860, but it'll be in a separate PR.

PR #3896, opened 2025-02-20
Add Innr SP 242 plug quirk

mgjbroadbent:
## Proposed change

V2 Quirk for the inner SP-242 plug that updates the MeteringCluster to use the existing MeteringClusterInnr. This adds a divisor for the summed energy usage to report correct usage.

## Additional information

Tested in HA as a custom quirk against an SP-242.

## Checklist

Related: #4009

### DIFF
--- a/zhaquirks/innr/innr_sp242_plug.py
+++ b/zhaquirks/innr/innr_sp242_plug.py
@@ -1,0 +1,11 @@
+"""Innr SP 242 plug."""
+
+from zigpy.quirks.v2 import QuirkBuilder
+
+from zhaquirks.innr import INNR, MeteringClusterInnr
+
+(
+    QuirkBuilder(INNR, "SP 242")
+    .replaces(MeteringClusterInnr, endpoint_id=1)
+    .add_to_registry()
+)


TheJulianJES:
Thanks for the PR! Apparently there were some changes in newer firmware versions that cause issues with our existing quirks, as newer firmware versions now use a different divisor. Reference: https://github.com/Koenkk/zigbee-herdsman-converters/issues/8712

I'll have to come back to this to see how we can fix this for all firmware versions ideally, as not everyone upgrades.
If you upgrade your SP-242, it might just work without a quirk at all now.

AlexanderLeporiday:
If you update to the latest firmware you also need to update Z2M to a newer version that has the new converter. You then need to reconfigure the device to load the new converter as the old values will not work for the latest firmware. The latest firmware actually provides its divisors on setup, whereas previously they were hard-coded into the converter.
It appears the converters are cached even through updates so reconfiguring the device will load the correct values. It will reset them back to factory default so any customizations will need to be reapplied such as reporting.

At this time no code change should be needed if you update to the latest versions of firmware and z2m.

snellejelle99:
> Thanks for the PR! Apparently there were some changes in newer firmware versions that cause issues with our existing quirks, as newer firmware versions now use a different divisor. Reference: [Koenkk/zigbee-herdsman-converters#8712](https://github.com/Koenkk/zigbee-herdsman-converters/issues/8712)
> 
> I'll have to come back to this to see how we can fix this for all firmware versions ideally, as not everyone upgrades. If you upgrade your SP-242, it might just work without a quirk at all now.

Hi Julian,

I'm running into this problem with my SP240's as well.
Is there any update on this PR/issue. or a way to disable quirks on these devices?

PR #3897, opened 2025-02-21
Add Schneider Electric smoke alarms

uvNikita:
## Proposed change
Add support for Schneider Electric Smoke Alarms.

Reference to the original issue https://github.com/zigpy/zha-device-handlers/pull/1705

Specs: [ZB Spec - Smoke Alarm  -110422.pdf](https://github.com/user-attachments/files/18905404/ZB.Spec.-.Smoke.Alarm.-110422.pdf)

image

## Additional information

There are two issues from the logs:

- Home Assistant complains when "y" is set as a duration unit ("Entity is using native unit of measurement 'y' which is not a valid unit for the device class ('duration') it is using; expected one of ['min', 'd', 'h', 's', 'ms'];"
- "Test mode" switch should probably be under "Diagnostics", but then it fails with "does not have an entity class mapping" error similar to https://github.com/zigpy/zha/issues/291

## Checklist


uvNikita:
Successfully tested the "test mode" of the smoke alarm. Toggling the switch activates smoke alarm sound.

However, after around 15 seconds, test alarm stops, but the switch entity remains in "on" state. Going into "Manage zigbee device" and re-reading `se_test_mode` attribute clears the toggle back to "off" state.

So my guess is that ZHA doesn't get updated attribute state. @TheJulianJES is there something I can do to fix this?

uvNikita:
I added `reporting_config` to both switches (`test_mode` and `no_disturb_mode`) with `min_interval=15, max_interval=300, reportable_change=1`.

This fixed the issue, now `test_mode` resets after 15 seconds. I hope this change will not have too big impact on the battery drain of the device?

zhaquirks/schneiderelectric/smoke.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+    .switch(
+        cluster_id=SEAlarmConfiguration.cluster_id,
+        endpoint_id=20,
+        attribute_name=SEAlarmConfiguration.AttributeDefs.se_test_mode.name,
+        translation_key="test_mode",
+        fallback_name="Test mode",
+        reporting_config=ReportingConfig(
+            min_interval=15,
+            max_interval=300,
```
TheJulianJES: Do we need `max_interval=300`? That would mean this attribute is reported at least every 300 seconds.

I feel like we can set this to `0` (so no periodic reporting) or just increase it a lot more.
Probably fine for `min_interval` as well (minimum delay between reports), since this shouldn't change often.
So, only `reportable_change=1`, others 0. (Or increase max_interval a lot, e.g. to `10800` to make sure we eventually get the updated attribute, even if we missed a report once. This might also be nice for "do not disturb" mode, especially if they can also be toggled from the device.)

See [ZCL R8](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf) page 72 and 73 (line 2695) for more information on reporting config.

zhaquirks/schneiderelectric/smoke.py
```diff
@@ -0,0 +1,174 @@
...
+    .switch(
+        cluster_id=SEAlarmConfiguration.cluster_id,
+        endpoint_id=20,
+        attribute_name=SEAlarmConfiguration.AttributeDefs.se_no_disturb_mode.name,
+        translation_key="no_disturb_mode",
+        fallback_name="No disturb mode",
+        reporting_config=ReportingConfig(
+            min_interval=15,
+            max_interval=300,
+            reportable_change=1,
```
TheJulianJES: Same comment as above.

zhaquirks/schneiderelectric/smoke.py
```diff
@@ -0,0 +1,174 @@
...
+        ),
+    )
+    .switch(
+        cluster_id=SEAlarmConfiguration.cluster_id,
+        endpoint_id=20,
+        attribute_name=SEAlarmConfiguration.AttributeDefs.se_no_disturb_mode.name,
+        translation_key="no_disturb_mode",
+        fallback_name="No disturb mode",
```
TheJulianJES: Can we call this "Do not disturb mode" or "Silence alarm" or something like that?
"No disturb mode" sounds a bit like something is missing or wrong grammatically

zhaquirks/schneiderelectric/smoke.py
uvNikita: Fixed

zhaquirks/schneiderelectric/smoke.py
uvNikita: Fixed

zhaquirks/schneiderelectric/smoke.py
uvNikita: Fixed

uvNikita:
Do you have any comments regarding the following issues?

> - Home Assistant complains when "y" is set as a duration unit ("Entity is using native unit of measurement 'y' which is not a valid unit for the device class ('duration') it is using; expected one of ['min', 'd', 'h', 's', 'ms'];"
> - "Test mode" switch should probably be under "Diagnostics", but then it fails with "does not have an entity class mapping" error similar to
https://github.com/zigpy/zha/issues/291

TheJulianJES:
> - Home Assistant complains when "y" is set as a duration unit ("Entity is using native unit of measurement 'y' which is not a valid unit for the device class ('duration') it is using; expected one of ['min', 'd', 'h', 's', 'ms'];"

You'll need to remove the device class for that entity for now. It's a HA thing that years isn't allowed for the duration device class. We'd need an accepted architecture proposal, similar to this, before we can make changes to HA: https://github.com/home-assistant/architecture/discussions/1206

> - "Test mode" switch should probably be under "Diagnostics", but then it fails with "does not have an entity class mapping" error 

We've generally not categorized switch entities as "diagnostic", as you're actively "configuring" the device with them. Diagnostics is generally reserved for read-only sensors (and alike), but "Test mode" might be a bit of an exception here though. I'd still keep it in configuration for now. If we want to adjust it, we can add a new discovery schema to ZHA here: https://github.com/zigpy/zha/blob/695475a927451524475bdcef90a25177727e2345/zha/application/discovery.py#L206-L215

zhaquirks/schneiderelectric/smoke.py
```diff
@@ -0,0 +1,174 @@
...
+        fallback_name="Alarm sound mode",
+        enum_class=SEAlarmSoundMode,
+    )
+    .sensor(
+        cluster_id=SEAlarmConfiguration.cluster_id,
+        endpoint_id=20,
+        attribute_name=SEAlarmConfiguration.AttributeDefs.se_lifetime.name,
+        translation_key="lifetime",
+        fallback_name="Lifetime",
+        device_class=NumberDeviceClass.DURATION,
```
TheJulianJES: Per the comments above, let's remove this for now, as HA doesn't accept "y" (years) for the duration device class at the moment.
```suggestion
```

TheJulianJES (approved):
I think this looks good now. Thanks!

PR #3901, opened 2025-02-23
Add NodOn SIN-4-RS-20 and SIN-4-RS-20_PRO config options

ikruglov:
## Proposed change
This PR adds the following configuration options for NodOn roller shutter SIN-4-RS-20 and PRO version

- calibration_vertical_run_time_up - Set vertical run time up of the roller shutter
- calibration_vertical_run_time_down - Set vertical run time down of the roller shutter
- calibration_rotation_run_time_up - Set rotation run time up of the roller shutter
- calibration_rotation_run_time_down - Set rotation run time down of the roller shutter

All these settings allow fine-tune calibration if necessary. 

## Additional information
This is a port of https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/nodon.ts#L16

## Checklist

### DIFF
--- a/zhaquirks/nodon/roller_shutter.py
+++ b/zhaquirks/nodon/roller_shutter.py
@@ -18,22 +18,50 @@ class NodOnWindowCovering(WindowCovering, CustomCluster):
         """Attribute definitions."""
 
         # IDs 0x0001-0x0004 redefined below
-        window_covering_type = WindowCovering.AttributeDefs.window_covering_type                            # 0x0000
-        number_of_actuations_lift = WindowCovering.AttributeDefs.number_of_actuations_lift                  # 0x0005
-        number_of_actuations_tilt = WindowCovering.AttributeDefs.number_of_actuations_tilt                  # 0x0006
-        config_status = WindowCovering.AttributeDefs.config_status                                          # 0x0007
-        current_position_lift_percentage = WindowCovering.AttributeDefs.current_position_lift_percentage    # 0x0008
-        current_position_tilt_percentage = WindowCovering.AttributeDefs.current_position_tilt_percentage    # 0x0009
-        installed_open_limit_lift = WindowCovering.AttributeDefs.installed_open_limit_lift                  # 0x0010
-        installed_closed_limit_lift = WindowCovering.AttributeDefs.installed_closed_limit_lift              # 0x0011
-        installed_open_limit_tilt = WindowCovering.AttributeDefs.installed_open_limit_tilt                  # 0x0012
-        installed_closed_limit_tilt = WindowCovering.AttributeDefs.installed_closed_limit_tilt              # 0x0013
-        velocity_lift = WindowCovering.AttributeDefs.velocity_lift                                          # 0x0014
-        acceleration_time_lift = WindowCovering.AttributeDefs.acceleration_time_lift                        # 0x0015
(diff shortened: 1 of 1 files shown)


zhaquirks/nodon/roller_shutter.py
```diff
@@ -0,0 +1,154 @@
...
+from zigpy.zcl.clusters.closures import WindowCovering
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+NODON = "NodOn"
+
+
+class NodOnWindowCovering(WindowCovering, CustomCluster):
+    """NodOn custom WindowCovering cluster."""
+
+    class AttributeDefs(BaseAttributeDefs):
```
TheJulianJES, line 17: Is there a reason to not inherit from `WindowCovering.AttributeDefs` here?
If that's done, we could remove all the copied attributes from zigpy below.

zhaquirks/nodon/roller_shutter.py
```diff
@@ -0,0 +1,154 @@
...
+    .replaces(NodOnWindowCovering)
+    .number(
+        attribute_name=NodOnWindowCovering.AttributeDefs.calibration_vertical_run_time_up.name,
+        cluster_id=NodOnWindowCovering.cluster_id,
+        min_value=0,
+        max_value=65535,
+        step=10,
+        unit=UnitOfTime.MILLISECONDS,
+        device_class=NumberDeviceClass.DURATION,
+        initially_disabled=True,
```
TheJulianJES, line 113: I think we should removed `initially_disabled=True` everywhere.

TheJulianJES (commented):
Seems good otherwise.

zhaquirks/nodon/roller_shutter.py
TheJulianJES: Ah, the other attributes have the same attribute id. Hmm.. 
Does it not work if both are included? There should be a difference because of the manufacturer id. But it might not work with zigpy cache and some other zigpy parts.

zhaquirks/nodon/roller_shutter.py
ikruglov: Hi. I don't have a strong opinion but these parameters are to fine-tune or override settings after internal calibration procedure. So, they are not much useful by default. Hence, I think it makes sense to keep them disabled by default.

zhaquirks/nodon/roller_shutter.py
ikruglov: Yes, the attributes have same IDs. Happy to inherit and override if there is an easy way.

ikruglov: No, it doesn't work of if I just inherit. It was my first approach and it didn't work.

TheJulianJES: Ideally, zigpy should be able to have both defined at the same time, since the original attribute still exists on the cluster, as long as the manufacturer code isn't provided.
Unfortunately, zigpy does not support this yet and it will require some bigger changes.

This approach might be fine for now, since we don't seem to use the attributes that are overridden by the manufacturer ones.

zhaquirks/nodon/roller_shutter.py
TheJulianJES: Ok, seems fine.

TheJulianJES (approved):
Thanks!

PR #3905, opened 2025-02-26
Add external switch type setting to various Candeo devices

candeodevelopment:
## Proposed change

Adds specific Quirks for:

Candeo C205 Switch Module
Candeo C-ZB-SM205-2G Switch Module
Candeo C203 Dimmer Module
Candeo C204 Dimmer Module
Candeo C-ZB-DM204 Dimmer Module

## Additional information

We've been providing Quirks to our customers for some time via our own Github repository, we're hoping now to have them officially directly integrated in order to alleviate customers needing to manually add them to their ZHA configuration.

## Checklist

This is our first submission to the project, so hopefully we're doing things correctly.  

Please let us know if not!


zhaquirks/candeo/candeo_c205_switch_module_V2.py
```diff
@@ -0,0 +1,51 @@
...
+    attr_config = {AttributeDefs.external_switch_type.id: CandeoSwitchType.Toggle}
+
+    async def apply_custom_configuration(self, *args, **kwargs):
+        """Apply Custom Configuration."""
+        await self.write_attributes(self.attr_config, manufacturer=0x1224)
+
+
+(
+    QuirkBuilder(CANDEO, "C205")
+    .replaces(CandeoBasicCluster)
```
puddly: I'm noticing a lot of repetition between these quirks. Do the devices differ in functionality or features? If not, you could use `.applies_to()` to share a definition.

zhaquirks/candeo/candeo_c205_switch_module_V2.py
```diff
@@ -0,0 +1,51 @@
...
+        external_switch_type = ZCLAttributeDef(
+            id=0x8803,
+            type=CandeoSwitchType,
+            zcl_type=DataTypeId.uint8,
+            is_manufacturer_specific=True,
+        )
+
+    attr_config = {AttributeDefs.external_switch_type.id: CandeoSwitchType.Toggle}
+
+    async def apply_custom_configuration(self, *args, **kwargs):
```
puddly: Is the custom configuration required for the device to function? Or is it resetting it to a default state?

The reason I ask is that we have a separate system for defining binding and attibute reporting. If the attribute write isn't entirely required, if would be best to drop it to remove the need for unit testing this code.

zhaquirks/candeo/candeo_c205_switch_module_V2.py
candeodevelopment: They are different devices with different features so we'd prefer to have a separate quirk for each of them if possible. We'll likely add additional functionality to each quirk at a later date.

This particular group of quirks just adds the ability to configure the type of switch connected to the input.

zhaquirks/candeo/candeo_c205_switch_module_V2.py
candeodevelopment: This attribute write is to initialise the state of the input to momentary for our dimmer modules and toggle for our switch modules.

puddly: What state is the device in if this attribute isn't written?

candeodevelopment: From the factory the default is momentary for dimmer modules and toggle for switch modules.

The setting doesn't get defaulted after a factory reset of the module though, so we choose to explicitly set it to the default during pairing to ensure the module is in factory default state after pairing.

dhc25:
@TheJulianJES @puddly 
Can we please get an update on this?
Is there any specific reason that is stopping this PR getting merged?
Thank you

TheJulianJES:
There's really a lot of duplicated code. It should be combined for the most part.
If different functionality for the quirks is to be added later, it should be split up then.

"External Switch Type" should also be "External switch type" due to HA naming conventions.

A small test should also be added for the custom configuration.

Although, `apply_custom_configuration` will also cause the attribute to be written every time the device is reconfigured, not only when paired. I don't think this is a good idea, as that generally only sets up binding and attribute reporting again and shouldn't change settings on the device. We poll some devices on startup, I don't think we reconfigure them, but I'd still avoid doing this.
The user already has the attribute exposed as a config entity in the UI, so why exactly do we need to reset this every time when using ZHA?

candeodevelopment:
> `apply_custom_configuration` will also cause the attribute to be written every time the device is reconfigured

I took inspiration from https://github.com/zigpy/zha-device-handlers/blob/e4844d48549d5a16ed288bb59283a30937ea5591/zhaquirks/philips/__init__.py#L77 

Is there a preferred way to set an initial attribute value only during pairing instead then?

> "External Switch Type" should also be "External switch type" due to HA naming conventions.

OK, we'll get that changed.

> A small test should also be added for the custom configuration.

We're not sure what you mean by this?

> so why exactly do we need to reset this every time when using ZHA

This attribute write is to initialise the state of the input to momentary for our dimmer modules and toggle for our switch modules.

From the factory the default is momentary for dimmer modules and toggle for switch modules.

The setting doesn't get defaulted after a factory reset of the module though, so we choose to explicitly set it to the default during pairing to ensure the module is in factory default state after pairing.

dhc25:
@TheJulianJES Thanks for your feedback.
Can you please let us know how to proceed based on @candeodevelopment feedback above?

puddly:
Hi @candeodevelopment, I'm going to reply to you here for visibility.

> The setting doesn't get defaulted after a factory reset of the module though, so we choose to explicitly set it to the default during pairing to ensure the module is in factory default state after pairing.

A factory reset *must* reset the device back to the factory state. If it does not, I would argue it is a firmware bug. We do not support running code on first join in ZHA right now but I'll update you when I add support for this feature soon.

In the mean time, please implement the suggestions @TheJulianJES made earlier.

candeodevelopment:
Thanks @puddly, we made some changes to the PR, can you let us know if they are OK now please?

1) Changed string per HA naming conventions
2) Moved common code to module to reduce duplication
3) Removed apply_custom_configuration
4) Replaced individual quirk files with common file for all modules

TheJulianJES:
I think the PR looks good now.

By the way, for sharing common entities/custom replaced clusters, you can add a base quirk (without matching any models nor adding it to the registry) and then use that base quirk in specific quirks.
I don't think you need it here, but something you might wanna use in the future if you add more device specific options, but have the "external switch type" shared by all quirks, for example.

```python
# base quirk that will be used as a basis for other device specific quirks
base_quirk = (
    QuirkBuilder()
    .sensor(...)  # this will be cloned to both quirks below
    .switch(...)  # and this too
)


(
    base_quirk.clone()  # inherit everything from base quirk (replaced custom clusters, custom entities, ...)
    .applies_to("manufacturer", "model1")
    .switch(...)  # specific to model1
    .add_to_registry()
)


(
    base_quirk.clone()  # inherit everything from base quirk (replaced custom clusters, custom entities, ...)
    .applies_to("manufacturer", "model2")
    .sensor(...)  # specific to model2
    .add_to_registry()
)
```

We'll get docs ready at some point...

candeodevelopment:
That's great news @TheJulianJES !

Thanks for the pointers on the base quirk concept, that looks useful.

Since the two quirks here are essentially the same but the second needs a specific endpoint_id, is there was a way to do something like this:

```

base_quirk = (
    QuirkBuilder()
    .enum(with_common_stuff)  
)

(
    base_quirk.clone()  
    .applies_to()
    .enum_from_the_base_quirk(add_some_additional_stuff)
    .add_to_registry()
)
```

Or is that wishful thinking?

zhaquirks/candeo/candeo_modules_V2.py
```diff
@@ -0,0 +1,38 @@
...
+    .applies_to(CANDEO, "HK-LN-DIM-A")
+    .applies_to(CANDEO, "C204")
+    .applies_to(CANDEO, "C-ZB-DM204")
+    .applies_to(CANDEO, "C205")
+    .replaces(CandeoBasicCluster)
+    .enum(
+        attribute_name=CandeoBasicCluster.AttributeDefs.external_switch_type.name,
+        enum_class=CandeoSwitchType,
+        cluster_id=CandeoBasicCluster.cluster_id,
+        initially_disabled=False,
```
TheJulianJES: `initially_disabled` is already `False` by default, so I'll just remove this.
```suggestion
```

zhaquirks/candeo/candeo_modules_V2.py
```diff
@@ -0,0 +1,38 @@
...
+
+(
+    QuirkBuilder(CANDEO, "C-ZB-SM205-2G")
+    .replaces(CandeoBasicCluster, endpoint_id=11)
+    .enum(
+        attribute_name=CandeoBasicCluster.AttributeDefs.external_switch_type.name,
+        enum_class=CandeoSwitchType,
+        cluster_id=CandeoBasicCluster.cluster_id,
+        endpoint_id=11,
+        initially_disabled=False,
```
TheJulianJES: `initially_disabled` is already `False` by default, so I'll just remove this.
```suggestion
```

TheJulianJES (approved):
Thanks, lgtm!

TheJulianJES:
> Since the two quirks here are essentially the same but the second needs a specific endpoint_id, is there was a way to do something like this:

Not at the moment.

Issue #3907, opened 2025-02-26
[Device Support Request] Sonoff TRVZB expose external temperature sensor


panther7:
### Problem description

With latest firmware 1.2.1 was added expose external temperature sensor.

### Solution description

Solved here https://github.com/Koenkk/zigbee-herdsman-converters/pull/8873

Hnzti:
I’m trying to update the firmware of my Sonoff TRVZB through ZHA (Zigbee Home Automation), but the OTA update fails with the error: Nepodařilo se provést akci update/install. Update was not successful: <Status.FAILURE: 1>. 

Has anyone else experienced this issue or found a solution?

bwims:
Bump!

The solution was merged 3 weeks ago.   How do I achieve the added functionality?

Thanks
B.

bwims:
![Image](https://github.com/user-attachments/assets/fb1ad30d-ed3e-47df-90a9-26f0b4579a08)

panther7:
Done

bwims:
But how do I achieve the functionality?  Nothing new has appeared in home assistant.  Is it ZHA quirk or what?

panther7:
It was relesed #3909 via [HA 2025.5.0](https://www.home-assistant.io/changelogs/core-2025.5).

bwims:
HA 2025.5.0 is what I needed! Thanks!

Issue #3906, opened 2025-02-26
[Device Support Request] Sonoff TRVZB hysteresis


panther7:
### Problem description

With latest firmware 1.3.0 was added hysteresis settings.

### Solution description

Solved here https://github.com/Koenkk/zigbee-herdsman-converters/pull/8837

PR #3909, opened 2025-02-27
Expose Sonoff TRVZB attributes for 1.3 firmware

danielkleinert:
## Proposed change
This change exposes new attributes of the current firmware.

Values are adopted from zigbee-herdsman:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/8837
https://github.com/Koenkk/zigbee-herdsman-converters/pull/8873

I am unsure about the naming. [Hysteresis](https://en.wikipedia.org/wiki/Hysteresis) could be better than "Temperature control accuracy". 

## Checklist

### DIFF
--- a/zhaquirks/sonoff/trvzb.py
+++ b/zhaquirks/sonoff/trvzb.py
@@ -70,6 +70,21 @@ class CustomSonoffCluster(CustomCluster):
             type=t.uint8_t,
         )
 
+        external_temperature_sensor_enable = ZCLAttributeDef(
+            id=0x600E,
+            type=t.uint8_t,
+        )
+
+        external_temperature_sensor_value = ZCLAttributeDef(
+            id=0x600D,
+            type=t.int16s,
+        )
+
+        temperature_control_accuracy = ZCLAttributeDef(
+            id=0x6011,
+            type=t.int16s,
+        )
+
     @property
     def _is_manuf_specific(self):
         return False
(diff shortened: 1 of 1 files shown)


photomoose:
@danielkleinert I’m still waiting for confirmation from Sonoff, but I'm starting to think the "external temperature sensor" attribute is a bit field rather than a simple on/off. See my comment here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/8873#issuecomment-2688708314

danielkleinert:
@photomoose Thanks for letting me know. I'll try to map that once you get the confirmation on the semantics.

panther7:
issues #3906, #3907

panther7:
Hi, why still not merged? Thanks

bwims:
When this is merged, which version of HA will it be distributed with?  Or will it be possible to simply upgrade the zha component?

Thanks!

TheJulianJES (commented):
So, due to how v2 quirk entities are currently created, they'll be force created, even for older firmware versions that don't support the attributes.

It's not a big issue, so I think we can go ahead with this.

zhaquirks/sonoff/trvzb.py
```diff
@@ -121,5 +136,33 @@ def _is_manuf_specific(self):
...
+        step=0.2,
+        unit=UnitOfTemperature.CELSIUS,
+        multiplier=0.01,
+        translation_key="temperature_control_accuracy",
+        fallback_name="Temperature control accuracy",
+    )
+    .switch(
+        CustomSonoffCluster.AttributeDefs.external_temperature_sensor_enable.name,
+        CustomSonoffCluster.cluster_id,
+        translation_key="external_temperature_sensor_enable",
```
TheJulianJES: I'll remove the `_enable` suffix from this `translation_key`. More in line with other translation keys in ZHA/quirks and HA then.
```suggestion
        translation_key="external_temperature_sensor",
```

TheJulianJES (commented):
Also added the `device_class=NumberDeviceClass.TEMPERATURE` for the two new number entities.

For the name of the first one, I think "accuracy" is fine. Might be a little easier to understand.

TheJulianJES (approved):
This looks good now. Thanks!

PR #3346, opened 2024-09-03
Add initial Sonoff smart water valve quirk

fgsch:
## Proposed change

Add quirk for Sonoff SWV.

## Additional information

This exposes the `water_valve_state` attribute in the custom cluster and ~~adds a sensor to HA with the same information.~~
EDIT: The enum sensor will not be added with this initial PR. We'd like two get two different binary sensors to parse the same attribute. Since quirks v2 can't do this yet, we'll need to do this work in a future PR.
This initial PR will just expose the attribute.

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

PR #3910, opened 2025-02-28
Add Sonoff water valve leak and supply sensors

fgsch:
## Proposed change

Add sensors for water leaks and supply problems.
While here rename EwelinkCluster to match other Sonoff quirks.

## Additional information

Follow-up from https://github.com/zigpy/zha-device-handlers/pull/3346. 
Supersedes https://github.com/zigpy/zha/pull/189.

## Checklist

### DIFF
--- a/zhaquirks/sonoff/swv.py
+++ b/zhaquirks/sonoff/swv.py
@@ -38,7 +38,7 @@ class CustomSonoffCluster(CustomCluster):
     .binary_sensor(
         CustomSonoffCluster.AttributeDefs.water_valve_state.name,
         CustomSonoffCluster.cluster_id,
-        device_class=BinarySensorDeviceClass.PROBLEM,
+        device_class=BinarySensorDeviceClass.MOISTURE,
         attribute_converter=lambda x: x & ValveState.Water_Leakage,
         unique_id_suffix="water_leak_status",
         reporting_config=ReportingConfig(
--- a/zhaquirks/sonoff/swv.py
+++ b/zhaquirks/sonoff/swv.py
@@ -1,9 +1,10 @@
 """Sonoff SWV - Zigbee smart water valve."""
 
 from zigpy.quirks import CustomCluster
-from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2 import QuirkBuilder, ReportingConfig
 from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
 import zigpy.types as t
+from zigpy.zcl import foundation
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 
@@ -20,6 +21,7 @@ class CustomSonoffCluster(CustomCluster):
     """Custom Sonoff cluster."""
 
     cluster_id = 0xFC11
+    manufacturer_id_override: t.uint16_t = foundation.ZCLHeader.NO_MANUFACTURER_ID
 
     class AttributeDefs(BaseAttributeDefs):
         """Attribute definitions."""
@@ -28,10 +30,6 @@ class CustomSonoffCluster(CustomCluster):
             id=0x500C,
             type=ValveState,
         )
-
-    @property
-    def _is_manuf_specific(self):
-        return False
 
 
 (
@@ -43,6 +41,9 @@ class CustomSonoffCluster(CustomCluster):
         device_class=BinarySensorDeviceClass.PROBLEM,
         attribute_converter=lambda x: x & ValveState.Water_Leakage,
         unique_id_suffix="water_leak_status",
+        reporting_config=ReportingConfig(
+            min_interval=30, max_interval=900, reportable_change=1
+        ),
         translation_key="water_leak",
         fallback_name="Water leak",
     )
--- a/zhaquirks/sonoff/swv.py
+++ b/zhaquirks/sonoff/swv.py
@@ -57,4 +57,4 @@ class CustomSonoffCluster(CustomCluster):
         fallback_name="Water supply",
     )
     .add_to_registry()
-)  # fmt: skip
+)
--- a/zhaquirks/sonoff/swv.py
+++ b/zhaquirks/sonoff/swv.py
@@ -2,6 +2,7 @@
 
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
 import zigpy.types as t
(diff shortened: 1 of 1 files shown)


fgsch:
I will move this to open later today when I can test it, but FYI @TheJulianJES .

fgsch:
Question: Is there any way to define reporting for the `water_valve_state` attribute and not at the sensor level? 
Or does that not work and need to be defined for each separately?

fgsch:
@TheJulianJES Can you let me know the answer to my question above? 
Also, do you know why the sensors come as PROBLEM instead of their name on the UI?

Screenshot 2025-03-16 at 21 49 06

TheJulianJES:
That is expected if the sensor has a state class due to the way `fallback_name` is implemented in HA.
It's only an issue when the translation key isn't present in the translations. It will work fine when we PR the changes to Home Assistant. See this for more information: https://github.com/zigpy/zigpy/issues/1523#issuecomment-2564106284

fgsch:
> That is expected if the sensor has a state class due to the way `fallback_name` is implemented in HA. It's only an issue when the translation key isn't present in the translations. It will work fine when we PR the changes to Home Assistant. See this for more information: [zigpy/zigpy#1523 (comment)](https://github.com/zigpy/zigpy/issues/1523#issuecomment-2564106284)

Thanks for the information. I thought that might be related but wasn't sure.

Regarding:

>Question: Is there any way to define reporting for the water_valve_state attribute and not at the sensor level?
Or does that not work and need to be defined for each separately?

Do you happen to know the answer? I'd like to finish this PR.
Thank you again.

TheJulianJES:
Only possible to do directly with the sensor function at the moment.

fgsch:
> Only possible to do directly with the sensor function at the moment.

Thanks. I will update this and open It for review.

fgsch:
Ok, I think this is ready for review now. PTAL.

fgsch:
@TheJulianJES,

> That is expected if the sensor has a state class due to the way `fallback_name` is implemented in HA. It's only an issue when the translation key isn't present in the translations. It will work fine when we PR the changes to Home Assistant. See this for more information: [zigpy/zigpy#1523 (comment)](https://github.com/zigpy/zigpy/issues/1523#issuecomment-2564106284)

Does this apply to the entity ID as well? I can see both have `binary_sensor.garden_tap_valve_problem`.

abmantis:
> @TheJulianJES,
> 
> > That is expected if the sensor has a state class due to the way `fallback_name` is implemented in HA. It's only an issue when the translation key isn't present in the translations. It will work fine when we PR the changes to Home Assistant. See this for more information: [zigpy/zigpy#1523 (comment)](https://github.com/zigpy/zigpy/issues/1523#issuecomment-2564106284)
> 
> Does this apply to the entity ID as well? I can see both have `binary_sensor.garden_tap_valve_problem`.

Yes. The entity will be generated from the name, which will be the fallback_name you specified when it gets integrated in HA.

abmantis (approved):

TheJulianJES (approved):
Thanks!

zhaquirks/sonoff/swv.py
```diff
@@ -28,13 +31,30 @@ class AttributeDefs(BaseAttributeDefs):
...
-
 
 (
     QuirkBuilder("SONOFF", "SWV")
-    .replaces(EwelinkCluster)
+    .replaces(CustomSonoffCluster)
+    .binary_sensor(
+        CustomSonoffCluster.AttributeDefs.water_valve_state.name,
+        CustomSonoffCluster.cluster_id,
+        device_class=BinarySensorDeviceClass.PROBLEM,
```
TheJulianJES: Ah, actually. Can we use `BinarySensorDeviceClass.MOISTURE` here? Does that make sense for this device?

That's the device class we use for (IAS Zone) leak sensors in ZHA. If that works, we could also drop the `translation_key` if the default name ("Moisture") is clear enough.

fgsch: "Moisture" for leak seems odd IMO, but I will give it a try and report back.

TheJulianJES: Leak sensors generally use `BinarySensorDeviceClass.MOISTURE` [in HA](https://github.com/search?q=repo%3Ahome-assistant%2Fcore%20BinarySensorDeviceClass.MOISTURE&type=code).
The entity icon will show as a water droplet (crossed through or not) and be "dry" or "wet".
I'm not sure where a leak sensor is positioned on this device, but I think the device class is fitting here.

abmantis: I agree that `MOISTURE` is the proper class here. Also [from the docs](https://developers.home-assistant.io/docs/core/entity/binary-sensor/): `On means wet, Off means dry.`

TheJulianJES: Will apply this suggestion to get it merged:
```suggestion
        device_class=BinarySensorDeviceClass.MOISTURE,
```

zhaquirks/sonoff/swv.py
```diff
@@ -28,13 +31,30 @@ class AttributeDefs(BaseAttributeDefs):
...
+    .binary_sensor(
+        CustomSonoffCluster.AttributeDefs.water_valve_state.name,
+        CustomSonoffCluster.cluster_id,
+        device_class=BinarySensorDeviceClass.MOISTURE,
+        attribute_converter=lambda x: x & ValveState.Water_Leakage,
+        unique_id_suffix="water_leak_status",
+        reporting_config=ReportingConfig(
+            min_interval=30, max_interval=900, reportable_change=1
+        ),
+        translation_key="water_leak",
```
TheJulianJES, line 47: If we didn't have a translation key for this, the name would be "Moisture" due to the device class.
I think we can leave the "Water leak" name here though, since it's a bit more explicit for this device.

zhaquirks/sonoff/swv.py
```diff
@@ -28,13 +31,30 @@ class AttributeDefs(BaseAttributeDefs):
...
+        CustomSonoffCluster.AttributeDefs.water_valve_state.name,
+        CustomSonoffCluster.cluster_id,
+        device_class=BinarySensorDeviceClass.PROBLEM,
+        attribute_converter=lambda x: x & ValveState.Water_Shortage,
+        unique_id_suffix="water_supply_status",
+        translation_key="water_supply",
+        fallback_name="Water supply",
+    )
     .add_to_registry()
 )  # fmt: skip
```
TheJulianJES: This should not be needed here, as the quirk is now longer, so it'll format itself properly.
```suggestion
)
```

TheJulianJES (approved):
Thanks again!

fgsch:
Thank you for updating and merging this PR.

PR #3912, opened 2025-03-01
Add Sonoff SNZB-02D LCD temp/humidity sensor

fgsch:
## Proposed change

Add quirks for Sonoff SNZB-02D for firmware 1.0.9 (stock firmware AFAIK).

Supersedes #2916.

## Additional information

UI changes:

Screenshot 2025-03-01 at 18 56 32

## Checklist

### DIFF
--- a/zhaquirks/sonoff/snzb02d.py
+++ b/zhaquirks/sonoff/snzb02d.py
@@ -1,0 +1,111 @@
+"""Sonoff SNZB-02D - Zigbee LCD smart temperature humidity sensor."""
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import PERCENTAGE, UnitOfTemperature
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, DataTypeId, ZCLAttributeDef
+
+
+class TemperatureUnit(t.enum16):
+    """Temperature unit."""
+
+    Celsius = 0
+    Fahrenheit = 1
+
+
+class CustomSonoffCluster(CustomCluster):
+    """Sonoff custom cluster."""
+
+    cluster_id = 0xFC11
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        temperature_threshold_high = ZCLAttributeDef(
+            id=0x0003,
+            type=t.int16s,
+        )
+
+        temperature_threshold_low = ZCLAttributeDef(
+            id=0x0004,
+            type=t.int16s,
+        )
+
+        humidity_threshold_low = ZCLAttributeDef(
+            id=0x0005,
+            type=t.uint16_t,
+        )
+
+        humidity_threshold_high = ZCLAttributeDef(
+            id=0x0006,
+            type=t.uint16_t,
+        )
+
+        temperature_unit = ZCLAttributeDef(
+            id=0x0007,
+            type=TemperatureUnit,
+            zcl_type=DataTypeId.uint16,
+        )
+
+    @property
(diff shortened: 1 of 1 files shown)


elupus (approved):
Nice!

fgsch:
And similar to above but using "Comfort humidity/temperature min/max":

Screenshot 2025-03-02 at 11 31 31

In case anyone cares.

fgsch:
Now that the new firmware is available, I will open a separate PR to expose the new entities.

elupus (approved):

fgsch:
@TheJulianJES any chance you can have a look at this?

zhaquirks/sonoff/snzb02d.py
```diff
@@ -0,0 +1,111 @@
...
+
+        temperature_unit = ZCLAttributeDef(
+            id=0x0007,
+            type=TemperatureUnit,
+            zcl_type=DataTypeId.uint16,
+        )
+
+    @property
+    def _is_manuf_specific(self):
+        return False
```
TheJulianJES: I think using the `manufacturer_id_override` for this class should also work instead of overriding `_is_manuf_specific` now:
```python
    # remove manufacturer id for cluster
    manufacturer_id_override: t.uint16_t = foundation.ZCLHeader.NO_MANUFACTURER_ID
```

with the import at the top:
```python
```

zhaquirks/sonoff/snzb02d.py
```diff
@@ -0,0 +1,111 @@
...
+        multiplier=0.01,
+        translation_key="comfort_humidity_max",
+        fallback_name="Comfort humidity max",
+    )
+    .enum(
+        CustomSonoffCluster.AttributeDefs.temperature_unit.name,
+        TemperatureUnit,
+        CustomSonoffCluster.cluster_id,
+        translation_key="temperature_unit",
+        fallback_name="Temperature unit",
```
TheJulianJES: For a Tuya sensor, we use `display_unit` as the translation key and `Display unit` as the `fallback_name`/English translated name. We might wanna use this here, too.

TheJulianJES (commented):
Should be good to go after those two small comments. Nice work!

fgsch:
Thank you for reviewing it and the comments. 

I will test the changes tonight and push an update.

zhaquirks/sonoff/snzb02d.py
fgsch: Done.

zhaquirks/sonoff/snzb02d.py
fgsch: This is done too.

fgsch:
PTAL

zhaquirks/sonoff/snzb02d.py
```diff
@@ -0,0 +1,109 @@
...
+(
+    QuirkBuilder("SONOFF", "SNZB-02D")
+    .replaces(CustomSonoffCluster)
+    .number(
+        CustomSonoffCluster.AttributeDefs.comfort_temperature_min.name,
+        CustomSonoffCluster.cluster_id,
+        min_value=-10,
+        max_value=60,
+        step=0.1,
+        unit=UnitOfTemperature.CELSIUS,
```
TheJulianJES, line 63: Also using `device_class=NumberDeviceClass.TEMPERATURE` might make sense for all of these entities.
I think only then will HA apply unit conversion when using Fahrenheit as the HA unit.

We currently only use it for very few config entities (both in quirks and ZHA directly), but looking at Home Assistant Core, I think we should use it everywhere.
Let's leave it like this for now to make sure we want to apply unit conversion everywhere. If so, we'll eventually have a big PR for both projects that sets the device class everywhere.

TheJulianJES (approved):

PR #3920, opened 2025-03-04
Add Third Reality switch "count down time"

hwzolin:
## Checklist

### DIFF
--- a/zhaquirks/thirdreality/switch.py
+++ b/zhaquirks/thirdreality/switch.py
@@ -1,118 +1,41 @@
 """Third Reality switch devices."""
 
-from zigpy.profiles import zha
-from zigpy.quirks import CustomDevice
-from zigpy.zcl.clusters.general import Basic, Groups, Identify, OnOff, Ota, Scenes
-from zigpy.zcl.clusters.homeautomation import Diagnostic
+from typing import Final
 
-from zhaquirks import PowerConfigurationCluster
-from zhaquirks.const import (
-    DEVICE_TYPE,
-    ENDPOINTS,
-    INPUT_CLUSTERS,
-    MODELS_INFO,
-    OUTPUT_CLUSTERS,
-    PROFILE_ID,
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ThirdRealitySwitchCluster(CustomCluster):
+    """Third Reality's temperature and humidity sensor private cluster."""
+
+    cluster_id = 0xFF02
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Define the attributes of a private cluster."""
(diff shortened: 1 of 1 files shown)


zhaquirks/thirdreality/switch.py
```diff
@@ -1,118 +1,41 @@
...
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ThirdRealitySwitchCluster(CustomCluster):
+    """Third Reality's temperature and humidity sensor private cluster."""
```
TheJulianJES:
```suggestion
    """Third Reality's switch private cluster."""
```

zhaquirks/thirdreality/switch.py
```diff
@@ -1,118 +1,41 @@
...
+    .replaces(ThirdRealitySwitchCluster)
+    .number(
+        attribute_name=ThirdRealitySwitchCluster.AttributeDefs.count_down_time.name,
+        min_value=0,
+        max_value=65535,
+        step=1,
+        unit=UnitOfTime.SECONDS,
+        cluster_id=ThirdRealitySwitchCluster.cluster_id,
+        translation_key="count_down_time",
+        fallback_name="Count Down Time",
```
TheJulianJES:
```suggestion
        fallback_name="Count down time",
```

zhaquirks/thirdreality/switch.py
```diff
@@ -1,118 +1,41 @@
...
+    .replaces(ThirdRealitySwitchCluster)
+    .number(
+        attribute_name=ThirdRealitySwitchCluster.AttributeDefs.count_down_time.name,
+        min_value=0,
+        max_value=65535,
+        step=1,
+        unit=UnitOfTime.SECONDS,
+        cluster_id=ThirdRealitySwitchCluster.cluster_id,
+        translation_key="count_down_time",
+        fallback_name="Count down time",
```
TheJulianJES, line 38: So what does "Count down time" do roughly? Is this a delay before turning on/off?
"Turn on delay" might be a better name then?

Also, Z2M seems to have two attributes (seemingly not exposed to HA): `backOn` (id: 0x1) and `backOff` (id: 0x2). Should we also have the id 0x2 attribute here?

zhaquirks/thirdreality/switch.py
```diff
@@ -1,118 +1,41 @@
...
+    .add_to_registry()
 )
-from zhaquirks.thirdreality import THIRD_REALITY
-
-
-class CustomPowerConfigurationCluster(PowerConfigurationCluster):
-    """Custom PowerConfigurationCluster."""
-
-    MIN_VOLTS = 2.1
-    MAX_VOLTS = 3.0
```
TheJulianJES, line 24: Also, the previous implementation had a `CustomPowerConfigurationCluster` that replaced the percentage reported by the device by one calculated from minimum and maximum voltage.

Is that no longer needed? So the device reports the percentage value itself properly?

PR #3923, opened 2025-03-04
Add Third Reality soil moisture sensor correction options

hwzolin:
## Checklist

### DIFF
--- a/zhaquirks/thirdreality/soil_moisture_sensor.py
+++ b/zhaquirks/thirdreality/soil_moisture_sensor.py
@@ -5,8 +5,8 @@ from typing import Final
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.v2 import QuirkBuilder
 import zigpy.types as t
-from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 from zigpy.zcl.clusters.measurement import SoilMoisture
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 
 class ThirdRealitySoilMoistureCluster(CustomCluster):
--- a/zhaquirks/thirdreality/soil_moisture_sensor.py
+++ b/zhaquirks/thirdreality/soil_moisture_sensor.py
@@ -1,0 +1,72 @@
+"""Third Reality soil moisture sensor devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ThirdRealitySoilMoistureCluster(CustomCluster):
+    """Third Reality's soil moisture sensor private cluster."""
+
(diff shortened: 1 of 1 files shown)


hwzolin:
@TheJulianJES 
Hello, the code has not been merged for such a long time. Do I need to make any modifications?

zhaquirks/thirdreality/soil_moisture_sensor.py
```diff
@@ -0,0 +1,74 @@
...
+(
+    QuirkBuilder("Third Reality, Inc", "3RSM0147Z")
+    .replaces(ThirdRealitySoilMoistureCluster)
+    .replaces(SoilMoisture)
+    .number(
+        attribute_name=ThirdRealitySoilMoistureCluster.AttributeDefs.temperature_correction_celsius.name,
+        min_value=-10,
+        max_value=10,
+        multiplier=0.01,
+        # unit=UnitOfTemperature.CELSIUS,
```
TheJulianJES, line 48: Why is `unit` for the sensors commented out?

zhaquirks/thirdreality/soil_moisture_sensor.py
```diff
@@ -0,0 +1,74 @@
...
+            id=0x0033,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
+
+        temperature_correction_celsius: Final = ZCLAttributeDef(
+            id=0x0031,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
```
TheJulianJES, line 30: Why is the correction separate for Celsius and Fahrenheit? Does the sensor have some kind of display?
I'm assuming it only transmits data (in Celsius) via the ZCL spec `TemperatureMeasurement` cluster, right?
Home Assistant would then convert the values to Fahrenheit if chosen by the user.

TheJulianJES (commented):
Assuming this device has no display, so only HA gets the data, having calibration options is a bit weird, since this can kind of already be done in Home Assistant by adding a template sensor to the same device via the UI (and defining a static offset there, or even more advanced stuff).

Since this applies device level correction, it may be fine though. The other comments below are still valid though.

Also, in the future, please make sure to use proper PR titles. As an example: "Add Third Reality moisture sensor correction options".
Just saying "Modify the private cluster of XXX" does not make it clear about what device this is and what the "modified private cluster" allows for.

TheJulianJES:
Why did you close all of the PRs? Some still need to be reviewed and this one has changes pending, some of which (like changing the title) are also relevant for the others.

3reality-support:
@TheJulianJES Hello, I will be responsible for the PR of our company in the future. Currently, there are quite a few products that need to be submitted, and we plan to submit one PR at a time so that the same mistake will not be repeated

zhaquirks/thirdreality/soil_moisture_sensor.py
3reality-support: Because some of our customers use Celsius and some use Fahrenheit, if we only correct Celsius, customers who use Fahrenheit will need to convert the value they want to correct to Celsius and then write it

Issue #3581, opened 2024-12-04
[BUG]If the unit is set, the operation information will not be displayed in the logbook after the control is set for the number component of Quirks v2


hwzolin:
### Bug description

I added two UI components based on Quirks v2 in the code, the type is Number. After repeated testing, I found that component B, which has a set unit, has no information recorded in the logbook after control. However, if you don't have a unit set for component A, you can see the record information you just operated in the logbook after waiting for a while.

Here's my full code, check it out:

```
"""Third Reality temperature and humidity sensor devices."""

from typing import Final

from zigpy.quirks import CustomCluster
from zigpy.quirks.v2 import QuirkBuilder
from zigpy.quirks.v2.homeassistant import UnitOfTemperature
import zigpy.types as t
from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef


class ThirdRealityTemperatureAndHumidityCluster(CustomCluster):
    """Third Reality's temperature and humidity sensor private cluster."""

    cluster_id = 0xFF01

    class AttributeDefs(BaseAttributeDefs):
        """Define the attributes of a private cluster."""

        temperature_correction_fahrenheit: Final = ZCLAttributeDef(
            id=0x0033,
            type=t.int16s,
            is_manufacturer_specific=True,
        )

        temperature_correction_celsius: Final = ZCLAttributeDef(
            id=0x0031,
            type=t.int16s,
            is_manufacturer_specific=True,
        )

        humidity_correction: Final = ZCLAttributeDef(
            id=0x0032,
            type=t.int16s,
            is_manufacturer_specific=True,
        )


(
    QuirkBuilder("Third Reality, Inc", "3RTHS24BZ")
    .also_applies_to("Third Reality, Inc", "3RTHS0224Z")
    .replaces(ThirdRealityTemperatureAndHumidityCluster)
    .number(
        attribute_name=ThirdRealityTemperatureAndHumidityCluster.AttributeDefs.temperature_correction_celsius.name,
        min_value=-1000,
        max_value=1000,
        step=1,
        # unit=UnitOfTemperature.CELSIUS,
        cluster_id=ThirdRealityTemperatureAndHumidityCluster.cluster_id,
        translation_key="temperature_correction_celsius",
        # fallback_name="Correction/100",
        fallback_name="A",
    )
    .number(
        attribute_name=ThirdRealityTemperatureAndHumidityCluster.AttributeDefs.temperature_correction_fahrenheit.name,
[...]

PR #3926, opened 2025-03-04
Add Third Reality watering kit options

hwzolin:
## Checklist

### DIFF
--- a/zhaquirks/thirdreality/watering_kit.py
+++ b/zhaquirks/thirdreality/watering_kit.py
@@ -1,0 +1,57 @@
+"""Third Reality water leak devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ThirdRealityWateringKitCluster(CustomCluster):
+    """Third Reality's water leak sensor private cluster."""
+
+    cluster_id = 0xFFF2
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Define the attributes of a private cluster."""
+
+        water_duration: Final = ZCLAttributeDef(
+            id=0x0000,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+
+        water_interval: Final = ZCLAttributeDef(
+            id=0x0001,
+            type=t.uint8_t,
+            is_manufacturer_specific=True,
+        )
+
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RWK0148Z")
+    .replaces(ThirdRealityWateringKitCluster)
+    .number(
+        attribute_name=ThirdRealityWateringKitCluster.AttributeDefs.water_duration.name,
+        min_value=1,
+        max_value=1800,
+        step=1,
+        unit=UnitOfTime.SECONDS,
+        cluster_id=ThirdRealityWateringKitCluster.cluster_id,
+        translation_key="water_duration",
+        fallback_name="Water Duration",
+    )
+    .number(
+        attribute_name=ThirdRealityWateringKitCluster.AttributeDefs.water_interval.name,
+        min_value=0,
+        max_value=30,
+        step=1,
+        unit=UnitOfTime.DAYS,
+        cluster_id=ThirdRealityWateringKitCluster.cluster_id,
+        translation_key="water_interval",
+        fallback_name="Water Interval",
+    )
+    .add_to_registry()
+)
--- a/zhaquirks/thirdreality/watering_kit.py
+++ b/zhaquirks/thirdreality/watering_kit.py
@@ -1,4 +1,5 @@
-"""Third Reality water leak devices."""
+"""Third Reality watering kit devices."""
+
 
 from typing import Final
 
@@ -10,7 +11,8 @@ from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 
 class ThirdRealityWateringKitCluster(CustomCluster):
-    """Third Reality's water leak sensor private cluster."""
+    """Third Reality's watering kit private cluster."""
+
 
     cluster_id = 0xFFF2
 
--- a/zhaquirks/thirdreality/watering_kit.py
+++ b/zhaquirks/thirdreality/watering_kit.py
@@ -1,6 +1,5 @@
 """Third Reality watering kit devices."""
 
-
 from typing import Final
 
 from zigpy.quirks import CustomCluster
@@ -12,7 +11,6 @@ from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 class ThirdRealityWateringKitCluster(CustomCluster):
     """Third Reality's watering kit private cluster."""
-
 
     cluster_id = 0xFFF2
 
--- a/zhaquirks/thirdreality/watering_kit.py
+++ b/zhaquirks/thirdreality/watering_kit.py
@@ -41,7 +41,7 @@ class ThirdRealityWateringKitCluster(CustomCluster):
         unit=UnitOfTime.SECONDS,
         cluster_id=ThirdRealityWateringKitCluster.cluster_id,
         translation_key="water_duration",
-        fallback_name="Water Duration",
+        fallback_name="Water duration",
     )
     .number(
         attribute_name=ThirdRealityWateringKitCluster.AttributeDefs.water_interval.name,
@@ -51,7 +51,7 @@ class ThirdRealityWateringKitCluster(CustomCluster):
         unit=UnitOfTime.DAYS,
         cluster_id=ThirdRealityWateringKitCluster.cluster_id,
         translation_key="water_interval",
-        fallback_name="Water Interval",
+        fallback_name="Water interval",
     )
     .add_to_registry()
 )
--- a/zhaquirks/thirdreality/watering_kit.py
+++ b/zhaquirks/thirdreality/watering_kit.py
@@ -7,6 +7,7 @@ from zigpy.quirks.v2 import QuirkBuilder
 from zigpy.quirks.v2.homeassistant import UnitOfTime
 import zigpy.types as t
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
 
 
 class ThirdRealityWateringKitCluster(CustomCluster):
@@ -38,6 +39,7 @@ class ThirdRealityWateringKitCluster(CustomCluster):
         min_value=1,
         max_value=1800,
         step=1,
+        device_class=NumberDeviceClass.DURATION,
         unit=UnitOfTime.SECONDS,
         cluster_id=ThirdRealityWateringKitCluster.cluster_id,
         translation_key="water_duration",
@@ -48,6 +50,7 @@ class ThirdRealityWateringKitCluster(CustomCluster):
         min_value=0,
         max_value=30,
         step=1,
+        device_class=NumberDeviceClass.DURATION,
         unit=UnitOfTime.DAYS,
         cluster_id=ThirdRealityWateringKitCluster.cluster_id,
         translation_key="water_interval",
--- a/zhaquirks/thirdreality/watering_kit.py
+++ b/zhaquirks/thirdreality/watering_kit.py
@@ -5,9 +5,9 @@ from typing import Final
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.v2 import QuirkBuilder
 from zigpy.quirks.v2.homeassistant import UnitOfTime
+from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
 import zigpy.types as t
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
-from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
 
 
 class ThirdRealityWateringKitCluster(CustomCluster):


zhaquirks/thirdreality/watering_kit.py
```diff
@@ -0,0 +1,57 @@
...
+    .replaces(ThirdRealityWateringKitCluster)
+    .number(
+        attribute_name=ThirdRealityWateringKitCluster.AttributeDefs.water_duration.name,
+        min_value=1,
+        max_value=1800,
+        step=1,
+        unit=UnitOfTime.SECONDS,
+        cluster_id=ThirdRealityWateringKitCluster.cluster_id,
+        translation_key="water_duration",
+        fallback_name="Water Duration",
```
TheJulianJES: Due to HA naming conventions, only the first letter of an entity name should be upper-case (except for abbreviations). So this should be "Water duration":
```suggestion
        fallback_name="Water duration",
```

(`fallback_name` will automatically be added to https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/strings.json when this is merged to HA, so it will be translated in to other languages automatically where the capitalization might be done differently.)

zhaquirks/thirdreality/watering_kit.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    .number(
+        attribute_name=ThirdRealityWateringKitCluster.AttributeDefs.water_interval.name,
+        min_value=0,
+        max_value=30,
+        step=1,
+        unit=UnitOfTime.DAYS,
+        cluster_id=ThirdRealityWateringKitCluster.cluster_id,
+        translation_key="water_interval",
+        fallback_name="Water Interval",
```
TheJulianJES: Same here:
```suggestion
        fallback_name="Water interval",
```

zhaquirks/thirdreality/watering_kit.py
```diff
@@ -0,0 +1,57 @@
+"""Third Reality water leak devices."""
```
TheJulianJES:
```suggestion
"""Third Reality watering kit devices."""

```

zhaquirks/thirdreality/watering_kit.py
```diff
@@ -0,0 +1,57 @@
...
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ThirdRealityWateringKitCluster(CustomCluster):
+    """Third Reality's water leak sensor private cluster."""
```
TheJulianJES:
```suggestion
    """Third Reality's watering kit private cluster."""

```

TheJulianJES (commented):
I'll push the suggested changes myself.

Please keep in mind the correct capitalization needed for HA entities for future PRs:
Due to HA naming conventions, only the first letter of an entity name should be upper-case (except for abbreviations). So it should be "Water duration" instead of "Water Duration", for example.

TheJulianJES (commented):
Also, please use more descriptive PR titles for future PRs.
Examples for this PR:
- "Add Third Reality watering kit options"
- "Add Third Reality watering kit entities"
- "Add Third Reality watering kit configuration"

zhaquirks/thirdreality/watering_kit.py
```diff
@@ -0,0 +1,57 @@
...
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RWK0148Z")
+    .replaces(ThirdRealityWateringKitCluster)
+    .number(
+        attribute_name=ThirdRealityWateringKitCluster.AttributeDefs.water_duration.name,
+        min_value=1,
+        max_value=1800,
+        step=1,
+        unit=UnitOfTime.SECONDS,
```
TheJulianJES, line 43: We can use the `NumberDeviceClass.DURATION` device class for HA here.
```suggestion
        device_class=NumberDeviceClass.DURATION,
        unit=UnitOfTime.SECONDS,
```

zhaquirks/thirdreality/watering_kit.py
```diff
@@ -0,0 +1,57 @@
...
+        cluster_id=ThirdRealityWateringKitCluster.cluster_id,
+        translation_key="water_duration",
+        fallback_name="Water duration",
+    )
+    .number(
+        attribute_name=ThirdRealityWateringKitCluster.AttributeDefs.water_interval.name,
+        min_value=0,
+        max_value=30,
+        step=1,
+        unit=UnitOfTime.DAYS,
```
TheJulianJES, line 54: We can use the `NumberDeviceClass.DURATION` device class for HA here.
```suggestion
        device_class=NumberDeviceClass.DURATION,
        unit=UnitOfTime.DAYS,
```

hwzolin:
> I'll push the suggested changes myself.
> 
> Please keep in mind the correct capitalization needed for HA entities for future PRs: Due to HA naming conventions, only the first letter of an entity name should be upper-case (except for abbreviations). So it should be "Water duration" instead of "Water Duration", for example.

OK

zhaquirks/thirdreality/watering_kit.py
```diff
@@ -0,0 +1,57 @@
+"""Third Reality watering kit devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
```
TheJulianJES, line 10: and we then need to import it:
```suggestion
from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
```
(pre-commit will fix the order in a second)

hwzolin:
> Also, please use more descriptive PR titles for future PRs. Examples for this PR:
> 
> * "Add Third Reality watering kit options"
> * "Add Third Reality watering kit entities"
> * "Add Third Reality watering kit configuration"

OK

zhaquirks/thirdreality/watering_kit.py
hwzolin: OK

TheJulianJES (approved):
I think this looks good now, thanks!

hwzolin:
> I think this looks good now, thanks!

Thank you, too!

TheJulianJES:
It was mentioned that the device has an IasZone cluster. What does that do? Report low battery? @hwzolin 

ZHA doesn't support the low battery bit for the IasZone cluster, so if that's the only function, we can disable discovery of default entities for the IasZone cluster by adding this:
```python
.prevent_default_entity_creation(endpoint_id=1, cluster_id=IasZone.cluster_id)
```

An alternative is to "remove" the cluster entirely (from ZHA's point of view) via:
```python
.removes(IasZone.cluster_id)
```

hwzolin:
> It was mentioned that the device has an IasZone cluster. What does that do? Report low battery? @hwzolin
> 
> ZHA doesn't support the low battery bit for the IasZone cluster, so if that's the only function, we can disable discovery of default entities for the IasZone cluster by adding this:
> 
> ```python
> .prevent_default_entity_creation(endpoint_id=1, cluster_id=IasZone.cluster_id)
> ```
> 
> An alternative is to "remove" the cluster entirely (from ZHA's point of view) via:
> 
> ```python
> .removes(IasZone.cluster_id)
> ```

Hello, the setting is like this: our watering kit, in the process of watering, if it detects that there is no water, then it will trigger an alarm of no water. This alert also needs to be reflected in the ZHA.
The following is the display of our company APP：
- ias zone is yes：
ias_zone_yes
ias_zone_yes_app
- ias zone is no:
ias_zone_no
ias_zone_no_app

TheJulianJES:
Ah ok, thanks! That'll already work as expected then.

mguaylam:
@hwzolin is there reporting on the custom cluster (0xfff2)?
When I change the value on the device, there is no reporting. I tried to configure it with : `min_interval=0`, `max_interval=900`, `reportable_change=1` to no avail.

How do you know in the app the user has changed the device values?
I also made changes to watering time : `min_value=10`, `max_value=7200`, `step=10` if that's ok with you, the manual seem to say it can go as far as `7200 seconds`.

See https://github.com/zigpy/zha-device-handlers/pull/3983

hwzolin:
> @hwzolin is there reporting on the custom cluster (0xfff2)? When I change the value on the device, there is no reporting. I tried to configure it with : `min_interval=0`, `max_interval=900`, `reportable_change=1` to no avail.
> 
> How do you know in the app the user has changed the device values? I also made changes to watering time : `min_value=10`, `max_value=7200`, `step=10` if that's ok with you, the manual seem to say it can go as far as `7200 seconds`.
> 
> See #3983

@mguaylam Hello and thanks for your feedback. In order to reduce battery drain, the watering duration has been changed to 1 - 1800 seconds, depending on the settings of our new version of the device.

@TheJulianJES However, the user mentioned the issue about the report after the device modification is completed. I'm also curious about what the ZHA platform will do. Is it possible to display the value of the change in real time in the log on the right side of the page?
But I found a problem before, and raised an issue to the official, visible link: https://github.com/zigpy/zha-device-handlers/issues/3581
I wonder if there is a solution to this problem?

The following image is a screenshot of the water duration command and the response of the device:
image
image

TheJulianJES:
> the issue about the report after the device modification is completed. I'm also curious about what the ZHA platform will do. Is it possible to display the value of the change in real time in the log on the right side of the page?

To update the entity in Home Assistant, you can send a "Zigbee attribute report" for that attribute in the device firmware.
ZHA / Home Assistant will process that and update the entity state. This is the same as sending attribute reports on the "temperature" or "humidity" clusters for sensors, for example.

> But I found a problem before, and raised an issue to the official, visible link: https://github.com/zigpy/zha-device-handlers/issues/3581
> I wonder if there is a solution to this problem?

Replied in the issue. I think I managed to reproduce this, but it's mostly only a visual glitch, from what I can tell.
Found the same issue with the Matter integration.. Clicking on the number entity, the logbook section at the bottom never loads, just an endless spinner. Weird...

The entity should still work fine elsewhere, from what I can see. So, I still recommend setting correct units and hope this can be fixed in Home Assistant at some point.

hwzolin:
> > the issue about the report after the device modification is completed. I'm also curious about what the ZHA platform will do. Is it possible to display the value of the change in real time in the log on the right side of the page?
> 
> To update the entity in Home Assistant, you can send a "Zigbee attribute report" for that attribute in the device firmware. ZHA / Home Assistant will process that and update the entity state. This is the same as sending attribute reports on the "temperature" or "humidity" clusters for sensors, for example.
> 
> > But I found a problem before, and raised an issue to the official, visible link: #3581
> > I wonder if there is a solution to this problem?
> 
> Replied in the issue. I think I managed to reproduce this, but it's mostly only a visual glitch, from what I can tell. Found the same issue with the Matter integration.. Clicking on the number entity, the logbook section at the bottom never loads, just an endless spinner. Weird...
> 
> The entity should still work fine elsewhere, from what I can see. So, I still recommend setting correct units and hope this can be fixed in Home Assistant at some point.

OK, thanks for taking the time to look into this. Although the problem is not solved, you cleared up my confusion.

mguaylam:
> @mguaylam Hello and thanks for your feedback. In order to reduce battery drain, the watering duration has been changed to 1 - 1800 seconds, depending on the settings of our new version of the device.

1. @hwzolin so if I understand, there has been a change in the firmware to make max : 1800s? It says in the manual that the maximum is 7200s. photo_2025-03-26_11-43-48


2. About the reporting issue, it seem to be on the firmware side. The device need to report the change of the value to ZHA. If I change the numbers on the device, ZHA is not made aware of the change.

It works on the OnOff cluster : `2025-03-26 11:54:54.424 DEBUG (MainThread) [zha] Emitting event cluster_handler_attribute_updated with data ClusterAttributeUpdatedEvent(attribute_id=0, attribute_name='on_off', attribute_value=<Bool.false: 0>, cluster_handler_unique_id='b4:0e:06:0f:ff:e0:22:17:1:0x0006', cluster_id=6, event_type='cluster_handler_event', event='cluster_handler_attribute_updated') (1 listeners)`

But not on your custom cluster (0xfff2), I can't see any response from the device. You need to do a ZigBee attribute report.

https://github.com/user-attachments/assets/c009e0f4-f026-4b13-bb2f-9472b1f9a1b6

hwzolin:
> > @mguaylam Hello and thanks for your feedback. In order to reduce battery drain, the watering duration has been changed to 1 - 1800 seconds, depending on the settings of our new version of the device.
> 
> 1. @hwzolin so if I understand, there has been a change in the firmware to make max : 1800s? It says in the manual that the maximum is 7200s. photo_2025-03-26_11-43-48
> 2. About the reporting issue, it seem to be on the firmware side. The device need to report the change of the value to ZHA. If I change the numbers on the device, ZHA is not made aware of the change.
> 
> It works on the OnOff cluster : `2025-03-26 11:54:54.424 DEBUG (MainThread) [zha] Emitting event cluster_handler_attribute_updated with data ClusterAttributeUpdatedEvent(attribute_id=0, attribute_name='on_off', attribute_value=<Bool.false: 0>, cluster_handler_unique_id='b4:0e:06:0f:ff:e0:22:17:1:0x0006', cluster_id=6, event_type='cluster_handler_event', event='cluster_handler_attribute_updated') (1 listeners)`
> 
> But not on your custom cluster (0xfff2), I can't see any response from the device. You need to do a ZigBee attribute report.
> 
>  IMG_5027.MOV

@mguaylam  Hello, for your first question. Yes, that's what we're set up right now. You can check the latest electronic version of the user manual on our official website (link is: https://tr-common-us-prod.s3.us-west-1.amazonaws.com/product/manual/SmartWateringKit.pdf)

And your second question, I've already fed back to our developers. We will optimize this in future firmware releases.

Thank you for your attention, I believe that your valuable suggestions will help us make our products better.

hwzolin:
> > @mguaylam Hello and thanks for your feedback. In order to reduce battery drain, the watering duration has been changed to 1 - 1800 seconds, depending on the settings of our new version of the device.
> 
> 1. @hwzolin so if I understand, there has been a change in the firmware to make max : 1800s? It says in the manual that the maximum is 7200s. photo_2025-03-26_11-43-48
> 2. About the reporting issue, it seem to be on the firmware side. The device need to report the change of the value to ZHA. If I change the numbers on the device, ZHA is not made aware of the change.
> 
> It works on the OnOff cluster : `2025-03-26 11:54:54.424 DEBUG (MainThread) [zha] Emitting event cluster_handler_attribute_updated with data ClusterAttributeUpdatedEvent(attribute_id=0, attribute_name='on_off', attribute_value=<Bool.false: 0>, cluster_handler_unique_id='b4:0e:06:0f:ff:e0:22:17:1:0x0006', cluster_id=6, event_type='cluster_handler_event', event='cluster_handler_attribute_updated') (1 listeners)`
> 
> But not on your custom cluster (0xfff2), I can't see any response from the device. You need to do a ZigBee attribute report.
> 
>  IMG_5027.MOV

Hello, regarding the issue you mentioned, "About the reporting issue, it seems to be on the firmware side. The device needs to report the change of the value to ZHA. If I change the numbers on the device, ZHA is not made aware of the change."

Currently, we have released the firmware version 1.00.40. You can OTA to this version and experience the updated features.

mguaylam:
@hwzolin I have updated my device.

I now get reporting! But I can't seem to be able to configure it. Is it on purpose?

image

When I try to configure the reports it says unsupported_attributes.

Also, the IAS sensor always report OK even if my device has no water in it. Is this normal? When should it return not ok?
image

TheJulianJES:
If reporting is set by the device, we should not try to configure it.
Does the device report on its own successfully now?

mguaylam:
Yes. I'll test without configuration but if so, I'll remove reporting configuration.

PR #3930, opened 2025-03-04
Add auto close time to Sonoff SWV-BSP Smart Water Valve

micz:
I added these features to the Sonoff SWV-BSP Smart Water Valve (see #3298):

- Water Leakage binary sensor
- Water Shortage binary sensor
- `on_time` attribute, representing the number of seconds after opening before the valve automatically closes.
- A custom behaviour for the On command:
   1. If `on_time == 0` it's a normal On command (`0x01`)
   2. If `on_time != 0` it sends a on_with_timed_off (`0x42`), with the `on_time` numbers of seconds sets in the attribute

immagine immagine

This is my first attempt at creating a Quirk.
Any advice on whether I'm following the correct way would be appreciated.
Thank you!

## Checklist

### DIFF
--- a/zhaquirks/heiman/hs1sa_efa.py
+++ b/zhaquirks/heiman/hs1sa_efa.py
@@ -119,6 +119,7 @@ class CustomHeimanCluster(CustomCluster):
     QuirkBuilder()
     .applies_to("HEIMAN", "HS1SA-EF-3.0")
     .applies_to("HEIMAN", "HS1SA-E-PLUS")
+    .friendly_name(manufacturer="HEIMAN", model="HS1SA-E-PLUS")  # Used by newer fw
     .replaces(CustomHeimanCluster)
     .exposes_feature(SIREN_BASIC)
     .change_entity_metadata(


zhaquirks/sonoff/swv.py
```diff
@@ -28,13 +36,126 @@ class AttributeDefs(BaseAttributeDefs):
...
+        await asyncio.sleep(delay + 1)
+        await self.endpoint.on_off.read_attributes(["on_off"])
+
+
+def is_water_shortage(valve_state: ValveState) -> bool:
+    """Check if the valve state indicates water shortage."""
+    return valve_state in {
+        ValveState.Water_Shortage,
+        ValveState.Water_Shortage_And_Leakage,
+    }
```
fgsch: You can simplify this and use `& ValveState.Water_Shortage` instead.

zhaquirks/sonoff/swv.py
```diff
@@ -28,13 +36,126 @@ class AttributeDefs(BaseAttributeDefs):
...
+
+
+def is_water_leakage(valve_state: ValveState) -> bool:
+    """Check if the valve state indicates water leakage."""
+    return valve_state in {
+        ValveState.Water_Leakage,
```
fgsch: You can simplify this and use `& ValveState.Water_Leakage` instead.

fgsch:
FWIW, I have a similar PR minus the `on_time` in https://github.com/zigpy/zha-device-handlers/pull/3910.

micz:
Thank you @fgsch, I've updated the code.
The `on_time` is the feature I need as a failsafe if the Zigbee connection fails.
What do you think about it? Have you tried it?

I didn't see your PR, I'm sorry.
I saw that you changed the Cluster name, do you think it's better to change it here also?
Thank you!

fgsch:
> Thank you @fgsch, I've updated the code. 

👍 

> The `on_time` is the feature I need as a failsafe if the Zigbee connection fails. What do you think about it? Have you tried it?

I haven't tried it yet. 

IIUC you are using the attribute for the automatic valve close if there is no water together with the timing irrigation command for this purpose? Is this correct?

> I didn't see your PR, I'm sorry. I saw that you changed the Cluster name, do you think it's better to change it here also? Thank you!

No worries. I changed the name to match other devices. Might as well do it here but really up to you.

micz:
> IIUC you are using the attribute for the automatic valve close if there is no water together with the timing irrigation command for this purpose? Is this correct?

I'm sorry I don't understand your question.

I added the `on_time` attribute to let the user decide how much time set for the auto close (0 = no autoclose).
Then, I use that attribute in the command to determine whether it needs to be changed to `0x42`.

I added also the two alarms, because I found your PR #3346 that added the attributes and I read that the other PRs in ZHA needed for the `attribute_converter` were done, but I didn't found your last PR, so I added them here.

fgsch:
>[..]
> I'm sorry I don't understand your question.
> 
> I added the `on_time` attribute to let the user decide how much time set for the auto close (0 = no autoclose). Then, I use that attribute in the command to determine whether it needs to be changed to `0x42`.

That attribute, 0x5011, is used to turn the valve off if there is no water for the specified amount of time (as per https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sonoff.ts#L1447).

This PR uses this attribute as a timer, which is later used with command 0x42 for timed irrigation (as per https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sonoff.ts#L329-L363).

Effectively, this is changing what 0x5011 is used for, assuming I understood correctly. 
Maybe there is a better way to implement this that does not (ab)use existing attributes.

micz:
Now I understand.
It's my first quirk so any help in more than appreciated! 😄

My idea was to keep a single switch but let the user set a timeout and then send a 42 instead of a 01.

I found no way to add and store that value. Trying to figure it out I found that parameter, but I didn't get it was used for water shortage.
I tried to find a solution in other quirks v2, but I didn't find anything.
In z2m you could send a different command issuing a 42, but in ZHA it's different. If I got it right.

Could you point me in the right direction of adding an optional timeout in the device config?
Thank you!

fgsch:
>[..]
>
> Could you point me in the right direction of adding an optional timeout in the device config? Thank you!

I'm not sure. Perhaps one of the devs could help.

micz:
@fgsch, are you the reviewer who needs to respond based on the label?
Thank you.

TheJulianJES:
No, it's a label for me (or other reviewers) that an answer/review/help is needed.
I'll come back to this when I find some time.

micz:
Ok, thank you.

micz:
I've updated the PR for version 0.0.136, considering the merging of #3910.

LethalWhisper:
I've already tested your quirk with my 2 Sonoff SWV and it works perfectly fine :)

In my opinion it is ready for merge :)

I am currently trying to implement the few missing features that you may find in z2m, although it's not that useful.

zhaquirks/sonoff/swv.py
```diff
@@ -31,10 +38,87 @@ class AttributeDefs(BaseAttributeDefs):
...
+    @property
+    def _is_manuf_specific(self):
+        return False
+
+
+class SwvOnOff(NoReplyMixin, CustomCluster, OnOff):
+    """SWV On Off Cluster."""
+
+    cluster_id = OnOff.cluster_id
+    cmd_values = OnOff.commands_by_name.values()
```
puddly:
```suggestion
```

zhaquirks/sonoff/swv.py
```diff
@@ -31,10 +38,87 @@ class AttributeDefs(BaseAttributeDefs):
...
+                CustomSonoffCluster.AttributeDefs.on_time.id
+            )
+            if on_time_value is not None and on_time_value != 0:
+                command = self.server_commands[0x42]  # on_with_timed_off
+                kwargs["on_off_control"] = 0x00
+                kwargs["on_time"] = on_time_value
+                kwargs["off_wait_time"] = 1
+                self.create_catching_task(self._turn_off_later(on_time_value))
+
+        return await self.request(
```
puddly: You should call `return await super().command(*args, **kwargs)` instead, since this method is effectively proxying the call to `command`.

zhaquirks/sonoff/swv.py
```diff
@@ -31,10 +38,87 @@ class AttributeDefs(BaseAttributeDefs):
...
+            self._is_manuf_specific or command.is_manufacturer_specific
+        ):
+            manufacturer = self._manufacturer_id
+        if command_id == 0x01:
+            base_cluster = self.endpoint.in_clusters[CustomSonoffCluster.cluster_id]
+            on_time_value = base_cluster.get(
+                CustomSonoffCluster.AttributeDefs.on_time.id
+            )
+            if on_time_value is not None and on_time_value != 0:
+                command = self.server_commands[0x42]  # on_with_timed_off
```
puddly:
```suggestion
                command = self.ServerCommandDefs.on_with_timed_off
```

What does this line do? `command` isn't used anywhere.

zhaquirks/sonoff/swv.py
```diff
@@ -31,10 +38,87 @@ class AttributeDefs(BaseAttributeDefs):
...
+            typing.Coroutine: A coroutine that sends the command and waits for the response.
+
+        """
+        command = self.server_commands[command_id]
+
+        if manufacturer is None and (
+            self._is_manuf_specific or command.is_manufacturer_specific
+        ):
+            manufacturer = self._manufacturer_id
+        if command_id == 0x01:
```
puddly:
```suggestion
        if command_id == self.ServerCommandDefs.on.id:
```

zhaquirks/sonoff/swv.py
```diff
@@ -31,10 +38,87 @@ class AttributeDefs(BaseAttributeDefs):
...
+            manufacturer=manufacturer,
+            expect_reply=expect_reply,
+            tsn=tsn,
+            **kwargs,
+        )
+
+    async def _turn_off_later(self, delay):
+        """We are not receiving the auto off event, so we force an update."""
+        await asyncio.sleep(delay + 1)
+        await self.endpoint.on_off.read_attributes(["on_off"])
```
puddly: Does the device not report the attribute change on its own, when it turns back off?

puddly (changes requested):
I've added a few code quality comments. I do have a question about this functionality as a whole, however: it looks like this PR basically adds a software-only "auto close time" configurable entity to the device and then intercepts `OnOff:on` commands, substituting them with `OnOff:on_with_timed_off`.

Since `OnOff:on_with_timed_off` is a standard ZCL cluster command, it feels odd to change the behavior of this one device. We should maybe instead consider implementing this as an action for all devices with `OnOff` cluster support? @TheJulianJES thoughts?

TheJulianJES:
> Since `OnOff:on_with_timed_off` is a standard ZCL cluster command, it feels odd to change the behavior of this one device. We should maybe instead consider implementing this as an action for all devices with `OnOff` cluster support? @TheJulianJES thoughts?

Yeah, I don't think this implementation is optimal. If needed, this can already be done using a HA automation that sends a cluster command. `homeassistant.update_entity` can be called to poll the entity after the delay if it doesn't report its state.

The "`on_time`" (`0x5011`) attribute also seems to be used by the firmware to determine when the device should turn off if there's no water. See Z2M for reference: https://github.com/Koenkk/zigbee-herdsman-converters/blob/4275931105788ec04b5b671640a2c86efa97f131/src/devices/sonoff.ts#L1954-L1962
It seems like we're misusing that attribute for local storage here?

fcrozat:
> > Since `OnOff:on_with_timed_off` is a standard ZCL cluster command, it feels odd to change the behavior of this one device. We should maybe instead consider implementing this as an action for all devices with `OnOff` cluster support? @TheJulianJES thoughts?
> 
> Yeah, I don't think this implementation is optimal. If needed, this can already be done using a HA automation that sends a cluster command. `homeassistant.update_entity` can be called to poll the entity after the delay if it doesn't report its state.
> 
> The "`on_time`" (`0x5011`) attribute also seems to be used by the firmware to determine when the device should turn off if there's no water. See Z2M for reference: https://github.com/Koenkk/zigbee-herdsman-converters/blob/4275931105788ec04b5b671640a2c86efa97f131/src/devices/sonoff.ts#L1954-L1962 It seems like we're misusing that attribute for local storage here?

This is a different feature from the device. 

https://github.com/Koenkk/zigbee-herdsman-converters/blob/4275931105788ec04b5b671640a2c86efa97f131/src/devices/sonoff.ts#L278 function is how timed / cyclic irrigation is managed by Z2M, using '0x5008' attribute

PaulBol:
> This is a different feature from the device.
> 
> https://github.com/Koenkk/zigbee-herdsman-converters/blob/4275931105788ec04b5b671640a2c86efa97f131/src/devices/sonoff.ts#L278 function is how timed / cyclic irrigation is managed by Z2M, using '0x5008' attribute

@fcrozat I think your are referring to something else. This branch is about sending an on_with_timed_off command (0x42) with a parameter that tells the valve to close after a certain amount of time. The review questions were how to send the command and where to store this interval.

I would love to see the cyclic irrigation options implemented in ZHA as well. But that seems to be out of scope for this branch.

fcrozat:
Sorry, I wasn't clear enough. I wanted to expand a bit @TheJulianJES answer:

the way timed irrigation is handled on Z2M is simply through a 1 cycle of cycled irrigation (if I read the code right) and not (mis)-using "on_time" (0x5011).

micz:
This is my first attempt at creating a quirk, so I'm really open to help and suggestions. :relaxed:
What I needed was a failsafe mechanism to close the valve even if a network error occurs when the close command is sent.

Looking at z2m code, if I get it right, I saw that it's possible to send the `OnOff:on_with_timed_off` that will close the valve after `on_time` seconds.

I wanted to send this command using the already available switch if the `on_time` is > 0 and not using an automation to send a cluster command (that I didn't know it was possible before reading the [comment](https://github.com/zigpy/zha-device-handlers/pull/3930#issuecomment-3082440220) here).

I checked the cyclic code in z2m, but it was too difficult to translate and test in a real case scenario.

If using the `on_time` in this way is incorrect, how is it possible to achieve the failsafe implementation I described?

Thank you for your support.

micz:
@puddly @TheJulianJES Do I have to fix the requested changes, or is it necessary to verify if this is the right approach as stated [here](https://github.com/zigpy/zha-device-handlers/pull/3930#issuecomment-3082440220)?

puddly:
I just realized that ZHA doesn't actually implement the HA Valve platform! I think, since this is a standard ZCL command and theoretically is supported by many other valves in a standard way, we could just globally implement this for all valves, since this functionality likely isn't useful for other device types.

@TheJulianJES thoughts?

MattWestb:
Have Matter implanting valve control in its cluster library ?
If yes implanting it as they have done as its a part pf the upcoming ZCL 9 then its being released.

ausfas:
Hi @micz , I had to dig a lot to come to this PR ( and thanks for proposing the changes) as these are the changes I have been looking for:
Water Shortage binary sensor and Auto Close Timer, State when Power etc. When I installed the Sonoff valve it somehow picked these features (not sure from where and how) but in later instances its falling to default Sonoff valve with just on /off and sensors but not the additional binary sensor and the slider. I don't like the very basic function as Zigbee can crash and I don't want the valve to remain open so I need to have water shortage as off as an absolute must. 

I hope your changes are implemented as otherwise I need to use your code an implement it as Custom Quirk as I don't want the existing basic valve features at all. 

Some users are saying its because of old Firmware so updating it to 0x00001004 to see if additional features appear for ZHA or not.

micz:
> Some users are saying its because of old Firmware so updating it to 0x00001004 to see if additional features appear for ZHA or not

You need to update the firmware to use those features and also this PR as a quirk.

Hedda:
> I just realized that ZHA doesn't actually implement the HA Valve platform! I think, since this is a standard ZCL command and theoretically is supported by many other valves in a standard way, we could just globally implement this for all valves, since this functionality likely isn't useful for other device types.

For reference, there is now an open pull request in the zha repo that will add support for HA's Valve platform to ZHA, see:

* https://github.com/zigpy/zha/pull/673

Also be aware of these related feature requests and  discussions (which in turn links to other indirectly related discussions):

* https://github.com/zigpy/zha-device-handlers/issues/4340

* https://github.com/zigpy/zha/issues/528

> Have Matter implanting valve control in its cluster library ? If yes implanting it as they have done as its a part pf the upcoming ZCL 9 then its being released.

ZCL R8 spec does not actually support valves in any way, but the Matter 1.3 spec does, so support might come in ZCL R9

ausfas:
Unfortunately, no update to add cluster command to pass on timer off to valve:
  - action: zha.issue_zigbee_cluster_command
            data:
              ieee: "{{ ieee }}"
              endpoint_id: 1
              cluster_type: in
              command_type: server
              cluster_id: 6
              command: 66
              params:
                on_off_control: 0
                on_time: "{{ secs | int }}"
                off_wait_time: 1

Issue #3118, opened 2024-04-21
[Device Support Request] custom quirk for Tuya Zemismart blind cover motor TS0601 _TZE200_68nvbio9 with battery


stast1:
### Problem description

No battery charge display for Tuya Zemismart blind cover motor TS0601 _TZE200_68nvbio9

### Solution description

Add battery charge display

### Custom quirk

<details><summary>Custom quirk</summary>

```python
"""Tuya based cover and blinds."""
from typing import Optional, Union

from zigpy.profiles import zha
import zigpy.types as t
from zigpy.zcl import foundation
from zigpy.zcl.clusters.general import Basic, Groups, Ota, Scenes, Time

from zhaquirks import Bus
from zhaquirks.const import (
    DEVICE_TYPE,
    ENDPOINTS,
    INPUT_CLUSTERS,
    MODELS_INFO,
    OUTPUT_CLUSTERS,
    PROFILE_ID,
)
from zhaquirks.tuya import (
    TuyaManufacturerWindowCover,
    TuyaManufCluster,
    TuyaPowerConfigurationCluster,
    TuyaWindowCover,
    TuyaWindowCoverControl,
    TUYA_GET_DATA,
    TUYA_SET_DATA_RESPONSE,
    TUYA_DP_TYPE_VALUE,
)

TUYA_DP_ID_BATTERY = 0x0D

class TuyaZemismartManufacturerWindowCover(TuyaManufacturerWindowCover):
    """Manufacturer Specific Cluster for Zemismart cover device."""

    def handle_cluster_request(
        self,
        hdr: foundation.ZCLHeader,
        args: tuple[TuyaManufCluster.Command],
        *,
        dst_addressing: Optional[
            Union[t.Addressing.Group, t.Addressing.IEEE, t.Addressing.NWK]
        ] = None,
    ) -> None:
        """Handle cluster request."""
        # Tuya Zemismart Specific Cluster Commands
        if (
            hdr.command_id in (TUYA_GET_DATA, TUYA_SET_DATA_RESPONSE) 
            and args[0].command_id == TUYA_DP_TYPE_VALUE + TUYA_DP_ID_BATTERY
        ):
            self.endpoint.device.battery_bus.listener_event(
                "battery_change",
                args[0].data[4],
            )
        else:
            super().handle_cluster_request(hdr, args, dst_addressing=dst_addressing)

class TuyaZemismartSmartCover0601_battery(TuyaWindowCover):
[...]

PR #2404, opened 2023-05-18
Add battery reporting for Zemismart ZM85EL cover motors

michalmo:
## Proposed change
Zemismart ZM85EL series devices that I own report battery status as Tuya attribute `0x0D`. This PR adds optional support for reporting battery level to `TuyaManufacturerWindowCover` and moves the 3 manufacturer ids of these devices to a new quirk, with battery support enabled.

## Additional information
ZM85EL manufacturer ids are:

- `_TZE200_pw7mji0l` [ZM85EL-1Z](https://zigbee.blakadder.com/Zemismart_ZM85EL-1Z.html) (added in #1616) - I **do not** own a device with this id
- `_TZE200_68nvbio9` alternative id for ZM85EL-1Z bought from Zemismart store on Aliexpress (added in #1349) - I own this device
- `_TZE200_cf1sl3tj` [ZM85EL-2Z](https://zigbee.blakadder.com/Zemismart_ZM85EL-2Z.html) (added in #1652) - I own this device

These devices appeared in different quirks which have the same signatures and configuration.

## Checklist
- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

Related: #3118

PR #3936, opened 2025-03-07
Add Tuya cover TS0601 `_TZE200_ba69l9ol` variant

JeffResc:
## Proposed change
New cover model _TZE200_ba69l9ol added for Tuya TS0601

## Additional information
image
New Python file was tested as a custom quirk and works as intended.

## Checklist


JeffResc:
I'd also like to note that the battery quirk from #3118 seems to work well with this model, but would need the cover inverted

TheJulianJES:
> I'd also like to note that the battery quirk from https://github.com/zigpy/zha-device-handlers/issues/3118 seems to work well with this model, but would need the cover inverted

Can you link/post the exact quirk with which the battery works (but where it's wrongly inverted)?
So, is that from the "Custom quirk" section from the initial post in https://github.com/zigpy/zha-device-handlers/issues/3118 or maybe this PR https://github.com/zigpy/zha-device-handlers/pull/2404?

PR #3947, opened 2025-03-10
Add Tuya/NEO soil sensor NAS-STH02B2 `_TZE284_rqcuwlsa`

pdecat:
## Proposed change

This adds support for the [NEO NAS-STH02B2](https://www.zigbee2mqtt.io/devices/NAS-STH02B2.html) `_TZE284_rqcuwlsa`.

This was heavily inspired from this code: https://github.com/Koenkk/zigbee-herdsman-converters/blob/v23.4.0/src/devices/neo.ts#L369-L456

## Additional information

It works fine for all attributes but the electrical conductivity sensor which does not show in the entities:

image

In Z2M, it reports correctly (value changes when putting the device in different physical conditions):
image

## Checklist
* [ ]  The changes are tested and work correctly
* [x]  `pre-commit` checks pass / the code has been formatted using Black
* [ ]  Tests have been added to verify that the new code works

### DIFF
--- a/zhaquirks/tuya/builder/__init__.py
+++ b/zhaquirks/tuya/builder/__init__.py
@@ -20,6 +20,7 @@ from zigpy.zcl import foundation
 from zigpy.zcl.clusters.measurement import (
     PM25,
     CarbonDioxideConcentration,
+    ElectricalConductivity,
     FormaldehydeConcentration,
     IlluminanceMeasurement,
     RelativeHumidity,
@@ -63,6 +64,10 @@ BATTERY_VOLTAGES = {
 
 class TuyaCO2Concentration(CarbonDioxideConcentration, TuyaLocalCluster):
     """Tuya Carbon Dioxide concentration measurement."""
+
+
+class TuyaElectricalConductivity(ElectricalConductivity, TuyaLocalCluster):
+    """Tuya Electrical Conductivity measurement."""
 
 
 class TuyaFormaldehydeConcentration(FormaldehydeConcentration, TuyaLocalCluster):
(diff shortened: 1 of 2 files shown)


zhaquirks/tuya/builder/__init__.py
```diff
@@ -65,6 +66,10 @@ class TuyaCO2Concentration(CarbonDioxideConcentration, TuyaLocalCluster):
     """Tuya Carbon Dioxide concentration measurement."""
 
 
+class TuyaElectricalConductivity(ElectricalConductivity, TuyaLocalCluster):
```
pdecat, line 69: This was taken from https://github.com/zigpy/zha-device-handlers/issues/2565#issuecomment-1712541101

pdecat:
Not sure what's missing to make the electrical conductivity sensor working in ZHA.

TheJulianJES:
> Not sure what's missing to make the electrical conductivity sensor working in ZHA.

ZHA doesn't create any sensors for the `ElectricalConductivity` cluster at the moment. It can either be added to ZHA, so we can map it from the Tuya datapoint to a local/fake ZCL `ElectricalConductivity` cluster, or we can add a custom quirks v2 sensor for this device.
Since it's a standard ZCL cluster, we may want to do the former and add support for this in ZHA.

pdecat:
Hi @TheJulianJES, I've submitted https://github.com/zigpy/zha/pull/400 to try and implement the electrical conductivity cluster in ZHA. PTAL

It does make the new sensor appear:
image

Testing in progress...

pdecat:
There's probably some scaling to implement as the measured value seems way too high compared to [usual values](https://www.nrcs.usda.gov/sites/default/files/2022-10/Soil%20Electrical%20Conductivity.pdf) when the device is installed into my garden:

image

pdecat:
The Z2M unit seems to be in μS/cm, and [1 µS/cm = 0.001 dS/m](https://www.123calculus.com/en/conductivity-conversion-page-4-98-100.html), so the value either needs to be scaled or the unit changed. 
Also AFAICT, SI unit of conductivity is _[siemens per metre (S/m)](https://en.wikipedia.org/wiki/Electrical_resistivity_and_conductivity), so we may want to use that.
image

TheJulianJES (approved):
Looks good. Thanks! This should make it into the HA beta 2025.4.0b1 released later today.

If there's still scaling to be applied (although I think it should be fine now with µS/cm as the ZCL and ZHA default), please do it in a follow-up PR. We can merge that as a bugfix then.

PR #3958, opened 2025-03-12
Update script/setup to use uv

abmantis:
## Proposed change

Took a bit of inspiration from the HA script/setup script to improve this one a bit:

- Use `uv` instead of `pip`
- Check if we are already in a venv before trying to create and activate one. This allows using the script to update the requirements in an existing venv (instead of having to manually delete it first).

## Checklist


TheJulianJES:
Thanks, this looks good to me. We may wanna adjust our ZHA script too: https://github.com/zigpy/zha/blob/dev/script/setup
I think this approach is better than the ZHA one.

TheJulianJES (approved):
Thanks!

Issue #3942, opened 2025-03-09
[Device Support Request] ubisys devices device setup cluster


FlorianBruckner:
### Problem description

While ubisys devices work fine ootb, they support a cluster 0xfc00 for device setup (e.g. https://www.ubisys.de/downloads/ubisys-s1-technical-reference.pdf page 24). This cluster appears to be available on many ubisys devices (the documentation at https://www.ubisys.de/downloads/ubisys-input-configuration-integrators-guide.pdf mentions  C4, D1(-R), S1(-R), S2(-R), J1(-R) and LD6).

The attributes of this cluster are currently not available in the device UI in ZHA

### Solution description

A device handler that declares the device setup cluster for ubisys devices so I can easily configure the devices.

I am more than happy to contribute, but need some guidance.

As this device setup funcationality is available on many ubisys devices, I was looking for a way to provide this handler not just for one specific device but for many (potentially all ubisys devices). As far as I understand the documentation, though, a device handler needs to be created for every single device. Is there a way to declare a device handler for a number of devices, adding only the cluster declaration that we need to see this?

### Device signature

<details><summary>Device signature ubisys S1</summary>

```json

{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4338,
    "maximum_buffer_size": 127,
    "maximum_incoming_transfer_size": 127,
    "server_mask": 10752,
    "maximum_outgoing_transfer_size": 127,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x010a",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006"
      ],
      "output_clusters": []
    },
    "2": {
      "profile_id": "0x0104",
      "device_type": "0x0001",
      "input_clusters": [
        "0x0000",
        "0x0003"
      ],
      "output_clusters": [
        "0x0005",
        "0x0006",
        "0x0008"
      ]
    },
    "3": {
      "profile_id": "0x0104",
      "device_type": "0x0501",
      "input_clusters": [
        "0x0000",
        "0x0702",
        "0x0b04"
      ],
      "output_clusters": []
    },
    "200": {
[...]

PR #3959, opened 2025-03-12
ubisys S1, S2 WIP

FlorianBruckner:
## Proposed change
Manufacturer specific cluster for ubisys devices for device configuration as requested in #3942

## Additional information
see https://www.ubisys.de/wp-content/uploads/ubisys-s1-technical-reference.pdf and https://www.ubisys.de/downloads/ubisys-input-configuration-integrators-guide.pdf

## Checklist

### DIFF
--- a/zhaquirks/ubisys/ubisys.py
+++ b/zhaquirks/ubisys/ubisys.py
@@ -1,0 +1,38 @@
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.zcl.clusters.general import LevelControl
+from zigpy.quirks.v2 import CustomDeviceV2, CustomCluster
+from zigpy.zcl import ClusterType
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+class UbisysDevice(CustomDeviceV2):
+    """Ubisys device class """
+
+class UbisysManufacturerSpecificCluster(CustomCluster):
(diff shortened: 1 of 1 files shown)


FlorianBruckner:
Dear @TheJulianJES, Dear @puddly 

I have started with a device handler for my ubisys devices to allow for device configuration. Unfortunately I am stuck and hope to get some help from you.

The documentation (https://www.ubisys.de/wp-content/uploads/ubisys-s1-technical-reference.pdf) defines the attributes as 

```
This attribute is an array (ZCL data type 0x48) of 8-bit data (ZCL data type 0x08).
```

and

```
This attribute is an array (ZCL data type 0x48) of raw binary data (ZCL data type 0x41).
```

How would I declare these attributes and what would be the format (in the UI) for input and output? Or is this approach unfit for this purpose anyway and you would suggest a different one?

puddly:
The cluster commands and attributes need to be defined either way, regardless of how entities interface with them.

The ZCL array type define the data type dynamically so you can just specify the type as `Array` (there are other instances of this in the codebase).

Each array element seems to encompass two bits of configuration: disabling the input and inverting its logic level. I guess you can skip the input disabling part and just provide dropdowns to select the logic level? On the other hand, this can also easily be done in an automation.

I don't believe we have an API yet to process the value of an attribute bidirectionally, when creating a v2 quirk entity.

PR #3961, opened 2025-03-14
Add ELIT Scandinavia IHC Link quirk

kimtnielsen:
## Checklist


zhaquirks/elit/ihc_link.py
```diff
@@ -0,0 +1,116 @@
...
+            cluster = endpoint.in_clusters.get(MultistateInput.cluster_id)
+            if cluster:
+                await cluster.bind()
+                await cluster.configure_reporting(
+                    MultistateInput.AttributeDefs.present_value.id, 0, 0, 0
+                )
+
+            cluster = endpoint.in_clusters.get(OnOff.cluster_id)
+            if cluster:
+                await cluster.bind()
```
TheJulianJES, line 57: Is binding and manually configuring reporting really required here?

We already bind the `MultistateInput` cluster by default and set up reporting for the `present_value` attribute in ZHA: https://github.com/zigpy/zha/blob/695475a927451524475bdcef90a25177727e2345/zha/zigbee/cluster_handlers/general.py#L383-L392

zhaquirks/elit/ihc_link.py
```diff
@@ -0,0 +1,116 @@
...
+            and (action := ACTION_TYPE.get(value)) is not None
+        ):
+            event_args = {VALUE: value}
+            self.listener_event(ZHA_SEND_EVENT, action, event_args)
+
+
+(
+    QuirkBuilder(ELIT, "IHC Link")
+    .device_class(EHCLinkDevice)
+    .skip_configuration()
```
TheJulianJES, line 79: Ah, I see you're skipping the ZHA configuration here. I don't think we should do that.

Can you provide the full device signature to see if there are any clusters that perhaps shouldn't be bound (which I'm guessing this is why you've added this)?

kimtnielsen: Yes, there are really many endpoints (34 endpoints in worst case) and therefore we have to skip the automatic bindings, otherwise the device will work very slow and in some cases the binding table runs full. For the endpoints with MultistateInput, only this attribute needs binding. 
This device enables/disables endpoints when it powered up the first time (is static after joining), depending on what is connected to it wired databusses. So it can have either 32 input endpoints, or 16 inputs + 8 outputs, or 16 outputs.

PR #3967, opened 2025-03-18
Fix wrong Tuya DPToAttributeMapping imports

abmantis:
## Proposed change

These modules were importing the wrong `DPToAttributeMapping` class, since the builder uses the one from `mcu`.

## Checklist

### DIFF
--- a/tests/test_tuya_builder.py
+++ b/tests/test_tuya_builder.py
@@ -18,7 +18,6 @@ from zhaquirks.const import BatterySize
 from zhaquirks.tuya import (
     TUYA_QUERY_DATA,
     TUYA_SET_TIME,
-    DPToAttributeMapping,
     TuyaCommand,
     TuyaData,
     TuyaDatapointData,
@@ -42,7 +41,7 @@ from zhaquirks.tuya.builder import (
     TuyaTemperatureMeasurement,
     TuyaValveWaterConsumedNoInstDemand,
 )
-from zhaquirks.tuya.mcu import TuyaMCUCluster, TuyaOnOffNM
+from zhaquirks.tuya.mcu import DPToAttributeMapping, TuyaMCUCluster, TuyaOnOffNM
 from zhaquirks.tuya.tuya_sensor import NoManufTimeTuyaMCUCluster
 
 ZCL_TUYA_SET_TIME = b"\x09\x12\x24\x0d\x00"
--- a/zhaquirks/tuya/tuya_rain.py
+++ b/zhaquirks/tuya/tuya_rain.py
@@ -6,14 +6,9 @@ from zigpy.quirks.v2.homeassistant.sensor import SensorDeviceClass, SensorStateC
 import zigpy.types as t
 from zigpy.zcl.clusters.security import IasZone
 
-from zhaquirks.tuya import (
-    TUYA_CLUSTER_ID,
-    BatterySize,
-    DPToAttributeMapping,
-    TuyaLocalCluster,
-)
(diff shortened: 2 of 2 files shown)


abmantis:
I actually wanted to get rid of the "duplicated" DPToAttributeMapping classes, since currently they are different (the one in `mcu` has an extra `dp_converter`, but then they both go into `_dp_to_attributes`. It works, but is a bad practice and breaks proper typing.

I had 2 options in mind, but both could break custom quirks:

1. rename the one in `tuya.mcu` to `MCUDPToAttributeMapping`, make it a subclass of `tuya.DPToAttributeMapping`, keeping only the `dp_converter` attribute in it (since the others would come from the super class).
2. delete `tuya.mcu.DPToAttributeMapping` and keep only `tuya.DPToAttributeMapping`, adding the `dp_converter` to it.

Unfortunatelly, both would break any custom quirk importing `tuya.mcu.DPToAttributeMapping`.

Any suggestions?

TheJulianJES:
> I actually wanted to get rid of the "duplicated" DPToAttributeMapping classes, since currently they are different (the one in `mcu` has an extra `dp_converter`, but then they both go into `_dp_to_attributes`. It works, but is a bad practice and breaks proper typing.

Yes, we definitely want to do this at some point. IMO the whole non-MCU Tuya module really needs to be cleaned up (very old Tuya v1 quirks use custom clusters defined there with `TuyaManufCluster`, which should be directly in the quirks using `dp_to_attribute`). If that's all done, we might be able to move everything modern that's left in the MCU module(?)

There are also multiple Tuya clusters that should be merged (e.g. `TuyaMCUCluster` and `TuyaNewManufCluster` probably).

> I had 2 options in mind, but both could break custom quirks:
> 1. rename the one in `tuya.mcu` to `MCUDPToAttributeMapping`, make it a subclass of `tuya.DPToAttributeMapping`, keeping only the `dp_converter` attribute in it (since the others would come from the super class).

Hmm, I think the `tuya.mcu` one would be used by more custom quirks. Not sure though..

> 2. delete `tuya.mcu.DPToAttributeMapping` and keep only `tuya.DPToAttributeMapping`, adding the `dp_converter` to it.
> 
> Unfortunatelly, both would break any custom quirk importing `tuya.mcu.DPToAttributeMapping`.

In that case, we should be able to add an import to the MCU module to allow custom quirks to import from both `zhaquirks.tuya` and `zhaquirks.tuya.mcu`. Both will reference the same `DPToAttributeMapping` then, so we shouldn't break anything:

```python
# when keeping `DPToAttributeMapping` in non-MCU module, add this import to MCU module:
from zhaquirks.tuya import DPToAttributeMapping  # noqa: F401
```

We might also be able to leave `DPToAttributeMapping` in the MCU module and add an import for custom quirks compatibility to the non-MCU Tuya module, though that might break due to circular imports then depending on how/what the MCU module is importing from the non-MCU Tuya module.

```python
# when keeping in MCU module, import in non-MCU module:
from zhaquirks.tuya.mcu import DPToAttributeMapping  # noqa: F401
```

Ultimately, it doesn't really matter where we put the `DPToAttributeMapping`. Having it in the non-MCU module might be cleaner import-wise. A lot of the more modern Tuya code is in the MCU module though.
We can still move it later if we eventually get around to cleaning up a lot of the other old Tuya stuff.

TheJulianJES (approved):
Thanks. `DPToAttributeMapping` can likely be consolidated with what's mentioned above, but this is an easy fix, so merging this already.

Issue #3585, opened 2024-12-06
[BUG] Aqara E1 TRV thermostat quirk does not expose external temperature input


mwhapples:
### Bug description

The Aqara E1 TRV allows sending it a external temperature sensor reading. The existing quirk in zhaquirks/xiaomi/aqara/thermostat_agl001.py does not expose this functionality. An example of how this should be sent can be found in zigbee-herdsman-converters and the functionality from that does work in zigbee2mqtt. It would be good if this functionality could be added to the quirk so it can be used with ZHA.

### Steps to reproduce

In home-assistant, view details of the Aqara E1 device. Look through entities exposed. See that external temperature input is not exposed.

### Expected behavior

The external temperature input should be exposed and when set it should send the value to the Aqara E1 TRV.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Device signature

<details><summary>Device signature</summary>

```json
[Paste the device signature here]
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
[Paste the diagnostic information here]
```

</details>

### Logs

<details><summary>Logs</summary>

```python
[Paste the logs here]
```

</details>

### Additional information

Refer to zigbee-herdsman-converters for a working implementation of this feature.

mwhapples:
Lost interest due to no responses, moved on to other solutions. The GitHub action to clean stuff up just confirms lack of interest and wanting to clear issues from being listed without fixing.

cl445:
For anyone tracking this issue, there are now two complementary PRs working towards full AGL001 feature parity with zigbee2mqtt:

**#4463** by @Elyasin - External temperature sensor support (addresses this issue directly)
**#4604** by @cl445 - Heartbeat parsing, type-safe enums, device temperature cluster

Both PRs have passing CI and are ready for review. Together they would significantly improve the Aqara E1 TRV experience in ZHA.

Hedda:
> For anyone tracking this issue, there are now two complementary PRs working towards full AGL001 feature parity with zigbee2mqtt:
> 
[...]

PR #2802, opened 2023-12-05
Add external temperature sensor support to Aqara E1 thermostat

mb1248:
## Proposed change
Support for external temperature sensor.

## Additional information
This is a copy of the not merged part from #2109. Discussion starts under #2687 

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

Related: #3585

PR #3974, opened 2025-03-23
Add support for Aqara external temperature sensor

Antoninj:
## Proposed change
Support for external temperature sensor.

## Additional information

This is a copy of https://github.com/zigpy/zha-device-handlers/pull/2802 + the missing unit tests

## Checklist

Related: #3585

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,7 +456,7 @@ The tests use the [pytest](https://docs.pytest.org/en/latest/) framework.
 To get set up, you need install the test dependencies:
 
 ```bash
-pip install -r requirements_test_all.txt
+pip install -r requirements_test.txt
 ```
 
 ### Running the tests
--- a/README.md
+++ b/README.md
@@ -396,6 +396,8 @@ replacement = {
             INPUT_CLUSTERS: [BinaryInput.cluster_id],
             OUTPUT_CLUSTERS: [BinaryInput.cluster_id, Groups.cluster_id],
         },
+    },
+}
 ```
 
 You can see that we have replaced `ElectricalMeasurement.cluster_id` from endpoint 1 in the `signature` dict with the name of our cluster that we created: `ElectricalMeasurementCluster` and on endpoint 2 we replaced `AnalogInput.cluster_id` with the implementation we created for that: `AnalogInputCluster`. This instructs Zigpy to use these `CustomCluster` derivatives instead of the normal cluster definitions for these clusters and this is why this part of the quirk is called `replacement`.
@@ -456,7 +458,7 @@ The tests use the [pytest](https://docs.pytest.org/en/latest/) framework.
 To get set up, you need install the test dependencies:
 
 ```bash
-pip install -r requirements_test.txt
+pip install -r requirements_test_all.txt
 ```
 
 ### Running the tests
--- a/zhaquirks/xiaomi/aqara/thermostat_agl001.py
+++ b/zhaquirks/xiaomi/aqara/thermostat_agl001.py
@@ -413,7 +413,7 @@ class AqaraThermostatSpecificCluster(XiaomiAqaraE1Cluster):
             )
         super()._update_attribute(attrid, value)
 
-    def aqaraHeader(self, counter: int, params: bytearray, action: int) -> bytearray:
+    def aqara_header(self, counter: int, params: bytearray, action: int) -> bytearray:
         """Create Aqara header for setting external sensor."""
         header = bytes([0xAA, 0x71, len(params) + 3, 0x44, counter])
         integrity = 512 - sum(header)
@@ -444,7 +444,7 @@ class AqaraThermostatSpecificCluster(XiaomiAqaraE1Cluster):
                 params += bytes([0x00, 0x01, 0x00, 0x55])
                 params += temperatureBuf
 
-                attrs[SENSOR_ATTR_NAME] = self.aqaraHeader(0x12, params, 0x05) + params
+                attrs[SENSOR_ATTR_NAME] = self.aqara_header(0x12, params, 0x05) + params
 
             elif attr_def and attr_def.id == SENSOR:
                 # set internal/external temperature sensor
@@ -468,10 +468,10 @@ class AqaraThermostatSpecificCluster(XiaomiAqaraE1Cluster):
 
                     attrs1 = {}
                     attrs1[SENSOR_ATTR_NAME] = (
-                        self.aqaraHeader(0x12, params1, 0x04) + params1
+                        self.aqara_header(0x12, params1, 0x04) + params1
                     )
                     attrs[SENSOR_ATTR_NAME] = (
-                        self.aqaraHeader(0x13, params2, 0x04) + params2
+                        self.aqara_header(0x13, params2, 0x04) + params2
                     )
 
                     await super().write_attributes(attrs1, manufacturer)
@@ -530,10 +530,10 @@ class AqaraThermostatSpecificCluster(XiaomiAqaraE1Cluster):
 
                     attrs1 = {}
                     attrs1[SENSOR_ATTR_NAME] = (
-                        self.aqaraHeader(0x12, params1, 0x02) + params1
+                        self.aqara_header(0x12, params1, 0x02) + params1
                     )
                     attrs[SENSOR_ATTR_NAME] = (
-                        self.aqaraHeader(0x13, params2, 0x02) + params2
+                        self.aqara_header(0x13, params2, 0x02) + params2
                     )
 
                     await super().write_attributes(attrs1, manufacturer)
--- a/zhaquirks/xiaomi/aqara/thermostat_agl001.py
+++ b/zhaquirks/xiaomi/aqara/thermostat_agl001.py
@@ -56,6 +56,8 @@ SENSOR_ATTR_NAME = "sensor_attr"
 
 XIAOMI_CLUSTER_ID = 0xFCC0
 
+SENSOR_ID = bytearray.fromhex("00158d00019d1b98")
+
 DAYS_MAP = {
     "mon": 0x02,
     "tue": 0x04,
@@ -426,26 +428,25 @@ class AqaraThermostatSpecificCluster(XiaomiAqaraE1Cluster):
         self, attributes: dict[str | int, Any], manufacturer: int | None = None
     ) -> list:
         """Write attributes to device with internal 'attributes' validation."""
-        sensor = bytearray.fromhex("00158d00019d1b98")
-        attrs = {}
+        attrs: dict[str | int, Any] = {}
 
         for attr, value in attributes.items():
             # implemented with help from https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/xiaomi.js
+            attr_def = self.find_attribute(attr)
 
-            if attr == SENSOR_TEMP:
+            if attr_def and attr_def.id == SENSOR_TEMP:
                 # set external sensor temp. this function expect value to be passed multiplied by 100
                 temperatureBuf = bytearray.fromhex(
                     self._float_to_hex(round(float(value)))[2:]
                 )
 
-                params = sensor
+                params = SENSOR_ID
                 params += bytes([0x00, 0x01, 0x00, 0x55])
                 params += temperatureBuf
 
-                attrs = {}
                 attrs[SENSOR_ATTR_NAME] = self.aqaraHeader(0x12, params, 0x05) + params
 
-            elif attr == SENSOR:
+            elif attr_def and attr_def.id == SENSOR:
                 # set internal/external temperature sensor
                 device = bytearray.fromhex(
                     f"{self.endpoint.device.ieee}".replace(":", "")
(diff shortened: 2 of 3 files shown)


zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -388,6 +395,8 @@ class AqaraThermostatSpecificCluster(XiaomiAqaraE1Cluster):
             SCHEDULE_SETTINGS: ("schedule_settings", ScheduleSettings, True),
             SENSOR: ("sensor", t.uint8_t, True),
             BATTERY_PERCENTAGE: ("battery_percentage", t.uint8_t, True),
+            SENSOR_TEMP: ("sensor_temp", t.uint32_t, True),
+            SENSOR_ATTR: (SENSOR_ATTR_NAME, t.LVBytes, True),
```
TheJulianJES, line 401: All these attributes should really be converted to the new zigpy `AttributeDefs` style. Search `AttributeDefs` in this repo or the zigpy repo for examples.
You don't have to do it in this PR, but it would be nice.

We can also access the name and id then: `AqaraThermostatSpecificCluster.AttributeDefs.sensor_temp.name` or `.id`.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +411,167 @@ def _update_attribute(self, attrid, value):
             )
         super()._update_attribute(attrid, value)
 
+    def aqaraHeader(self, counter: int, params: bytearray, action: int) -> bytearray:
```
TheJulianJES: This should be snake_case, so `aqara_header`.
Or maybe `wrap_with_aqara_header`? Something like that?

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +411,167 @@ def _update_attribute(self, attrid, value):
...
+                temperatureBuf = bytearray.fromhex(
+                    self._float_to_hex(round(float(value)))[2:]
+                )
+
+                params = sensor
+                params += bytes([0x00, 0x01, 0x00, 0x55])
+                params += temperatureBuf
+
+                attrs = {}
+                attrs[SENSOR_ATTR_NAME] = self.aqaraHeader(0x12, params, 0x05) + params
```
TheJulianJES: Is `attrs` intentionally set to `{}` again? It's already set at the beginning of the method. Depending on the order of the `attributes` parameter, the others wouldn't be written (although it's rare that this method gets multiple attributes to write at the same time).

 If it's not intentional, this could be rewritten as: 
```python
attrs = {SENSOR_ATTR_NAME: self.aqara_header(0x12, params, 0x05) + params}
```
If not, well, the `attrs = {}` can be removed here.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +411,167 @@ def _update_attribute(self, attrid, value):
...
+    def _float_to_hex(self, f):
+        """Convert float to hex."""
+        return hex(struct.unpack("<I", struct.pack("<f", f))[0])
+
+    async def write_attributes(
+        self, attributes: dict[str | int, Any], manufacturer: int | None = None
+    ) -> list:
+        """Write attributes to device with internal 'attributes' validation."""
+        sensor = bytearray.fromhex("00158d00019d1b98")
+        attrs = {}
```
TheJulianJES: I think it'd be nice to type this. Should be  `attrs: dict[str | int, Any] = {}` then.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +411,167 @@ def _update_attribute(self, attrid, value):
...
+                            0x00,
+                            0x00,
+                            0x00,
+                            0x00,
+                            0x00,
+                            0x00,
+                            0x00,
+                            0x00,
+                        ]
+                    )
```
TheJulianJES: I believe you can just do `params1 += bytes(12)`.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +411,167 @@ def _update_attribute(self, attrid, value):
...
```
TheJulianJES: Here as well:
I believe you can just do `params2 += bytes(12)`.

TheJulianJES: Here as well:
I believe you can just do `params2 += bytes(12)`.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +411,167 @@ def _update_attribute(self, attrid, value):
...
+        self, attributes: dict[str | int, Any], manufacturer: int | None = None
+    ) -> list:
+        """Write attributes to device with internal 'attributes' validation."""
+        sensor = bytearray.fromhex("00158d00019d1b98")
+        attrs = {}
+
+        for attr, value in attributes.items():
+            # implemented with help from https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/xiaomi.js
+
+            if attr == SENSOR_TEMP:
```
TheJulianJES: Technically, the `attributes` can either be the id or the name. Generally, it's the id, so this works, but it might be nice to accept both the id and name for `SENSOR` and `SENSOR_TEMP`.

If you were to implement the new-style `AttributeDefs`, you can use `AqaraThermostatSpecificCluster.AttributeDefs.sensor_temp.name` or `AqaraThermostatSpecificCluster.AttributeDefs.sensor_temp.id` to access name/id.

Or you can also use `find_attribute` to get the `ZCLAttributeDef` then. No need to check both id and name then. I think this should work then:
```python
        attrs: dict[str | int, Any] = {}
        for attr, value in attributes.items():
            attr_def = self.find_attribute(attr)
            if attr_def == AqaraThermostatSpecificCluster.AttributeDefs.sensor_temp:
                # do stuff
            else:
                attrs[attr] = value
        return await super().write_attributes(attrs, manufacturer)
```

(This might be streamlined in zigpy in the future.)

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +411,167 @@ def _update_attribute(self, attrid, value):
...
+
+    def _float_to_hex(self, f):
+        """Convert float to hex."""
+        return hex(struct.unpack("<I", struct.pack("<f", f))[0])
+
+    async def write_attributes(
+        self, attributes: dict[str | int, Any], manufacturer: int | None = None
+    ) -> list:
+        """Write attributes to device with internal 'attributes' validation."""
+        sensor = bytearray.fromhex("00158d00019d1b98")
```
TheJulianJES: I guess we don't need to call this every time here. Probably enough to just move it outside of the method.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +411,167 @@ def _update_attribute(self, attrid, value):
...
+
+                    attrs1 = {}
+                    attrs1[SENSOR_ATTR_NAME] = (
+                        self.aqaraHeader(0x12, params1, 0x04) + params1
+                    )
+                    attrs[SENSOR_ATTR_NAME] = (
+                        self.aqaraHeader(0x13, params2, 0x04) + params2
+                    )
+
+                    await super().write_attributes(attrs1, manufacturer)
```
TheJulianJES: So these need to be written in separate calls?

README.md
```diff
@@ -456,7 +456,7 @@ The tests use the [pytest](https://docs.pytest.org/en/latest/) framework.
 To get set up, you need install the test dependencies:
 
 ```bash
-pip install -r requirements_test_all.txt
+pip install -r requirements_test.txt
```
TheJulianJES: Thanks for noticing. Let's keep this in a separate PR though.
(Also, `script/setup` should also do this.)

TheJulianJES (commented):
Thanks for adding the tests. I think this looks pretty good already.

README.md
Antoninj: ok, I'll revert no problem

zhaquirks/xiaomi/aqara/thermostat_agl001.py
Antoninj: I honestly have no idea, I bootstrapped this PR from existing work from someone else and mostly focused on addding unit tests tbh

Antoninj:
@TheJulianJES thanks for the detailed code review, I've addressed most of the remarks.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
Antoninj: I hear you but I honestly don't have much time to spend on this PR atm and multiple people have been emailing me to track progress on this change since it's a highly requested feature for the quirk.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
MichelFR: They need to be in separate calls otherwise its starts to misbehave

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -402,6 +413,136 @@ def _update_attribute(self, attrid, value):
...
+            # implemented with help from https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/xiaomi.js
+            attr_def = self.find_attribute(attr)
+
+            if attr_def and attr_def.id == SENSOR_TEMP:
+                # set external sensor temp. this function expect value to be passed multiplied by 100
+                temperatureBuf = bytearray.fromhex(
+                    self._float_to_hex(round(float(value)))[2:]
+                )
+
+                params = SENSOR_ID
```
NJLangley, line 443: This seems to break the quirk for me. I can set the TRV into external sensor mode once and set the temp, but all subsequent calls fail and it seems that the max size of the constructed message is exceeded.

It's because the `SENSOR_ID` variable is not copied, so appending more bytes to the `params` variable is modifying `SENSOR_ID`, which means it's then bigger to start with the next time it's run. Changing the line to 
```
params = bytearray(SENSOR_ID)
```
seems to fix it for me. Might want the tests to check the length of the message args for two subsequent calls to verify this?

NJLangley:
I was having some issues with this, as it worked for the first call to set the mode to external sensor and then set the temp, but for all subsequent calls until the ZHA integrations was reloaded then failed. Digging through the stack trace for the error suggests the message exceeded the size of the window used for Zigbee comms. 

The original quirk from #2802 does work, and I've left a comment on the line that needs fixing. Looks like it's not copying the byte array before building the message from it, so it's them broken when it builds the next message

zhaquirks/xiaomi/aqara/thermostat_agl001.py
Elyasin: This is quite some time ago. I would like to help in this case.

I found that this PR is not up to date with the upstream.
In the upstream I see that the AttributeDefs are already defined.
So it is, I believe, straight forward to do this change.

Unfortunately, I don't know how I can collaborate on this PR.

@Antoninj - I believe you can do that quite easily by doing the following:

1. Sync the fork (dev branch) with your the upstream (on GitHub you can do that easily)
2. Pull the updated dev branch to your local machine (e.g. `git checkout dev` and then `git pull`)
3. Now you can rebase the branch `aqara_external_temperature_sensor` onto  dev branch (or merge dev branch into your branch)
4. You will have merge conflicts, which are easily resolved:
  - The test file is new and is not affected
  - The file `thermostat_agl001.py` is already in the new style and you should keep the code of your branch, except the changes of the attributes. These are already in the new style and you need to add those two lines I believe:
```
        sensor_temp: Final = ZCLAttributeDef(
            id=SENSOR_TEMP, type=t.uint32_t, is_manufacturer_specific=True
        )
        sensor_attr: Final = ZCLAttributeDef(
            id=SENSOR_ATTR, type=t.LVBytes, is_manufacturer_specific=True
        )
```

The tests run fine when I tried it.

However, there is the comment of NJLangley, which needs to be addressed. I am trying to look into that. As I am still quite "new" I try to learn und understand what is meant by the comment.

Elyasin: If you feel that you do not have time, I could replicate this PR and do the changes.
Let me know if that is ok for you.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
Elyasin: I tried the suggestion you mentioned and wrote a test for this to avoid regression.

I hope to revceive an answer from @Antoninj in regard to collaborate on this PR or if I re-create the PR with the changes.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
Elyasin: My guess is that two separate calls is necessary, otherwise it would not work correctly.
I conclude that there is nothing to do then.

Elyasin:
I created a new PR that continues Antonin's work with the review comments taken into account. I added two test cases and did little refactoring, which I believe helps read the code better.

https://github.com/zigpy/zha-device-handlers/pull/4463

PR #3977, opened 2025-03-24
Fix typing of Tuya `DPToAttributeMapping` classes

abmantis:
## Proposed change

Previously, the `DPToAttributeMapping` classes were duplicated, with the one in `zhaquirks.tuya.mcu` having an extra attribute. Both were stored in `_dp_to_attributes`, despite the typing using only the one in `zhaquirks.tuya`.

This change fixes that by making the class in the `zhaquirks.tuya.mcu` module inherit from the one in `zhaquirks.tuya`. The `__init__` method needs to be specified instead of using `dataclass` to keep backwards compatibility due to the order of the attributes being different now.

## Checklist

### DIFF
--- a/zhaquirks/tuya/mcu/__init__.py
+++ b/zhaquirks/tuya/mcu/__init__.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 from collections.abc import Callable
-import dataclasses
 import datetime
 from typing import Any
 
@@ -19,6 +18,7 @@ from zhaquirks.tuya import (
     TUYA_MCU_VERSION_RSP,
     TUYA_SET_DATA,
     TUYA_SET_TIME,
+    DPToAttributeMapping as DpToAttributeMappingBase,
     EnchantedDevice,  # noqa: F401
     NoManufacturerCluster,
     PowerOnState,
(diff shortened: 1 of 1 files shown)


abmantis:
@TheJulianJES as follow-up to https://github.com/zigpy/zha-device-handlers/pull/3967#issuecomment-2745418669 , I ended up keeping `dp_converter` in the `mcu` module class, since it is only used by that module and could be misleading to have it in the top module. I think this way it is clearer what exists and is used where, but let me know what you think!

abmantis:
Hey @TheJulianJES , I just noticed that you added `needs reviewer answer` label. Is this missing something?

TheJulianJES:
There was something else I wanted to try when I added that label, but I'll just merge this for now, since it's blocking your other PRs.

On another note, we might be able to just force kwargs usage for `DPToAttributeMapping` (except for the first argument(s)) by adding `*` between the args, so we can maybe consolidate both classes.

TheJulianJES (approved):
Thanks!

abmantis:
> On another note, we might be able to just force kwargs usage for `DPToAttributeMapping` (except for the first argument(s)) by adding `*` between the args, so we can maybe consolidate both classes.

Won't that break existing quirks that are not using kwargs?

TheJulianJES:
Yes, but I'm not sure if that's even done in custom quirks. If we do it in our code base, custom quirks will also do that, but if we've always used kwargs for the converter attributes, custom quirks are likely to be the same.

abmantis:
> Yes, but I'm not sure if that's even done in custom quirks. If we do it in our code base, custom quirks will also do that, but if we've always used kwargs for the converter attributes, custom quirks are likely to be the same.

We do it in our code base: https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/ts0601_rcbo.py#L389-L389

Issue #3188, opened 2024-06-04
[Device Support Request] SONOFF SNZB-03P


olikami:
### Problem description

- Motion is always stuck in detected
- Dim and bright is not reported

### Solution description

Make motion work properly, add the dim and bright values

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4742, maximum_buffer_size=82, maximum_incoming_transfer_size=255, server_mask=11264, maximum_outgoing_transfer_size=255, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0107",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0020",
        "0x0406",
        "0x0500",
        "0xfc57"
      ],
      "output_clusters": [
        "0x0003",
        "0x0019"
      ]
    }
  },
  "manufacturer": "eWeLink",
  "model": "SNZB-03P",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2024.5.5",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.12.2",
    "docker": true,
    "arch": "aarch64",
    "timezone": "Europe/Zurich",
    "os_name": "Linux",
    "os_version": "6.6.28-haos-raspi",
    "supervisor": "2024.05.1",
    "host_os": "Home Assistant OS 12.3",
    "docker_version": "25.0.5",
    "chassis": "embedded",
    "run_as_root": true
  },
  "custom_components": {
    "zha_toolkit": {
      "documentation": "https://github.com/mdeweerd/zha-toolkit",
[...]

PR #3979, opened 2025-03-25
Add Sonoff SNZB-03P motion sensor 

panther7:
## Proposed change
- add Sonoff SNZB-03P motion sensor (#3188)
- author of script: @BillyBlaze (modified by @panther7)

## Checklist

### DIFF
--- a/zhaquirks/sonoff/snzb03p.py
+++ b/zhaquirks/sonoff/snzb03p.py
@@ -1,0 +1,61 @@
+"""Sonoff SNZB-03P - Zigbee motion sensor."""
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import EntityPlatform, EntityType, NumberDeviceClass, QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+import zigpy.types as t
+from zigpy.zcl.clusters.measurement import OccupancySensing
+from zigpy.zcl.clusters.security import IasZone
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef, ZCLHeader
+
+
+class LastIlluminationState(t.enum8):
+    """Last measureed state of illumination enum."""
+
+    Dark = 0x00
+    Light = 0x01
+
+
+class SonoffIlluminationCluster(CustomCluster):
+    """Custom Sonoff illumination cluster."""
+
+    cluster_id = 0xFC11
+    manufacturer_id_override: t.uint16_t = ZCLHeader.NO_MANUFACTURER_ID
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        last_illumination_state = ZCLAttributeDef(
+            id=0x2001,
+            type=LastIlluminationState,
+            access="rp",
+            is_manufacturer_specific=True,
+        )
+
+
+(
+    QuirkBuilder("eWeLink", "SNZB-03P")
+    .replaces(SonoffIlluminationCluster)
+    .removes(IasZone.cluster_id)  #  IasZone cluster 0x0500: remove motion detection
+    .number(
+        OccupancySensing.AttributeDefs.ultrasonic_o_to_u_delay.name,
+        OccupancySensing.cluster_id,
+        min_value=5,
+        max_value=60,
+        step=1,
+        unit=UnitOfTime.SECONDS,
+        device_class=NumberDeviceClass.DURATION,
+        translation_key="motion_timeout",
+        fallback_name="Presence detection timeout",
+    )
+    .enum(
+        SonoffIlluminationCluster.AttributeDefs.last_illumination_state.name,
+        LastIlluminationState,
+        SonoffIlluminationCluster.cluster_id,
+        entity_platform=EntityPlatform.SENSOR,
+        entity_type=EntityType.STANDARD,
+        translation_key="last_illumination_state",
+        fallback_name="Last illumination state",
+    )
+    .add_to_registry()
+)


zhaquirks/sonoff/snzb03p.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @property
+    def _is_manuf_specific(self):
+        return False
+
+
+(
+    QuirkBuilder("eWeLink", "SNZB-03P")
+    .replaces(SonoffIlluminationCluster)
+    .removes(IasZone.cluster_id)  #  IasZone cluster 0x0500: remove motion detection
```
TheJulianJES, line 39: So the `IasZone` cluster doesn't do anything at all?
Can we use `.prevent_default_entity_creation(endpoint_id=1, cluster_id=IasZone.cluster_id)` here instead?
It's a new functionality that doesn't entirely remove the cluster for us, just disables the auto discovery for ZHA entities on that cluster.

zhaquirks/sonoff/snzb03p.py
```diff
@@ -0,0 +1,64 @@
...
+        last_illumination_state = ZCLAttributeDef(
+            id=0x2001,
+            type=LastIlluminationState,
+            access="rp",
+            is_manufacturer_specific=True,
+        )
+
+    @property
+    def _is_manuf_specific(self):
+        return False
```
TheJulianJES: I think we can use `manufacturer_id_override: t.uint16_t = foundation.ZCLHeader.NO_MANUFACTURER_ID` instead. See this PR as an example: https://github.com/zigpy/zha-device-handlers/pull/3910/files#diff-6c9212ca1af0b4604a20edb89247ceb6d30b010041e81fde812465a2f73a3ccbR24

zhaquirks/sonoff/snzb03p.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+(
+    QuirkBuilder("eWeLink", "SNZB-03P")
+    .replaces(SonoffIlluminationCluster)
+    .removes(IasZone.cluster_id)  #  IasZone cluster 0x0500: remove motion detection
+    .number(
+        OccupancySensing.AttributeDefs.ultrasonic_o_to_u_delay.name,
```
TheJulianJES, line 41: Ah, interesting. This is a default attribute. It's also used by the Hue motion sensors, though it's reported as an "unsupported attribute" if we read it during pairing, for some reason, so I've not added that as a config entity for now. We could add the entities via quirks v2 for both sensors, but we might be able to add this as a general entity in ZHA. I'll have to come back to this.

zhaquirks/sonoff/snzb03p.py
TheJulianJES: Z2M seems to specify a manufacturer code: https://github.com/Koenkk/zigbee-herdsman-converters/blob/216ba1a56f1b21fda7de7204839f0fb838df0b85/src/devices/sonoff.ts#L1162
Do we even need this at all? By default, zigpy will use the manufacturer code for the device if the attribute is marked as `is_manufacturer_specific` (it is) and if the cluster is in the manufacturer specific range (it is).

I guess this is only really needed if we explicitly poll/read the attribute ourselves though (through the "Manage Zigbee device -> Clusters" debug menu).

panther7:
Thanks for huge feedback.

This PR is only request with [originial code](https://github.com/zigpy/zha-device-handlers/issues/3188#issuecomment-2585694831) from @BillyBlaze and with (little) mods.

I have zero experience with Python and device handlers, bit i will try it.

zhaquirks/sonoff/snzb03p.py
panther7: Can not be used.

Replace with `manufacturer_id_override: t.uint16_t = foundation.ZCLHeader.NO_MANUFACTURER_ID` is broken:

- eWeLink SNZB-03P is always detected
- Last illumination state is Unavailable




With `_is_manuf_specific(self): return False` is right:

zhaquirks/sonoff/snzb03p.py
panther7: If removed, then same issue like https://github.com/zigpy/zha-device-handlers/pull/3979#discussion_r2013693307

zhaquirks/sonoff/snzb03p.py
panther7: Yes, cluster is always "unavailable".

panther7:
@TheJulianJES What next now? If i use some from your suggest, some functions wrong.

TheJulianJES:
Leave it like this for now. I'll come back to it.

zhaquirks/sonoff/snzb03p.py
TheJulianJES: > eWeLink SNZB-03P is always detected
> Last illumination state is Unavailable

Adding `manufacturer_id_override` on the `SonoffIlluminationCluster` won't influence the motion entity/attribute, since it's on another cluster.

And last illumination sensor being unavailable also can't be a result of adding that. Did you see any errors in the logs when adding `manufacturer_id_override` to that cluster, similar to how it's done in the linked PR above?

Maybe you forgot to import `from zigpy.zcl import foundation` at the top?
Please try adding `manufacturer_id_override: t.uint16_t = foundation.ZCLHeader.NO_MANUFACTURER_ID` again, making sure the import exists.

zhaquirks/sonoff/snzb03p.py
TheJulianJES: I meant potentially removing the whole `.number()` method for `ultrasonic_o_to_u_delay`.
It won't cause other entities to break if removed like that.

Perhaps we can keep this though. I'll have to check back with how we want to implement config support for this entity in ZHA.

TheJulianJES:
Just in case you haven't seen it, I've answered in this comment thread: https://github.com/zigpy/zha-device-handlers/pull/3979#discussion_r2067554011

zhaquirks/sonoff/snzb03p.py
panther7: Done

zhaquirks/sonoff/snzb03p.py
panther7: If I remove `ultrasonic_o_to_u_delay`, then configuration for `Presence detection timeout` will stop works.

panther7:
Any news? @TheJulianJES 

This PR waiting few months,
[issue](https://github.com/zigpy/zha-device-handlers/issues/3188) is older than 1 year,
we can move it?

curiousYL:
I there anything to do to help here?

cur1ou5soul:
Anything to do to help here?

tjl10:
@panther7 there's a label on this saying "waiting for changes" - but I think you made the requested changes on 10th June. Can you remove the label so that @TheJulianJES notices and can re-review the code?

panther7:
I drop all [ZHA](https://www.home-assistant.io/integrations/zha/) intergation and moved all to [MQTT](https://www.home-assistant.io/integrations/mqtt/) because i can't wait ~~one year~~ 8 months for approve PR.

Btw, MQTT have better support with more devices.

tjl10:
> I drop all [ZHA](https://www.home-assistant.io/integrations/zha/) intergation and moved all to [MQTT](https://www.home-assistant.io/integrations/mqtt/) because i can't wait ~one year~ 7 months for approve PR.
> 
> Btw, MQTT have better support with more devices.

Fair. I'm happily using the custom quirk but every time HA restarts the log warns me I should submit it to the project, so it would be nice to get this over the line.

I'm happy to take over the PR (create a new one) if any further changes are required and you don't want to work on it anymore. It all works though so let's see what the outcome of the review is.

panther7:
Replaced #4628

^ @TheJulianJES

PR #3983, opened 2025-03-26
Change min value, max value, step and try to implement reporting for ThirdReality Watering kit

mguaylam:
## Proposed change
Fix the values of the watering time and try to implement reporting.

## Additional information
Fixes #3926.

## Checklist

### DIFF
--- a/zhaquirks/thirdreality/watering_kit.py
+++ b/zhaquirks/thirdreality/watering_kit.py
@@ -3,12 +3,12 @@
 from typing import Final
 
 from zigpy.quirks import CustomCluster
-from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2 import QuirkBuilder, ReportingConfig
 from zigpy.quirks.v2.homeassistant import UnitOfTime
 from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
 import zigpy.types as t
+from zigpy.zcl.clusters.security import IasZone
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
-
 
 class ThirdRealityWateringKitCluster(CustomCluster):
     """Third Reality's watering kit private cluster."""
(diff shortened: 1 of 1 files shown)


mguaylam:
Hey there @hwzolin, once you fix the reporting for `0xFFF2` in the firmware, can you let us know so we can merge this?
@TheJulianJES should we wait for ThirdReality to implement it? The number `min` and `increment` needed to be fixed anyway.

TheJulianJES:
We should hold back with the reporting until it's implemented in the firmware.
I guess `min` and `increment` can be changed though.

`increment` only really influences this if the number entity mode is in `SLIDER` mode, which I believe most quirks v2 entities are currently due to https://github.com/zigpy/zha/issues/364.
If a number entity is in `BOX` mode, I don't think that `increment` does anything. A service/action call can always be used to set a number manually otherwise, even if in `SLIDER` mode.

hwzolin:
> We should hold back with the reporting until it's implemented in the firmware. I guess `min` and `increment` can be changed though.
> 
> `increment` only really influences this if the number entity mode is in `SLIDER` mode, which I believe most quirks v2 entities are currently due to [zigpy/zha#364](https://github.com/zigpy/zha/issues/364). If a number entity is in `BOX` mode, I don't think that `increment` does anything. A service/action call can always be used to set a number manually otherwise, even if in `SLIDER` mode.

@mguaylam Ok, I'll let you know as soon as we develop a new firmware that supports the feature.

mguaylam:
Just received a new firmware that seem to report changes from the devices.
But I guess it's not configurable, just hard coded @hwzolin ?

image

For the numbers, are service calls able to ignore the increment value and just send arbitrary numbers?
In either box and slider @TheJulianJES, I seem to have an increment value taken into account, I can't input anything else which is what we want no?

I'm wondering in which case we should use box and slider, I guess for now those choices are ok because they look cute together :
image

@hwzolin when should the water alert (IAS) turn on/off? My device has no water but does not display an alert. Is that normal?

@TheJulianJES I'm wondering is we still should do reporting the way it's been implemented here.

Hedda:
@3reality-support any updates on this?

3reality-support:
Hello, I have two questions. Firstly, why do we need to modify the min value and step, and what are the current issues? Secondly, why do we need to implement config report? This attribute is not like temperature, it requires a minimum interval and a maximum interval. The water duration and water interval are fixed and do not require the cooperation of config

TheJulianJES (changes requested):
Since reporting doesn't need to be configured by us (and is directly set in firmware), we should at least remove that from the PR.

But I believe the `min`/`step` size changes are unnecessary as well. If the device supports a water duration of 1 second, which might be applicable for some use-cases, we should support that.

One change that might be useful to implement would be setting `mode="box"`, but that should be done in a new PR now.

TheJulianJES:
Closing this PR as per the above comment.  There don't seem to be any issues with the current quirk.

PR #3986, opened 2025-03-26
Decompose `attribute_name` tuple to new Tuya DP mapping list format

abmantis:
## Proposed change

~~Requires #3977.~~

As previously discussed in #3643, this converts the old tuple in DPToAttributeMapping to the new list of DPToAttributeMapping.

For this, the `dp_converter` also needs to be wrapped and a new attributes_to_dp_converters attribute was introduced in TuyaMCUCluster.

In a follow-up PR I'll add support to the TuyaQuirkBuilder for the new `dp_converter`. Eventually, after migrating all existing clusters and quirks to the new format we could change the deprecation info logs in DPToAttributeMapping to warnings, so we can clean up the code after some time.

## Checklist


zhaquirks/tuya/tuya_trv.py
```diff
@@ -696,7 +696,8 @@ async def write_attributes(
                     TuyaPresetMode.Heat: Thermostat.SystemMode.Heat,
                     TuyaPresetMode.Off: Thermostat.SystemMode.Off,
                 }[x],
-                dp_converter=lambda x: {
+                dp_converter=lambda x,
+                _: {  # TODO: add a new dp_converter argument to tuya_dp_multi and remove from this
```
abmantis: To keep this PR small, I plan to add support for this in the builder in a separate PR. For now, it needs the extra argument since this DP has 2 attributes mapped.
If this is not fine, I can do the Builder changes in this PR.

abmantis: The changes were smaller than I anticipated, so updated this one.

zhaquirks/tuya/__init__.py
```diff
@@ -1456,6 +1456,46 @@ class DPToAttributeMapping:
...
+    ):
+        """Init DPToAttributeMapping."""
+        self.ep_attribute = ep_attribute
+        self.attribute_name = attribute_name
+        self.converter = converter
+        self.endpoint_id = endpoint_id
+
+        if not isinstance(attribute_name, str):
+            _LOGGER.info(
+                "Using tuple attribute_name is deprecated, please multiple "
```
TheJulianJES: The verb, likely "use", is missing here.

zhaquirks/tuya/__init__.py
```diff
@@ -1456,6 +1456,46 @@ class DPToAttributeMapping:
...
+                "DPToAttributeMapping instances instead. %s",
+                attribute_name,
+            )
```
TheJulianJES: Also, we may want to expand this message to include a message that (custom) ZHA quirks are affected and need to be updated. Maybe something like the following?

"Using a tuple for `attribute_name` is deprecated, please update ZHA quirks to use multiple `DPToAttributeMapping` instances instead. Affected tuple: %s"

Also, we may want to bump this to `warning`? I think `WARNING` is the default log level on new HA installs?
And we should also make sure that we don't use this in our codebase anymore.

EDIT: Ah, if we still use this in our own codebase, maybe just `debug` instead, for now..
Some old installs still have `INFO` as the default log level and I don't think we should log our own issues.

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -49,6 +52,12 @@ def __init__(
         """Init method for compatibility with previous quirks using positional arguments."""
         super().__init__(ep_attribute, attribute_name, converter, endpoint_id)
         self.dp_converter = dp_converter
+        if dp_converter:
+            _LOGGER.info(
+                "DPToAttributeMapping with dp_converter is deprecated, use TuyaQuirksBuilder "
+                "(or TuyaMCUCluster.attributes_to_dp_converters) instead. attribute_name: %s",
+                attribute_name,
+            )
```
TheJulianJES: It's `TuyaQuirkBuilder`.

TheJulianJES (commented):
There were/are still some issues I have with this PR, which isn't helped by the fact that I have to re-read the old Tuya code over and over again every time I come back to this PR. That's not your fault though 😄 

I'll just leave my couple of existing review comments here, for now. Still have to come back to this.
I've not forgotten about it at least. Sorry for the long wait on this one.

zhaquirks/tuya/mcu/__init__.py
abmantis: also changed the log to debug, to match the suggestion in https://github.com/zigpy/zha-device-handlers/pull/3986#discussion_r2067603895

abmantis:
The test failure seems to be some hickup in the CI

abmantis:
Hey @TheJulianJES did you had the chance to get back to this?

zhaquirks/tuya/__init__.py
```diff
@@ -1496,6 +1496,46 @@ class DPToAttributeMapping:
...
+        endpoint_id: int | None = None,
+    ):
+        """Init DPToAttributeMapping."""
+        self.ep_attribute = ep_attribute
+        self.attribute_name = attribute_name
+        self.converter = converter
+        self.endpoint_id = endpoint_id
+
+        if not isinstance(attribute_name, str):
+            _LOGGER.warning(
```
TheJulianJES: Until our own quirks are updated, let's only use the debug log level (`ts0601_rcbo.py` uses this a lot):
```suggestion
            _LOGGER.debug(
```

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -219,20 +244,19 @@ def from_cluster_data(self, data: TuyaClusterData) -> list[TuyaCommand]:
...
                 args = []
-                if isinstance(mapping.attribute_name, tuple):
+                for dp_attr in self._dp_to_attributes[dp]:
+                    if dp_attr.attribute_name == data.cluster_attr:
+                        args.append(val)
+                        continue
                     endpoint = self.endpoint
-                    if mapping.endpoint_id:
+                    if dp_attr.endpoint_id:
                         endpoint = endpoint.device.endpoints[mapping.endpoint_id]
```
TheJulianJES: Should the second line use `dp_attr.endpoint_id` for getting the endpoint? 
```suggestion
                    if dp_attr.endpoint_id:
                        endpoint = endpoint.device.endpoints[dp_attr.endpoint_id]
```

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -195,6 +205,21 @@ class ServerCommandDefs(TuyaNewManufCluster.ServerCommandDefs):
...
+        else:
+            # convert from legacy DP2AttributeMapping with attribute_name tuple to new
+            # DP2AttributeMapping with single attribute_name
+            self._attributes_to_dp_converters = {}
+            for dp, mappings in self.dp_to_attribute.items():
+                if not isinstance(mappings, list):
+                    mappings = [mappings]
+                for dp_mapping in mappings:
+                    if hasattr(dp_mapping, "dp_converter") and dp_mapping.dp_converter:
+                        self._attributes_to_dp_converters[dp] = dp_mapping.dp_converter
```
TheJulianJES, line 234: Can we add a comment here that the `hasattr` check is needed, as we could also have a "non-MCU" `DPToAttributeMapping` without the attribute being present? Something like this maybe:

```suggestion
                    # DPToAttributeMapping from the base Tuya module doesn't have `dp_converter`
                    # only the MCU DPToAttributeMapping has dp_converter, so check hasattr before
                    if hasattr(dp_mapping, "dp_converter") and dp_mapping.dp_converter:
                        self._attributes_to_dp_converters[dp] = dp_mapping.dp_converter
```

(This whole double Tuya + Tuya MCU `DPToAttributeMapping` is really annoying...)

zhaquirks/tuya/mcu/__init__.py
TheJulianJES: Also note this line is not covered by existing tests, unfortunately.

zhaquirks/tuya/__init__.py
```diff
@@ -1579,8 +1619,10 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
         self._dp_to_attributes: dict[int, list[DPToAttributeMapping]] = {
-            dp: attr if isinstance(attr, list) else [attr]
-            for dp, attr in self.dp_to_attribute.items()
+            dp: [x for mapping in mappings for x in mapping.decompose_attributes()]
+            if isinstance(mappings, list)
+            else mappings.decompose_attributes()
+            for dp, mappings in self.dp_to_attribute.items()
```
TheJulianJES: This is not very easy to read. Can we clean this up a bit? Or add some comments?

Would something like this be better, or not? 😄 
```suggestion
        # Normalize dp_to_attribute: map each DP to a flat list of single attr mappings
        # This decomposes old tuple-based mappings
        self._dp_to_attributes: dict[int, list[DPToAttributeMapping]] = {
            dp: [
                decomposed
                # Normalize to list (dp_to_attribute allows single mapping or list)
                for mapping in (mappings if isinstance(mappings, list) else [mappings])
                # Decompose tuple-based mappings into individual single-attribute mappings
                for decomposed in mapping.decompose_attributes()
            ]
            for dp, mappings in self.dp_to_attribute.items()
        }
```

Or something like this to be even more explicit?
```suggestion
        # Normalize dp_to_attribute: map each DP to a flat list of single attr mappings
        # This decomposes old tuple-based mappings
        self._dp_to_attributes: dict[int, list[DPToAttributeMapping]] = {}
        for dp, mappings in self.dp_to_attribute.items():
            # Normalize to list (dp_to_attribute allows single mapping or list)
            if not isinstance(mappings, list):
                mappings = [mappings]
            # Flatten: decompose any tuple-based mappings into individual mappings
            self._dp_to_attributes[dp] = [
                decomposed
                for mapping in mappings
                for decomposed in mapping.decompose_attributes()
            ]
```

TheJulianJES (commented):
The backwards compatibility is really annoying to work with, but I think it's necessary for now.. Some parts of the Tuya code could probably use some more comments/docstrings.

There's one question or possible issue (`dp_attr.endpoint_id` not being used). Otherwise, I think this is fine..?

TheJulianJES:
This will cause a very small conflict, but should be easy to fix then:
- https://github.com/zigpy/zha-device-handlers/pull/4649/

zhaquirks/tuya/__init__.py
abmantis: Yeah, this was too confusing indeed. I've switched to your explicit for-loop since it is easier to read here.

zhaquirks/tuya/mcu/__init__.py
abmantis: Added a test too.

TheJulianJES (approved):
I think this looks good (enough 😄). Thanks!!
I'll merge this for HA 2026.3.0. We need to get some patch releases out before that due to the big zigpy migration in 2026.2.x.

I can address a couple of Copilots about typing and test coverage in a future PR.

abmantis:
Thanks @TheJulianJES !

PR #3989, opened 2025-03-30
Add temperature and humidity offset to Sonoff SNZB-02D

fgsch:
## Proposed change

Add temperature and humidity calibration to Sonoff SNZB-02D.

## Additional information

N/A.

## Checklist

### DIFF
--- a/zhaquirks/sonoff/snzb02d.py
+++ b/zhaquirks/sonoff/snzb02d.py
@@ -121,7 +121,7 @@ class CustomSonoffCluster(CustomCluster):
         min_value=-50,
         max_value=50,
         step=0.1,
-        unit=TemperatureUnit,
+        unit=UnitOfTemperature.CELSIUS,
         multiplier=0.01,
         translation_key="temperature_calibration",
         fallback_name="Temperature calibration",
--- a/zhaquirks/sonoff/snzb02d.py
+++ b/zhaquirks/sonoff/snzb02d.py
@@ -1,7 +1,7 @@
 """Sonoff SNZB-02D - Zigbee LCD smart temperature humidity sensor."""
 
 from zigpy.quirks import CustomCluster
-from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2 import NumberDeviceClass, QuirkBuilder
 from zigpy.quirks.v2.homeassistant import PERCENTAGE, UnitOfTemperature
 import zigpy.types as t
 from zigpy.zcl import foundation
@@ -70,6 +70,7 @@ class CustomSonoffCluster(CustomCluster):
         min_value=-10,
         max_value=60,
         step=0.1,
+        device_class=NumberDeviceClass.TEMPERATURE,
         unit=UnitOfTemperature.CELSIUS,
         multiplier=0.01,
         translation_key="comfort_temperature_min",
@@ -81,6 +82,7 @@ class CustomSonoffCluster(CustomCluster):
         min_value=-10,
         max_value=60,
         step=0.1,
+        device_class=NumberDeviceClass.TEMPERATURE,
         unit=UnitOfTemperature.CELSIUS,
         multiplier=0.01,
         translation_key="comfort_temperature_max",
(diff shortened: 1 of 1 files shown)


zhaquirks/sonoff/snzb02d.py
```diff
@@ -105,5 +115,27 @@ class AttributeDefs(BaseAttributeDefs):
         translation_key="display_unit",
         fallback_name="Display unit",
     )
+    .number(
+        CustomSonoffCluster.AttributeDefs.temperature_calibration.name,
+        CustomSonoffCluster.cluster_id,
+        min_value=-50,
+        max_value=50,
+        step=0.1,
+        unit=TemperatureUnit,
```
TheJulianJES: You should not use the enum `TemperatureUnit` for the `unit` — it's expecting a `str`.
The offset is likely in `°C`, so use `UnitOfTemperature.CELSIUS` instead.

zhaquirks/sonoff/snzb02d.py
```diff
@@ -105,5 +115,27 @@ class AttributeDefs(BaseAttributeDefs):
...
+        translation_key="temperature_calibration",
+        fallback_name="Temperature calibration",
+    )
+    .number(
+        CustomSonoffCluster.AttributeDefs.humidity_calibration.name,
+        CustomSonoffCluster.cluster_id,
+        min_value=-50,
+        max_value=50,
+        step=0.1,
+        unit=PERCENTAGE,
```
TheJulianJES, line 141: Hmm, technically, the unit for the calibration won't be a percentage, but percentage points.
Don't really think HA uses `pp` as a unit for entities like this though. I do see at least one HA entity using "%", so this is likely *fine*.

TheJulianJES (commented):
Also, any issues with this PR still, since it's still in draft? Seems like it should work for the most part.

zhaquirks/sonoff/snzb02d.py
fgsch: Updated.

fgsch:
> Also, any issues with this PR still, since it's still in draft? Seems like it should work for the most part.

Nop, sorry, I just been busy with life. I will try to wrap up this PR this week.

fgsch:
PTAL.

zhaquirks/sonoff/snzb02d.py
```diff
@@ -105,5 +115,27 @@ class AttributeDefs(BaseAttributeDefs):
...
+        multiplier=0.01,
+        translation_key="humidity_calibration",
+        fallback_name="Humidity calibration",
+    )
```
TheJulianJES: There's also `NumberDeviceClass.TEMPERATURE` and `NumberDeviceClass.HUMIDITY` for device classes.
I believe using the `TEMPERATURE` device class would cause the temperature entity to be converted to °F if using that, instead of °C.

I think we should set these device classes for calibration entities too, though we apparently don't set device classes for all existing ZHA number entities (and Tuya v2 quirks config entities?).

TheJulianJES (commented):
Hopefully the last thing, I think we're good otherwise.
I can also add it tomorrow or so (hopefully before the beta release) if you don't get around to it.

zhaquirks/sonoff/snzb02d.py
TheJulianJES: Added with [`ef69b21` (#3989)](https://github.com/zigpy/zha-device-handlers/pull/3989/commits/ef69b215b8c12170cde35cd757795fdc25414503), for all temperature/humidity entities.

TheJulianJES:
I've also changed the temperature/humidity "calibration" wording to "offset", as that's what we use in a similar Telink quirk.
Commit for that change: [`7269e84` (#3989)](https://github.com/zigpy/zha-device-handlers/pull/3989/commits/7269e8446b8905cf0624aacf770cf218e96b2973)

As I've also renamed the attribute name (upon which the unique id is based on by default), you'll likely get new entities if you were using the previous version of this quirk as a custom quirk.

TheJulianJES (approved):
I think this should be good to go now. Thanks for the PR!

fgsch:
Sorry, I missed your comments and suggestions.

Thank you for doing those changes and merging the PR.

PR #3999, opened 2025-04-02
Add Yandex devices: YNDX-00530, 00531, 00532, 00534, 00535, 00537, 00538, 00591, 00592

Nullcaller:
## Proposed change

Add support for manufacturer-specific features of the following Yandex devices:

- YNDX-00530 Dimmer
- YNDX-00531 Single-gang Wired Switch
- YNDX-00532 Double-gang Wired Switch
- YNDX-00531 Single-gang Wireless Switch
- YNDX-00532 Double-gang Wireless Switch
- YNDX-00537 Single Relay
- YNDX-00538 Double Relay
- YNDX-00591 Curtain Motor
- YNDX-00592 Curtain Motor

(YNDX-00534 and YNDX-00535 wireless switch quirks are only really there for automation triggers' sake)

## Additional information

I've used code from Koenkk/zigbee-herdsman-converters#8345 and Koenkk/zigbee-herdsman-converters#9017 as reference for cluster attributes and commands.

Known issues (fixed):
- ~~Dimmer is pretty much untested. I'll test it when I have time.~~
- ~~Some of the endpoints added in replacement signatures are 'fake' and only work if the device is set into a 'decoupled' mode, if the endpoints are used in HA in any way when the device is not in the 'decoupled' mode, HA hangs until the requests time out.~~
- ~~I haven't been able to find the manufacturer code for the curtain motor as of yet~~

Known issues (wontfix):
- ~~The led_indicator attribute is weird, I don't think it's working as it should on all devices. (Seems to be Yandex's problem)~~
- Various buggy behavior that has to do with adding new devices. Probably Yandex's fault, because device adding pretty much always works flawlessly if you only start scanning for new devices in HA like 5 seconds after initiating device reset. And in the rare cases it doesn't, power cycling both the device and HA always helps, for some reason.
  - When adding devices, they get stuck in "Configuring" stage for a long while, but eventually get added (a restart of Home Assistant helps)
  - If `SKIP_CONFIGURATION: True` is added to replacement signature, devices don't get stuck in "Configuring" stage, but attribute reporting doesn't work
  - After adding double-gang switch or double relay, attribute reporting may not work for the second endpoint, which somehow gets fixed after pressing 'Reconfigure' on device page in HA

## Checklist

### DIFF
--- a/zhaquirks/yandex/__init__.py
+++ b/zhaquirks/yandex/__init__.py
@@ -1,0 +1,190 @@
+"""Common code for Yandex devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+import zigpy.types as t
+from zigpy.zcl.foundation import (
+    BaseAttributeDefs,
+    BaseCommandDefs,
+    ZCLAttributeDef,
+    ZCLCommandDef,
+)
+
+### CONSTANTS ###
+
+
+YANDEX = "Yandex"
+YANDEX_CLUSTER_ID = 0xFC03
+YANDEX_MANUFACTURER_CODE = 0x140A
+
+
+### TYPES ###
+
+
+class YandexType_SwitchMode(t.enum8):
+    """Wired switches: gang mode."""
+
+    Control_Relay = 0x00
+    Up_Decoupled = 0x01
+    Decoupled = 0x02
+    Down_Decoupled = 0x03
+
+
+class YandexType_SwitchType(t.enum8):
+    """Relays: connected switch type."""
+
+    Rocker = 0x00
+    Button = 0x01
+    Decoupled = 0x02
+
+
+class YandexType_PowerType(t.enum8):
+    """Wired devices: power level."""
+
+    High = 0x00
+    Medium = 0x01
+    Low = 0x02
+
+
+class YandexType_LedIndicator(t.basic.enum8):
+    """Enable/disable LED indicator."""
+
+    Enabled = 0
+    Disabled = 1
+
+
+class YandexType_Interlock(t.basic.enum8):
+    """Double relay only: enable/disable interlock mode."""
+
+    Disabled = 0
+    Enabled = 1
+
+
+class YandexType_ButtonMode(t.enum8):
+    """Dimmer only: button presses interpretation."""
+
+    General = 0x00
+    Alternative = 0x01
+
+
+### ATTRIBUTE DEFINITIONS ###
+
+
+YANDEX_ATTRIBUTE_SWITCH_MODE = ZCLAttributeDef(
+    id=0x0001,
+    name="switch_mode",
+    type=YandexType_SwitchMode,
+    access="rw",
+)
+YANDEX_ATTRIBUTE_SWITCH_TYPE = ZCLAttributeDef(
+    id=0x0002,
+    name="switch_type",
+    type=YandexType_SwitchType,
+    access="rw",
+)
+YANDEX_ATTRIBUTE_POWER_TYPE = ZCLAttributeDef(
+    id=0x0003,
+    name="power_type",
+    type=YandexType_PowerType,
+    access="rw",
+)
+YANDEX_ATTRIBUTE_LED_INDICATOR = ZCLAttributeDef(
+    id=0x0005,
+    name="led_indicator",
+    type=YandexType_LedIndicator,
+    access="rw",
+)
+YANDEX_ATTRIBUTE_INTERLOCK = ZCLAttributeDef(
+    id=0x0007,
+    name="interlock",
+    type=YandexType_Interlock,
+    access="rw",
+)
+YANDEX_ATTRIBUTE_BUTTON_MODE = ZCLAttributeDef(
+    id=0x0008,
+    name="button_mode",
+    type=YandexType_ButtonMode,
+    access="rw",
+)
+
+
+### COMMAND DEFINITIONS ###
+
+
+YANDEX_COMMAND_SWITCH_MODE = ZCLCommandDef(
+    id=0x01,
+    name="switch_mode",
+    schema={"value": YandexType_SwitchMode},
+    direction=False,
+    is_manufacturer_specific=True,
+)
+YANDEX_COMMAND_SWITCH_TYPE = ZCLCommandDef(
+    id=0x02,
+    name="switch_type",
+    schema={"value": YandexType_SwitchType},
+    direction=False,
+    is_manufacturer_specific=True,
+)
+YANDEX_COMMAND_POWER_TYPE = ZCLCommandDef(
+    id=0x03,
+    name="power_type",
+    schema={"value": YandexType_PowerType},
+    direction=False,
+    is_manufacturer_specific=True,
+)
+YANDEX_COMMAND_LED_INDICATOR = ZCLCommandDef(
+    id=0x05,
+    name="led_indicator",
+    schema={"value": YandexType_LedIndicator},
+    direction=False,
+    is_manufacturer_specific=True,
+)
+YANDEX_COMMAND_INTERLOCK = ZCLCommandDef(
+    id=0x07,
+    name="interlock",
+    schema={"value": YandexType_Interlock},
+    direction=False,
+    is_manufacturer_specific=True,
+)
+YANDEX_COMMAND_BUTTON_MODE = ZCLCommandDef(
+    id=0x08,
+    name="button_mode",
+    schema={"value": YandexType_ButtonMode},
+    direction=False,
+    is_manufacturer_specific=True,
+)
+
+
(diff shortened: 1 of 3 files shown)


Nullcaller:
I think the biggest issue with this quirk at this point is the ability to toggle the OnOff clusters in 'fake' endpoints (which only become 'real' in decoupled mode) from HA. **If someone can point me in the general direction of a solution for this problem, that'd be greatly appreciated.** Some of my inklings of a thought of an idea:

- Is there a way to make endpoints/clusters/ha entities read-only?
- Is there a way to make replacement signatures dynamic, i.e. dependent on an attribute value?

Nullcaller:
A little bit of an update: I fixed the 'fake' OnOff clusters. The issue was, on Yandex devices in decoupled mode, the event prosucers for decoupled mode switch presses are actually _output_ OnOff clusters, not input. I discovered this by looking at how wireless switches are paired with HA, i.e. their resultant signature. I'll produce a commit with fixed signatures in 3-7 business days. Then I'll try to convert it all to proper automation triggers, and to a v2 quirk if I find sufficient documentation on the matter.

Nullcaller:
As this quirk is quite basic, I don't really know whether it needs any tests. There is nothing functionally new to test here. It's all just wrapper classes, attributes and commands. I *have* tested it with pretty much all devices, though. It does work.

Nullcaller:
For anyone wanting to use the quirk before it is merged, I've compiled the quirk into a single file and made a gist: https://gist.github.com/Nullcaller/746f3a44ff694043c1141a84e66c4483

Yevgenium:
Could you please fix the quirk for YNDX-00591? Currently, it only displays the states "opening" or "closing." The rest of the time, it displays "unknown."

```{
  "node_descriptor": {
    "logical_type": 1,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4911,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 255,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 255,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0202",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0102",
        "0x0b05"
      ],
      "output_clusters": [
        "0x0003",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "Yandex",
  "model": "YNDX-00591",
  "class": "yndx.YandexCurtainMotor"
}

**Comment by Nullcaller** [comment]
@Yevgenium, yup, will do when I have the time. What I think is happening, is since I'm replacing the curtain motor cluster with a custom cluster, the attribute reporting for it does not get set up when you add the device with the quirk enabled.

**In the meanwhile, if you re-add the device with the quirk disabled, and then re-enable the quirk, it should work fine.**

I've also been trying to just generally improve support for YNDX-00591, it's supposed to have attributes for reading/setting speed, opening and closing limits and other stuff in the same cluster, but so far all I've been able to do with those attributes is get my motor to think it's closed when it's open and vice-versa, and somehow it doesn't want to go back to the proper way ¯\\\_(ツ)\_/¯

So, in short, I'm working on it

**Comment by Nullcaller** [comment]
I think I've fixed it. I've also updated the gist: https://gist.github.com/Nullcaller/746f3a44ff694043c1141a84e66c4483

Can everyone having issues test the fix, please?

**Comment by Nullcaller** [comment]
I think I'm right about what the problem is, but attempting to understand how to manually setup attribute reporting has been...

Let's just say, this is the reason I don't like Python. Or modern very-high-level programming paradigms, for that matter.

"Somewhere in the nebulous space that is code, which is neither here nor there because it is everywhere, there's a reflector which references a mixin created via an annotation born out of thin air, that changes how all things shall behave. To stumble upon it is great fortune, though not to do so is also fine. At the end of the day, all is governed by _The Vibe_"

If I just use a bog-standard, library window covering cluster, the code for which hints at no attribute reporting setup whatsoever, attribute reporting works. If I make a custom cluster, however, then it magically ceases to work.

Magic giveth, magic taketh away.

_But!_ If I then also subclass my custom cluster from window covering cluster _and_ subclass the attribute and command definition classess off of the window covering cluster's attribute and command definition classess respectively, attribute reporting works once more. Hooray, I guess.

_Bububut!_ If I then want to remove some of the attrubutes by setting them to `None` in _my (goddamn!)_ subclass, because the device doesn't actually work with those attributes, meaning HA will never be able to glean anything meaningful from them, let alone write new values for those attributes, then attribute reporting, once again, ceases to work.

Magic giveth, magic taketh away.

Well, lots of pointless, useless, no good attributes it is, I guess.

**Comment by Nullcaller** [comment]
Oh, and by the way, I added YNDX-00592 to the matcher thingamajig, should work with the quirk now, if anybody's wondering. I'll add the thermostat when I have time.

File: zhaquirks/yandex/yndx_0059x.py [hunk 1]
```diff
@@ -0,0 +1,149 @@
...
+
+class YandexWindowCovering(CustomCluster, WindowCovering):
+    """Yandex-flavor window covering cluster."""
+
+    manufacturer_id_override = YANDEX_MANUFACTURER_CODE_2
+
+    class AttributeDefs(WindowCovering.AttributeDefs):
+        """Attribute definitions."""
+
+        window_covering_type: Final = ZCLAttributeDef(
```
puddly, line 54: Why is `window_covering_type` redefined? If the Yandex window covering cluster is the same as the normal ZCL `WindowCovering` cluster but with a few extra attributes, it should be enough to just add those exact attributes.

`YANDEX_ATTRIBUTE_UNK_FFFD` is the standard `ZCL_REPORTING_STATUS_ATTR`, part of the spec.

Nullcaller:
And if anybody's having the same issue as my stupid ass, where HA considers curtains closed when they're open and vice versa, there is an attribute called `window_covering_mode` in the wondow covering cluster, set it to `0`. If the first bit of it is set to `1`, it reverses the motor direction, and unlike most attributes on the Yandex curtain motor, this one you can actually write to. And somehow even facory resetting the device through the basic cluster does not reset this attribute. Or that's somehow just the default value, which is even more confusing.

zhaquirks/yandex/yndx_0059x.py
```diff
@@ -0,0 +1,149 @@
...
+            id=0x02, schema={}, direction=Direction.Client_to_Server
+        )
+        go_to_lift_percentage: Final = ZCLCommandDef(
+            id=0x05,
+            schema={"percentage_lift_value": t.uint8_t},
+            direction=Direction.Client_to_Server,
+        )
+
+
+class YandexCurtainMotor(CustomDevice):
```
puddly, line 91: Inheriting from `CustomDevice` is the quirks "v1" API and something that we're migrating away from, in favor of the simpler and more powerful quirks V2 API. For this device, you can do something like:

- `.replaces(YandexWindowCovering, endpoint_id=1)` to replace the ZCL standard `WindowCovering` with a subclass.
- `.write_attr_button(` to create buttons that write attributes
- `.number(` to create numerical input entities

And so on. You can see [an example](https://github.com/zigpy/zha-device-handlers/blob/20a07776b156ad80d0bf2f09f983db1703fbe403/zhaquirks/thirdreality/garage_door_tilt_sensor.py#L51) here.

zhaquirks/yandex/__init__.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Slow = 6
+    Normal = 9
+    Fast = 12
+
+
+### ATTRIBUTE DEFINITIONS ###
+
+
+YANDEX_ATTRIBUTE_SWITCH_MODE = ZCLAttributeDef(
```
puddly, line 84: Centrally defining all attributes and then importing them, as opposed to just directly declaring them in the cluster attribute definitions, makes things harder to read. If you move them into their respective device module, it will make the code more self-contained and we can merge this PR device-by-device instead of waiting for all Yandex devices to be supported at once within this one PR.

Nullcaller: Thanks for all the comments! I'll try to apply all of your suggestions as soon as I can, so that this PR can get merged.

Yevgenium:
> Can everyone having issues test the fix, please?

Now the states are displayed correctly. Thank you.

Nullcaller:
I converted all of this to v2 quirks and compartmentalized the code, and opened #4843 as the first of PRs intended to replace this one. If and when it gets merged, I'll open more.

PR #4016, opened 2025-04-18
Update ts0601_sensor.py with _TZE200_w6n8jeuu

yucombinator:
Add `_TZE200_w6n8jeuu` to list of Tuya TS0601 temperature sensors

## Proposed change
Add a new Tuya device id to the TS0601 quirk.

_TZE200_w6n8jeuu is the device id I see in ZHA, but paring it with ZHA had no usable entities. Adding the device id to a local quirk file fixed the issue for me.  After my change my sensor entities are detected and works:

Screenshot 2025-04-17 at 5 11 52 PM

## Additional information

Link to the device on aliexpress: https://www.aliexpress.us/item/3256807102148050.html?spm=a2g0o.order_list.order_list_main.22.52b0180242jdlY&gatewayAdapt=glo2usa

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [x] Tests have been added to verify that the new code works

PR #4019, opened 2025-04-21
Add Tuya temp sensor `_TZE200_w6n8jeuu` variant

yucombinator:
Add `_TZE200_w6n8jeuu` to list of Tuya TS0601 temperature sensors

## Proposed change
Add a new Tuya device id to the TS0601 quirk.

_TZE200_w6n8jeuu is the device id I see in ZHA, but paring it with ZHA had no usable entities. Adding the device id to a local quirk file fixed the issue for me.  After my change my sensor entities are detected and works:

Screenshot 2025-04-17 at 5 11 52 PM

## Additional information

Link to the device on aliexpress: https://www.aliexpress.us/item/3256807102148050.html?spm=a2g0o.order_list.order_list_main.22.52b0180242jdlY&gatewayAdapt=glo2usa

## Checklist

### DIFF
--- a/zhaquirks/tuya/tuya_sensor.py
+++ b/zhaquirks/tuya/tuya_sensor.py
@@ -118,6 +118,7 @@ class NoManufTimeTuyaMCUCluster(TuyaMCUCluster):
     .applies_to("_TZE200_eanjj2pa", "TS0601")
     .applies_to("_TZE200_ydrdfkim", "TS0601")
     .applies_to("_TZE284_locansqn", "TS0601")
+    .applies_to("_TZE200_w6n8jeuu", "TS0601")
     .tuya_temperature(dp_id=1, scale=10)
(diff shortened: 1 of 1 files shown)


yucombinator:
I re-created the PR from https://github.com/zigpy/zha-device-handlers/pull/4016 but now based on the dev branch

TheJulianJES:
@yucombinator Can you send a full screenshot of the device page (including configuration entities) with the new v2 quirk? I'm wondering if this was perhaps added to the wrong base quirk, as the initial PR had it in a different spot.

Also, mind double checking that the local quirk you've used is the exact same as the patch here?

EDIT: This might be fully correct. I was confused by a similar Tuya sensor PR. I'd still appreciate a screenshot of the full device page in HA (including config entities) and you double checking that your custom quirk is the same is this patch.

PR #4023, opened 2025-04-23
Keep overriden functions async

puddly:
## Proposed change

Some functions, like `read_attributes`, are now async in zigpy but are overridden by a few quirks and replaced with sync overloads. This is functionally equivalent but results in ZHA test warnings:

```
Warning: coroutine 'Cluster.read_attributes' was never awaited
```

I've also converted the quirk for the affected device into a v2 quirk, since it did nothing but replace a single cluster.

## Checklist

### DIFF
--- a/zhaquirks/eurotronic/__init__.py
+++ b/zhaquirks/eurotronic/__init__.py
@@ -133,19 +133,19 @@ class ThermostatCluster(CustomCluster, Thermostat):
 
         return success, error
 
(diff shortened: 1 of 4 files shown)


TheJulianJES (approved):
Nice catch! Seems like we also have an old quirk doing this for `write_attributes`:
https://github.com/zigpy/zha-device-handlers/blob/e4844d48549d5a16ed288bb59283a30937ea5591/zhaquirks/eurotronic/__init__.py#L136-L151

PR #4031, opened 2025-04-28
Add Sonoff smart plug S60ZBTPF quirk

Gatsby-2024:
## Checklist

### DIFF
--- a/zhaquirks/sonoff/s60zbtpf.py
+++ b/zhaquirks/sonoff/s60zbtpf.py
@@ -1,7 +1,6 @@
 """Sonoff S60ZBTPF - Zigbee Smart Plug."""
 
 from typing import Final
-
 from zigpy import types
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.v2 import QuirkBuilder
@@ -10,11 +9,15 @@ from zigpy.quirks.v2.homeassistant import (
     UnitOfElectricPotential,
     UnitOfPower,
 )
-from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
 import zigpy.types as t
-from zigpy.zcl import ClusterType, foundation
-import zigpy.zcl.foundation as zcl_f
+from zigpy.zcl import foundation
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
+from zha.application.platforms.number.const import NumberMode
+from zigpy.zcl.clusters.general import OnOff
+from zigpy.zcl import ClusterType
+from zigpy.quirks import CustomDevice
+import zigpy.zcl.foundation as zcl_f
 
 
 class SonoffCluster(CustomCluster):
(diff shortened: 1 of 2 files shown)


zhaquirks/sonoff/s60zbtpf.py
```diff
@@ -0,0 +1,190 @@
...
+        return await super()._read_attributes(
+            attribute_ids,
+            *args,
+            manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID,
+            **kwargs,
+        )
+
+    @property
+    def _is_manuf_specific(self):
+        return False
```
TheJulianJES, line 88: We should be able to set `manufacturer_id_override` for the cluster instead.
See as an example: https://github.com/zigpy/zha-device-handlers/blob/a941a8968ece77df01defe8bbee337c939e04b82/zhaquirks/sonoff/swv.py#L24

zhaquirks/sonoff/s60zbtpg.py
```diff
@@ -0,0 +1,190 @@
...
+        return await super()._read_attributes(
+            attribute_ids,
+            *args,
+            manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID,
+            **kwargs,
+        )
+
+    @property
+    def _is_manuf_specific(self):
+        return False
```
TheJulianJES, line 88: Same comment as above here:

We should be able to set `manufacturer_id_override` for the cluster instead. See as an example: https://github.com/zigpy/zha-device-handlers/blob/a941a8968ece77df01defe8bbee337c939e04b82/zhaquirks/sonoff/swv.py#L24

zhaquirks/sonoff/s60zbtpf.py
```diff
@@ -0,0 +1,190 @@
...
+    cluster_id = 0xFC11
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        # network_led = ZCLAttributeDef(
+        #     name="network_led",
+        #     id=0x0001,
+        #     type=t.Bool,
+        # )
```
TheJulianJES: Does this attribute not work? If so, it can be removed. We shouldn't add commented code.
If the attribute does work, we can define it here (so just uncomment). We don't have to necessarily add a config entity for all attributes (although it might make sense for this attribute).

zhaquirks/sonoff/s60zbtpf.py
```diff
@@ -0,0 +1,190 @@
...
+(
+    QuirkBuilder("SONOFF", "S60ZBTPF")
+    .replaces(SonoffCluster)
+    # .enum(
+    #     SonoffCluster.AttributeDefs.network_led.name,
+    #     SonoffNetworkLedSetType,
+    #     0xFC11,
+    #     translation_key = "network_led",
+    #     fallback_name = "Network led",
+    # )
```
TheJulianJES: Related to the comment above regarding the `network_led` attribute:
Does this not work? Should this be a `switch` instead?

zhaquirks/sonoff/s60zbtpg.py
```diff
@@ -0,0 +1,190 @@
...
+        unit=UnitOfPower.WATT,
+        multiplier=0.001,
+        translation_key="ac_power_max_overload",
+        device_class=NumberDeviceClass.POWER,
+        fallback_name="AC power max overload",
+    )
+    .add_to_registry()
+)
+
+# class SonoffS60zb(CustomDevice):
```
TheJulianJES: Please remove this commented line.

TheJulianJES (commented):
The quirks look very similar to me. If they are indeed basically the same, we should combine them in one file, e.g. `sonoff/plug.py` and add the different variants for models in one quirk using `applies_to`.

Gatsby-2024:
@TheJulianJES I have already made the changes.

PR #3821, opened 2025-02-01
Add support for Sinope TH1134ZB-HC thermostat

claudegel:
## Proposed change

Add support for the new TH1134ZB-HC Sinope thermostat. Added many attributes and support for weather icon control on thermostat display.

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

PR #4042, opened 2025-05-02
Transfer all Sinopé devices to quirks V2

claudegel:
## Proposed change

Transfer all Sinopé devices to quirks V2 and add support for some new devices.
Add many new attributes in manufacturer cluster.

## Additional information

Some manufacturer cluster attributes are already supported in ZHA but attributes values are not the same. ex. attribute 0x0002 for keypad_lockout in manufacturer cluster have three values. Unlocked = 0x00, Locked = 0x01, Partial_lock = 0x02. In ZHA there is 5 values  Unlock, lock1, lock2, lock3 and lock4 which have no meaning for Sinopé devices. 
Same for on_led_color which have only the color used in Neviweb. We should be able to add mode color.
They should be included in this quirks as they are only available for Sinopé devices.
This PR replace PR #3821.

## Checklist


claudegel:
Look like there is a bug with some quirks:
AssertionError: Manufacturer-model pair 'Sinope Technologies' 'RM3250ZB' is shared by multiple quirks: ['/home/runner/work/zha-device-handlers/zha-device-handlers/zhaquirks/sinope/switch.py:444', '/home/runner/work/zha-device-handlers/zha-device-handlers/zhaquirks/sinope/switch.py:399']
The to devices have the same SKU RM3250ZB but there cluster are different.
Can I just rename one from QuirkBuilder(SINOPE, "RM3250ZB") to QuirkBuilder(SINOPE, "RM3250ZB-V2") to fix this. I have many devices like this and it was ok in the previous quirks.

TheJulianJES:
Quirks v2 no longer match by the whole ep/cluster signature. They only match by manufacturer and model.

If you really need to distinguish two identically named devices depending on their cluster signature, I believe you should be able to do it something like this:
```python
sinope_device_1_signature = {
    MODELS_INFO: [(SINOPE, "RM3250ZB")],
    ENDPOINTS: {
        1: {
            PROFILE_ID: zha_p.PROFILE_ID,
            DEVICE_TYPE: zha_p.DeviceType.ON_OFF_OUTPUT,
            INPUT_CLUSTERS: [
                Basic.cluster_id,
                DeviceTemperature.cluster_id,
                Identify.cluster_id,
                Groups.cluster_id,
                Scenes.cluster_id,
                OnOff.cluster_id,
                Metering.cluster_id,
                ElectricalMeasurement.cluster_id,
                Diagnostic.cluster_id,
                SINOPE_MANUFACTURER_CLUSTER_ID,
            ],
            OUTPUT_CLUSTERS: [
                Identify.cluster_id,
                Groups.cluster_id,
                Ota.cluster_id,
            ],
        }
    },
}

# and when calling the v2 QuirkBuilder methods, add:
.filter(signature_matches(sinope_device_1_signature))
# ...
```

That *should* additionally match by signature, though you ideally don't want to do this.
The tests will likely still fail, as we haven't needed this behavior so far. If we need to go this route, I can adjust the tests to ignore quirks with a specified `filter` though.

If this is about preventing an entity to be created for one firmware version, there's no good way to do that at the moment.
Currently, v2 quirks entities are force-created, even if the attribute value cannot be read during pairing, for cases when the attribute isn't exposed on the device.
Ideally, we'd add an additional parameter that can be enabled to only create specified entities if the attribute is actually present in zigpy cache, so exposed by the device. This would work for older devices returning `None" and newer ones actually returning the attribute.

We do have v2 quirks that also expose the entity on older firmware versions though, even if it's not supported.
There's no big harm in that, but we'll try to fix that in the future.
Related issue: https://github.com/zigpy/zha/issues/373

TheJulianJES:
> Can I just rename one from QuirkBuilder(SINOPE, "RM3250ZB") to QuirkBuilder(SINOPE, "RM3250ZB-V2") to fix this. I have many devices like this and it was ok in the previous quirks.

So, to answer that: No, if the model name of the device is "RM3250ZB", then the "RM3250ZB-V2" quirk wouldn't match it. Only the first one with "RM3250ZB" would match it.

claudegel:
If I keep the two as QuirkBuilder(SINOPE, "RM3250ZB") it load correctly and I can see both device in ZHA. But it is shared-ci / Run tests Python 3.12 that fail with AssertionError

TheJulianJES:
No, it's the same manufacturer and model. Only one of the quirks will apply. Keep both as `RM3250ZB` and remove one. Everything will seem work fine. Now, add that quirk back and remove the other `RM3250ZB`. Everything will still work fine.
It's because only one of the quirks will ever work.
I'd have to see if it's the one added first (i.e. further up in the file) or the one further below, but that doesn't really matter. Only one of them will apply if they have the same manufacturer and model.

v2 quirks only match by manufacturer and model and you're adding two quirks with the same manufacturer and model.
ZHA will only apply one of them, since they're meant for the exact same device.
The test just informs you that you're doing something that you likely don't want to do.

v1 quirks had the `signature`. v2 quirks only have manufacturer and model to match against by default, no `signature`.
The `# SimpleDescriptor` comments are just comments, they don't do anything.

TheJulianJES:
The great thing about v2 quirks is that we don't need multiple quirk variants if firmware updates or other variants changed the device signature ever so slightly.
Like, to v2 quirks, it doesn't matter if there's an additional `Time` cluster on one of the endpoints after a firmware update.

For v1 quirks, you'd have to add another quirk, with a new `signature` and have two quirks in the end if there are different variants of devices (e.g. with different firmware).
With v2 quirks, you only have one quirk, as it only matches the device by the manufacturer and model name. The rest of the signature (endpoints and clusters) are completely ignored.

claudegel:
Both device expose there signature for cluster but one device have more cluster than the old one:
```
    # <SimpleDescriptor(endpoint=1, profile=260,
    # device_type=2, device_version=0,
    # input_clusters=[0, 3, 4, 5, 6, 1794, 2820, 2821, 65281]
    # output_clusters=[3, 4, 25]>
    QuirkBuilder(SINOPE, "RM3250ZB")
```
```
    # <SimpleDescriptor(endpoint=1, profile=260,
    # device_type=2, device_version=0,
    # input_clusters=[0, 2, 3, 4, 5, 6, 1794, 2820, 2821, 65281]
    # output_clusters=[3, 4, 25]>
    QuirkBuilder(SINOPE, "RM3250ZB")
```
But they share the same QuirkBuilder(SINOPE, "RM3250ZB") and same device_version=0.

Can I put only the device quirks that have all cluster like the second one above. Does ZHA will load the first one above if one cluster is missing, 0x0002

zhaquirks/sinope/light.py
```diff
@@ -237,199 +258,262 @@ def _get_command_from_action(
...
+    QuirkBuilder(SINOPE, "SW2500ZB")
+    .applies_to(SINOPE, "SW2500ZB-G2")
+    .replaces_endpoint(1, device_type=zha_p.DeviceType.ON_OFF_LIGHT)
+    .adds(Basic, endpoint_id=1)
+    .adds(Identify, endpoint_id=1)
+    .adds(Groups, endpoint_id=1)
+    .adds(Scenes, endpoint_id=1)
+    .adds(OnOff, endpoint_id=1)
+    .adds(Metering, endpoint_id=1)
+    .adds(Diagnostic, endpoint_id=1)
```
TheJulianJES, line 319: You shouldn't be doing this in v2 quirks. Only add clusters if they're missing from the device signature by default.
This won't do anything, since they're most likely all already exposed by the device itself.

Like, all of the `.adds(...)` should be completely removed here.

TheJulianJES:
To re-iterate on the differences vs v1 quirks and v2 quirks:
v1 quirks had a `signature` and `replacement`. If the `signature` matches a device, it'll present the `replacement` signature to ZHA. So, if you want to replace a cluster, you'd just replace it in the `replacement` signature, keeping everything the same.

For v2 quirks, it works nothing like that. They match by manufacturer and model only. Now, if I want to replace a cluster only, I just do this:
```python
(
    QuirkBuilder("IKEA", "TRADFRI control outlet")
    .replaces(MyCustomCluster, endpoint_id=1) # the endpoint_id=1 is optional for ep 1, as 1 is the default
    .add_to_registry()
)
```

All other clusters on the device are left untouched. For example, all devices have to expose a `Basic` cluster on at least one endpoint, but we don't care about that here, as we don't need to modify it. So we don't do anything with that.
Only the `MyCustomCluster` is replaced with one that has further attribute definitions, for example.

The example IKEA control outlet above also has an `OnOff` cluster exposed by default. We don't touch it, because that basic functionality already works fine.

--- 

You can kind of consider v2 quirks as the difference of v1 quirks when comparing the `signature` and `replacement`.
All clusters that you don't care about are exactly the same in `signature` and `replacement` for v1 quirks.
So, for v2 quirks, you don't do anything with those clusters.

But if you want to replace/remove/add a cluster in a v1 quirk, you change that in the `replacement`.
Considering the difference between the `signature` and `replacement`, you'll then want to use v2 quirk equivalents like `.replaces(cluster)`,`.remove(cluster_id)`, or `.adds(cluster)`.

Is that a better explanation?

zhaquirks/sinope/light.py
claudegel: I'm doing this only in the light.py because I had to add line
`.replaces_endpoint(1, device_type=zha_p.DeviceType.ON_OFF_LIGHT)`
and this was removing all cluster that I didn't .replace
```
    .replaces(CustomDeviceTemperatureCluster)
    .replaces(LightManufacturerCluster)
```
Adding those lines fix the problem. 
In Sinopé light DevicesType are reported wrong so I need to do that .replace to change the DeviceType. This work ok in V2.

claudegel:
So I will keep only one copy of QuirkBuilder(SINOPE, "RM3250ZB") and make sure that it contain all .number, .enum, .sensor that are present in the other QuirkBuilder(SINOPE, "RM3250ZB").
I'll do it for all duplicate I have mostly in thermostat.py, switch.py and sensor.py.
What happens for the device that is missing one cluster ? The related .number, .sensor etc will be ignored or just grey out in the device card ?

claudegel:
There is an import error in test_sinope.py. Is it you who fix this ?

Also there is a AssertionError: Translation key 'valve_countdown' is shared by quirks with different fallback names:
Do I need to pay Tuya :)

claudegel:
Also where can I add missing french translation.

TheJulianJES:
> What happens for the device that is missing one cluster ? The related .number, .sensor etc will be ignored or just grey out in the device card ?

Currently, the entity will still be shown for all versions of that device, but not work if the attribute isn't supported
In the future, we can hopefully not show it depending on if the attribute value can be read during pairing. But since it's just one entity, it's not that big of a concern. Especially since the user can just disable the entity (or update device firmware to make it appear).

> There is an import error in test_sinope.py. Is it you who fix this ?

Try to remove the `parametrize` line and use the `zigpy_device_from_v2_quirk` fixture instead with the device manufacturer and model to test, so something like this for `test_sinope.py`:
```python
async def test_sinope_device_temp(zigpy_device_from_v2_quirk):
    """Test that device temperature is multiplied."""
    device = zigpy_device_from_v2_quirk("Sinope Technologies", "RM3500ZB")
    # continue as before...
```

> Also there is a AssertionError: Translation key 'valve_countdown' is shared by quirks with different fallback names:
Do I need to pay Tuya :)

So, the translation keys and English values from `fallback_name` will eventually make its way into Home Assistant here: [homeassistant/components/zha/strings.json#L1045-L1047](https://github.com/home-assistant/core/blob/cad2d72ed98f8ae145fc13ac056420c6cdca3cd4/homeassistant/components/zha/strings.json#L1045-L1047)
And one Tuya quirk unfortunately used the `valve_countdown` translation key with the value of "Irrigation time".
Since that would conflict with your translation key and fallback name, the test fails.

So, we either have to change that Tuya quirk, or your quirk has to use a different translation key.
Changing the Tuya one(s) to something like `irrigation_time` probably makes more sense... I'll come back to this.

> Also where can I add missing french translation.

They'll be added by community members when the ZHA bump PR with this quirks version is merged into HA.
They can be modified on the HA's Lokalise page (after a ZHA bump PR is merged). For more info, see this: https://developers.home-assistant.io/docs/translations/

For custom quirks, you can't modify the translations.
Also, there's also an issue with local quirks that quirks v2 entities with a device class will not show the `fallback_name` for the entity name, but instead always show the device class name, e.g. just "Temperature" for that device class.

This will automatically be "fixed" when the translations are available inside HA (so when the PR + bump PRs to HA are merged). For more info on that issue, see here: https://github.com/zigpy/zigpy/issues/1523#issuecomment-2564106284

claudegel:
Look like there is only the Valve countdown collision with Tuya that we need to fix. maybe I can change it to closing countdown. I hope nobody have used this one :)

claudegel:
@TheJulianJES There is one last test that do not work but this one I don't know how to fix. Can you help ?

TheJulianJES:
The Codecov project check doesn't really matter.
I think it just fails, as the overall number of lines was drastically reduced with the v2 quirks, thus also causing the whole project coverage percentage to be lower, failing the check.

All lines of the v2 quirks are covered, so that's good.

claudegel:
Thank you. check it and let me know if we can merge to V2

claudegel:
I've found my last bug for dimmer, DM2500ZB. Now all entities are up and running
Thank you for your help

claudegel:
@TheJulianJES , I've few questions:
- For number is it possible to force display as a number instead of a slider. I have possible value from 0 to 60.
- For number do we need to put a unit= ???, I've a number which is not in seconds, celsius or percent. Just no unit of measure. Can I omit unit= or it will use a default unit.
- for IAS 0x0002 ZoneStatus I have different values for different status. This attribute is not in the quirks and It is set as binary_sensor, on / off. Is it possible to make a select even if that attribute is uint16 in order to list the error message instead of a value. or anything else that work.

claudegel:
Ok I've found how to make an enum out of t.uint16_t
I just put t.enum_factory(t.uint16_t, "manufacturer_specific") and it work.
```
class ZoneStatus(t.enum_factory(t.uint16_t, "manufacturer_specific")):
    """IAS zone status."""

    Ok = 0x0030
    Connector_1 = 0x0031
    Connector_2 = 0x0032
    Low_battery = 0x0038
    Connector_low_bat = 0x003A
```
Now can you tell me how to limit a value. For example I receive the value for battery_percentage_remaining which goes to 200% with new battery. How can I set the value red from the attribute not to go above 100%

TheJulianJES:
> For number is it possible to force display as a number instead of a slider. I have possible value from 0 to 60.

Not possible via quirks v2 entities, yet. There's a request for it here: https://github.com/zigpy/zha/issues/364

> For number do we need to put a unit= ???, I've a number which is not in seconds, celsius or percent. Just no unit of measure. Can I omit unit= or it will use a default unit.

You can omit the unit and no unit will be shown in HA.
Even if you want to use a unit that's not already [built-in here](https://github.com/zigpy/zigpy/blob/dev/zigpy/quirks/v2/homeassistant/__init__.py), you can just define a custom string, e.g. `unit="°C"` (though that's built-in).

> * for IAS 0x0002 ZoneStatus I have different values for different status. This attribute is not in the quirks and It is set as binary_sensor, on / off. Is it possible to make a select even if that attribute is uint16 in order to list the error message instead of a value. or anything else that work.

An `IasZone` entity should already be created by ZHA, though that will always trigger for bit 1 and bit 2.
If you need to disable the default entity to create your own entities, you can do something like this:
```python
.prevent_default_entity_creation(endpoint_id=1, cluster_id=IasZone.cluster_id)  # disable ZHA entity
.binary_sensor(
    IasZone.AttributeDefs.zone_status.name,  # though you may want to use your own attribute definition here
    IasZone.cluster_id,
    unique_id_suffix="ias_1",  # to not conflict with unique ID of other entity, as only the attr name is used by default
    translation_key="leak_status",
    fallback_name="Leak status",
    attribute_converter=lambda x: x & 0b00000001,  # use only bit 1
)
.binary_sensor(
    IasZone.AttributeDefs.zone_status.name,  # though you may want to use your own attribute definition here
    IasZone.cluster_id,
    unique_id_suffix="ias_2",  # to not conflict with unique ID of other entity, as only the attr name is used by default
    translation_key="tamper_status",
    fallback_name="Tamper status",
    `,  # use only bit 2
)
.add_to_registry()
```
If you have multiple quirks v2 entities for the same attribute, as might be possible for `zone_status`, you'll need to add a `unique_id_suffix` to each entity like shown above to not conflict in HA, as it's normally just the attribute name, so it would be `zone_status` for both entities (but we specified a different `unique_id_suffix` above, so it's fine). 

The same might also apply if you have the same quirks v2 entity for multiple endpoint ids, but I can't fully remember right now.

Above, I've also shown how you can use `attribute_converter` to only use some bits of the attribute value.


> Ok I've found how to make an enum out of t.uint16_t I just put t.enum_factory(t.uint16_t, "manufacturer_specific") and it work.
> 
> ```
> class ZoneStatus(t.enum_factory(t.uint16_t, "manufacturer_specific")):
>     """IAS zone status."""
> 
>     Ok = 0x0030
>     Connector_1 = 0x0031
>     Connector_2 = 0x0032
>     Low_battery = 0x0038
>     Connector_low_bat = 0x003A
> ```


Eh, I don't think you should/want to do that generally.
There's also a `zcl_type` you can set on attribute definitions, though there might still be issues when reporting needs to be set up for that. But you can have the zigpy (enum) `type` and the `zcl_type` being the integer type.

Like, the above way shows how to make separate binary sensors for that (might need to keep the type as an int?).
If you want it to be one read-only sensor entity, you can keep using `.enum` and just add `entity_platform=EntityPlatform.SENSOR` to use the sensor platform, instead of the default select platform.

> Now can you tell me how to limit a value. For example I receive the value for battery_percentage_remaining which goes to 200% with new battery. How can I set the value red from the attribute not to go above 100%

ZHA already creates a default entity for `battery_percentage_remaining`, so you shouldn't need to create that on your own.

ZCL specifies that it's `percentage = value / 2`. So, 100 % is reported as value `200`.
I think you should be able to add `multiplier=0.5` on your sensor or number entity to divide the number by 2.

Otherwise, `attribute_converter=lambda x: x / 2` should also be possible. You can also do more sophisticated calculations using `attribute_converter` (e.g. call a "proper method").

TheJulianJES:
I've not had a proper look at everything changed in this PR, but if you have multiple quirks that use duplicated entities, you might be able to add a base quirk without any model info, and then clone that base quirk multiple times with added model info and more custom entities.

See this example, but try to ignore all the Tuya stuff:
https://github.com/zigpy/zha-device-handlers/blob/5556090557f8e10e65cda4b36db5dfa45911cef4/zhaquirks/tuya/tuya_valve.py#L226-L351

`gx02_base_quirk` is the base quirk. Its entities will be cloned to the other two quirks below (the first for `_TZE200_sh1btabb` only and the second for `_TZE200_a7sghmms` and others). 
That way, the entities added to the base `gx02_base_quirk` quirk don't have to be copied to the two other quirks.

Doing it this way isn't always helpful, but it might be for this PR (if you have very similar, but slightly differing devices).

claudegel:
It there a way to express data like battery voltage as integer not float. 100 instead of 100.0

puddly:
You can use `suggested_display_precision=0` when creating a sensor. It defaults to 1.

claudegel:
The sensor for battery_percentage_remaining is not created by my quirk but come from ZHA.
So I didn't create the sensor and I can't add suggested_display_precision=0. Another option?
But it's good to know for my quirks. Thanks

claudegel:
Using suggested_display_precision=0 give me the following error:
QuirkBuilder.sensor() got an unexpected keyword argument 'suggested_display_precision' and I'm not sure it give the result it should. I still get value as 20.0 instead of 20 for battery percent remaining. Do I need to import something in sensor.py ?

claudegel:
I've finished to review all sensor, number etc to avoid duplicate with ZHA quirks V2. I also moved all common sensor etc to sinope_base_quirk for thermostat.py. let me know how to fix suggested_display_precision error.

claudegel:
I've a question. For tank level sensor LM4110-ZB. if the probe is disconneted the sensor gauge_angle will report a value of -2. It there a way we can catch this error in the sensor.py code ?

TheJulianJES:
> Using suggested_display_precision=0 give me the following error:
QuirkBuilder.sensor() got an unexpected keyword argument 'suggested_display_precision'

We only added this for HA 2025.5.0 (and later). The minimum zigpy version required by zha-quirks doesn't have this additional parameter yet, so the tests in this repo will fail until I bump that version (and this PR being updated/rebased on top of that then).

However, it should work locally..

> I've a question. For tank level sensor LM4110-ZB. if the probe is disconneted the sensor gauge_angle will report a value of -2. It there a way we can catch this error in the sensor.py code ?

We might be able to add an `attribute_converter` to that sensor entity:
```python
attribute_converter=lambda x: None if x == -2 else x
```

That would just return `None` if the value is `-2`. That *should*  show "Unavailable" (or "Unknown"?) in the HA UI if the value is `-2`.

zhaquirks/sinope/switch.py
```diff
@@ -294,517 +372,460 @@ class SinopeTechnologiesMeteringCluster(CustomCluster, Metering):
...
+        state_class=SensorStateClass.MEASUREMENT,
+        unit=PERCENTAGE,
+        device_class=SensorDeviceClass.BATTERY,
+        reporting_config=ReportingConfig(
+            min_interval=30, max_interval=43200, reportable_change=1
+        ),
+        translation_key="battery_percentage_remaining",
+        fallback_name="Battery percentage remaining",
+        entity_type=EntityType.DIAGNOSTIC,
+    )
```
TheJulianJES, line 672: Why is this needed? If you remove this, does the built-in ZHA battery percentage sensor not appear?

zhaquirks/sinope/switch.py
```diff
@@ -294,517 +372,460 @@ class SinopeTechnologiesMeteringCluster(CustomCluster, Metering):
...
+        state_class=SensorStateClass.TOTAL_INCREASING,
+        unit=UnitOfEnergy.WATT_HOUR,
+        device_class=SensorDeviceClass.ENERGY,
+        reporting_config=ReportingConfig(
+            min_interval=59, max_interval=1799, reportable_change=60
+        ),
+        translation_key="current_summ_delivered",
+        fallback_name="Current summ delivered",
+        entity_type=EntityType.STANDARD,
+    )
```
TheJulianJES, line 514: Same for this. ZHA should already create an entity for this `Metering` cluster, so we shouldn't add a custom entity.
Does the ZHA entity not appear for some reason?

claudegel:
I've added attribute_converter=lambda x: None if x == -2 else x, in sensor.py
.sensor(  # battery percent
If I remove it the battery percent does not show up in that diagnostic section.
.sensor(  # Current summ delivered
yours is not working. see two screen capture: with my Current summ delivered in quirks and one without. Your summ delivered is not working and if I remove mine, both are not working
summ with mine
summ without mine

TheJulianJES:
> I've added attribute_converter=lambda x: None if x == -2 else x, in sensor.py

Ok, and that works as expected now (showing "unavailable" if the probe is not connected)?

> .sensor(  # battery percent
If I remove it the battery percent does not show up in that diagnostic section.
.sensor(  # Current summ delivered
yours is not working. see two screen capture: with my Current summ delivered in quirks and one without. Your summ delivered is not working and if I remove mine, both are not working

Hmm, we have to get this fixed. Were you able to try this on HA 2025.6.0 betas? They already appear very stable to me. If so, does the behavior change on those versions?

Also, do the default ZHA entities work if you use the (integrated) v1 quirks, or no quirks for the Sinope devices at all? (you'd have to remove the integrated quirks to test "no quirks" at all)
Like, did these entities already not show up with the v1 quirks, even though the clusters are default ZCL and present on the devices?

claudegel:
- For the switch if I remove my quirk, switch.py and restart HA the Current summ delivered does not work.
- same for the valve .sensor(  # battery percent does not work.
- Restarting HA with the quirk switch.py back in strangely both are working for the current summ delivered. My battery percent is working again.
- For your current summ delivery for switch I've noticed that sometime it work and after a restart of HA it don't work. Mine is always ok. :)
- For the sensor I'll check tomorrow as I need to put it back in HA. Was working on an update in another custom_component about that device.
- I also add to change something in thermostat.py to be able to set an enum entities for PumpStatus
```
class PumpStatus(t.enum_factory(t.uint8_t, "manufacturer_specific")):
    """Pump protection status."""

    Off = 0x00
    On = 0x01
```
if I live it as:
```
class PumpStatus(t.uint8_t):
    """Pump protection status."""

    Off = 0x00
    On = 0x01
```
I get an error:
```
TypeError: 'type' object is not iterable
2025-06-02 11:20:01.889 ERROR (MainThread) [homeassistant.components.zha.helpers] Error while adding entity from entity data: EntityData(entity=<zha.application.platforms.select.ZCLEnumSelectEntity object at 0x7f7d39919950>, device_proxy=<homeassistant.components.zha.helpers.ZHADeviceProxy object at 0x7f7d3992d6d0>, group_proxy=None)
  File "/srv/homeassistant/lib/python3.13/site-packages/homeassistant/components/zha/helpers.py", line 1169, in async_add_entities
    entities_to_add.append(entity_class(entity_data))
                           ~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/srv/homeassistant/lib/python3.13/site-packages/homeassistant/components/zha/select.py", line 55, in __init__
    super().__init__(entity_data, **kwargs)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/srv/homeassistant/lib/python3.13/site-packages/homeassistant/components/zha/entity.py", line 45, in __init__
    meta = self.entity_data.entity.info_object
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/functools.py", line 1039, in __get__
    val = self.func(instance)
  File "/srv/homeassistant/lib/python3.13/site-packages/zha/application/platforms/select.py", line 232, in info_object
    options=self._attr_options,
            ^^^^^^^^^^^^^^^^^^
AttributeError: 'ZCLEnumSelectEntity' object has no attribute '_attr_options'
```

TheJulianJES:
> if I live it as: [...] I get an error:

Yeah, because `t.uint8_t` is an int type, not an enum type. You'd need to change the class to inherit from `t.enum8` instead.
Then, the `zcl_type=t.uint8_t` can be defined in the `ZCLAttributeDef`.
Some other quirks still use `t.enum_factory(t.uint8_t)` instead, but we should no longer do that, as it's not really correct.

However, it doesn't make a ton of sense to define an enum for `Off`/`On`. IMO, the class should be removed and you should just keep using `type=t.uint8_t` in the `ZCLAttributeDef`.
Then, use a `binary_sensor` or `switch`, depending on if it's read-only or also writeable.

TheJulianJES:
For the missing default ZHA entities, there should be some more info when enabling debug logs in `configuration.yaml` like this:
```yaml
logger:
  default: info
  logs:
    homeassistant.core: debug
    homeassistant.components.zha: debug
    bellows.zigbee.application: debug
    bellows.ezsp: debug
    zha: debug
    zigpy: debug
    zigpy_deconz.zigbee.application: debug
    zigpy_deconz.api: debug
    zigpy_xbee.zigbee.application: debug
    zigpy_xbee.api: debug
    zigpy_zigate: debug
    zigpy_znp: debug
    zhaquirks: debug
```
(`zha: debug` alone should be enough)

Please try to grab debug logs, without your custom quirks `battery_percentage_remaining` and `current_summ_delivered` sensors. Possible important lines are `<ATTR_NAME> is not supported - skipping <XXX> entity creation`.
Please also download diagnostics information and drag and drop them into a GitHub comment to upload them.
The diagnostics information should not contain anything sensitive, but the debug logs might.

IIRC the `current_summ_delivered` sensor needs `unit_of_measurement` on be specified on the `Metering` cluster. This should be done by the device, but the diagnostics information will contain the current state of the zigpy attribute cache for that device, so give us clarity on if the attribute is present on the device (or if zigpy think it is (or not)).

The `battery_percentage_remaining` sensor needs the entity to be marked as a non-mains device in the NodeDescriptor.

Also, are you on HA [2025.6.0b4](https://github.com/home-assistant/core/releases/tag/2025.6.0b4) now and the issues still continue?

And one last thing, please try to remove(!) and re-pair the device once (without your custom quirks sensors being present, but with the v2 quirks, and maybe also without after that). Removing the device makes sure the zigpy attribute cache in the database is cleaned for that device, so it's properly re-interviewed. It will also remove all your HA entities though.

Please also grab diagnostics information after that, so we can possibly compare them.

claudegel:
OK I confirm that attribute_converter=lambda x: None if x == -2 else x return None if the sensor is disconnected.
sensor

claudegel:
for t.enum_factory(t.uint8_t) I've set it back to t.uint8_t
I've replace the .enum with a .switch like this:
```
    .switch(  # Pump protection status
        attribute_name=SinopeTechnologiesManufacturerCluster.AttributeDefs.pump_protection_status.name,
        cluster_id=SinopeTechnologiesManufacturerCluster.cluster_id,
        off_value: 0
        on_value: 1
        translation_key="pump_protection_status",
        fallback_name="Pump protection status",
    )
```
Do I need to add anything else to the .switch ?
Now I'm going to play with logs...

claudegel:
With my switch.py quirk removed when I restart HA i got this
`[zha.application.platforms.sensor] current_summ_delivered is not supported - skipping SmartEnergySummation entity creation`
`instantaneous_demand is not supported - skipping SmartEnergyMetering entity creation`
`on_intensity is not supported - skipping SinopeDimmerOnLevelConfigurationEntity entity creation`
and many more
[home-assistant.log.zip](https://github.com/user-attachments/files/20596428/home-assistant.log.zip)

claudegel:
Now I'm on HA 2025.5.3, how do I try 2025.6.0b4

claudegel:
here is my log withthe switch.py back in
[home-assistant.log](https://github.com/user-attachments/files/20596534/home-assistant.log)

claudegel:
Since HA 2025.6.2 suggested_display_precision is working. Thank's

claudegel:
So do we need to add anything to make that PR going ?

claudegel:
Closure countdown was actually a timer to temporary disable flow alarm action and report. So a change the name to better reflect its action, alarm_disable_countdown and I adjust the limit values from 0 to 86400 seconds, 0 = off

zhaquirks/sinope/thermostat.py
```diff
@@ -7,64 +7,90 @@
...
+            try:
+                await self.configure_reporting(
+                    attribute=attr_id,
+                    min_interval=min_i,
+                    max_interval=max_i,
+                    reportable_change=change,
+                )
+                self.debug(f"Reporting configured for attr {hex(attr_id)}")
+            except Exception as e:
+                self.debug(f"Reporting configuration fail for attr {hex(attr_id)}: {e}")
```
TheJulianJES, line 52: If there's a quirks v2 entity for the attribute you're looking to configure reporting, you can provide `reporting_config` for that entity and ZHA will configure reporting for the underlying attribute.
See this example: https://github.com/zigpy/zha-device-handlers/blob/bf73aa94193e8162d85997d311f23e04cd1a25a0/zhaquirks/sonoff/swv.py#L50-L52

In the future, we likely want a separate quirks v2 method for configuring reporting, so we can also do this for attributes that are not directly used by a quirks v2 entity.

zhaquirks/sinope/thermostat.py
```diff
@@ -325,414 +477,451 @@ class SinopeTechnologiesElectricalMeasurementCluster(
...
+    )
+    .sensor(  # Gfci status
+        attribute_name=SinopeTechnologiesManufacturerCluster.AttributeDefs.gfci_status.name,
+        cluster_id=SinopeTechnologiesManufacturerCluster.cluster_id,
+        state_class=SensorStateClass.MEASUREMENT,
+        reporting_config=ReportingConfig(
+            min_interval=10, max_interval=3600, reportable_change=1
+        ),
+        translation_key="gfci_status",
+        fallback_name="Gfci status",
```
TheJulianJES, line 763: Since this is an abbreviation, it can be "GFCI status". See [this](https://developers.home-assistant.io/docs/core/entity/#entity-naming:~:text=Entity%20names%20should%20start%20with%20a%20capital%20letter%2C%20the%20rest%20of%20the%20words%20are%20lower%20case%20(unless%20it%27s%20a%20proper%20noun%20or%20a%20capitalized%20abbreviation%20of%20course).).

zhaquirks/sinope/thermostat.py
```diff
@@ -325,414 +477,451 @@ class SinopeTechnologiesElectricalMeasurementCluster(
...
+    sinope_base_quirk.clone()
+    .applies_to(SINOPE, "TH1300ZB")
+    .replaces(SinopeTechnologiesElectricalMeasurementCluster)
+    .enum(  # Config air floor mode
+        attribute_name=SinopeTechnologiesManufacturerCluster.AttributeDefs.air_floor_mode.name,
+        cluster_id=SinopeTechnologiesManufacturerCluster.cluster_id,
+        enum_class=FloorMode,
+        translation_key="air_floor_mode",
+        fallback_name="Air floor mode",
+        entity_type=EntityType.CONFIG,
```
TheJulianJES, line 710: Let's always keep `translation_key` and `fallback_name` as the last arguments, same as in other v2 quirks.

Also, `entity_type` is often `CONFIG` by default for a lot of these entities, since quirks v2 entities were mostly made to create config entities, so we shouldn't need to explicitly set this in most places.

zhaquirks/sinope/thermostat.py
TheJulianJES: Ah, I've seen you already used `reporting_config` elsewhere.

zhaquirks/sinope/thermostat.py
```diff
@@ -325,414 +477,451 @@ class SinopeTechnologiesElectricalMeasurementCluster(
...
+        fallback_name="Time format",
+        entity_type=EntityType.CONFIG,
+    )
+    .number(  # eco delta setpoint
+        attribute_name=SinopeTechnologiesManufacturerCluster.AttributeDefs.eco_delta_setpoint.name,
+        cluster_id=SinopeTechnologiesManufacturerCluster.cluster_id,
+        step=1,
+        min_value=-128,
+        max_value=100,
+        unit=UnitOfTemperature.CELSIUS,
```
TheJulianJES, line 533: We can add the `NumberDeviceClass.TEMPERATURE` device class for these. See [this](https://developers.home-assistant.io/docs/core/entity/number/#available-device-classes:~:text=NumberDeviceClass.TEMPERATURE).

I believe one difference it makes is converting from °C to °F if you're using imperial units (and the device class is set).

zhaquirks/sinope/light.py
TheJulianJES: The issue with `replaces_endpoint` should have been fixed with https://github.com/zigpy/zigpy/pull/1602. Can you confirm if you can now remove these `adds` lines?

zhaquirks/sinope/switch.py
TheJulianJES: Can we go back to this: We've not had a custom battery percentage sensor with quirks v1, so I'm assuming the default ZCL one from ZHA was shown? This should still be the case, so why is a custom one added here?

1. Was there a battery percentage sensor for the v1 quirks? Or was it also missing there?
2. If you remove the custom "battery percentage remaining" sensor, does the default ZCL/ZHA battery percentage remaining sensor show up? Does it work? Is it unavailable? Does it not show up at all?

TheJulianJES (commented):
Sorry for the wait on this. It's just a huge PR. Let's leave it like this for now, but it would have been nicer if we had multiple PRs, ideally even per platform (separately done for light/sensor/switch/thermostat), like this:

1. PR: Small changes like importing a lot of things from `foundation` directly.
2. PR: Replacing v1 quirks with v2 quirks, without adding any entities.
3. PR: Adding new attributes.
4. PR: Adding quirks v2 entities.

This is a huge diff and makes it hard to review. If one comment is added somewhere, it also needs to be changed in a lot of other places. If these were done one after another (and also separated per platform), this would be much easier to do and review.

But again, leave it like this now. Added a few comments and questions. Some apply to other places as well. Also note that I did not go over all of the updated code yet.

claudegel:
Thank you for your review. Let me go through your comment and I'll see what I can do. For sure I can change some parts to add your suggestion. 
I'm also really busy with my other components that I'm porting to V2 config_flow etc.

zhaquirks/sinope/thermostat.py
claudegel: It's only for the manufacturer cluster 0xff01 because many users that pear new devices that was not paired with the Neviweb server from Sinopé don't have there cluster configured so I implement this to make sure new device are correctly configured for the manufacturer cluster. For other cluster I let ZHA do it. But also Sinopé have added some attributes specific for there devices in other clusters. Does ZHA will be able to configure the reporting for those attributes if needed ? If yes I'll check to use reporting_config.

zhaquirks/sinope/thermostat.py
claudegel: Will replace with GFCI status, thanks

zhaquirks/sinope/thermostat.py
claudegel: I'll refactror to have those two at the end. For the other do you need a specific order?

zhaquirks/sinope/thermostat.py
claudegel: Thanks for the link, I'll use this NumberDeviceClass.TEMPERATURE

claudegel:
For the battery percentage remaining I need to check tonight
Same for the light.py .adds
I let you know

zhaquirks/sinope/thermostat.py
claudegel: About  entity_type=EntityType.CONFIG, can you tell me if it is the default for .enum, .switch, and .number Only .sensor need to be defined as the default should be EntityType.DIAGNOSTIC ?

claudegel:
I'm working on all your suggestions. Can you tell me if, for example I have a temperature attribute like min_temperature_limit which is expressed as temp * 100, 3 oC = 300 in the cluster/attribute
Is it possible to see the value in .number as 3 oC (converted from 300) but if I set it to 5 oC it send 500. I see that I can use attribute_converter but I don't know if it is a one way converter or a two way

zhaquirks/sinope/thermostat.py
TheJulianJES: Search the method you want the default for in this file: https://github.com/zigpy/zigpy/blob/dev/zigpy/quirks/v2/__init__.py (or just search for `entity_type`).
It's also *fine* to keep to explicitly set them, but it's still mostly unnecessary, so I'd avoid it.

(and I promise we'll have proper docs for this one day)

TheJulianJES:
> Can you tell me if, for example I have a temperature attribute like min_temperature_limit which is expressed as temp * 100, 3 oC = 300 in the cluster/attribute
Is it possible to see the value in .number as 3 oC (converted from 300) but if I set it to 5 oC it send 500. I see that I can use attribute_converter but I don't know if it is a one way converter or a two way

You can provide the `multiplier` argument for the `number` method, e.g. setting it to `100`. You can also use it as a divisor by providing `0.01`. I think that should do what you're asking?
`attribute_converter` is generally only needed for more complex operations and not implemented for the number platform, for example.

zhaquirks/sinope/switch.py
```diff
@@ -173,638 +310,622 @@ class SinopeManufacturerCluster(CustomCluster):
...
-                OUTPUT_CLUSTERS: [],
-            },
-        },
-    }
+        status: Final = ZCLAttributeDef(
+            id=0x0200, type=ValveStatus, access="r", is_manufacturer_specific=True
+        )
+        unit_of_measure: Final = ZCLAttributeDef(
+            id=0x0300, type=UnitOfMeasure, access="r", is_manufacturer_specific=True
+        )
```
TheJulianJES, line 486: You had that issue with the ZHA `instantaneous_demand` and `current_summ_delivered` sensor not being created, right? I think this might be the issue. You're overriding the ZCL `status` and `unit_of_measure` attributes here.

A device can have two attributes with the same ID if the manufacturer code differs.
So, by default, the `Metering` cluster has the `status` and `unit_of_measure` attributes with `manufacturer_code=None`.

After recent zigpy changes, it's also possible to define another attribute with the same ID as a ZCL one but a different `manufacturer_code` (`is_manufacturer_specific` just automatically determines the manufacturer code to use then).

However, here, the attribute name is the same. It's unclear what this should do. Should this replace the base ZCL attributes with a different datatype? That doesn't really work nicely.

Are these actually manufacturer specific attributes? If so, just rename them to not collide with the ZCL ones. E.g. `status_mf` and `unit_of_measure_mf`.

Is there some technical reference documentation available for this Sinope device?

TheJulianJES: Even with the existing v1 quirks, I think there's an issue here. The Sinope switches `SP2600ZB` + `SP2610ZB` use the `SinopeTechnologiesMeteringCluster` cluster with the weirdly redefined attributes.
The `status` attribute has `ValveStatus` as the type. That seems incorrect for a switch.

Home Assistant Core 2026.2.x (with 2026.3.x having improvements and releasing later today) essentially differentiate attributes using `(attribute ID, manufacturer code)` internally, instead of just the attribute ID.
So, for the valves, it would be really good to know if they need the manufacturer code set when reading/writing and getting attribute reports for them. If they do, they should just be renamed so that the names `status` and `unit_of_measure` are not used, since they're already used by the ZCL attributes.

But if it's only the data type that's changed and they should use no manufacturer code, that's really weird and unexpected. I hope that's not the case. If you have any technical docs for these devices, or can provide logs of you reading those two attributes when a quirk defines it with `is_manufacturer_specific=False`, with `is_manufacturer_specific=True`, and if there's no custom attribute definition at all on Core 2026.2.x, that would be amazing.

claudegel: Unfortunately there is no technical reference documentation available from Sinopé. So I'll rename the 0x0200 and 0x0300 as you describe. ValveStatus is because the valve device are included in the switch.py. ZHA should add valve platform as HA did. Then the valve will go out of switch.py to valve.py. Sinopé use the Metering cluster  attr 0x0200 to define the valve staus open, close, close with alarm system on.

There is another problem with the V2 quirks, If we pair a new Sinopé device directly with the V2 quirks. The device behave weirdly and do not function properly. The fix if to remove the device from ZHA, remove my V2 quirks from config/zhaquirks. Restart HA and repair the device using the V1 quirks. Then re add the V2 quirks and restart HA to have the device work perfectly in ZHA V2. Look like ZHA V2 do not detect correctly the manufacturer custom cluster. This is what preventing me to finalize the V2 quirks as if we merge we won't have any way to pair new devices to ZHA. Unless I find a way to make a dummy quirks that will help pairing new devices correctly.

claudegel: I can provide a cluster scan of the valve if you want

TheJulianJES: > There is another problem with the V2 quirks, If we pair a new Sinopé device directly with the V2 quirks. The device behave weirdly and do not function properly. 

Hmm, that's not good. It would be really cool to get the Sinope v2 quirks running and right into ZHA.
v1 and v2 quirks work very similar under-the-hood, so I wonder if there's an issue with some of the added custom entities (for which the attributes are auto-read during pairing), e.g. if some datatype is wrong.

If you have some time, these things would be really helpful):
- Diagnostics files for all possible states of the device / quirk variants. Can you download a diagnostics file from the HA device page for:
  - A device you paired with the built-in v1 quirk (fully working but no custom entities).
  - Switch to the v2 quirk with the device paired with the v1 quirk, and also provide those diagnostics.
  - And ideally also remove the device and pair it with the v2 quirk, so download diagnostics when it's in a broken state.
- Enable debug logs when pairing the device with the v2 quirk and post them here if they contain no sensitive information, or email them to me if you're not sure (email is on my GH profile).
- Home Assistant Core 2026.3.0b0 (beta) released today. It includes newer ZHA and zigpy versions. If you could grab diagnostics + logs + reproduce the issue on that version, it would be really nice. 2026.2.3 should also be fine though.
- Lastly, can you state what's broken with the device(s)? (e.g. is binding or attribute reporting for standard clusters maybe not set up properly, so changes from device to HA aren't reported correctly?)

claudegel: The main problem is with the reporting that do not work. So many data is missing. If somebody manually change setpoint on a thermostat it is not reported to ZHA. Only a restart will populate the data once.
I'll do my best to grab as much log I can.

TheJulianJES: Ok, that would be really helpful. The issue could be that something is wrong with the datatypes for the attributes. If there's a quirks v2 entity for an attribute, ZHA will try to read that attribute during pairing.
If it gets an unexpected datatype, I think initialization for the entire cluster, endpoint, or device does not continue (including binding + attribute reporting).

The easiest way to catch this is probably: Remove device from ZHA and install v2 quirk(s). Enable debug logs (via the UI in HA) and freshly pair the device with the v2 quirk, so it's "broken" afterwards. Then, send me the debug logs (via email). There's likely going to be some error in them. If not, all other steps (diagnostics + more logs, ...) from above will help. But the easiest thing to check first is debug logs when pairing via a v2 quirk.

PR #4044, opened 2025-05-03
Add dedicated quirk for Tuya 24GHz mmWave radar `_TZE204_gkfbdvyx`

Bruners:
## Proposed change

This commits adds support for Tuya compatible mmWave 24Ghz ceiling mounted human presence sensor. The unit seems to be in line with _TZE204_ya4ft0w so they could possible be merged if tested. This unit does not like sensitivity set to 0.

## Checklist

Related: #4072


fweffort:
Any chance we get something for this one as well? 
_TZE204_lbbg34rj

Bruners:
> Any chance we get something for this one as well? _TZE204_lbbg34rj

Would need some more information about the device for that to be possible, you could try to add the identifying string to any of the different models and test for yourself. 

Just add this to under any of the many tuy motion sensors in zhaquirks/tuya/tuya_motion.py
`.applies_to("_TZE204_lbbg34rj", "TS0601")`
and load the custom quirk file to be loaded in HA.

After adding the line in the custom quirk file, you need to restart HA and possibly remove and add the device again. The sensors are different in what data point they use so it is impossible to do without the device on hand.

Does it have any labels or anything on the "bottom" of the sensor?

caevv:
Can this be merged?

PersistentCloud:
I've tested this as a custom quirk in ZHA and it works fine. Can someone tell us, when it will be merged?

Bruners:
> Any chance we get something for this one as well? _TZE204_lbbg34rj

Not sure if you have figured out your device yet, but in z2m your device seems to be matched with _TZE204_ex3rcdha 

https://github.com/zigpy/zha-device-handlers/blob/500e74e16a655114b4983ea91df4be712a78d535/zhaquirks/tuya/tuya_motion.py#L1524

zhaquirks/tuya/tuya_motion.py
```diff
@@ -85,6 +85,14 @@ class TuyaPresenceStateV02(t.enum8):
     Gigantic = 0x05
 
 
+class TuyaPresenceStateV03(t.enum8):
+    """Tuya presence state enum, variation 03."""
+
+    Unoccupied = 0x00
+    Presence = 0x01
+    Move = 0x02
```
TheJulianJES: Can we call this `Movement`?

```suggestion
    Movement = 0x02
```

zhaquirks/tuya/tuya_motion.py
```diff
@@ -1611,3 +1618,112 @@ class TuyaMotionDetectionMode(t.enum8):
...
+    # 1: presence="presence" state=true triggers only with presence sensitivity on low
+    # 2: presence="move" state=true
+    .tuya_enum(
+        dp_id=1,
+        attribute_name="presence",
+        enum_class=TuyaPresenceStateV03,
+        entity_platform=EntityPlatform.SENSOR,
+        entity_type=EntityType.STANDARD,
+        translation_key="presence_state",
+        fallback_name="Presence state",
```
TheJulianJES, line 1639: Hmm, we have multiple other sensors calling "this" `human_motion_state`/"Human motion state" and one `motion_state`/"Motion state". `presence_state` is probably more accurate though.
The `attribute_name` is what matters for unique ID generation in HA, but since this is the only sensor where `dp_id=1` is for thus presence/motion state, we can keep this for now.

We may wanna streamline this at some point though. Presence state is probably the most accurate for this though.

zhaquirks/tuya/tuya_motion.py
```diff
@@ -1611,3 +1618,112 @@ class TuyaMotionDetectionMode(t.enum8):
...
+        attribute_name="distance",
+        type=t.uint16_t,
+        state_class=SensorStateClass.MEASUREMENT,
+        device_class=SensorDeviceClass.DISTANCE,
+        unit=UnitOfLength.METERS,
+        multiplier=0.1,
+        entity_type=EntityType.STANDARD,
+        translation_key="target_distance",
+        fallback_name="Target distance",
+    )
```
TheJulianJES, line 1695: Would it make sense to clone `base_tuya_motion.clone()` for DP IDs 2, 3, 4, and 9? Or are the values too different? The values for `base_tuya_motion` for reference: https://github.com/zigpy/zha-device-handlers/blob/500e74e16a655114b4983ea91df4be712a78d535/zhaquirks/tuya/tuya_motion.py#L182-L234

zhaquirks/tuya/tuya_motion.py
```diff
@@ -1611,3 +1618,112 @@ class TuyaMotionDetectionMode(t.enum8):
...
+        min_value=1,
+        max_value=10,
+        step=1,
+        translation_key="presence_sensitivity",
+        fallback_name="Presence sensitivity",
+    )
+    .tuya_illuminance(dp_id=103)
+    .tuya_number(
+        dp_id=105,
+        attribute_name="presence_timeout",
```
TheJulianJES: Unfortunately, all the Tuya motion sensors all kind of do some things differently. Previously, we used `presence_timeout` for some sensors, but later unified them all under `fading_time`/"Fading time".
However, the `attribute_name` needed to stay the same, so some sensors kept `presence_timeout`. Otherwise, you'd get a new HA entity.

If we're adding a new sensor, we should use `fading_time` for the `attribute_name` as well (as we're doing with the translation key and fallback name).
This is also something we're looking to improve in the future, as I personally prefer "Presence timeout" over "Fading time", but let's use "Fading time"/`fading_time` for everything, for now, to stay consistent.

TheJulianJES (commented):
Ok, finally got to this. This looks good for the most part. Three small things:
- `move` -> `movement`
- `presence_timeout` -> `fading_time` (for consistency) 
- and maybe consider if we wanna clone `base_tuya_motion` for the first DPs: https://github.com/zigpy/zha-device-handlers/pull/4044/files#r2408042537. Not a must, as the values don't really match, just a suggestion.

zhaquirks/tuya/tuya_motion.py
Bruners: This sensor model (i have two) does not like sensitivity set to 0, it seems to crash and stops responding until it is reset and paired again. i don't know how much the different steps and values do to the sensor or if they work at all in between values but this allegedly what the Tuya app says for this sensor. 

My best guess is to not copy the base and i don't see a reason to do it unless it can be verified that it is actually working correctly with those numbers. In the long run with more sensors with different capabilites coming all the time, it is not better for readability or code maintenance to continue to use that base.

Bruners:
I’ve finally acquired a Tuya Zigbee hub to make development a bit easier.
However, I noticed that dp 104 is documented by Tuya as an enum named man_state. In our quirk, dp 104 is currently mapped to OccupancySensing.occupancy using x == 1:

```
.tuya_dp(
    dp_id=104,
    ep_attribute=TuyaOccupancySensing.ep_attribute,
    attribute_name=OccupancySensing.AttributeDefs.occupancy.name,
    converter=lambda x: x == 1,
)
```

Tuya defines the following values for man_state:

```
class TuyaManState(t.enum8):
    """Tuya TuyaManState"""
    nobody = 0x00
    exist = 0x01
    nobody5min = 0x02
    nobody10min = 0x03
    nobody15min = 0x04
    nobody30min = 0x05
    nobody60min = 0x06
    exist5min = 0x07
    exist10min = 0x08
    exist15min = 0x09
    exist30min = 0x0A
    exist60min = 0x0B
```

Quick change could be to change the converter

```
converter=lambda v: 1 if v in {0x01, 0x07, 0x08, 0x09, 0x0A, 0x0B} else 0,
```

Not sure what theese is to be used for in the real world or if it should be corrected to enum?
If to be implemented as enum what attributes/keys/names should be used?

Bruners:
```    
.tuya_enum(
    # State_None or Exist
    dp_id=104,
    attribute_name="human_motion_state",
    enum_class=TuyaHumanMotionStateV3,
    entity_platform=EntityPlatform.SENSOR,
    translation_key="human_motion_state",
    fallback_name="Human motion state",
)
```

mots:
> I’ve finally acquired a Tuya Zigbee hub to make development a bit easier. However, I noticed that dp 104 is documented by Tuya as an enum named man_state. In our quirk, dp 104 is currently mapped to OccupancySensing.occupancy using x == 1:
> 
> ```
> .tuya_dp(
>     dp_id=104,
>     ep_attribute=TuyaOccupancySensing.ep_attribute,
>     attribute_name=OccupancySensing.AttributeDefs.occupancy.name,
>     converter=lambda x: x == 1,
> )
> ```
> 
> Tuya defines the following values for man_state:
> 
> ```
> class TuyaManState(t.enum8):
>     """Tuya TuyaManState"""
>     nobody = 0x00
>     exist = 0x01
>     nobody5min = 0x02
>     nobody10min = 0x03
>     nobody15min = 0x04
>     nobody30min = 0x05
>     nobody60min = 0x06
>     exist5min = 0x07
>     exist10min = 0x08
>     exist15min = 0x09
>     exist30min = 0x0A
>     exist60min = 0x0B
> ```
> 
> Quick change could be to change the converter
> 
> ```
> converter=lambda v: 1 if v in {0x01, 0x07, 0x08, 0x09, 0x0A, 0x0B} else 0,
> ```
> 
I think this explains the behavior I've been seeing while using this as custom quirk. The occupancy sensor clears while the enum state string keeps working - see here: 
unnamed

I'll try checking against that array and see if that fixes things - for now I've just been using a template sensor against the string in HA.

Leo1906:
> I've tested this as a custom quirk in ZHA and it works fine. Can someone tell us, when it will be merged?

You said you got this working as a custom quirk - I am having a hard time trying to create a custom quirck out of this PR myself on the latest stable HA version.
If you still got the code around, would you mind sharing it please? :)
So I can use the device already until this PR gets hopefully merged :)
Thank you all for contributing to this project!

PR #4047, opened 2025-05-06
Return datetimes for Tuya timestamp sensors

puddly:
## Proposed change

Currently, we perform an unnecessary round-trip conversion of timestamps from integers, to datetimes, back to integers (to pass to ZHA), and then back to datetimes in ZHA for formatting. This PR modifies the few quirks v2 entities to use datetime objects directly.

Unrelated, but I noticed that the current code did not actually work as intended because `datetime.replace` is not an in-place method. Thus, we currently just return the local time directly. This PR fixes this behavior.

In the end, it will allow us to remove this workaround: https://github.com/zigpy/zha/blob/ecb2b9b85999bb76d852fa1177cbaa934e9eae92/zha/application/discovery.py#L218-L220

## Additional information

It requires a corresponding PR to ZHA. (insert link here)

## Checklist

### DIFF
--- a/tests/test_tuya_valve.py
+++ b/tests/test_tuya_valve.py
@@ -181,7 +181,7 @@ async def test_giex_02_quirk(zigpy_device_from_v2_quirk, model, manuf, use_minut
 async def test_giex_functions():
     """Test various Giex Valve functions."""
     assert zhaquirks.tuya.tuya_valve.giex_string_to_td("12:01:05,3") == 43265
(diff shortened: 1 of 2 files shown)


TheJulianJES (approved):
Quirk changes look good

puddly:
I think this is ready to go, we can migrate the ZHA PR commits to the library bump.

TheJulianJES:
Nice! Wasn't sure as I hadn't seen the ZHA PR/code, but good if that's ready as well.

PR #4056, opened 2025-05-09
Support Frient devices

puddly:
## Proposed change

This adds quirks for the vast majority of Frient devices and migrates the existing quirks to v2 in the process.

## Checklist


zhaquirks/develco/air_quality.py
```diff
@@ -1,241 +1,90 @@
...
+    .replaces(DevelcoPowerConfiguration, endpoint_id=38)
+    .sensor(
+        attribute_name=DevelcoVOCMeasurement.AttributeDefs.measured_value.name,
+        cluster_id=DevelcoVOCMeasurement.cluster_id,
+        endpoint_id=38,
+        device_class=SensorDeviceClass.VOLATILE_ORGANIC_COMPOUNDS,
+        state_class=SensorStateClass.MEASUREMENT,
+        unit=CONCENTRATION_PARTS_PER_BILLION,
+        fallback_name="VOC Level",
+        unique_id_suffix="voc_level",
```
TheJulianJES: I'm aware the fallback name isn't used in HA and no translation key is specified (meaning we use the device class name: "Volatile organic compounds"), but maybe we should still rename that to "VOC level" to match HA conventions.
```suggestion
        fallback_name="VOC level",
        unique_id_suffix="voc_level",
```
The IKEA VINDSTYRKA quirk uses "VOC index" and "voc_index", but it's intentionally using `SensorDeviceClass.AQI`, as its measurement has no unit. We could still use "index" in the name here, too. Doesn't really matter.

zhaquirks/develco/intelligent_keypad.py
```diff
@@ -0,0 +1,39 @@
...
+        ),
+    )
+    .binary_sensor(
+        endpoint_id=44,
+        cluster_id=IasZone.cluster_id,
+        attribute_name=IasZone.AttributeDefs.zone_status.name,
+        device_class=BinarySensorDeviceClass.TAMPER,
+        attribute_converter=lambda value: bool(value & IasZone.ZoneStatus.Tamper),
+        unique_id_suffix="tamper",
+        translation_key="tamper",
```
TheJulianJES: The `translation_key` should be removed here, as we want to use the device class name: "Tamper".
```suggestion
```
(The `fallback_name` always needs to be provided, but won’t be used by HA in this case.)

zhaquirks/develco/smart_siren.py
```diff
@@ -0,0 +1,58 @@
...
+    .applies_to("frient A/S", "SIRZB-110")
+    # Create a tamper sensor
+    .binary_sensor(
+        endpoint_id=43,
+        cluster_id=IasZone.cluster_id,
+        attribute_name=IasZone.AttributeDefs.zone_status.name,
+        device_class=BinarySensorDeviceClass.TAMPER,
+        attribute_converter=lambda value: bool(value & IasZone.ZoneStatus.Tamper),
+        unique_id_suffix="tamper",
+        translation_key="tamper",
```
TheJulianJES: The `translation_key` should be removed here, as we want to use the device class name: "Tamper".
```suggestion
```
(The `fallback_name` always needs to be provided, but won’t be used by HA in this case.)

In other quirks, we also often set `entity_type=EntityType.STANDARD,` for a tamper sensor. But since really seems to be a secondary function here, diagnostic is probably just fine.

TheJulianJES (commented):
Just did a quick review and it seems good. Only a couple of minor comments regarding the removal of the tamper translation key to use the HA device class name.

zhaquirks/develco/io_module.py
```diff
@@ -0,0 +1,53 @@
...
+        unique_id_suffix="117-6",
+        new_fallback_name="COM 2",
+        new_translation_key="frient_com2",
+    )
+    # And the two inputs
+    .change_entity_metadata(
+        endpoint_id=112,
+        cluster_id=BinaryInput.cluster_id,
+        unique_id_suffix="112-15",
+        new_fallback_name="IN1",
```
TheJulianJES, line 28: Is this supposed to be "IN1" or should it be "IN 1 (and so on for the others)?
(since it's "COM 1" and "COM 2")

zhaquirks/develco/io_module.py
```diff
@@ -0,0 +1,53 @@
...
+        new_fallback_name="COM 1",
+        new_translation_key="frient_com1",
+    )
+    .change_entity_metadata(
+        endpoint_id=117,
+        cluster_id=OnOff.cluster_id,
+        unique_id_suffix="117-6",
+        new_fallback_name="COM 2",
+        new_translation_key="frient_com2",
+    )
```
TheJulianJES: General comment regarding this (I'll follow-up somewhere else soon):
Now, it's not trivial to automatically generate `strings.json` for HA with the `new_fallback_name` and `new_translation_key` (if populated), as we don't know which platform the entities are for.

We may want to include the `Platform` enum (switch, light, ...) in `ChangedEntityMetadata`, but I'll check again soon. Just looking at this, it could be a switch or a binary sensor, for example. It would also make that more clear I guess.

zhaquirks/develco/io_module.py
```diff
@@ -0,0 +1,53 @@
...
+    # And the two inputs
+    .change_entity_metadata(
+        endpoint_id=112,
+        cluster_id=BinaryInput.cluster_id,
+        unique_id_suffix="112-15",
+        new_fallback_name="IN1",
+        new_translation_key="frient_in1",
```
TheJulianJES: I'd also prefer using `frient_in_1` etc. (so, a `_` before the suffix). That's what we seem to do for the few (Tuya) devices that use translation keys / names in a similar fashion.

zhaquirks/develco/air_quality.py
puddly: Sure! I think we should invest in custom pylint code to enforce these rules via CI.

puddly: Or maybe just have zigpy be strict with it and throw an error?

zhaquirks/develco/io_module.py
puddly: It's intentional, the device itself has labels for these pins and uses this exact format: `IN1` and `COM 1`.

zhaquirks/develco/io_module.py
puddly: Can we have ZHA itself generate these translation strings with a CLI tool? We should enforce all new quirk contributions to attach device diagnostics so that we can do it automatically.

zhaquirks/develco/air_quality.py
TheJulianJES: I think having zigpy throw an error might be too strict, since there are some times where we don't want to enforce this (due to abbreviations or special/custom names).
We could also just add a test that goes through all `EntityMetadata` and the `ChangedEntityMetadata` and verifies the `fallback_name`s, but it would still easily allow us to opt-out certain names. I'll get back to this soon.

puddly: Right. I *think* we can do it with https://pylint.pycqa.org/en/latest/development_guide/how_tos/custom_checkers.html, allowing quirks in this repo to opt out on a case-by-case basis with a narrow `# noqa` directive.

zhaquirks/develco/io_module.py
TheJulianJES: Yeah, I'll try and do this soon. For now, I've just been using this: [home-assistant/core@`dev...TheJulianJES:core:tjj/zha_translation_key_generation`#diff-5d97af3392](https://github.com/home-assistant/core/compare/dev...TheJulianJES:core:tjj/zha_translation_key_generation#diff-5d97af3392f844cfcf573642b8ae8a0cc9d605b107865aafea51083d089c7ba0R13-R37) 😅 

It's very rudimentary and only for quirks v2 strings, except for the new `ChangedEntityMetadata`, since we need way more info than just the metadata for getting the correct platform.

I can also include a section for the device diagnostics in the quirks PR template soon.

TheJulianJES (approved):
nice work on this as well 😄

PR #220, opened 2019-12-22
Add lumi.plug - US version

dmulcahey:
Fixes #218

PR #4057, opened 2025-05-09
Add Tuya/Zemismart 3 phase power meter variants

abmantis:
## Proposed change

Add Zemismart SPM02 3 phase power meter.

Product link: https://www.zemismart.com/products/spm02-d2tz
Spec in Z2M: https://github.com/Koenkk/zigbee-herdsman-converters/blob/da540f0285a6d4fe6ec09e6a77024db3dcf534b1/src/devices/tuya.ts#L11389

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_power.py
+++ b/zhaquirks/tuya/ts0601_power.py
@@ -314,7 +314,7 @@ class Tuya3PhaseElectricalMeasurement(ElectricalMeasurement, TuyaLocalCluster):
         dp_id=110,
         attribute_name="energy_produced_ph_a",
         type=t.uint32_t,
-        converter=lambda x: x / 100,
+        divisor=100,
         state_class=SensorStateClass.TOTAL_INCREASING,
         device_class=SensorDeviceClass.ENERGY,
         unit=UnitOfEnergy.KILO_WATT_HOUR,
--- a/zhaquirks/tuya/ts0601_power.py
+++ b/zhaquirks/tuya/ts0601_power.py
@@ -2,11 +2,13 @@
 
 from collections.abc import ByteString
 
-from zigpy.quirks.v2 import SensorDeviceClass, SensorStateClass
+from zigpy.quirks.v2 import EntityType, SensorDeviceClass, SensorStateClass
 from zigpy.quirks.v2.homeassistant import (
+    PERCENTAGE,
     UnitOfElectricCurrent,
     UnitOfEnergy,
     UnitOfPower,
+    UnitOfTime,
 )
 import zigpy.types as t
 from zigpy.zcl.clusters.general import LevelControl, OnOff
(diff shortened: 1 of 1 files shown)


ajbathe:
I see some problem with "active power":

With this quirk, I get an unknown entity "Scheinleistung" (apparent power), but I guess this meaning is "total power". I get power of all three phases but not the sum of them (as with my quirk).

In line 317, shouldn't be 'converter=lambda x: x / 100' just 'divisor=100'? Is this converter needed here?
[zemismart_spm02-d2tz.py.gz](https://github.com/user-attachments/files/20272921/zemismart_spm02-d2tz.py.gz)

abmantis:
> I see some problem with "active power":
> 
> With this quirk, I get an unknown entity "Scheinleistung" (apparent power), but I guess this meaning is "total power". I get power of all three phases but not the sum of them (as with my quirk).
> 
> In line 317, shouldn't be 'converter=lambda x: x / 100' just 'divisor=100'? Is this converter needed here? [zemismart_spm02-d2tz.py.gz](https://github.com/user-attachments/files/20272921/zemismart_spm02-d2tz.py.gz)

Apparent Power is added automatically for any device that has the Electrical Measurement cluster. 

Regarding the sum, I am not sure if we should add "virtual" attributes, since the user can use an helper for that. If we should, then I can add it though.

> In line 317, shouldn't be 'converter=lambda x: x / 100' just 'divisor=100'? Is this converter needed here?

Yeah, probably bad copy-paste.

TheJulianJES (approved):
Looks good, thanks!

TheJulianJES:
> Apparent Power is added automatically for any device that has the Electrical Measurement cluster.

Yeah, this is something that shouldn't happen I think. Maybe create an issue for that on the ZHA repo.

TheJulianJES:
> Regarding the sum, I am not sure if we should add "virtual" attributes, since the user can use an helper for that. If we should, then I can add it though.

It might be fine to add for this quirk, since that's easy enough to do. But generally, we don't add "custom entities" in quirks that we completely create ourselves (so ones that are not offered by the device at all).
A template helper entity can always be used in HA for these.

But for a simple sum, I think it's probably fine to add to this quirk (in a follow-up PR) if you really want to have it.

abmantis:
I just noticed that I misinterpreted the comment regarding totals, because the quirk already has a DP attribute for total power.
It does not have an entity because there is no mapping for `total_active_power`.
I'm considering adding it in ZHA  with `_skip_creation_if_no_attr_cache = True`. Or should I add a `tuya_sensor` instead?

TheJulianJES:
Hmm, not quite sure. We should really have a sensor for `total_active_power` in ZHA though...
We will have to remove this though: https://github.com/zigpy/zha-device-handlers/blob/5f1f14b95d352ff1cf0cbd34ba86c496f259d2f5/zhaquirks/xiaomi/__init__.py#L305-L309

That's the wrong attribute being updated (with the sum of all past consumption in Wh). Has been in place since https://github.com/zigpy/zha-device-handlers/pull/220 apparently, but it shouldn't be used anywhere.

We could also remap the attribute to the EM cluster for the sum, but I think adding a sensor in ZHA might be the best option. We should really support that attribute natively.

abmantis:
> Hmm, not quite sure. We should really have a sensor for `total_active_power` in ZHA though... We will have to remove this though:
> 
> https://github.com/zigpy/zha-device-handlers/blob/5f1f14b95d352ff1cf0cbd34ba86c496f259d2f5/zhaquirks/xiaomi/__init__.py#L305-L309
> 
> That's the wrong attribute being updated (with the sum of all past consumption in Wh). Has been in place since #220 apparently, but it shouldn't be used anywhere.
> 
> We could also remap the attribute to the EM cluster for the sum, but I think adding a sensor in ZHA might be the best option. We should really support that attribute natively.

New entity added in https://github.com/zigpy/zha/pull/458 
Xiaomi clean up in https://github.com/zigpy/zha-device-handlers/pull/4095

PR #4060, opened 2025-05-10
Add Tuya temperature sensor `_TZE284_yjjdcqsq` variant

Axeman72:
## Proposed change
  Added support for a new TS601 Temperature / humidity sensor with different signature

## Additional information
I buy the same sensors from the same vendor. Last batch (same shape, same part number, same box) arrived with the _TZE284_yjjdcqsq instead of _TZE204_yjjdcqsq or _TZE200_yjjdcqsq.
I changed this file on my Home Assistant staging environment to confirm that the sensor works like the previous ones by adding only the "applies_to" line, and it works. 

## Checklist


TheJulianJES (approved):
Thanks!

PR #4080, opened 2025-05-20
Change name of requirements file in readme

girardof:
## Proposed change
The file requirements_test_all.txt was renamed into requirements_test.txt, this commit change the name of the file in the README file.

## Checklist


puddly (approved):

TheJulianJES (approved):
Thanks!
Ideally, you'd just use `script/setup` (which also uses `uv` instead of `pip`), but this is already an improvement over the existing docs.

In general, the README is pretty outdated for the most part and really needs to be updated. It's on my list to eventually do, but we'll see when that is..

PR #4081, opened 2025-05-20
Add support for Schneider electric FLS/AIRLINK/4

girardof:
## Proposed change

Add support for [Schneider electric FLS/AIRLINK/4](https://github.com/zigpy/zha-device-handlers/wiki/Schneider-Electric#flsairlink4-)

## Additional information

[ZB Spec - Wireless Switch - 110422.pdf](https://github.com/user-attachments/files/20350200/ZB.Spec.-.Wireless.Switch.-.110422.pdf)

This is my first time working with this library, and I have a few questions. If someone with more experience could help me, I would be very grateful:

- I'm unsure how to test the changes. I have a Home Assistant instance running and one FLS/AIRLINK/4 switch, so I could in theory edit the code directly within my Home Assistant setup. But, if there’s a better approach, I’m open to suggestions.

- The following cluster IDs aren't implemented:
  - 0x0003
  - 0x0004
  - 0x0005
  - 0x0006
  - 0x0008
  - 0x0019
  - 0x0102

Should they be implemented as well?

## Checklist


TheJulianJES:
> I'm unsure how to test the changes. I have a Home Assistant instance running and one FLS/AIRLINK/4 switch, so I could in theory edit the code directly within my Home Assistant setup. But, if there’s a better approach, I’m open to suggestions.

Search "custom quirks zha". You can set up a folder and just drop this modified v2 quirk file in that. Those custom quirks will be preferred over the built-in ones.

> The following cluster IDs aren't implemented: [...] Should they be implemented as well?

Those are all standard ZCL (Zigbee Cluster Library) clusters. They are already defined in zigpy and all entities should work as expected. Not quite sure what you mean with "implemented" here.
If there are custom attributes present on those clusters as well, you can create a class inheriting from both the original cluster (e.g. `WindowCovering` and `CustomCluster`), then use `replaces` as well.

SE seems to offer additional attributes on these default clusters, since other quirks also replace them. E.g. there are `SEBasic` and `SEOnOff` clusters here: https://github.com/zigpy/zha-device-handlers/blob/81108c4af789a20d4c45d6c51757c71260a12229/zhaquirks/schneiderelectric/__init__.py#L18-L132
You can also add the `replaces` for them to your (custom) quirk, then try the "Manage Zigbee device" menu to read the custom attributes and see if they're available for your device as well.

girardof:
Hi @TheJulianJES, thank you for taking the time to write your answer. I managed to use a custom quirks on my installation of HA. As you mentioned `SEOnOff` and `SEBasic` are needed for the quirk to work so I added them in a new commit.

After the changes implemented in the second commit the buttons are visible and using the physical switch toggle the controls in HA

Screenshot 2025-05-24 at 23 25 29

girardof:
https://github.com/zigpy/zha-device-handlers/pull/4130 should be merged first, then SESpecific can be renamed into SESwitchConfiguration in the PR

girardof:
Unfortunately this PR is not ready to be merge, I can see device in home assistant and it's state (on/off). When pressing physically on the device the state in Home assistant change but when triggering the button from assistant I encounter the following issue: 

```
Failed to perform the action switch/turn_on. Failed to turn on: <Status.UNSUPPORTED_CLUSTER: 195>
```

I'm still trying to find a solution; if anyone has any ideas about what might be causing the issue, I'm all ears.

PR #4084, opened 2025-05-21
Use manufacturer ID override for Sonoff ZBminiR2

AnatolyRugalev:
This PR updates zbminir2 to use new manufacturer id override. Cleaning things up before adding missing attributes to this quirk.


TheJulianJES (approved):
Thanks!

Just to confirm, you did test this with a physical device and it still works as expected?

AnatolyRugalev:
> Just to confirm, you did test this with a physical device and it still works as expected?

Yes

PR #4086, opened 2025-05-23
Add Senoro door sensor `_TZE200_ytx9fudw`

ChristianGr1974:
## Proposed change
Support for the window/door contact from Senoro.

## Additional information
Currently not supported.
Tested with hardware device as external quirk.

## Checklist


zhaquirks/senoro/senorowin.py
```diff
@@ -0,0 +1,43 @@
...
+        fallback_name="Opening state",
+        entity_type=EntityType.STANDARD,
+        entity_platform=EntityPlatform.SENSOR,
+    )
+    .tuya_switch(
+        dp_id=16,
+        attribute_name="alarm",
+        entity_type=EntityType.STANDARD,
+        translation_key="alarm",
+        fallback_name="Tamper Alarm",
```
TheJulianJES: To follow HA naming conventions, only the first word can start with an uppercase character (excluding abbreviations). So, this should be "Tamper alarm".

```suggestion
        fallback_name="Tamper alarm",
```

TheJulianJES: Since the "Tamper alarm" is a switch (and not a binary sensor), what's the difference in device behavior when the switch is on or off?

zhaquirks/senoro/senorowin.py
```diff
@@ -0,0 +1,43 @@
...
+        enum_class=OpeningStateEnum,
+        translation_key="opening_state",
+        fallback_name="Opening state",
+        entity_type=EntityType.STANDARD,
+        entity_platform=EntityPlatform.SENSOR,
+    )
+    .tuya_switch(
+        dp_id=16,
+        attribute_name="alarm",
+        entity_type=EntityType.STANDARD,
```
TheJulianJES: Also, should the "Tamper alarm" not be a configuration entity (default for quirks v2 switches) instead?

zhaquirks/senoro/senorowin.py
```diff
@@ -0,0 +1,43 @@
...
+from zhaquirks.tuya import BatterySize
+from zhaquirks.tuya.builder import TuyaQuirkBuilder
+
+
+class OpeningStateEnum(t.enum8):
+    """Enum for opening state."""
+
+    open = 0
+    closed = 1
+    tilted = 2
```
TheJulianJES: Let's make these uppercase to follow other enums:

```suggestion
    Open = 0
    Closed = 1
    Tilted = 2
```

Due to some ZHA limitations, these are not translateable for now, but we'll work on changing that in the future.
Related PR: https://github.com/zigpy/zha/pull/86

zhaquirks/senoro/__init__.py
```diff
@@ -0,0 +1 @@
+"""Module for Senoro quirks implementations."""
```
TheJulianJES: Since this is a (white-label) Tuya device, let's remove this and put the quirk file (named something like  `tuya_door.py`) in the `tuya` folder.

zhaquirks/senoro/senorowin.py
```diff
@@ -0,0 +1,43 @@
...
+        dp_id=2,
+        battery_type=BatterySize.CR2032,
+        battery_qty=3,
+    )
+    .tuya_enum(
+        dp_id=101,
+        attribute_name="opening_state",
+        enum_class=OpeningStateEnum,
+        translation_key="opening_state",
+        fallback_name="Opening state",
```
TheJulianJES: Let's move `translation_key` and `fallback_name` parameters to the bottom, to follow the style with how other quirks v2 entities are defined.

TheJulianJES: Also, should we only call this "Opening" or "Door" (removing the "state" part) to better match with other entity names in HA?

TheJulianJES (commented):
Thanks for the PR. I've added some comments.

zhaquirks/senoro/senorowin.py
TheJulianJES: If this is supposed to be a binary sensor, we should also set:
```python
        device_class=BinarySensorDeviceClass.TAMPER,
        fallback_name="Tamper",
```
No translation key is required then, as the default device class name will be used, which is "Tamper".
The fallback name will not be used, but still needs to be present.

zhaquirks/senoro/senorowin.py
ChristianGr1974: Senoro.Win is a window/door contact that can report the open/closed/tilt status (low-power device). At the same time, it offers burglary protection in the form of an alarm. This is sent via Tuya data point 16 (boolean). If the alarm is triggered, the device switches to a normal device and you can write _boolean_ _false_ to data point 16 to deactivate the alarm again.

zhaquirks/senoro/senorowin.py
ChristianGr1974: Honestly I'm not very familiar with the ZHA integration, I just started with it and tried to integrate this device. 
I did this by trial and error with a custom quirk under _/config/custom_zha_quirks_ in Home-Assistant. The only important thing here is that you need to be able to read and write boolean on Tuya datapoint 16.

zhaquirks/senoro/senorowin.py
ChristianGr1974: Done

zhaquirks/senoro/senorowin.py
ChristianGr1974: Done

zhaquirks/senoro/__init__.py
ChristianGr1974: Done

zhaquirks/senoro/senorowin.py
ChristianGr1974: Done

ChristianGr1974: Done

zhaquirks/senoro/senorowin.py
TheJulianJES: Ah, okay. That's interesting. I guess this is fine then.

Just curious, what do you mean by "If the alarm is triggered, the device switches to a normal device"?
Like, the sensor won't make any sounds or lights when the alarm is on, right?

zhaquirks/senoro/senorowin.py
TheJulianJES: Yeah, all good. I've never come across something where you can turn on/off the tamper alarm yourself.

I think a switch for this should be fine then. No device class that's really applicable then. We can maybe set some alarm icon (which could be done [here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/icons.json) after a quirks bump PR with the quirk is merged to HA).

zhaquirks/senoro/senorowin.py
ChristianGr1974: You can see the device here: www.senoro.de
It is primarily a door/window sensor that sends the Open/Close/Tilt statuses. The device is battery operated. If a burglar shakes the window or tries to force it open, the device registers this by vibration. The device then makes a very high-frequency sound via a siren (like a smoke detector). From this moment on, it switches to a different mode and allows writing to data point 16 (not possible before) in order to switch off the siren (the alarm) again.

zhaquirks/senoro/senorowin.py
ChristianGr1974: Sounds good!

zhaquirks/tuya/tuya_door.py
```diff
@@ -0,0 +1,43 @@
...
+        translation_key="opening",
+        fallback_name="Opening",
+    )
+    .tuya_switch(
+        dp_id=16,
+        attribute_name="alarm",
+        entity_type=EntityType.STANDARD,
+        translation_key="alarm",
+        fallback_name="Tamper alarm",
+    )
```
TheJulianJES: Since this switch cannot be turned on (when it's off) to turn on the alarm manually, we should think about doing the following instead:

Have a binary sensor entity (for dp 16) (with `BinarySensorDeviceClass.TAMPER` to show nicely in HA) and also provide a "write attribute button" to reset the dp back to `0`/clear.
Implementing the button might require us to map the dp with `tuya_dp` and just use normal `binary_sensor` and `write_attr_button` (instead of the Tuya variants), or just use the mapped dp from `tuya_binary_sensor` then. That might work as well, but I'll take another look at this, since it's not exactly easy.

I think the approach with a binary sensor and button to reset might be a bit nicer than a switch entity though (that you cannot turn on).

I'll come back to this, but the beta cutoff for 2025.6.0 is already tomorrow, so we likely won't make that, as there are still some other things I have to take a look at.

ChristianGr1974: Have you had time to think about this?

senorosoftware: @TheJulianJES fyi

TheJulianJES: Can we remove the "Tamper alarm" from this initial PR? We can hopefully get the other stuff into 2025.7.0 then.
I've not had much time yet, unfortunately. You can immediately create another PR with the same "Tamper alarm" switch, so it's not forgotten about.

When I get to it, I can try to push/comment the button/binary sensor alternative. But if you can get that working, that'd be great too.

senorosoftware: @TheJulianJES 
ok, so I removed the tamper alarm switch. But currently, if the alarm turns on, you can't turn it off. The device activates the alarm for three minutes, then it turns off, or you have to remove the batteries...

TheJulianJES: Ok, I'll come back to you with the button and binary sensor entity for turning off the tamper alarm.

TheJulianJES (approved):
Again, I'll come back to you regarding the binary sensor and button for the tamper alarm.
But for now, this looks good.

felixmueller:
@TheJulianJES, @ChristianGr1974 So HA 2025.7 will contain the first Sonoro support for ZHA without a possibility to disable the alarm?
Any ETA for this feature to arrive?
Thanks!

TheJulianJES:
For anyone who has the sensor, please test this PR and let me know if it works as expected:
- https://github.com/zigpy/zha-device-handlers/pull/4155

The PR adds a binary sensor entity for viewing the current alarm state and a button dedicated to resetting the alarm state.

PR #4088, opened 2025-05-24
Bump zigpy to 0.80.0

TheJulianJES:
## Proposed change
Bump zigpy to [0.80.0](https://github.com/zigpy/zigpy/releases/tag/0.80.0) and adjust a bunch of quirks code to work properly.

## Additional information
0.79.0 includes `suggested_display_precision`, which we need for some v2 quirks.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ readme = "README.md"
 license = {text = "Apache-2.0"}
 requires-python = ">=3.12"
 dependencies = [
-    "zigpy>=0.76.0",
+    "zigpy>=0.80.0",
 ]
 
 [tool.setuptools.packages.find]
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -13,5 +13,5 @@ pytest-sugar
 pytest-timeout
 pytest-asyncio
 pytest>=7.1.3
-zigpy>=0.76.0
+zigpy>=0.80.0
 ruff==0.0.261
--- a/tests/test_tuya_clusters.py
+++ b/tests/test_tuya_clusters.py
@@ -252,7 +252,7 @@ def test_tuya_cluster_request(
     """Test cluster specific request."""
 
     hdr = zcl_f.ZCLHeader.general(1, cmd_id, direction=zcl_f.Direction.Server_to_Client)
-    hdr.frame_control = hdr.frame_control.replace(disable_default_response=True)
+    hdr.frame_control = hdr.frame_control.replace(disable_default_response=False)
 
     with mock.patch.object(TuyaCluster, handler_name) as handler:
         handler.return_value = mock.sentinel.status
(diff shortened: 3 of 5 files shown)


TheJulianJES:
Tuya quirk tests are broken by this PR:
- https://github.com/zigpy/zigpy/pull/1596

As mentioned, it's a breaking change 😄 
I wonder if we actually broke some Tuya quirks since HA 2025.5.0.. But it seems to be limited to the tests.
I'll take a look now.

TheJulianJES:
@puddly If you wanna have a look at this, I'd appreciate that. Otherwise, I can try to take another shot at this tomorrow, maybe.

From a brief look, one thing that's broken seems to be (the constructor in) `TuyaData` due to the new deserialization(?):
https://github.com/zigpy/zha-device-handlers/blob/f9898ee812046cb26c11995347d82f059f16bfb3/zhaquirks/tuya/__init__.py#L151-L234

And well, we obviously can't/shouldn't modify the existing object like this anymore.

TheJulianJES:
Ugh, the main issue with the Tuya stuff was this: [`6ce82c9` (#4088)](https://github.com/zigpy/zha-device-handlers/pull/4088/commits/6ce82c90f7b7392fffffcf2aebc333b1b463cf72)

Since we're creating a new object (in the new deserialization) and need to pass the args here, this obviously removed them when creating a new one. I looked over this like ten times and was wondering why it didn't work, especially since most changes are for the `IntStruct`.

Me accidentally switching the `disable_default_response` in [`72aa24a` (#4088)](https://github.com/zigpy/zha-device-handlers/pull/4088/commits/72aa24a0ca948b08b51a2e67e5b6f63606660112) caused a completely different issue with accessing uninitialized variables to occur, fixed here: [`c21d03a` (#4088)](https://github.com/zigpy/zha-device-handlers/pull/4088/commits/c21d03a71a7962b9c6987c8fc20bb51a65b52e71), but that only further added to the confusion in the meantime.. 😄 

Now, the issues left are XBee tests erroring out (and hanging) and some weird Tuya test data.

puddly:
Thanks for taking a look! Let me see if I can figure out why `__new__` was overridden like this, very strange.

TheJulianJES:
So apparently removing the ``__new__`` override actually causes the other Tuya tests to break, even with the older zigpy.
Both the one small test snippet in `test_tuya_quirkbuilder` and all others in `tuya_test_rcbo` (that are failing in this PR).

puddly:
`TuyaData` and `Data` do some weird stuff, I think a lot of the functionality is used in a single spot and this code can be greatly simplified. I'll post an update in a bit.

puddly:
The Tuya code is very fragile. I'm having a hard time making sense of how or why it works 😅.

The current Tuya DP <-> ZCL attributes and commands mapping may be a use case for direct ZHA entity management and creation, if we ever want to go down that route.

TheJulianJES:
> The Tuya code is very fragile. I'm having a hard time making sense of how or why it works 😅.

Oh, yeah... We've generally tried to not touch too much of the old Tuya code, since there are very few tests for the old Tuya quirks, and there are unfortunately also some very popular custom quirks that depend on how everything currently *"works"*.

The goal is to switch as many (upstream v1 and custom) Tuya quirks over to v2, since they take priority and also all depend on `TuyaQuirkBuilder`, allowing us to eventually swap out quite a big portion of the whole DP <-> ZCL conversion below, still relying on the `Bus` system (with DP writes not waiting on any confirmation, causing race conditions when writing multiple DPs at the same time).

TheJulianJES:
The changes all look good to me. Thanks, @puddly!

PR #4095, opened 2025-05-29
Remove incorrect total power attribute forwarding from Xiaomi

abmantis:
## Proposed change

Follow up from https://github.com/zigpy/zha-device-handlers/pull/4057#issuecomment-2917151144 to remove an incorrect mapping from energy to total power.
Since there is currently no entity mapped for total_active_power, this change should not affect any existing entities.

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/__init__.py
+++ b/zhaquirks/xiaomi/__init__.py
@@ -128,6 +128,17 @@ class XiaomiQuickInitDevice(XiaomiCustomDevice, QuickInitDevice):
 class XiaomiCluster(CustomCluster):
     """Xiaomi cluster implementation."""
 
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        super().__init__(*args, **kwargs)
+        # Previously, this cluster was wrongly setting the total_active_power attribute,
+        # which was not added to HA.
+        # Since it is now added to HA and the incorrect value could be set, we need to
+        # reset it.
+        self._update_attribute(
+            ElectricalMeasurement.AttributeDefs.total_active_power.id, None
+        )
+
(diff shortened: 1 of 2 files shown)


zhaquirks/xiaomi/__init__.py
```diff
@@ -128,6 +128,17 @@ class XiaomiQuickInitDevice(XiaomiCustomDevice, QuickInitDevice):
...
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        super().__init__(*args, **kwargs)
+        # Previously, this cluster was wrongly setting the total_active_power attribute,
+        # which was not added to HA.
+        # Since it is now added to HA and the incorrect value could be set, we need to
+        # reset it.
+        self._update_attribute(
+            ElectricalMeasurement.AttributeDefs.total_active_power.id, None
+        )
```
TheJulianJES: I don't think we can do this in the `XiaomiCluster`. We'll update some attribute id there, instead of on the EM cluster.
Even if we accessed the EM cluster from here, it wouldn't be present on all Xiaomi devices (i.e. on sensors).
I think we can just move this to the Xiaomi `ElectricalMeasurementCluster` class though, since that's used in all Xiaomi plugs(?)
There, we're already in the right cluster, so it's also guaranteed to be present on the device.

Also, it would be really nice if we can get a short test for this. But I'm realizing this might be a bit awkward to test for, since we'll have to listen early enough to catch the `attribute_cleared` when creating the device.
I guess we could also just patch `_update_attribute` before creating the test device otherwise. Hopefully, I can come back to this if needed.

TheJulianJES (approved):
Looks good now. Thanks!

PR #4109, opened 2025-06-06
Add Third Reality dual plug and single plug settings

3reality-support:
## Proposed change

 This PR added adaptation code for dual_plug and added 2 additional private clusters
 1. reset_summation_delivered---reset the accumulated power of the plug
 2. on_to_off_delay---off delay, when you set the attribute to 10 and the current status is on, it will change to off after 10 seconds
 3. off_to_on_delay---on delay, similar to the above
 4. If both on_to_off_delay and off_to_of_delay are set, the plug will loop on and off

## Additional information

2ca2370e55bad991f254209e0406c61
a0886f3197e45043ea09cbb4b9e4023
b15b302acac803df5e1e733339e9b99

## Checklist


3reality-support:
@TheJulianJES Please help to review it

puddly:
@3reality-support Some comments, applicable to all PRs.

1. When submitting a PR, please fill out the PR template and explain what your pull request implements in both the title and PR body. In this PR, for example, it's clear that you're modifying the private clusters, but what exactly are you modifying and why?
2. Please comment your code and explain what these private attributes do in comments in the code.
   - What happens if a user sets `on_to_off_delay`? What about `off_to_on_delay` at the same time?
   - Why is `reset_summation_delivered` an attribute instead of a command? What do the different values do?
3. When creating fallback strings, please set them to user-presentable strings. `Reset summation delivered2`, for example, is not something that we can show.

3reality-support:
> > @3reality-support Some comments, applicable to all PRs.
> > 
> > 1. When submitting a PR, please fill out the PR template and explain what your pull request implements in both the title and PR body. In this PR, for example, it's clear that you're modifying the private clusters, but what exactly are you modifying and why?
> > 2. Please comment your code and explain what these private attributes do in comments in the code.
> >    
> >    * What happens if a user sets `on_to_off_delay`? What about `off_to_on_delay` at the same time?
> >    * Why is `reset_summation_delivered` an attribute instead of a command? What do the different values do?
> > 3. When creating fallback strings, please set them to user-presentable strings. `Reset summation delivered2`, for example, is not something that we can show.
> 
> I completely agree with the other points, but I don't quite understand the third one. This attribute is to reset the Summation delivered. Can you help recommend a fallback name?

3reality-support:
@puddly  @TheJulianJES

3reality-support:
@puddly @TheJulianJES Please help to check

3reality-support:
@TheJulianJES

zhaquirks/thirdreality/plug_dual.py
```diff
@@ -0,0 +1,117 @@
...
+        endpoint_id=2,
+        min_value=0,
+        max_value=65535,
+        step=1,
+        mode="box",
+        unit=UnitOfTime.SECONDS,
+        device_class=NumberDeviceClass.DURATION,
+        translation_key="on_to_off_delay_ep2",
+        fallback_name="Turn off delay",
+    )
```
TheJulianJES: The `fallback_name` will be used as the English translation for the `translation_key` when this makes its way into HA. Here, the translation keys are different for the two endpoints, but the fallback name is the same.
We should just change the `fallback_name` to either be "Turn off delay 1" and "Turn off delay 2" or be something like "Turn off delay left" and "Turn off delay right".

In the future, we'll try and add a postfix if there are entities named in the same way on multiple endpoints. For now, it's fine to explicitly make the translation key and fallback name different. Other integrations in HA also do this.

Please also adjust all other occurrences.

TheJulianJES: Also, have you checked that the two entities each for the dual plug both show up in HA?
I thought the unique IDs might conflict, as they're just based on the attribute name, but we might also include the endpoint. I'd have to check.

If both appear, this is fine. Otherwise, you can include `unique_id_suffix="on_to_off_delay_1"` (and `_2`) for the entities to assign a different UID.

zhaquirks/thirdreality/plug_dual.py
```diff
@@ -0,0 +1,117 @@
...
+        off_to_on_delay: Final = ZCLAttributeDef(
+            id=0x0002,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RDP01072Z")
+    .also_applies_to("Third Reality, Inc", "3RWP01073Z")
```
TheJulianJES: Let's use `applies_to`. Both do the same, but we've switched to just using `applies_to` in most quirks.

zhaquirks/thirdreality/plug_dual.py
```diff
@@ -0,0 +1,117 @@
...
+        fallback_name="Reset right summation delivered ep2",  # ep2 is right
+    )
+    .number(
+        attribute_name=ThirdRealityPlugCluster.AttributeDefs.on_to_off_delay.name,
+        cluster_id=ThirdRealityPlugCluster.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        step=1,
+        mode="box",
```
TheJulianJES: `mode=box` is fine to keep, but we don't actually set that mode in ZHA yet.
So, it's expected that this does nothing (for now). There's an open issue in the ZHA repo about this.

zhaquirks/thirdreality/plug_dual.py
```diff
@@ -0,0 +1,117 @@
...
+        translation_key="reset_summation_delivered_ep1",
+        fallback_name="Reset left summation delivered",  # ep1 is left
+    )
+    .write_attr_button(
+        attribute_name=ThirdRealityPlugCluster.AttributeDefs.reset_summation_delivered.name,
+        attribute_value=0x01,
+        cluster_id=ThirdRealityPlugCluster.cluster_id,
+        endpoint_id=2,
+        translation_key="reset_summation_delivered_ep2",
+        fallback_name="Reset right summation delivered ep2",  # ep2 is right
```
TheJulianJES: This is similar to my [previous comment](https://github.com/zigpy/zha-device-handlers/pull/4109/files#r2202279603), let's just include either left/right or 1/2, not both and not something abbreviated like "ep2".

TheJulianJES (commented):
Otherwise, I think this looks good. We'll come back to this when the last few things are addressed.
Lastly, there's no reason to ping us every couple of days. We'll try to review PRs as soon as possible.

zhaquirks/thirdreality/plug_dual.py
3reality-support: Do you mean that the translation_key for ep1 and ep2 can be the same.But there was an error on my end
3129505c582599ab562f5ca2761bc44

3reality-support:
> Otherwise, I think this looks good. We'll come back to this when the last few things are addressed. Lastly, there's no reason to ping us every couple of days. We'll try to review PRs as soon as possible.

Sorry, I got it. Also, do I need to @ you every time I make modifications. Or quote your comment

3reality-support:
@TheJulianJES

3reality-support:
@TheJulianJES

zhaquirks/thirdreality/plug_dual.py
```diff
@@ -0,0 +1,113 @@
...
+    QuirkBuilder("Third Reality, Inc", "3RDP01072Z")
+    .applies_to("Third Reality, Inc", "3RWP01073Z")
+    .replaces(ThirdRealityPlugCluster, endpoint_id=1)
+    .replaces(ThirdRealityPlugCluster, endpoint_id=2)
+    .write_attr_button(
+        attribute_name=ThirdRealityPlugCluster.AttributeDefs.reset_summation_delivered.name,
+        attribute_value=0x01,  # 1 reset summation delivered
+        cluster_id=ThirdRealityPlugCluster.cluster_id,
+        endpoint_id=1,
+        translation_key="reset_summation_delivered1",
```
TheJulianJES: Please add an underscore before the `1` at the end. Do this for all translation keys.

In this case, it might make even more sense to rename the `translation_key` to `reset_summation_delivered_left`, since that's what the `fallback_name` (translation value) will be. If the translation were to include `1`  or `2`, we should mention that in the translation key. But if it's left or right, we should instead mention that.

TheJulianJES (commented):
We'll likely automatically add a postfix number to all entities named the same in the future: https://github.com/zigpy/zha/pull/525

I'll come back to this shortly. For clarity, you can still use different translation keys with left and right though. Those won't be impacted by the change then, since they'll be named differently.

zhaquirks/thirdreality/plug_dual.py
3reality-support: OK. Modified, please check

3reality-support: @TheJulianJES

TheJulianJES (commented):
I've combined these two quirks into one file/module now, as the manufacturer cluster is exactly the same at the moment.
They can be split apart again if the manufacturer cluster is changed for one device only.

I've also added `mode="box"`, even though that was likely already used due to the big `max_value` in HA. I've also removed `step=1`, as that's the default, and we don't really need it with the "box" mode.
Lastly, I changed some names/translations.

PR #4114, opened 2025-06-09
Add support for additional Legrand wireless switches

nzapponi:
## Proposed change
Added support for additional Legrand wireless switches, including NLT, NLD and NLW.

## Additional information
This is mostly to receive battery status for all wireless switches.

## Checklist

### DIFF
--- a/zhaquirks/legrand/wirelessswitch.py
+++ b/zhaquirks/legrand/wirelessswitch.py
@@ -5,11 +5,17 @@ from zigpy.zcl.clusters.general import BinaryInput
 
 from zhaquirks.const import (
     BUTTON,
+    BUTTON_1,
+    BUTTON_2,
+    BUTTON_3,
     COMMAND,
     COMMAND_MOVE,
     COMMAND_OFF,
     COMMAND_ON,
     COMMAND_STOP,
+    DIM_DOWN,
+    DIM_UP,
+    ENDPOINT_ID,
     LONG_PRESS,
     LONG_RELEASE,
     PARAMS,
--- a/zhaquirks/legrand/wirelessswitch.py
+++ b/zhaquirks/legrand/wirelessswitch.py
@@ -1,4 +1,4 @@
-"""Module for Legrand wireless radiant switch."""
+"""Module for Legrand wireless switches (Radiant, NLD, NLT, NLW)."""
 
 from zigpy.quirks.v2 import QuirkBuilder
 from zigpy.zcl.clusters.general import BinaryInput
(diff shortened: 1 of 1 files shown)


nzapponi:
@TheJulianJES how can I get this merged in? Thanks!

TheJulianJES:
Wait for a review

shimon88:
i tried to use your wirelessswitch file in my zha custom quirk folder to try and get my NLT remote switches to be identified correctly (they read the wrong SW version like they are NLF for some reason, but it doesnt seems to work. and the BinaryInput theat was always Off is now Unavailable.

any idea?

nzapponi:
Is the SW version something you experienced before my updates already? I never had any issues with NLT SW version.
It's normal for BinaryInput to be unavailable, as my quirk removes it (it's useless, really).

shimon88:
not saying your quirk broke the version, i thought it might fix it :)
it seems as zha identifies Bticino K4003CWI as NLF instead of NLT for some reason.

i added z2m as a repo for ota updates and NLT has 81 already and i would like to update.

nzapponi:
Can you share your device's signature please? FWIW, I've got a Legrand 574291 (equivalent of BTicino K4003C) and it's correctly showing up as NLT)

shimon88:
Hi,

K4003C is not NLT. NLT is for remote switches and K4003C is a connected switch with mains power.

The diagnostics from ZHA are attached
[zha-5a50d99752605c051fc2afcb73902cd4- Legrand  Remote switch-9dede244b5daf84bd7c7460067a20436.json](https://github.com/user-attachments/files/20771278/zha-5a50d99752605c051fc2afcb73902cd4-.Legrand.Remote.switch-9dede244b5daf84bd7c7460067a20436.json)

nzapponi:
Yeah sorry, it's not the equivalent. But Legrand's 574291 is definitely an NLT and shows up as such.
Very weird btw - our devices' signatures are exactly the same, and yet mine is recognised as NLT and yours isn't. Can you share a screenshot of the NLF thing?

shimon88:
I meant that the switch is identified and "wireless switch" which means it is identified as NLT, but the FW version does not match the latest in z2m repo.
now all versions are Unknown so maybe there is another issue

shimon88:
BTW what is the current version your remote switch has?

nzapponi:
image

shimon88:
This is weird. I will try to remove custom quirks I have that adds the led control switches. 

Anyway, any idea why the firmware becomes unknown for all legrand switches after few hours from system restart?

nzapponi:
How did you configure OTA? I don't get the unknown thing tbh - mine are showing consistently. 

It may be down to the OTA config and the radio you're using, I've seen reports of less successful set ups with different antennas.

shimon88:
like that:

zigpy_config:
  ota:
    extra_providers:
      - type: ikea
      - type: z2m

shimon88:
seems like i have config issue as you suspected.
now i see several pending fw updates for legrand devices.
but none of them succseed. all fail on timeout (error 148)

nzapponi:
Yeah that happens often - the way I managed is by disconnecting and reconnecting the antenna, pressing the button right after initiating the update in HA, and potentially even re-pair the button.
No need to remove it from HA, just Add a new device and put the button back into pairing mode. HA will retain all the config (automations, scenes, etc) but the reconfiguration and interview process means that right after pairing, I was able to start the firmware update immediately. It doesn't always work, though. You may need to try a few times, replace the battery, etc...

nzapponi:
FWIW (unrelated to your original ask), I also set up group binding on the remote switches, and removed any config/automation in HA. The benefit is much lower latency, and the ability for the switch to work even if HA / the Zigbee coordinator goes down.

TheJulianJES (commented):
Since these quirks all seem to be exactly the same other than the model name of the devices, we can just use `applies_to` for the other model names.
Similar to this: https://github.com/zigpy/zha-device-handlers/blob/d475f76f2feead387c2bd240f0dfd3e184d7a91e/zhaquirks/tuya/ty0201.py#L7-L9

~~I can update the PR in a few minutes.~~ EDIT: See comment below.

TheJulianJES (commented):
Actually, are the device automation triggers correct? They should be different for the double/triple remotes, since they have more buttons, right?

nzapponi:
Good point, they're probably not right. Do you have an example on how to set up device automation triggers that support multiple endpoints? Thanks!

nzapponi:
@TheJulianJES updated the device automation triggers. Would you mind giving it a look please?

nzapponi:
@TheJulianJES can we please merge this in?

puddly (approved):

TheJulianJES (approved):
This looks good now, thanks! Will be merged in time for HA 2025.9.0.

nzapponi:
Awesome thanks so much!

PR #4119, opened 2025-06-10
Replace ByteString

cdce8p:
## Proposed change
`typing.ByteString` and `collections.abc.ByteString` will be removed in Python `3.14`.
AFAICT the actual data type is `bytes` anyway, so just use that instead.
_The alternative would be the full type `bytes | bytearray | memoryview`._

Originally added in #3644 by @abmantis.

https://docs.python.org/3.13/library/typing.html#typing.ByteString

## Checklist


abmantis (approved):

TheJulianJES (approved):
Thanks

PR #4120, opened 2025-06-11
Add support for OWON PC321 3 phase power meter

zoic21:
## Proposed change

Add support of owon pc321

## Checklist

### DIFF
--- a/zhaquirks/owon/pc321.py
+++ b/zhaquirks/owon/pc321.py
@@ -87,6 +87,7 @@ class OwonPC321MeteringCluster(CustomCluster, Metering):
             id=0x4103, type=t.uint48_t, access="r", is_manufacturer_specific=True
         )
 
+
 (
     QuirkBuilder("OWON Technology Inc.", "PC321")
     .replaces(OwonPC321MeteringCluster)
--- a/zhaquirks/owon/pc321.py
+++ b/zhaquirks/owon/pc321.py
@@ -24,7 +24,7 @@ class OwonPC321MeteringCluster(CustomCluster, Metering):
 
     class AttributeDefs(Metering.AttributeDefs):
         """OwonPC321 Manufacturer Cluster Attributes."""
-        
+
         phase_1_power: Final = foundation.ZCLAttributeDef(
             id=0x2000, type=t.uint24_t, access="r", is_manufacturer_specific=True
         )
--- a/zhaquirks/owon/pc321.py
+++ b/zhaquirks/owon/pc321.py
@@ -1,7 +1,9 @@
 """Quirk for Owon PC321."""
 
+from typing import Final
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.v2 import QuirkBuilder, SensorDeviceClass, SensorStateClass
+from zigpy.zcl import foundation
 from zigpy.quirks.v2.homeassistant import (
     UnitOfElectricCurrent,
     UnitOfElectricPotential,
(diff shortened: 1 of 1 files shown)


zhaquirks/owon/pc321.py
```diff
@@ -0,0 +1,250 @@
...
+    """OwonPC321MeteringCluster cluster."""
+
+    cluster_id = 0x0702
+    name = "OwonPC321Metering"
+    ep_attribute = "smartenergy_metering"
+
+    attributes = Metering.attributes.copy()
+    attributes.update(
+        {
+            0x2000: ("phase_1_power", t.uint24_t, True),
```
puddly: We're moving away from this syntax for attribute definitions. Can you rewrite it to look like this? https://github.com/zigpy/zha-device-handlers/blob/5f1f14b95d352ff1cf0cbd34ba86c496f259d2f5/zhaquirks/sinope/sensor.py#L50-L74

zoic21:
Like that ?

zhaquirks/owon/pc321.py
```diff
@@ -0,0 +1,295 @@
...
+)
+import zigpy.types as t
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.smartenergy import Metering
+
+
+class OwonPC321MeteringCluster(CustomCluster, Metering):
+    """OwonPC321MeteringCluster cluster."""
+
+    cluster_id = 0x0702
```
TheJulianJES, line 21: The `cluster_id` line can be removed, as this is already set in the `Metering` class (which we're inheriting from).

zhaquirks/owon/pc321.py
```diff
@@ -0,0 +1,295 @@
...
+    name = "OwonPC321Metering"
+    ep_attribute = "smartenergy_metering"
```
TheJulianJES, line 23: The `ep_attribute` line can be removed, as this is already set in the `Metering` class (which we're inheriting from).

zhaquirks/owon/pc321.py
```diff
@@ -0,0 +1,295 @@
...
+    .replaces(OwonPC321MeteringCluster)
+    .sensor(
+        "phase_1_power",
+        OwonPC321MeteringCluster.cluster_id,
+        unit=UnitOfPower.WATT,
+        translation_key="active_power_phase_1",
+        fallback_name="Active power phase 1",
+        device_class=SensorDeviceClass.POWER,
+        state_class=SensorStateClass.MEASUREMENT,
+    )
```
TheJulianJES, line 107: We have some Tuya three phase power meter quirks [here](https://github.com/zigpy/zha-device-handlers/blob/6bc62472db42c0fd7c87e16aac2f22bd19ad07bd/zhaquirks/tuya/ts0601_power.py#L59-L411) and [here](https://github.com/zigpy/zha-device-handlers/blob/6bc62472db42c0fd7c87e16aac2f22bd19ad07bd/zhaquirks/tuya/ts0601_din_power.py#L254-L432).
They use phases A, B, and C, instead of 1, 2, and 3. I believe we should also use ABC here.
So, kind of making those custom entities similar as much as possible.

Salamandar:
Installing the custom quirk file actually gives values, but I'm pretty sure some aren't correct.

The current is way too high for the power (102A for 30W ?), but the values vary accordingly though.

We might have swapped around the phases (between the current and voltage sensors) but I'm not near this house at the moment so I can't fix that. I'll reply to this comment when i'm sure the setup is correct.

image

PR #4126, opened 2025-06-13
Add quirks for CTM Lyng outlets/relays and cooktop guard

uvNikita:
## Proposed change
CTM Lyng is a norwegian manufacturer which is very nice to work with.
They provided me with zigbee spec docs for their Tavrele 25A product.

The product mostly works out of the box, but there is a couple of manufacturer specific clusters (client `CTMDiagnosticsCluster` and server `CTMStoveGuardCluster`) that could be useful to have.

Once I'll get their stove guard sensor, I'll see how it can also be integrated.

## Additional information
[Outlet Zigbee Specs](https://github.com/user-attachments/files/20724763/ctm_mains_power_outlet_1.0.docx)
[Stove Guard Zigbee Specs](https://github.com/user-attachments/files/20758273/ctm_mkomfy_1.0.docx)

## Checklist


zhaquirks/ctm/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+
+    name = "CtmStoveGuard"
+    cluster_id = 0xFFC9
+
+    class AttributeDefs(CustomCluster.AttributeDefs):
+        """CTM Lyng stove guard cluster attribute definitions."""
+
+        ctm_alarm_status: Final = foundation.ZCLAttributeDef(
+            id=0x0001,
+            type=AlarmStatus,
```
puddly, line 114: For these devices, set `zcl_type=foundation.DataTypeId.uint8` and then you can have `AlarmStatus` be an `enum8` subclass.

For the vast majority of devices the ZCL type can be inferred from the Python type but there are many devices that do not follow the specification and use bare integers instead of enums.

uvNikita:
@puddly I set `zcl_type`  on enum attributes as you said.
I also added sever and client commands and a stoveguard support (not tested fully until I receive it, but should work in principle).

I have a couple of questions:

- For stoveguard I might also need to setup a Bus. Is there a way to do it in quirks v2?
- As you can see, their outlet has CTMStoveGuardCluster as a client cluster and stoveguard has it as a server cluster. Should I split them into two separate classes so that all attributes/commands are not present in both devices?
- I tried to add `paired_with` attribute from the client CTMStoveGuardCluster as an entity, but zha fails with `AssertionError` pointing to [assert cluster_handler](https://github.com/zigpy/zha/blob/a0fd5acf6b7676a147b390fa5eb7ac6a0e9e4a23/zha/application/discovery.py#L237). Reading the attribute manually from "Manage Zigbee device" works correctly.

uvNikita:
@TheJulianJES Could you look at this one when you have time?

I finally installed cooktop guard from CTM Lyng that pairs with their relay/outlet. Renamed stoveguard to cooktop guard.

I still have a few issues / questions:

- ~~Is there a way to setup bus in quirks v2?~~
- Should I split CTMCooktopGuardCluster into two CTMCooktopGuardServerCluster and CTMCooktopGuardClientCluster?
- `paired_with` / `cooktop_guard_sensor_address` entity (see commented lines) still fails with `AssertionError` pointing to [assert cluster_handler](https://github.com/zigpy/zha/blob/a0fd5acf6b7676a147b390fa5eb7ac6a0e9e4a23/zha/application/discovery.py#L237). It's probably because it's a client Cluster, but not sure how to fix this.
- ~~`alarm_status`, `battery_status` and `active_status` shows up and works correctly, but only if I update them manually. They don't seem to update automatically no matter what I do. AFAIK, this sensor "sleeps" until paired relay reports electrical usage to preserve battery. After that it wakes up and suppose to report temperature/status updates. I can confirm that manual reading works after there is electrical load, but no automatic updates. Maybe there is an obvious fix for that?~~

uvNikita:
I managed to fix reporting configuration for `ctm_alarm_status`, `ctm_battery_alarm` and `ctm_cooktop_active` attributes.

@puddly If I used `enum8` in `type=<enum>`, and `zcl_type=foundation.DataTypeId.uint8`, then I was getting the following error on reporting configuration (`INVALID_DATA_TYPE`):


```
2025-09-02 12:53:17.613 DEBUG (MainThread) [zigpy.zcl] [0x7703:1:0xffc9] Decoded ZCL frame: CtmMKomfyCluster:Configure_Reporting_rsp(status_records=[ConfigureReportingResponseRecord(status=<Status.INVALID_DATA_TYPE: 141>, direction=<ReportingDirection.SendReports: 0>, attrid=1), ConfigureReportingResponseRecord(status=<Status.INVALID_DATA_TYPE: 141>, direction=<ReportingDirection.SendReports: 0>, attrid=5)])
```

Changing type back to `t.uint8_t` solved the problem.

At this point everything is working pretty good, I only have a couple of questions left:

- Should I split CTMCooktopGuardCluster into two CTMCooktopGuardServerCluster and CTMCooktopGuardClientCluster?
- `paired_with` / `cooktop_guard_sensor_address` entity in the outlet device (see commented lines) still fails with `AssertionError` pointing to [assert cluster_handler](https://github.com/zigpy/zha/blob/a0fd5acf6b7676a147b390fa5eb7ac6a0e9e4a23/zha/application/discovery.py#L237). It's probably because it's a client Cluster, but not sure how to fix this.

TheJulianJES:
I'll get back to you on the rest, but the `zcl_type` causing issues with attribute reporting should be fixed with:
- https://github.com/zigpy/zigpy/pull/1650

uvNikita:
> I'll get back to you on the rest, but the `zcl_type` causing issues with attribute reporting should be fixed with:
> 
>     * [Use attribute zcl_type when configuring reporting. zigpy#1650](https://github.com/zigpy/zigpy/pull/1650)

That's great, then I changed it back to setting `zcl_type=foundation.DataTypeId.uint8` since it should be fixed in the recent version of zigpy.

uvNikita:
@TheJulianJES I rebased the PR on top of the current dev branch and:

- switched to new use `manufacturer_code` instead of `is_manufacturer_specific=True`
- fixed commands direction (client vs server)
- removed comment about client cluster sensor (cooktop_guard_sensor_address) since it's not supported by ZHA
- add some missing device classes

I think this should be ready to merge.

zhaquirks/ctm/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+        )
+        ctm_cooktop_temperature: Final = ZCLAttributeDef(
+            id=0x0003,
+            type=t.uint16_t,
+            access="r",
+            manufacturer_code=CTM_MANUF_CODE,
+        )
+        ctm_ambient_temperature: Final = ZCLAttributeDef(
+            id=0x0004,
+            type=t.uint8_t,
```
TheJulianJES, line 137: Z2M has this as `t.uint16_t`, not sure what is correct. Debug logs should include this when a report/read comes in.

TheJulianJES (commented):
Looks pretty good! Very small things:

https://github.com/zigpy/zha-device-handlers/pull/4126/changes/5c035f5bb890e4155ef086af675409d7625f15b5 dropped the `zcl_type` for an attribute that's seemingly an int on the ZCL side, but an enum on the zigpy side. I think it was correct before and it's broken now? Though it's really only an issue if you were to try and write to that attribute. For reading, zigpy should correctly parse it either way.

Also, the usual order for kwargs in quirks v2 entities is this:
- attribute_name, cluster_id, endpoint_id first; translation_key, fallback_name last.

Other feedback: 
- nit: name casing: "CTMDiagnostics" vs "CtmCooktopGuard"

zhaquirks/ctm/__init__.py
uvNikita: Here is log when I enabled debug logging and read the value of this attribute:

`DEBUG (MainThread) [zigpy.zcl] [0x7703:1:0xffc9] Decoded ZCL frame: CTMCooktopGuardCluster:ReadAttributesResponse(status_records=[ReadAttributeRecord(attrid=4, status=<Status.SUCCESS: 0>, value=TypeValue(type=uint8_t, value=28))])`

Looks like uint8_t is correct? And this is also what's specified in the spec docs in the first post.

uvNikita:
@TheJulianJES Thank you for the review, made all the requested changes.

uvNikita:
I rebased this PR on top of the current dev branch.

@TheJulianJES I think this one is ready for merge?

uvNikita:
- Changed to ZCL_CLUSTER_REVISION_ATTR
- Use manufacturer_code in commands
- Set access="rp"
- Convert EUI64 value to str
- Add state_class/device_class to sensors
- Rebased on top of the current dev.
 
Diagnostics files:
- [CTM Lyng mKomfy.json](https://github.com/user-attachments/files/30581734/zha-e6a6134e87a1487b579906f1f839c1fd-CTM.Lyng.mKomfy-9f09588ee4699d8489728d5cede38fb9.json)
- [CTM Lyng Tavlerele 25A.json](https://github.com/user-attachments/files/30581738/zha-e6a6134e87a1487b579906f1f839c1fd-CTM.Lyng.Tavlerele.25A-3ebeb58a02361c74c3ac1196a13d4001.json)

zhaquirks/ctm/outlet.py
```diff
@@ -0,0 +1,59 @@
...
+    .replaces(CTMDiagnosticsCluster)
+    .sensor(
+        attribute_name=CTMDiagnosticsCluster.AttributeDefs.ctm_reboot_counter.name,
+        cluster_id=CTMDiagnosticsCluster.cluster_id,
+        endpoint_id=1,
+        state_class=SensorStateClass.TOTAL_INCREASING,
+        entity_type=EntityType.DIAGNOSTIC,
+        initially_disabled=True,
+        translation_key="reboot_counter",
+        fallback_name="Reboot counter",
```
TheJulianJES: IMO "Reboot count" sounds better than both "Reboot counter" and "Rebooted count", but [we already have translations](https://github.com/home-assistant/core/blob/7289e79c4b860a783145d6d35a03dc2654861207/homeassistant/components/zha/strings.json#L1823-L1825) for `rebooted_count`: "Rebooted count", so let's use that:

```suggestion
        translation_key="rebooted_count",
        fallback_name="Rebooted count",
```

zhaquirks/ctm/outlet.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    .sensor(
+        attribute_name=CTMDiagnosticsCluster.AttributeDefs.ctm_button_0_click_counter.name,
+        cluster_id=CTMDiagnosticsCluster.cluster_id,
+        endpoint_id=1,
+        state_class=SensorStateClass.TOTAL_INCREASING,
+        entity_type=EntityType.DIAGNOSTIC,
+        initially_disabled=True,
+        translation_key="button_click_counter",
+        fallback_name="Button click counter",
```
TheJulianJES: Similar to above, let's use "count" everywhere, instead of "counter":
```suggestion
        translation_key="button_click_count",
        fallback_name="Button click count",
```

zhaquirks/ctm/cooktop_guard.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+    .sensor(
+        attribute_name=CTMDiagnosticsCluster.AttributeDefs.ctm_reboot_counter.name,
+        cluster_id=CTMDiagnosticsCluster.cluster_id,
+        endpoint_id=1,
+        state_class=SensorStateClass.TOTAL_INCREASING,
+        entity_type=EntityType.DIAGNOSTIC,
+        initially_disabled=True,
+        translation_key="reboot_counter",
+        fallback_name="Reboot counter",
```
TheJulianJES: See above comments:
```suggestion
        translation_key="rebooted_count",
        fallback_name="Rebooted count",
```

TheJulianJES (commented):
Two small things but I think this looks good otherwise. Thanks and sorry for the wait!
The reporting config mentioned by the bot might be worth addressing as well. I think `max_interval=3600`, so the device reports the attribute value at least every hour, might be nice.

uvNikita:
Done. I also renamed the attributes for consistency, or did you want me to only change translation_key/fallback_name?

PR #3575, opened 2024-12-03
Add entities for Schneider Electric Thermostats

uvNikita:
## Proposed change
Add most of the manufacturer specific attributes as Home Assistant entities (numbers, enums).

## Additional information
#1705 
https://github.com/zigpy/zha-device-handlers/pull/3473
https://github.com/Koenkk/zigbee2mqtt/issues/23739

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

PR #4132, opened 2025-06-18
Fix Schneider Electric thermostat number limits

uvNikita:
## Proposed change
This fixes limits for "Open window detection threshold" and "Boost amount" numbers according to the spec doc.

## Additional information
#3575 

## Checklist


TheJulianJES (approved):
thanks!

PR #3070, opened 2024-03-25
Add support for Aqara Ceiling Light T1M (acn032)

corvis:
## Proposed change

This quirk adds support for power on behavior configuration to Aqara Light T1M (ceiling light with RGB ring).
This device identifies itself as lumi.light.acn032.

## Additional information

Resolves #3066 

## Checklist

- [X] The changes are tested and work correctly
- [X] `pre-commit` checks pass / the code has been formatted using Black
- [X] Tests have been added to verify that the new code works

Issue #3066, opened 2024-03-24
[Device Support Request] Aqara Ceiling light T1M


corvis:
### Problem description

Device works fine in general both main light and rgb ring were identified automatically and work fine. However there is not power_outage_memory which is pretty important to me.

Device identifies itself as `lumi.light.acn032`.

### Solution description

I noticed zigbee2mqtt added [this recently](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7203/files). So I tried to replicate this:
![image](https://github.com/zigpy/zha-device-handlers/assets/5593615/58458fb6-f866-4c67-ac48-2943fbc36e90)

With a quirk like this:

```python
class LumiBasicCluster(CustomCluster, Basic):
    """Hue Motion Basic cluster."""
    manufacturer_id_override = 4447

    attributes = BasicCluster.attributes.copy()
    attributes[0xFF19] = ZCLAttributeDef(
        id=0xFF19, type=t.Bool, access=ZCLAttributeAccess.Write, is_manufacturer_specific=True,
        name="power_outage_memory"
    )


class LumiLightAcn032(XiaomiCustomDevice):
    """
    Aqara Light T1M (ceiling light with RGB ring).
    This quirk adds support for power on behavior by adding the power_outage_memory attribute, the rest works out
    of the box.

    It MIGHT also work for the Model T1 (acn031) but it has not been tested.
    """
    signature = {
        MODELS_INFO: [("Aqara", "lumi.light.acn032"), ],
        ENDPOINTS: {
            1: {
                PROFILE_ID: zha.PROFILE_ID,
                DEVICE_TYPE: zha.DeviceType.COLOR_DIMMABLE_LIGHT,
                INPUT_CLUSTERS: [
                    BasicCluster.cluster_id,  # 0x0000
                    Identify.cluster_id,  # 0x0003
                    Groups.cluster_id,  # 0x0004
                    Scenes.cluster_id,  # 0x0005
                    OnOff.cluster_id,  # 0x0006
                    LevelControl.cluster_id,  # 0x0008
                    Color.cluster_id,  # 0x0300
                    OppleClusterLightGen2.cluster_id,  # 0xFCC0 - manufacturer specific
                ],
                OUTPUT_CLUSTERS: [
                    Time.cluster_id,  # 0x000A
                    Ota.cluster_id,  # 0x0019
                ],
            },
            2: {
                PROFILE_ID: zha.PROFILE_ID,
[...]

PR #4136, opened 2025-06-18
Add Aqara T1M ceiling light support

evanjarrett:
## Proposed change
This quirk adds support for power on behavior configuration to Aqara Light T1M (ceiling light with RGB ring).
This device identifies itself as lumi.light.acn032.

## Additional information
This is an updated pull request based on #3070 and #3066 but updated to the new Quirks v2
This also allows the new option to be exposed in Home Assistant Configuration.

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/light_acn.py
+++ b/zhaquirks/xiaomi/aqara/light_acn.py
@@ -244,9 +244,9 @@ class LumiLightAcn014(XiaomiCustomDevice):
         AqaraLightT1M.AttributeDefs.power_on_state.name,
         LumiPowerOnStateMode,
         AqaraLightT1M.cluster_id,
+        endpoint_id=2,
         translation_key="power_on_state",
         fallback_name="Power on state",
-        endpoint_id=2,
     )
     .add_to_registry()
 )
--- a/zhaquirks/xiaomi/aqara/light_acn.py
+++ b/zhaquirks/xiaomi/aqara/light_acn.py
@@ -40,11 +40,11 @@ from zhaquirks.xiaomi import (
 )
 
 
-class LumiPowerOnBehaviorMode(t.enum8):
-    """Power on behavior mode."""
+class LumiPowerOnStateMode(t.enum8):
+    """Power on state mode."""
 
     On = 0x00
-    PreserveState = 0x01
+    LastState = 0x01
     Off = 0x02
 
 
@@ -57,16 +57,14 @@ class OppleClusterLight(XiaomiAqaraE1Cluster):
 
 
 class AqaraLightT1M(XiaomiAqaraE1Cluster):
-    """Add power on behavior management for Lumi devices."""
-
-    manufacturer_id_override = 4447
+    """Add power on state management for Lumi devices."""
 
     class AttributeDefs(BaseAttributeDefs):
         """Manufacturer specific attributes."""
 
-        power_on_behavior = ZCLAttributeDef(
+        power_on_state = ZCLAttributeDef(
             id=0x0517,
-            type=LumiPowerOnBehaviorMode,
+            type=LumiPowerOnStateMode,
             zcl_type=DataTypeId.uint8,
             access="rw",
             is_manufacturer_specific=True,
@@ -233,22 +231,22 @@ class LumiLightAcn014(XiaomiCustomDevice):
 (
     QuirkBuilder(AQARA, "lumi.light.acn032")
     .friendly_name(manufacturer="Aqara", model="Ceiling Light T1M")
-    .adds_endpoint(2, device_type=zha.DeviceType.COLOR_DIMMABLE_LIGHT)
+    .replaces_endpoint(2, device_type=zha.DeviceType.COLOR_DIMMABLE_LIGHT)
     .replaces(AqaraLightT1M)
     .replaces(AqaraLightT1M, endpoint_id=2)
     .enum(
-        AqaraLightT1M.AttributeDefs.power_on_behavior.name,
-        LumiPowerOnBehaviorMode,
+        AqaraLightT1M.AttributeDefs.power_on_state.name,
+        LumiPowerOnStateMode,
         AqaraLightT1M.cluster_id,
-        translation_key="power_on_behavior",
-        fallback_name="Power on behavior",
+        translation_key="power_on_state",
+        fallback_name="Power on state",
     )
     .enum(
-        AqaraLightT1M.AttributeDefs.power_on_behavior.name,
-        LumiPowerOnBehaviorMode,
+        AqaraLightT1M.AttributeDefs.power_on_state.name,
+        LumiPowerOnStateMode,
         AqaraLightT1M.cluster_id,
-        translation_key="power_on_behavior",
-        fallback_name="Power on behavior",
+        translation_key="power_on_state",
+        fallback_name="Power on state",
         endpoint_id=2,
     )
     .add_to_registry()
--- a/zhaquirks/xiaomi/aqara/light_acn.py
+++ b/zhaquirks/xiaomi/aqara/light_acn.py
@@ -2,6 +2,7 @@
 
 from zigpy import types as t
 from zigpy.profiles import zgp, zha
+from zigpy.quirks.v2 import QuirkBuilder
 from zigpy.zcl.clusters.general import (
     Alarms,
     AnalogInput,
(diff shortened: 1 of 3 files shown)


zhaquirks/xiaomi/aqara/light_acn.py
```diff
@@ -200,3 +228,28 @@ class LumiLightAcn014(XiaomiCustomDevice):
...
+        AqaraLightT1M.cluster_id,
+        translation_key="power_on_behavior",
+        fallback_name="Power on behavior",
+    )
+    .enum(
+        AqaraLightT1M.AttributeDefs.power_on_behavior.name,
+        LumiPowerOnBehaviorMode,
+        AqaraLightT1M.cluster_id,
+        translation_key="power_on_behavior",
+        fallback_name="Power on behavior",
```
TheJulianJES: Can we rename both to `power_on_state` and "Power on state"? We already have that very similar translation/string defined [here](https://github.com/home-assistant/core/blob/6641cb37998d0c70a9d9a2fef5a7a01cb4126090/homeassistant/components/zha/strings.json#L1176-L1178).

Also note that we'll try to add a suffix in ZHA to duplicate entity names in the future. So, no real need to do it here for now. But you'll need to test which entity is for what endpoint for now, unfortunately.

```suggestion
        translation_key="power_on_state",
        fallback_name="Power on state",
    )
    .enum(
        AqaraLightT1M.AttributeDefs.power_on_behavior.name,
        LumiPowerOnBehaviorMode,
        AqaraLightT1M.cluster_id,
        translation_key="power_on_state",
        fallback_name="Power on state",
```

zhaquirks/xiaomi/aqara/light_acn.py
```diff
@@ -200,3 +228,28 @@ class LumiLightAcn014(XiaomiCustomDevice):
             },
         }
     }
+
+
+(
+    QuirkBuilder(AQARA, "lumi.light.acn032")
+    .friendly_name(manufacturer="Aqara", model="Ceiling Light T1M")
+    .adds_endpoint(2, device_type=zha.DeviceType.COLOR_DIMMABLE_LIGHT)
```
TheJulianJES: For what is this needed? Is there a wrong `device_type` defined by default? 
If we want to change that, we should use `replaces_endpoint`.

There were some fixes to how it behaves: https://github.com/zigpy/zigpy/pull/1602 and also kinda https://github.com/zigpy/zigpy/pull/1603.

Or is the second endpoint actually not exposed by the device at all? If so, what (should) be on it then? Other than the Aqara config cluster.

zhaquirks/xiaomi/aqara/light_acn.py
```diff
@@ -45,6 +56,23 @@ class OppleClusterLight(XiaomiAqaraE1Cluster):
     }
 
 
+class AqaraLightT1M(XiaomiAqaraE1Cluster):
+    """Add power on behavior management for Lumi devices."""
+
+    manufacturer_id_override = 4447
```
TheJulianJES: Does the device provide another manufacturer id in the NodeDescriptor?

zhaquirks/xiaomi/aqara/light_acn.py
```diff
@@ -30,13 +32,22 @@
...
     XiaomiAqaraE1Cluster,
     XiaomiCustomDevice,
 )
 
 
+class LumiPowerOnBehaviorMode(t.enum8):
+    """Power on behavior mode."""
+
+    On = 0x00
+    PreserveState = 0x01
```
TheJulianJES: Can we use `LastState` here? That's similar to some config entities still in ZHA for Tuya plugs.
```suggestion
    LastState = 0x01
```

zhaquirks/xiaomi/aqara/light_acn.py
evanjarrett: Yes, this is what is in the node descriptor
```
  "node_descriptor": {
    "logical_type": 1,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4447,
    "maximum_buffer_size": 82,
    "server_mask": 11264,
  },
  ```

File: zhaquirks/xiaomi/aqara/light_acn.py (hunk 2 above)
Comment (evanjarrett) [reply zhaquirks/xiaomi/aqara/light_acn.py]: It appears I do not actually need this, so will remove.
I could have sworn the second endpoint didn't show up until I added it, but now it seems like its working fine without it.

Comment (TheJulianJES) [reply zhaquirks/xiaomi/aqara/light_acn.py]: Hmm. I hope we do not accidentally persist that added endpoint into the zigpy DB.
Mind removing and re-pairing the light to confirm?

But I've also seen some Aqara devices that randomly do not report some endpoints. I'd guess this would have been fixed in these newer products though.

File: zhaquirks/xiaomi/aqara/light_acn.py (hunk 3 above)
Comment (TheJulianJES) [reply zhaquirks/xiaomi/aqara/light_acn.py]: Ok, since the power on attribute is already marked as `is_manufacturer_specific`, zigpy will automatically use that manufacturer id then.
So, the override can be removed from what I can see.

File: zhaquirks/xiaomi/aqara/light_acn.py (hunk 2 above)
Comment (evanjarrett) [reply zhaquirks/xiaomi/aqara/light_acn.py]: I also wrote the unit test in mind to test each endpoint, and if I remove that, I can no longer test both

Comment (TheJulianJES) [reply zhaquirks/xiaomi/aqara/light_acn.py]: There is a parameter you can add to the unit test to add the additional endpoint to the test device. I can look it up later, but it's in the device from quirks v2 fixture.

Comment (evanjarrett) [reply zhaquirks/xiaomi/aqara/light_acn.py]: Yep, I removed the `adds/replaces`, removed and reset the device in HA, everything seems fine. (other than having to fix a few automations...)

Fixed the test to define endpoint_ids and everything is working now.

File: zhaquirks/xiaomi/aqara/light_acn.py [hunk 5]
```diff
@@ -200,3 +226,27 @@ class LumiLightAcn014(XiaomiCustomDevice):
...
+        translation_key="power_on_state",
+        fallback_name="Power on state",
+    )
+    .enum(
+        AqaraLightT1M.AttributeDefs.power_on_state.name,
+        LumiPowerOnStateMode,
+        AqaraLightT1M.cluster_id,
+        translation_key="power_on_state",
+        fallback_name="Power on state",
+        endpoint_id=2,
```
TheJulianJES: Ok, I'll just move up the `endpoint_id` argument, as we try to have `translation_key` and `fallback_name` as the last arguments to be somewhat consistent between quirks.
```suggestion
    .enum(
        AqaraLightT1M.AttributeDefs.power_on_state.name,
        LumiPowerOnStateMode,
        AqaraLightT1M.cluster_id,
        endpoint_id=2,
        translation_key="power_on_state",
        fallback_name="Power on state",
```

TheJulianJES (approved):
Everything looks good to me. Thanks!

Orange-GT3:
@evanjarrett - is there any way that this quirk can be extended to the [(Aqara Bulb T1)](https://www.aqarastore.com.au/collections/light/products/copy-of-aqara-led-light-bulb-tunable-white) which identifies as lumi.light.acn014?

I'm having issues with this randomly turning on at either full brightness or close to 0% brightness (reported by HA as 0%) and would like to be able to set the Power On default.

Orange-GT3:
The device signature is:

`

 {
   "node_descriptor": {
     "logical_type": 1,
     "reserved": 0,
     "aps_flags": 0,
     "frequency_band": 8,
     "mac_capability_flags": 142,
     "manufacturer_code": 4447,
     "maximum_buffer_size": 82,
     "server_mask": 11264,
   },
   "endpoints": {
     "1": {
       "profile_id": "0x0104",
       "device_type": "0x0102",
       "input_clusters": [
         "0x0000",
         "0x0003",
         "0x0004",
         "0x0005",
         "0x0006",
         "0x0008",
         "0x0300",
         "0xfcc0"
       ],
       "output_clusters": [
         "0x000a",
         "0x0019"
       ]
     }
   },
   "manufacturer": "Aqara",
   "model": "lumi.light.acn014",
 }

`

evanjarrett:
Each device needs its own quirk. That would require a separate pull request.

I don't have that device to test with, so I can't be the one to implement it.

PR #4153, opened 2025-06-29
Add Inovelli VZM32-SN mmWave switch support

InovelliUSA:
## Proposed change

Add Support for Inovelli VZM32-SN mmWave Switch

## Checklist

### DIFF
--- a/zhaquirks/inovelli/VZM32SN.py
+++ b/zhaquirks/inovelli/VZM32SN.py
@@ -1,0 +1,291 @@
+"""VZM32-SN mmWave Dimmer Switch."""
+"""Configuration entities need to:"""
+"""zha/zha/application/platforms/number/__init__.py"""
+
+from zigpy.profiles import zgp, zha
+from zigpy.profiles.zha import DeviceType
+from zigpy.quirks import CustomDevice
+from zigpy.zcl.clusters.general import (
+    Basic,
+    GreenPowerProxy,
+    Groups,
+    Identify,
+    LevelControl,
+    OnOff,
+    Ota,
+    Scenes,
+)
+from zigpy.zcl.clusters.homeautomation import Diagnostic, ElectricalMeasurement
+from zigpy.zcl.clusters.measurement import (
+    IlluminanceMeasurement,
+    OccupancySensing
+)
+from zigpy.zcl.clusters.smartenergy import Metering
+
+from zhaquirks.const import (
+    DEVICE_TYPE,
+    ENDPOINTS,
+    INPUT_CLUSTERS,
+    MODELS_INFO,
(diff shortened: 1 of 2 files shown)


puddly (changes requested):
Thanks for the PR!

We have new APIs for building quirks, clusters, and commands. I think they'll greatly simplify the way your quirk is currently written.

1. For the quirk itself, I think something like this should suffice and will match both variants of your device:
   ```python
   from zigpy.quirks.v2 import QuirkBuilder

   (
       QuirkBuilder("Inovelli", "VZM32-SN")
       .replaces(InovelliVZM32SNCluster, endpoint_id=1)
       .replaces(InovelliVZM32SNMMWaveCluster, endpoint_id=1)
       .device_automation_triggers(INOVELLI_AUTOMATION_TRIGGERS)
       .add_to_registry()
   )
   ```

2. Our ZCL cluster definition format has changed. Here's an example that rewrites a portion of your cluster definition:

   ```python
   class MMWaveLightOnPresenceBehavior(t.enum8):
       Option_one = 0x00
       Some_other_option = 0x01
   
   
   class MMwaveControlId(t.enum8):
       ...
   
   
   class InovelliVZM32SNCluster(InovelliCluster):
       """Inovelli VZM32-SN custom cluster."""
   
       name = "Inovelli VZM32SN Custom"
   
       class AttributeDefs(InovelliCluster.AttributeDefs):
           dimming_speed_up_local: Final = foundation.ZCLAttributeDef(
               id=0x0002,
               type=t.uint8_t,
               access="rw",
           )
   
           # For enum attributes that don't use an enum, please create one with all of the
           # known values and override the ZCL data type
           mmwave_light_on_presence_behavior: Final = foundation.ZCLAttributeDef(
               id=0x006E,
               zcl_type=foundation.DataTypeId.uint8,
               type=MMWaveLightOnPresenceBehavior,
               access="rw",
           )
   
       class ServerCommands(foundation.ServerCommandDefs):
           mmwave_control_command = foundation.ZCLCommandDef(
               id=0x00,
               schema={
                   # If the command schema includes an enumeration, break it out into a new
                   # type
                   "control_id": MMwaveControlId,
               },
               is_manufacturer_specific=True
           )
   ```

InovelliUSA:
@puddly we are actually working with @codyhackw to move everything over to the new format:

https://github.com/zigpy/zha-device-handlers/pull/3232

This is more of an "interim" change to make sure the device is supported when it comes out.

TheJulianJES:
https://github.com/zigpy/zha-device-handlers/pull/3232 is merged now, slated for HA 2025.8.0.
I guess we can close this then?

PR #4155, opened 2025-06-29
Add Senoro door sensor alarm

TheJulianJES:
## Proposed change
This adds a binary sensor entity for the Senoro door sensor that displays the current alarm state.
A button entity is also added that allows for resetting an activated tamper alarm.

**This still needs testing**

## Additional information
Follow-up to:
- https://github.com/zigpy/zha-device-handlers/pull/4086

Implementation details:
- The `tuya_switch` (and so on) methods do three things: (1) map the Tuya DP to a ZCL attribute, (2) create that ZCL attribute, and (3) create the switch entity metadata. Here, we call `tuya_dp_attribute` to (1) map the DP and (2) create the ZCL attribute. We call `binary_sensor` and `write_attr_button` methods directly for the mapped ZCL attribute.
- `translation_key` is omitted for the "Tamper" alarm, as we set the "tamper device class", [defaulting to a translated "Tamper" name in Home Assistant](https://github.com/home-assistant/core/blob/6d28b993444ea93c4d76ae000e46e22c84984ace/homeassistant/components/binary_sensor/strings.json#L293)
- `unique_id_suffix` is only kind of needed for the binary sensor and button entity. The unique ID for the HA entities is based on the `attribute_name` by default, which would conflict. However, unique IDs are still keyed by platform (`binary_sensor` vs `button` in ZHA), so they won't conflict for now. Still, we want to avoid this, so I've given them separate `unique_id_suffix`es.

## Checklist

### DIFF
--- a/zhaquirks/tuya/tuya_door.py
+++ b/zhaquirks/tuya/tuya_door.py
@@ -1,9 +1,10 @@
 """Senoro Window Sensor (TS0601)."""
 
-from zigpy.quirks.v2 import EntityPlatform, EntityType
+from zigpy.quirks.v2 import BinarySensorDeviceClass, EntityPlatform, EntityType
 import zigpy.types as t
+from zigpy.zcl import foundation
 
-from zhaquirks.tuya import BatterySize
+from zhaquirks.tuya import TUYA_CLUSTER_ID, BatterySize
 from zhaquirks.tuya.builder import TuyaQuirkBuilder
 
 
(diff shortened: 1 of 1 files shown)


senorosoftware:
@TheJulianJES 
This implementation seems to work. I tested  it with a device, There is a switch button that show "Tamper Alarm". The state of the button doesn't change if the alarm starts. The Manipulation field under "Sensor" is showing the correct state, if the alarm starts. Also the opening state and the battery works as expected. Resetting the alarm (turn off) via the "Reset Alarm" "Press" button works fine.

image

felixmueller:
@TheJulianJES 

Looks good for me, too. I don't have the "Tamper alarm" button, maybe that was just cached by HA @senorosoftware?

I didn't try if the "Reset alarm" control works because I don't want to trigger the alarm right now.

Will this update be also part of the next release used by HA 2025.7?

Thanks a lot!

sonoro_quirk

ChristianGr1974:
> @TheJulianJES
> 
> Looks good for me, too. I don't have the "Tamper alarm" button, maybe that was just cached by HA @senorosoftware?
> 
> I didn't try if the "Reset alarm" control works because I don't want to trigger the alarm right now.
> 
> Will this update be also part of the next release used by HA 2025.7?
> 
> Thanks a lot!
> 
> sonoro_quirk

@felixmueller 
You are right, the switch is cached. I didn‘t reconnected the device.

felixmueller:
@TheJulianJES 

Can this be merged now that the test from @senorosoftware was successful? Thanks!

ChristianGr1974:
@TheJulianJES 
When can this PR be completed?

PR #4187, opened 2025-07-15
Fix Tuya presence sensor "distance tracking" for `_ya4ft0w4`

rrooggiieerr:
The switch actually enables/disables distance tracking instead of a LED indicator. Renaming the switch to what it actually does prevents a lot of confusion.

## Checklist

### DIFF
--- a/zhaquirks/tuya/tuya_motion.py
+++ b/zhaquirks/tuya/tuya_motion.py
@@ -247,10 +247,10 @@ base_tuya_motion = (
     # 2, 3, 4, and 9 from base
     .tuya_switch(
         dp_id=101,
-        attribute_name="find_switch",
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
Thanks! We may also need to change this for other Tuya presence sensors(?)

Also note that this will create a new entity, since the `attribute_name` has changed (which we use for the entity unique ID in HA), but that's actually what we want here, since users might have already changed the default name to something else, e.g. "LED light", so they wouldn't see the correct name.
With this, we properly create a new entity that actually does what it says.

rrooggiieerr:
I'm not sure if other Tuya presence sensors need to be changed, I only have experience with the sensors covered by this change.

Issue #3990, opened 2025-03-30
[BUG] _TZE200_ya4ft0w4 Motion sensitivity and Presence sensitivity get reset


Itay1787:
### Bug description

Hi, I noticed that Motion sensitivity and Presence sensitivity reset every time on _TZE200_ya4ft0w4
I can configure it, and after about 15 minutes, it resets back to 0 - I just now tried to set it and the Presence sensitivity gets resat after one minute, the motion still set like I set it (for now...)
I tried configuring it through the "Configuration" in the device settings, and also through the "Manage Zigbee device"
It resets in both
In the past, I used to use the quirk where it would also reset, but less frequently
I would appreciate help with this

### Steps to reproduce

Change the Motion sensitivity and Presence sensitivity to a value greater than 0 
Then wait for it to reset

### Expected behavior

to stay on the set value that I set it

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

![Image](https://github.com/user-attachments/assets/10b9954b-ff2d-4913-8763-a561735715c5)

![Image](https://github.com/user-attachments/assets/8cd449ca-8656-4147-97f2-6f69d1633a33)

</details>

### Device signature

<details><summary>Device signature</summary>

```json

{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4660,
    "maximum_buffer_size": 108,
    "maximum_incoming_transfer_size": 0,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 0,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0100",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0x0400",
        "0x0406",
        "0xef00"
      ],
      "output_clusters": []
    }
  },
  "manufacturer": "_TZE200_ya4ft0w4",
  "model": "TS0601",
  "class": "zigpy.quirks.v2.CustomDeviceV2"
}

```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json

{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2025.3.3",
    "dev": false,
    "hassio": true,
[...]

PR #4188, opened 2025-07-15
Fix Tuya presence sensitivity config for `_ya4ft0w4`

rrooggiieerr:
## Proposed change
The multiplier for the presence sensitivity should be 1, or be deleted like I did.

See https://github.com/zigpy/zha-device-handlers/issues/3990#issuecomment-2902794465

## Checklist


TheJulianJES (approved):
Thanks!

PR #4193, opened 2025-07-17
Migrate all quirks to the new ZCL command and attribute structures

puddly:
## Proposed change

This PR migrates all quirks to use `ClientCommandDefs`, `ServerCommandDefs`, and `AttributeDefs` and drops the `direction` attribute (see https://github.com/zigpy/zigpy/pull/1614). This is an enormous diff but is mostly rote rewriting.

## Checklist


puddly:
@TheJulianJES  In the process of testing this change, I found a "new" class of bug related to `direction` that will appear once a new version of zigpy is released.`test_xiaomi_t1_door_sensor` is one example:

```python
# Device sends an attribute report
DEBUG    zigpy.zcl:__init__.py:267 [0x1234:1:0x0006] Received ZCL frame: b'\x18]\n\x00\x00\x10\x01'
DEBUG    zigpy.zcl:__init__.py:270 [0x1234:1:0x0006] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl<0x18>(frame_type=<FrameType.GLOBAL_COMMAND: 0>, is_manufacturer_specific=0, direction=<Direction.Server_to_Client: 1>, disable_default_response=1, reserved=0, *is_cluster=False, *is_general=True), tsn=93, command_id=10, *direction=<Direction.Server_to_Client: 1>)
DEBUG    zigpy.zcl:__init__.py:294 [0x1234:1:0x0006] Decoded ZCL frame: OnOff:Report_Attributes(attribute_reports=[Attribute(attrid=0x0000, value=TypeValue(type=Bool, value=<Bool.true: 1>))])

# Zigpy cannot match it to the correct cluster because `direction` is flipped in the ZCL header
DEBUG    zigpy.endpoint:endpoint.py:246 [0x1234:1] Message on unknown cluster 0x0006 for endpoint <CustomEndpoint id=1 in=[basic:0x0000, power:0x0001, identify:0x0003, opple_cluster:0xFCC0] out=[on_off:0x0006, ota:0x0019] status=<Status.ZDO_INIT: 1>>
```

The unit test explicitly creates a Zigbee packet object with `185D0A00001001` as the payload. Is it correct that the device sends this exact data? If so, this is a bit of a problem: the `OnOff` cluster is a client cluster, not a server cluster, so the `direction` sent by the device is wrong 😅. If not, then the unit test just needs to be slightly adjusted so all is well.

TheJulianJES:
I still have the Aqara T1 door sensor, so just triggered it on unmodified HA 2025.7.1. Here's the output:
```python
2025-07-17 21:08:53.345 DEBUG (MainThread) [bellows.ezsp.protocol] Received command incomingMessageHandler: {'type': <EmberIncomingMessageType.INCOMING_UNICAST: 0>, 'apsFrame': EmberApsFrame(profileId=260, clusterId=6, sourceEndpoint=1, destinationEndpoint=1, options=<EmberApsOption.APS_OPTION_ENABLE_ROUTE_DISCOVERY: 256>, groupId=0, sequence=115), 'lastHopLqi': 80, 'lastHopRssi': -80, 'sender': 0xBE48, 'bindingIndex': 255, 'addressIndex': 255, 'messageContents': b'\x18\xe3\n\x00\x00\x10\x01'}
2025-07-17 21:08:53.345 DEBUG (MainThread) [bellows.zigbee.application] Received incomingMessageHandler frame with [<EmberIncomingMessageType.INCOMING_UNICAST: 0>, EmberApsFrame(profileId=260, clusterId=6, sourceEndpoint=1, destinationEndpoint=1, options=<EmberApsOption.APS_OPTION_ENABLE_ROUTE_DISCOVERY: 256>, groupId=0, sequence=115), 80, -80, 0xBE48, 255, 255, b'\x18\xe3\n\x00\x00\x10\x01']
2025-07-17 21:08:53.346 DEBUG (MainThread) [zigpy.application] Received a packet: ZigbeePacket(timestamp=datetime.datetime(2025, 7, 17, 19, 8, 53, 346179, tzinfo=datetime.timezone.utc), priority=0, src=AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0xBE48), src_ep=1, dst=AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x0000), dst_ep=1, source_route=None, extended_timeout=False, tsn=115, profile_id=260, cluster_id=6, data=Serialized[b'\x18\xe3\n\x00\x00\x10\x01'], tx_options=<TransmitOptions.NONE: 0>, radius=0, non_member_radius=0, lqi=80, rssi=-80)
2025-07-17 21:08:53.347 DEBUG (MainThread) [zigpy.zcl] [0xBE48:1:0x0006] Received ZCL frame: b'\x18\xe3\n\x00\x00\x10\x01'
# notice direction=Direction.Server_to_Client below:
2025-07-17 21:08:53.348 DEBUG (MainThread) [zigpy.zcl] [0xBE48:1:0x0006] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl<0x18>(frame_type=<FrameType.GLOBAL_COMMAND: 0>, is_manufacturer_specific=0, direction=<Direction.Server_to_Client: 1>, disable_default_response=1, reserved=0, *is_cluster=False, *is_general=True), tsn=227, command_id=10, *direction=<Direction.Server_to_Client: 1>)
2025-07-17 21:08:53.349 DEBUG (MainThread) [zigpy.zcl] [0xBE48:1:0x0006] Decoded ZCL frame: OnOff:Report_Attributes(attribute_reports=[Attribute(attrid=0x0000, value=TypeValue(type=Bool, value=<Bool.true: 1>))])
2025-07-17 21:08:53.354 DEBUG (MainThread) [zigpy.zcl] [0xBE48:1:0x0006] Received command 0x0A (TSN 227): Report_Attributes(attribute_reports=[Attribute(attrid=0x0000, value=TypeValue(type=Bool, value=<Bool.true: 1>))])
2025-07-17 21:08:53.355 DEBUG (MainThread) [zigpy.zcl] [0xBE48:1:0x0006] Attribute report received: on_off=<Bool.true: 1>
2025-07-17 21:08:53.355 DEBUG (MainThread) [zha.zigbee.cluster_handlers] [0xBE48:1:0x0006_client]: cluster_handler[on_off] attribute_updated - cluster[On/Off] attr[on_off] value[Bool.true]
```

The issue with this device was that, when it gets sent Xiaomi/Aqara manufacturing codes during pairing, it behaves differently and doesn't use the `IasZone` cluster anymore, instead using an `OnOff` cluster that's not exposed during the interview.
It could very well be that the device actually uses `OnOff` as a server cluster and that the quirk just incorrectly assume it's a client one (like older Aqara door sensors).
We'd still want it to be a (`OnOff`) binary sensor for ZHA, so we could maybe fix the `direction` in the quirk or just expose a quirk v2 entity with the current `OnOff` UID?

puddly:
Thanks for the confirmation! So it's indeed a problem with the device. I assume other devices will have this problem too, we don't have unit tests for everything. Maybe we can provide a quirks v2 method to enable relaxed cluster direction matching for certain clusters?

TheJulianJES:
Yeah, that sounds like a good idea. Though I'd also consider allowing that for v1 quirks, maybe just set an attribute in `CustomCluster` to allow for that?

puddly:
For V1, I've broken out this functionality into an easily-overridable method:

```python
class MagnetT1(XiaomiCustomDevice):
    ...

    def _find_zcl_cluster(
        self, hdr: foundation.ZCLHeader, packet: t.ZigbeePacket
    ) -> Cluster:
        """Find a cluster for the packet."""
        assert packet.src_ep is not None
        endpoint = self.endpoints[packet.src_ep]

        # The `direction` field is incorrectly set, we should always route the packet
        # to client `OnOff` cluster
        if packet.cluster_id == OnOff.cluster_id:
            return endpoint.out_clusters[packet.cluster_id]

        return super()._find_zcl_cluster(hdr, packet)
```

For V2 we can make a helper to enable lax cluster matching.

puddly:
It looks like the API I mentioned above was not actually added to zigpy in today's release 😄. I will finish this up tomorrow, there are only XBee unit tests failing and the one Aqara device that needs this API.

TheJulianJES:
I'll take a last quick look in a few seconds

zhaquirks/tuya/mcu/__init__.py
```diff
@@ -355,6 +352,12 @@ def handle_mcu_connection_status(
 class TuyaOnOff(OnOff, TuyaLocalCluster):
     """Tuya MCU OnOff cluster."""
 
+    class AttributeDefs(OnOff.AttributeDefs):
+        """Cluster attributes."""
+
+    class ServerCommandDefs(OnOff.ServerCommandDefs):
+        """Server command definitions."""
+
```
TheJulianJES, line 360: Sometimes, `AttributeDefs` and `ServerCommandDefs` are added again, even though they should already be the same. Is this just for clarity?

TheJulianJES (approved):
Checked everything roughly and it looks good.
I expect this to break some custom quirks relying on these though, let's see 😄

zhaquirks/tuya/mcu/__init__.py
puddly: I think multiple base classes make the source of `AttributeDefs` ambiguous, at least to the type checker.

TheJulianJES: Ah ok, that makes sense.

PR #4198, opened 2025-07-19
Add Sonoff ZBM5 light switch

### LINKED ISSUE #4230: [Device Support Request] Sonoff ZBM5 (Specifically "Detached Relay" Support)
### Problem description

Sonoff ZBM5 + zigpy doesn't expose the "detached relay" options in Home Assistant

### Solution description

Quirk for now?

[zigbee2mqtt](https://www.zigbee2mqtt.io/devices/ZBM5-2C-120.html#detach-relay-mode-composite) has support.

### Screenshots/Video

Image

</details>

### Diagnostics information

_No response_

### Device signature

```json

```

</details>

### Logs

```python

[Paste the logs here]

```

</details>

### Custom quirk

```python

```

</details>

### Additional information

_No response_

MindFreeze:
## Proposed change

Add the detached relay and related entities for [ZBM5](https://sonoff.tech/en-eu/products/sonoff-switchman-zigbee-smart-wall-switch-zbm5-80-86-type?utm_source=ITEAD+Studio&utm_campaign=6a5bff62e4-EMAIL_CAMPAIGN_2025_04_15_08_53_COPY_01&utm_medium=email&utm_term=0_-827d8849b1-1347521457&goal=0_dc6f4826ce-6a5bff62e4-1347521457&mc_cid=6a5bff62e4&mc_eid=06d9149c0d&variant=46546052612337) ~~but there is a catch.
This switch has 2 modes because the neutral wire is optional. And if there is no neutral wire, detached relay is not supported. This can be determined from the "work_mode" sensor in this PR but I don't know how to conditionally add the detached relay entities only if `work_mode=1`~~
The Sonoff site says that detached relay is only supported when there is a Neutral wire provided but after testing with 2 dual switches, this seems to be false. It works fine without a Neutral. I only tested it with a 20W light though so it may not work as well with a smaller load.

## Checklist


zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,132 @@
...
+    cluster_id = 0xFC11
+
+    manufacturer_id_override = foundation.ZCLHeader.NO_MANUFACTURER_ID
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        external_trigger_mode = ZCLAttributeDef(
+            id=0x0016,
+            type=t.uint8_t,
```
puddly: You can specify `type=SonoffExternalSwitchTriggerType` and then `zcl_type=foundation.DataTypeId.uint8` to decouple the Python type from the low-level Zigbee data type.

TheJulianJES: Do we generally want to do this now? Here, the attribute is exposed via a quirks v2 (enum/select) entity already.
So, I'm not sure we if we should (or need to) deviate further from what the device actually exposes for the attributes.

I don't have a strong opinion either way, just curious what your opinion on this is.

puddly: Manufacturers using ints instead of enums is super common and is about the only place where we override the data type like this, I don't think there's a downside. It shows up correctly in the frontend as well.

Jolanrensen:
Hi! Thanks for taking the time to make the quirk :)

Personally I have the ZBM5-3C and -2C (Firmware: 0x00001003) and I'm running 2025.8 beta 1. I tried adding your version of the quirk to my Hass instance and while I am able to change the detach_relay_mask, the relay_x_detached only report "None".
I assume this is also why the 3 "Detach Relay" switches are disabled in the device.

Is there something I could test?

MindFreeze:
That was an old version. On the latest, you shouldn't see a mask. Only switches. And they are disabled by default. You have to enable them.

Jolanrensen:
@MindFreeze I meant in the "Manage Zigbee Device" screen, but gotcha, enabling the switches successfully allows me to toggle detach mode indeed!

I don't see anything being logged when I press one of the buttons in detached mode, though. Does that work on your end? Can you respond to any physical button presses?

MindFreeze:
Yes. On my switches it works as expected. The button toggles the switch in HA without changing the state of the physical relay

Jolanrensen:
> Yes. On my switches it works as expected. The button toggles the switch in HA without changing the state of the physical relay

So, you get 3 hass switches per sonoff button?
- Detach Relay x Switch 
- switch responding to physical button x
- switch corresponding to physical relay x

Might be slightly different on my end as I have the 120 version

MindFreeze:
> > Yes. On my switches it works as expected. The button toggles the switch in HA without changing the state of the physical relay
> 
> So, you get 3 hass switches per sonoff button?
> 
> * Detach Relay x Switch
> * switch responding to physical button x
> * switch corresponding to physical relay x
> 
> Might be slightly different on my end as I have the 120 version

No, I get 2 switches per button. 1 for the button/relay and 1 to detach the relay. It works as described in https://github.com/zigpy/zigpy/pull/1612#issuecomment-3122164643

Jolanrensen:
> No, I get 2 switches per button. 1 for the button/relay and 1 to detach the relay. It works as described in [zigpy/zigpy#1612 (comment)](https://github.com/zigpy/zigpy/pull/1612#issuecomment-3122164643)

Ah I see :) They have an "Opening" sensor that changes state whenever detached mode is enabled and a button is pressed. I also have an ZBMINIR2 where this works as expected. Unfortunately, there's no sensor for my ZBM5-3C-120, so there's no way to detect a button press in detached mode :/

I was kind of expecting that it worked similarly to my [Aqara Switch](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/xiaomi/aqara/switch_h1_double.py). This doesn't have a "sensor" in detached mdoe, but it simply fires events whenever a button is pressed or double-pressed, which you can hook into in automations.

sl33nyc:
FWIW, I successfully used this quirk from this PR after manually adding it to my 2025.8.0 installation.  My use case is to turn one of the two available button+relay into just a button, while the other button+relay remains button+relay.

Yeah, unlike the ZBMINIR2, which I also have, there isn't a separate Opening entity.

Screenshot 2025-08-07 at 11 05 22 AM

where:
* Garage Door events represent just button presses.
* Driveway events represent button presses + relay on/off.

:shipit: ❗

Jolanrensen:
@sl33nyc nice, but if you don't use the relay of the first button, then you don't even need detached mode, do you?

In proper detached mode, you should be able to handle the relays and buttons completely separately. 
For example, you should be able to press button 1, and, using an automation, only switch relay 2 without changing relay 1. Or you could double-press button 2, relay 2 stays unchanged, but you turn on a completely different light in the house.

In my case, that's what I'm trying to achieve. The 3 relays all have lights attached to them, but a combination of presses, or a doublepress, should control all lights in my living room at once.

MindFreeze:
These sonoff switches don't support that. I don't even see an output cluster for OnOff, so you can't bind it to a light, which is the original use case for detached relay.

Jolanrensen:
@MindFreeze That's a shame :/ But I think you're right. Sonoff gives as a usecase: _"With the detach relay enabled, smart lights won’t lose power when the wall switch is turned off, allowing them to remain smart at all times. *Supports connecting smart lights only in the neutral required mode."_ However, if you want to supply permanent power to a smart lightbulb anyway, why connect it to the smart switch in the first place? XD 
[Z2M](https://www.zigbee2mqtt.io/devices/ZBM5-3C-120.html#action-enum) also doesn't expose any actions (in contrast to my [aqara switch](https://www.zigbee2mqtt.io/devices/WS-K02E.html#action-enum)).

I wonder if there's a way to differentiate between "turned X" and "turned X triggered by action" in automations... That would allow me to build a workaround in detached mode. Because "turned X" means a button press, which doesn't change the relay, and "turned X triggered by action" means an action taken by home assistant, which does change the relay.
image

Anyway, I'll figure it out. Looks like the PR can achieve everything that's possible :) thanks! For me it can be merged too

0x55555555:
Hello,

I've applied this locally and tested Detached Relay mode works as expected.

Thanks,

disambiguationuk:
@MindFreeze are we still waiting for a review before merging?

MindFreeze:
> @MindFreeze are we still waiting for a review before merging?

I guess

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,166 @@
...
+        relay_2_detached = ZCLAttributeDef(
+            id=0x0FFB,
+            type=t.Bool,
+            is_manufacturer_specific=True,
+        )
+        relay_3_detached = ZCLAttributeDef(
+            id=0x0FFC,
+            type=t.Bool,
+            is_manufacturer_specific=True,
+        )
```
TheJulianJES: Hmm, so we have completely fake attributes, allowing us to expose them as quirks v2 entities. 
It might be better to adapt our quirks v2 entities to work for this (in zigpy + ZHA + quirks). Might be a bit difficult though.

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,166 @@
...
+    .adds(SonoffCluster)
+    .enum(
+        SonoffCluster.AttributeDefs.work_mode.name,
+        SonoffWorkMode,
+        SonoffCluster.cluster_id,
+        translation_key="work_mode",
+        fallback_name="Work mode",
+        entity_type=EntityType.DIAGNOSTIC,
+        entity_platform=EntityPlatform.SENSOR,
+        initially_disabled=True,
```
TheJulianJES: Let's move translation key and fallback name to the bottom, similar to all other v2 quirks:
```suggestion
        entity_type=EntityType.DIAGNOSTIC,
        entity_platform=EntityPlatform.SENSOR,
        initially_disabled=True,
        translation_key="work_mode",
        fallback_name="Work mode",
```

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+    .switch(
+        SonoffCluster.AttributeDefs.relay_1_detached.name,
+        SonoffCluster.cluster_id,
+        translation_key="relay_1_detached",
+        fallback_name="Detach Relay 1",
+        entity_type=EntityType.CONFIG,
+        initially_disabled=True,
```
TheJulianJES: Same here. We'll also need to lower-case "relay" to confirm with HA guidelines, as it's mid-sentence.
(The fallback name is the name we'll also use to add to Home Assistant's `strings.json` for translations.)

I'd also suggest changing the translation_key to be more in line with the name and existing translation keys. 
```suggestion
        entity_type=EntityType.CONFIG,
        initially_disabled=True,
        translation_key="detach_relay_1",
        fallback_name="Detach relay 1",
```
We could also use translation placeholders instead.

These entities should already be in the `CONFIG` category by default, since that's the default for quirks v2 switches. v2 quirks are mostly made for config entities.

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,166 @@
...
+    zbm_1c_quirk.clone()
+    .applies_to("SONOFF", "ZBM5-2C-80/86")
+    .applies_to("SONOFF", "ZBM5-2C-120")
+    .switch(
+        SonoffCluster.AttributeDefs.relay_2_detached.name,
+        SonoffCluster.cluster_id,
+        translation_key="relay_2_detached",
+        fallback_name="Detach Relay 2",
+        entity_type=EntityType.CONFIG,
+        initially_disabled=True,
```
TheJulianJES:
```suggestion
        entity_type=EntityType.CONFIG,
        initially_disabled=True,
        translation_key="detach_relay_2",
        fallback_name="Detach relay 2",
```

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,166 @@
...
+    zbm_2c_quirk.clone()
+    .applies_to("SONOFF", "ZBM5-3C-80/86")
+    .applies_to("SONOFF", "ZBM5-3C-120")
+    .switch(
+        SonoffCluster.AttributeDefs.relay_3_detached.name,
+        SonoffCluster.cluster_id,
+        translation_key="relay_3_detached",
+        fallback_name="Detach Relay 3",
+        entity_type=EntityType.CONFIG,
+        initially_disabled=True,
```
TheJulianJES:
```suggestion
        entity_type=EntityType.CONFIG,
        initially_disabled=True,
        translation_key="detach_relay_3",
        fallback_name="Detach relay 3",
```

zhaquirks/sonoff/zbm5.py
TheJulianJES: An issue here is that we'll (try to) read the `relay_x_detached` attributes in ZHA initialization code, as they're used by quirks v2 entities. We actually pass those reads to the device, which also isn't ideal.

The real `detach_relay_mask` is not read, so the device has to send it itself first, or the user has to attempt to change one switch. We currently do not have a proper method of manually adding attributes to be initialized if they're not directly used by quirks v2 entities, but it's something we want to add.

TheJulianJES (commented):
So, I'm not sure if we can/should go with the approach of fake attributes, just so we can expose them as switch entities to write to the real (single) bitmap attribute.

There's an issue with the incorrect attributes being initialized and reads of the fake attributes being passed to the device, whilst the real attribute is never initialized/read when pairing the device, as it's not directly used by a quirks v2 entity.

The direct use of attributes in entities is a pain-point for quirks v2 entities, yet *fine* for most devices. This isn't one of "most devices" though.

We might be able to have quirks v2 switch entities set a function that takes both the new switch state and existing attribute state to get the value ZHA should write.
We already have something similar for reading values: `attribute_converter`. Opened up https://github.com/zigpy/backlog/issues/68 to track this.

Matthias1111:
@MindFreeze thanks, for building that quirk, which is working for me as expected. 🙂

The only strange thing is, that pushing the softswitches in HA (not the physical ones) of the ZBM5, the quirk / detach relay does not take effect. But to be honest, for me there is no use case to use the virtual switches.

Best regards
Matthias

MindFreeze:
The quirk stopped working in the 2026.2 so I updated it.
Buttons in decoupled mode now don't change the switch state and only fire button press events that can be used as automation triggers.

Jolanrensen:
> The quirk stopped working in the 2026.2 so I updated it. Buttons in decoupled mode now don't change the switch state and only fire button press events that can be used as automation triggers.

That's cool! I've actually had some issues where turning the virtual switch "off" by pressing the physical button in decoupled mode would randomly turn it on again after a while

typhoon2099:
Seems to work for me. Does this support long and double press? I can only see short press as configurable via Home Assistant. I haven't looked into the ZHA events yet.

TheJulianJES:
The device only emits ZHA `toggle` events, so no long-press events exist. You'll need to use HA automations/blueprints for that, but they'll never be as good as if the device natively emitted long-press events.

typhoon2099:
Oh ffs. I wouldn't have bought the switches if I'd realised that. It's so hard to find devices that are both supported and advertise their functionality well.

I should be able to get away with swapping some light switches around in the house, and then reconfigure my automations. At this point I have three unique sets of double switches and one single switch, and they all require different quirks. I would happily swap them all out for the same model if I knew what I was buying beforehand...

TheJulianJES (commented):
Sorry for the wait here. I think this mostly good to go. No need to make any changes: see below.
**It would be ideal if someone can provide a diagnostics file for this device.** It can be downloaded from the HA device page and just attached here.

I've already made some changes here, that if good, can be pulled/pushed to this branch/PR: https://github.com/zigpy/zha-device-handlers/compare/dev...TheJulianJES:tjj/mindfreeze/zbm5_testing
The changes include:
- The fake attributes are moved to a separate `LocalDataCluster` that's completely isolated from the real `SonoffCluster`.
  - Not all existing quirks follow this pattern, but it's a bit nicer and completely prevents reads for fake attributes making it through to the real device.
- The writing redirect logic is slightly adjusted. A mix of being slightly cleaned up and being better aligned with newer zigpy versions.
- The tests have been reduced a bunch:
  - Previously, we tested pure definitions, but we don't really need to that. Testing only custom logic that's executed is enough. Pure definitions will be tested by the snapshots tests in ZHA (when device diagnostics are provided here, I'll add them there)
  - We also had tests for all the 1C, 2C, and 3C models. That's generally nice, but here, the logic is all the same, so IMO it's more readable to just test the 3C one. The other will work the same. If logic ever deviates between them, we can add them back.
- Directly using the natively emitted `toggle` event by ZHA for all client `OnOff` clusters
  - This follows the pattern of other quirks. The `zha_event` will be emitted for the `toggle` command on the correct endpoint, but we can just use that as the device trigger.
  - Note: If this doesn't work, the device doesn't expose the `OnOff` cluster (for some reason). Just doing `.adds(OnOff.cluster_id, cluster_type=cluster_type=ClusterType.Client)` should work then. `endpoint_id=x`) may also be needed.
  - (In the future, we'll have event entities for this as well)
- For the `LocalDataCluster`, we should populate the "fake" attributes. The recent zigpy release wipes the attribute cache after `Cluster.__init__`, so this will need some changes and we'll likely introduce `_DEFAULT_VALUES` for this.
  But it's already an existing issue and only really relevant when HA is restarted and the values are never touched. For switch entities, it likely doesn't even matter. They'll just be off by default.
  - We should also add a quirks v2 API method to read the real mask attribute during pairing. That will cause that cluster's `update_attribute` to fire, which would update the fake attributes with the settings from the device. This can technically already be done by using `apply_custom_configuration`, but it'd be a bit nicer to have a quirks v2 method for this.
  - We should make sure the default for the detached relays (fake) attributes is correct. Am I correct in assuming they should be `true` (attached) by default?

Do note that the changes aren't tested with a real device yet, but I think they should work. Diagnostics would be good to have for this device. Then we'll also know whether we need to add `.adds(OnOff.cluster_id, cluster_type=ClusterType.Client)` or not.

Eventually, we'll just do https://github.com/zigpy/backlog/issues/68 and then we can drop a good chunk of this quirk, but the v2 entities will stay the same (unique ID is determined by device + `attribute_name` by default), so we can just merge this for now, without breaking anything later.

zhaquirks/sonoff/zbm5.py
TheJulianJES: See this comment: https://github.com/zigpy/zha-device-handlers/pull/4198#pullrequestreview-3950589261
Should be addressed by the changes here: https://github.com/zigpy/zha-device-handlers/compare/dev...TheJulianJES:tjj/mindfreeze/zbm5_testing

typhoon2099:
Should have said by the way, I've tested the quirk in my HA and it appears to be working exactly as advertised. I've not run it for too long to notice any weird behaviour, but overall it seems fine.

Signature incoming.

typhoon2099:
```
{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4742,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 255,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 255,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0002",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0020",
        "0x0b05",
        "0xfc11",
        "0xfc57"
      ],
      "output_clusters": [
        "0x0019"
      ]
    },
    "2": {
      "profile_id": "0x0104",
      "device_type": "0x0002",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0xfc57"
      ],
      "output_clusters": []
    }
  },
  "manufacturer": "SONOFF",
  "model": "ZBM5-2C-80/86",
  "class": "zigpy.quirks.v2.CustomDeviceV2"
}
```

MindFreeze:
@TheJulianJES I tried your changes and it doesn't work. The detached relay switches are always off and can't be turned on even though it is in a detached state. Pressing the button in a detached state didn't produce any event.
Here is some more diagnostic too:
<details>

```json
{
  "data": {
    "version": 1,
    "ieee": "**REDACTED**",
    "nwk": "0x4C73",
    "manufacturer": "SONOFF",
    "model": "ZBM5-2C-80/86",
    "friendly_manufacturer": "SONOFF",
    "friendly_model": "ZBM5-2C-80/86",
    "name": "SONOFF ZBM5-2C-80/86",
    "quirk_applied": true,
    "quirk_class": "zigpy.quirks.v2.CustomDeviceV2",
    "exposes_features": [],
    "manufacturer_code": 4742,
    "power_source": "Battery or Unknown",
    "lqi": 88,
    "rssi": -78,
    "last_seen": "2026-03-16T15:00:54.539985+00:00",
    "available": true,
    "device_type": "EndDevice",
    "active_coordinator": false,
    "node_descriptor": {
      "logical_type": "EndDevice",
      "complex_descriptor_available": false,
      "user_descriptor_available": false,
      "reserved": 0,
      "aps_flags": 0,
      "frequency_band": 8,
      "mac_capability_flags": 128,
      "manufacturer_code": 4742,
      "maximum_buffer_size": 82,
      "maximum_incoming_transfer_size": 255,
      "server_mask": 11264,
      "maximum_outgoing_transfer_size": 255,
    },
    "endpoints": {
      "1": {
        "profile_id": 260,
        "device_type": {
          "name": "ON_OFF_OUTPUT",
          "id": 2
        },
        "in_clusters": [
          {
            "cluster_id": "0x0000",
            "endpoint_attribute": "basic",
            "attributes": [
              {
                "id": "0x0004",
                "name": "manufacturer",
                "zcl_type": "string",
                "value": "SONOFF"
              },
              {
                "id": "0x0005",
                "name": "model",
                "zcl_type": "string",
                "value": "ZBM5-2C-80/86"
              }
            ]
          },
          {
            "cluster_id": "0x0003",
            "endpoint_attribute": "identify",
            "attributes": []
          },
          {
            "cluster_id": "0x0004",
            "endpoint_attribute": "groups",
            "attributes": []
          },
          {
            "cluster_id": "0x0005",
            "endpoint_attribute": "scenes",
            "attributes": []
          },
          {
            "cluster_id": "0x0006",
            "attributes": [
              {
                "id": "0x0000",
                "name": "on_off",
                "zcl_type": "bool",
                "value": 0
              },
              {
                "id": "0x4003",
                "name": "start_up_on_off",
                "zcl_type": "enum8",
                "value": 255
              }
            ]
          },
          {
            "cluster_id": "0x0020",
            "endpoint_attribute": "poll_control",
            "attributes": [
              {
                "id": "0x0003",
                "name": "fast_poll_timeout",
                "zcl_type": "uint16",
                "value": 120
              }
            ]
          },
          {
            "cluster_id": "0x0b05",
            "endpoint_attribute": "diagnostic",
            "attributes": []
          },
          {
            "cluster_id": "0xfc11",
            "endpoint_attribute": null,
            "attributes": [
              {
                "id": "0x0019",
                "name": "detach_relay_mask",
                "zcl_type": "map8",
                "value": 2
              },
              {
                "id": "0x0ffa",
                "name": "relay_1_detached",
                "zcl_type": "bool",
                "value": 0
              },
              {
                "id": "0x0ffb",
                "name": "relay_2_detached",
                "zcl_type": "bool",
                "value": 1
              },
              {
                "id": "0x0ffc",
                "name": "relay_3_detached",
                "zcl_type": "bool",
                "value": 0
              },
              {
                "id": "0x0018",
                "name": "work_mode",
                "zcl_type": "enum8",
                "unsupported": true
              }
            ]
          },
          {
            "cluster_id": "0xfc57",
            "endpoint_attribute": "manufacturer_specific",
            "attributes": []
          }
        ],
        "out_clusters": [
          {
            "cluster_id": "0x0019",
            "endpoint_attribute": "ota",
            "attributes": [
              {
                "id": "0x0002",
                "name": "current_file_version",
                "zcl_type": "uint32",
                "value": 4100
              }
            ]
          }
        ]
      },
      "2": {
        "profile_id": 260,
        "device_type": {
          "name": "ON_OFF_OUTPUT",
          "id": 2
        },
        "in_clusters": [
          {
            "cluster_id": "0x0000",
            "endpoint_attribute": "basic",
            "attributes": []
          },
          {
            "cluster_id": "0x0003",
            "endpoint_attribute": "identify",
            "attributes": []
          },
          {
            "cluster_id": "0x0004",
            "endpoint_attribute": "groups",
            "attributes": []
          },
          {
            "cluster_id": "0x0005",
            "endpoint_attribute": "scenes",
            "attributes": []
          },
          {
            "cluster_id": "0x0006",
            "attributes": [
              {
                "id": "0x0000",
                "name": "on_off",
                "zcl_type": "bool",
                "value": 0
              },
              {
                "id": "0x4003",
                "name": "start_up_on_off",
                "zcl_type": "enum8",
                "value": 0
              }
            ]
          },
          {
            "cluster_id": "0xfc57",
            "endpoint_attribute": "manufacturer_specific",
            "attributes": []
          }
        ],
        "out_clusters": []
      }
    },
    "original_signature": {
      "manufacturer": "SONOFF",
      "model": "ZBM5-2C-80/86",
      "node_desc": {
        "logical_type": 2,
        "reserved": 0,
        "aps_flags": 0,
        "frequency_band": 8,
        "mac_capability_flags": 128,
        "manufacturer_code": 4742,
        "maximum_buffer_size": 82,
        "maximum_incoming_transfer_size": 255,
        "server_mask": 11264,
        "maximum_outgoing_transfer_size": 255,
      },
      "endpoints": {
        "1": {
          "profile_id": "0x0104",
          "device_type": "0x0002",
          "input_clusters": [
            "0x0000",
            "0x0003",
            "0x0004",
            "0x0005",
            "0x0006",
            "0x0020",
            "0x0b05",
            "0xfc57",
            "0xfc11"
          ],
          "output_clusters": [
            "0x0019"
          ]
        },
        "2": {
          "profile_id": "0x0104",
          "device_type": "0x0002",
          "input_clusters": [
            "0x0000",
            "0x0003",
            "0x0004",
            "0x0005",
            "0x0006",
            "0xfc57"
          ],
          "output_clusters": []
        }
      }
    },
    "zha_lib_entities": {
      "button": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "button",
            "class_name": "IdentifyButton",
            "translation_key": null,
            "translation_placeholders": null,
            "device_class": "identify",
            "state_class": null,
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "IdentifyClusterHandler",
                "generic_id": "cluster_handler_0x0003",
                "endpoint_id": 1,
                "cluster": {
                  "id": 3,
                  "name": "Identify",
                  "type": "server"
                },
                "id": "1:0x0003",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "command": "identify",
            "args": [
              5
            ],
            "kwargs": {}
          },
          "state": {
            "class_name": "IdentifyButton",
            "available": true
          }
        }
      ],
      "select": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "select",
            "class_name": "StartupOnOffSelectEntity",
            "translation_key": "start_up_on_off",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "OnOffClusterHandler",
                "generic_id": "cluster_handler_0x0006",
                "endpoint_id": 1,
                "cluster": {
                  "id": 6,
                  "name": "On/Off",
                  "type": "server"
                },
                "id": "1:0x0006",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": "on_off"
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "enum": "StartUpOnOff",
            "options": [
              "Off",
              "On",
              "Toggle",
              "PreviousValue"
            ]
          },
          "state": {
            "class_name": "StartupOnOffSelectEntity",
            "available": true,
            "state": "PreviousValue"
          }
        },
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "select",
            "class_name": "StartupOnOffSelectEntity",
            "translation_key": "start_up_on_off",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "OnOffClusterHandler",
                "generic_id": "cluster_handler_0x0006",
                "endpoint_id": 2,
                "cluster": {
                  "id": 6,
                  "name": "On/Off",
                  "type": "server"
                },
                "id": "2:0x0006",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": "on_off"
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 2,
            "available": true,
            "group_id": null,
            "enum": "StartUpOnOff",
            "options": [
              "Off",
              "On",
              "Toggle",
              "PreviousValue"
            ]
          },
          "state": {
            "class_name": "StartupOnOffSelectEntity",
            "available": true,
            "state": "Off"
          }
        }
      ],
      "sensor": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "LQISensor",
            "translation_key": "lqi",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": "measurement",
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": false,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "BasicClusterHandler",
                "generic_id": "cluster_handler_0x0000",
                "endpoint_id": 1,
                "cluster": {
                  "id": 0,
                  "name": "Basic",
                  "type": "server"
                },
                "id": "1:0x0000",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": null
          },
          "state": {
            "class_name": "LQISensor",
            "available": true,
            "state": 88
          }
        },
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "RSSISensor",
            "translation_key": "rssi",
            "translation_placeholders": null,
            "device_class": "signal_strength",
            "state_class": "measurement",
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": false,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "BasicClusterHandler",
                "generic_id": "cluster_handler_0x0000",
                "endpoint_id": 1,
                "cluster": {
                  "id": 0,
                  "name": "Basic",
                  "type": "server"
                },
                "id": "1:0x0000",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": "dBm"
          },
          "state": {
            "class_name": "RSSISensor",
            "available": true,
            "state": -78
          }
        },
        {
          "info_object": {
            "fallback_name": "Work mode",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "EnumSensor",
            "translation_key": "work_mode",
            "translation_placeholders": null,
            "device_class": "enum",
            "state_class": null,
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": false,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SonoffCluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": null
          },
          "state": {
            "class_name": "EnumSensor",
            "available": true,
            "state": null
          }
        }
      ],
      "switch": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "switch",
            "class_name": "Switch",
            "translation_key": "switch",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "OnOffClusterHandler",
                "generic_id": "cluster_handler_0x0006",
                "endpoint_id": 1,
                "cluster": {
                  "id": 6,
                  "name": "On/Off",
                  "type": "server"
                },
                "id": "1:0x0006",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": "on_off"
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null
          },
          "state": {
            "class_name": "Switch",
            "state": 0,
            "available": true
          }
        },
        {
          "info_object": {
            "fallback_name": "Detach relay 1",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "switch",
            "class_name": "ConfigurableAttributeSwitch",
            "translation_key": "detach_relay_1",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": false,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SonoffCluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "relay_1_detached",
            "invert_attribute_name": null,
            "force_inverted": false,
            "off_value": 0,
            "on_value": 1
          },
          "state": {
            "class_name": "ConfigurableAttributeSwitch",
            "available": true,
            "state": false,
            "inverted": false
          }
        },
        {
          "info_object": {
            "fallback_name": "Detach relay 2",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "switch",
            "class_name": "ConfigurableAttributeSwitch",
            "translation_key": "detach_relay_2",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": false,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SonoffCluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "relay_2_detached",
            "invert_attribute_name": null,
            "force_inverted": false,
            "off_value": 0,
            "on_value": 1
          },
          "state": {
            "class_name": "ConfigurableAttributeSwitch",
            "available": true,
            "state": true,
            "inverted": false
          }
        },
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "switch",
            "class_name": "Switch",
            "translation_key": "switch",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "OnOffClusterHandler",
                "generic_id": "cluster_handler_0x0006",
                "endpoint_id": 2,
                "cluster": {
                  "id": 6,
                  "name": "On/Off",
                  "type": "server"
                },
                "id": "2:0x0006",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": "on_off"
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 2,
            "available": true,
            "group_id": null
          },
          "state": {
            "class_name": "Switch",
            "state": 0,
            "available": true
          }
        }
      ],
      "update": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "update",
            "class_name": "FirmwareUpdateEntity",
            "translation_key": null,
            "translation_placeholders": null,
            "device_class": "firmware",
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "OtaClientClusterHandler",
                "generic_id": "cluster_handler_0x0019_client",
                "endpoint_id": 1,
                "cluster": {
                  "id": 25,
                  "name": "Ota",
                  "type": "client"
                },
                "id": "1:0x0019_client",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "supported_features": 7
          },
          "state": {
            "class_name": "FirmwareUpdateEntity",
            "available": true,
            "installed_version": "0x00001004",
            "in_progress": false,
            "update_percentage": null,
            "latest_version": null,
            "release_summary": null,
            "release_notes": null,
            "release_url": null
          }
        }
      ]
    },
    "neighbors": [],
    "routes": [
      {
        "dest_nwk": "0x0000",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0x0000"
      },
      {
        "dest_nwk": "0x9AB2",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0x9AB2"
      },
      {
        "dest_nwk": "0xF541",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0xF541"
      },
      {
        "dest_nwk": "0xDF53",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0xDF53"
      },
      {
        "dest_nwk": "0x14F4",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0x14F4"
      },
      {
        "dest_nwk": "0xF3A6",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0xF3A6"
      },
      {
        "dest_nwk": "0x411D",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0x411D"
      },
      {
        "dest_nwk": "0x4213",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0x8B2A"
      },
      {
        "dest_nwk": "0x15F0",
        "route_status": "Active",
        "memory_constrained": false,
        "many_to_one": false,
        "route_record_required": false,
        "next_hop": "0xF3A6"
      }
    ]
  },
  "issues": []
}
```

</details>

TheJulianJES:
Thanks for the diagnostics and testing! I've made two more tweaks:
- Actually sync back the local attributes after the mask is written (oops...)
- Add `OnOff` `Client` cluster to all endpoints
  - This usually works out-of-the-box, as ZHA's client cluster handlers emit events for all (client) clusters. However, the device either doesn't advertise that cluster entirely (firmware bug), or it actually sends `toggle` commands from a server cluster, which makes no sense (also a firmware bug). If this doesn't work, we can go back to a similar approach like you had in the quirk, possibly simplify it with an `EventableCluster`.

If you can test again, that'd be great:
- Diff: https://github.com/zigpy/zha-device-handlers/compare/dev...TheJulianJES:tjj/mindfreeze/zbm5_testing
- Quirk: https://github.com/TheJulianJES/zha-device-handlers/blob/tjj/mindfreeze/zbm5_testing/zhaquirks/sonoff/zbm5.py

jrfigura:
@MindFreeze & @TheJulianJES Could you have a look at my custom quirk? 
https://github.com/jrfigura/HomeAssistant/blob/main/zha_custom_quirks/sonoff_zbm5.py

I have build it with the help of Claude.ai before I discovered I don't have neutral wire in my socket. Can't test it

TheJulianJES:
Something like that would work if the three "detach attributes" were real. As far as I can see, the only real attribute on this device is a bitmap, where the individual bits can be flipped on or off to control the detached relay status.
Switches/enums would just set the whole attribute, overriding the state for other relays.

Also note that the `async_initialize_cluster_handler_specific` method should be `apply_custom_configuration` to work. But it's not actually needed at all if the device were to function like this, as ZHA reads all attributes for quirks v2 entities during pairing.

We'll just allow quirks v2 entities to directly control bitmaps in the ffuture. But if we could get this PR in for the next release, that would be nice. The approach with using local attributes (on a real cluster or a `LocalDataCluster`) is to work around this at the moment. Otherwise, the entities would need be directly defined in ZHA, which we're trying to not do anymore.
You can already do almost everything directly in quirks with `LocalDataCluster`s and we can silently replace what's happening in the background in a future release (as the unique IDs for the entities will stay the same).

MindFreeze:
Hopefully I'll be able to test again tomorrow.

@jrfigura You don't need a neutral wire for detached relay if the relay is kept on and the load is big enough to power the switch. A smart bulb in an off state is not enough though.

MindFreeze:
@TheJulianJES Latest code mostly works. I see an event and an "Opening" sensor. The only issue is that the initial state of the detached relay switch was not loaded but toggling works and on subsequent restarts it shows the correct state.

TheJulianJES:
Thanks for testing again! The useless "opening" entities should be removed now and the mask is also read during pairing/reconfiguration, updating the local attributes then.

tests/test_sonoff.py
```diff
@@ -0,0 +1,221 @@
...
+
+        assert local_listener.attribute_updates[0] == (relay_1_attr, False)
+        assert local_listener.attribute_updates[1] == (relay_2_attr, False)
+        assert local_listener.attribute_updates[2] == (relay_3_attr, False)
+
+
+async def test_sonoff_cluster_failed_write_does_not_propagate(
+    zigpy_device_from_v2_quirk,
+):
+    """Test that a failed mask write does not update local relay states."""
```
TheJulianJES, line 143: Note: Zigpy should likely not emit `AttributeWrittenEvent` when the status is not `SUCCESS` ([here](https://github.com/zigpy/zigpy/blob/190043d14476b190a77345c37248f87bd60a65c9/zigpy/zcl/__init__.py#L1346-L1378)). If that's adjusted in zigpy, we can remove the early return in the update handler and this entire test.

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,219 @@
...
+    Relay2 = 0b00000010
+    Relay3 = 0b00000100
+
+
+class SonoffCluster(CustomCluster):
+    """Custom Sonoff cluster for real device attributes."""
+
+    cluster_id = 0xFC11
+    ep_attribute = "sonoff_cluster"
+
```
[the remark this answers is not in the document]
TheJulianJES: This is fine.

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,219 @@
...
+        super().__init__(*args, **kwargs)
+        # TODO: Use _DEFAULT_VALUES when ready, this doesn't work in all circumstances
+        for attr_id in self._RELAY_BITS:
+            if attr_id not in self._attr_cache:
+                self._update_attribute(attr_id, t.Bool.true)
+
+    def update_relay_states(self, mask: int) -> None:
+        """Update individual relay states from a bitmap mask."""
+        for attr_id, bit in self._RELAY_BITS.items():
+            self._update_attribute(attr_id, bool(mask & bit))
```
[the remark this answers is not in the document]
TheJulianJES: This isn't consistent at the moment. `update_attribute` should likely convert the runtime type to the zigpy `Bool` type. Though currently, even restoring from the DB uses Python types, not Zigpy ones.
It doesn't really matter in practice but I'll keep this in mind for potentially adjusting in zigpy.

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,219 @@
...
+        AttributeDefs.relay_3_detached.id: SonoffDetachedRelayMask.Relay3,
+    }
+
+    def __init__(self, *args, **kwargs):
+        """Init with all relays attached by default."""
+        super().__init__(*args, **kwargs)
+        # TODO: Use _DEFAULT_VALUES when ready, this doesn't work in all circumstances
+        for attr_id in self._RELAY_BITS:
+            if attr_id not in self._attr_cache:
+                self._update_attribute(attr_id, t.Bool.true)
```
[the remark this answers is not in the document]
TheJulianJES: These shouldn't be shown now, really. But I've adjusted it now.

TheJulianJES (approved):
This should be good to go. Thanks!

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,219 @@
...
+
+        for attr, value in attributes.items():
+            bit = self._RELAY_BITS.get(self.find_attribute(attr).id)
+            if bit is not None:
+                if value:
+                    mask |= bit
+                else:
+                    mask &= ~bit
+
+        return await self.endpoint.sonoff_cluster.write_attributes({mask_attr_id: mask})
```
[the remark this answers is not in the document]
TheJulianJES: We don't want to pass these, since they're intended for the local cluster, but we then write to the real cluster.

zhaquirks/sonoff/zbm5.py
```diff
@@ -0,0 +1,219 @@
...
+        """Update individual relay states from a bitmap mask."""
+        for attr_id, bit in self._RELAY_BITS.items():
+            self._update_attribute(attr_id, bool(mask & bit))
+
+    async def write_attributes(
+        self,
+        attributes: dict[str | int | ZCLAttributeDef, Any],
+        **kwargs,
+    ) -> list:
+        """Translate per-relay writes into a mask write on real SonoffCluster."""
```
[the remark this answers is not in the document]
TheJulianJES: The return type will likely be adjusted soon, so I think we can keep this pretty raw. It's not really used by anything anyway.

renatoyamane:
Should it be working on 2026.3.4?

[zha-01JY94TZJP8M8Y8RT44B94DNG1-SONOFF ZBM5-1C-80_86-19fc4edc56a8dfc09460bb541005efee.json](https://github.com/user-attachments/files/26508703/zha-01JY94TZJP8M8Y8RT44B94DNG1-SONOFF.ZBM5-1C-80_86-19fc4edc56a8dfc09460bb541005efee.json)

image

TheJulianJES:
No, upgrade to 2026.4.0 or newer.

PR #4200, opened 2025-07-21
Add Tuya 213E ultrasonic water meter `_TZE200_ajlu4cud`

FlorianBruckner:
## Proposed change
while not a valve and just a meter, it is very similar to other devices defined in tuya_valve.py. It allows setting the reporting period, which should also be applicable to 214C, but as I don't have this device I didn't add it to the 214C quirk.

## Checklist


FlorianBruckner:
Before this is being closed automatically - how could we proceed with this? Do you see any risks with this or are there steps I have been missing?

PR #4207, opened 2025-07-27
Allow disabling Sonoff SWV "Water shortage auto-close" 

maga-1:
SWV has a function to turn off water automatically, when no or low water flow is detected. This is a problem for drip irrigation, for example. 
In Firmware version 1.0.4 a function was introduced to turn off this behavior:

* https://www.zigbee2mqtt.io/devices/SWV.html
* https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sonoff.ts

## Proposed change
I propose to add a switch to the existing quirk that allows to use this function from HA.

image

## Additional information

This PR is complementary to #3930. However, accessing "Auto close when water shortage" via ZHA commands is only possible with a modified quirk.

## Checklist


zhaquirks/sonoff/swv.py
```diff
@@ -56,5 +62,13 @@ class AttributeDefs(BaseAttributeDefs):
         translation_key="water_supply",
         fallback_name="Water supply",
     )
+    .switch(
+        CustomSonoffCluster.AttributeDefs.auto_close_when_water_shortage.name,
+        CustomSonoffCluster.cluster_id,
+        off_value=0,
+        on_value=30,
```
TheJulianJES, line 69: What does value `30` mean for the device? Like, what do other values do for example?

TheJulianJES: Ah, Z2M states "Automatically shut down the water valve after the water shortage exceeds 30 minutes".
That makes sense.

zhaquirks/sonoff/swv.py
```diff
@@ -56,5 +62,13 @@ class AttributeDefs(BaseAttributeDefs):
...
+        translation_key="auto_close_when_water_shortage",
+        fallback_name="Auto close when water shortage",
```
TheJulianJES: "Water shortage auto-close" would be slightly less text (and get rid of the "when"):
```suggestion
        fallback_name="Water shortage auto-close",
```

maga-1: Thanks, I think it is a good idea to make the label shorter. I have checked, now it fits on the panel.

zhaquirks/sonoff/swv.py
```diff
@@ -31,6 +31,12 @@ class AttributeDefs(BaseAttributeDefs):
             type=ValveState,
         )
 
+        auto_close_when_water_shortage = ZCLAttributeDef(
```
TheJulianJES:
```suggestion
        auto_close_water_shortage = ZCLAttributeDef(
```

zhaquirks/sonoff/swv.py
```diff
@@ -56,5 +62,13 @@ class AttributeDefs(BaseAttributeDefs):
         translation_key="water_supply",
         fallback_name="Water supply",
     )
+    .switch(
+        CustomSonoffCluster.AttributeDefs.auto_close_when_water_shortage.name,
```
TheJulianJES:
```suggestion
        CustomSonoffCluster.AttributeDefs.auto_close_water_shortage.name,
```

zhaquirks/sonoff/swv.py
```diff
@@ -56,5 +62,13 @@ class AttributeDefs(BaseAttributeDefs):
...
+        CustomSonoffCluster.cluster_id,
+        off_value=0,
+        on_value=30,
+        translation_key="auto_close_when_water_shortage",
```
TheJulianJES:
```suggestion
        translation_key="water_shortage_auto_close",
```

TheJulianJES (commented):
Ok, I'll change up a few names and then merge this.

TheJulianJES (approved):

maga-1:
Thank you for handling this PR!

TheJulianJES:
Thanks for creating it! 😄

PR #4208, opened 2025-07-27
update ruff

carlolars:
## Proposed change

Hi!

I did some cleaning up and changes to the develop environment. I figured since you now use uv then why not go all in.

- tox.ini is outdated and likely not used
- No need for extra pip-install (removed from README)
- Removed traces of isort, black and flake8 since using ruff instead
- Moved requirements into dev dependency-group in pyproject.toml, note that requirements_test.txt is probably still must exist
- pre-commit use tools from the local venv, this ensures the same versions are used without having to change in several places
- Fixed "warning: `TRY302` has been remapped to `TRY203`."
- Updated ruff to 0.12.5, removed selecting rule S320 as it has been removed
- Updated formatting and fixed linting

## Checklist


carlolars:
Meh that failed fast, and fixing it would need changes in zigpy/workflows

carlolars:
I'm closing this PR and splitting it in two parts, the first here #4241

PR #4214, opened 2025-07-30
Begin migrating simple v1 quirks to v2

puddly:
## Proposed change

This PR migrates a handful of very simple v1 quirks to v2 quirks: only those that replace clusters, modify the device type, add automation triggers, etc. We can do more in a future pass. My goal is to get rid of v1 quirks entirely and build out the APIs we need for v2 quirks to do that.

These quirks have *not* been tested and many will be broken. This is more or less a test that will be brought into a working state *eventually*.

## Checklist

### DIFF
--- a/zhaquirks/bitron/thermostat.py
+++ b/zhaquirks/bitron/thermostat.py
@@ -2,6 +2,7 @@
 
 from zigpy.profiles import zha
 from zigpy.quirks import CustomDevice
+from zigpy.quirks.v2 import QuirkBuilder
 from zigpy.zcl.clusters.general import (
     Basic,
     Identify,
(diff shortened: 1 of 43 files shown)


zhaquirks/centralite/cl_3310S.py
```diff
@@ -38,50 +27,17 @@ class AttributeDefs(BaseAttributeDefs):
...
+    .applies_to(CENTRALITE, "3310-S")
+    .applies_to(CENTRALITE, "3310")
+    .replaces(
+        replacement_cluster_class=PowerConfigurationCluster,
+        cluster_id=PowerConfigurationCluster.cluster_id,
+    )
+    .replaces(
+        replacement_cluster_class=SmartthingsRelativeHumidityCluster,
+        cluster_id=SMRT_THINGS_REL_HUM_CLSTR,
+    )
```
TheJulianJES, line 41: Providing `cluster_id` seems a bit unnecessary, since that's already set in the `Cluster` class itself?
```suggestion
    .replaces(replacement_cluster_class=PowerConfigurationCluster)
    .replaces(replacement_cluster_class=SmartthingsRelativeHumidityCluster)
```

zhaquirks/bitron/thermostat.py
```diff
@@ -36,54 +17,8 @@ class Av201032PowerConfigurationCluster(PowerConfigurationCluster):
...
-                    UserInterface.cluster_id,
-                    Diagnostic.cluster_id,
-                ],
-                OUTPUT_CLUSTERS: [Identify.cluster_id, Ota.cluster_id],
-            }
-        },
-    }
+(
+    QuirkBuilder(BITRON, "902010/32")
+    .replaces(replacement_cluster_class=Av201032PowerConfigurationCluster)
```
TheJulianJES, line 22: We also generally didn't specify the `replacement_cluster_class` kwarg for `replaces`.
I agree it's useful for everything, but not for the cluster that should be replaced/added IMO. Just a very long kwarg 😅 

It's used everywhere in this PR and I'm fine with it now, since we can always change that later.

zhaquirks/echostar/bell.py
```diff
@@ -19,70 +10,34 @@
...
-    ENDPOINTS,
-    INPUT_CLUSTERS,
-    MODELS_INFO,
-    OUTPUT_CLUSTERS,
-    PROFILE_ID,
     SHORT_PRESS,
 )
 
 MANUFACTURER = " Echostar"
 MODEL = "   Bell"
```
TheJulianJES, line 19: These constants are not used anymore and should be removed then:

```suggestion
```

zhaquirks/edpwithus/redy_plug.py
```diff
@@ -1,65 +1,17 @@
...
-                "output_clusters": [Ota.cluster_id],
-            }
-        }
-    }
+(
+    QuirkBuilder("EDP-WITHUS", "REDY")  # codespell:ignore
+    .replaces_endpoint(
+        endpoint_id=85,
+        profile_id=zha.PROFILE_ID,
+        device_type=zha.DeviceType.ON_OFF_PLUG_IN_UNIT,
```
TheJulianJES, line 13: side note: Weird that we replace `MAIN_POWER_OUTLET` with `ON_OFF_PLUG_IN_UNIT` in this quirk?
Should be kept for this, but do we not handle `MAIN_POWER_OUTLET` correctly somehow?

TheJulianJES (commented):
Couple of comments (for the first few quirks I looked at):

TheJulianJES:
Wait, I'm looking at the wrong PR...

TheJulianJES:
- https://github.com/zigpy/zha-device-handlers/pull/4499 replaces this.

zhaquirks/bitron/thermostat.py
puddly: I must have missed a few of these. I'll remove them.

puddly: Oh, never mind, I just saw your comment below 😄

PR #4215, opened 2025-07-30
Change datapoints for Tuya soil sensor `_TZE284_awepdiwi`

rwalker777:
## Checklist


rwalker777:
@TheJulianJES Without this the soil moisture for the device I added is not accurate.  It needs the multiplier.

PR #4233, opened 2025-08-04
Fix Tuya TH01Z temperature sensors not reporting negative temp

### LINKED ISSUE #4192: [BUG] _TZE200_qrztc3ev TS0601 quirk errors
### Bug description

_TZE200_qrztc3ev TS0601 temperature and humidity sensor is supported by the general tuya sensor quirk, but it have some errors (at least for this device).
 
1. The correct alarm enum is:
```python
class TuyaNousTempHumiAlarm(t.enum8):
    """Tuya temperature and humidity alarm enum."""

    Canceled = 0x00
    LowerAlarm = 0x01
    UpperAlarm = 0x02
```

2. The minimal accepted temperature alarm value is 0 instead of -20. You get error message if you try to set below 0.

### Steps to reproduce

1. Open the sensor configuration in ZHA
2. Set the alarm limits
3. See the result.

### Expected behavior

The alarm states are consistent with the limits are set.

### Screenshots/Video

</details>

### Diagnostics information

_No response_

### Device signature

```json

{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4742,
    "maximum_buffer_size": 74,
    "maximum_incoming_transfer_size": 404,
    "server_mask": 10752,
    "maximum_outgoing_transfer_size": 404,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0302",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0004",
        "0x0020",
        "0x0402",
        "0x0405",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE200_qrztc3ev",
  "model": "TS0601",
  "class": "zhaquirks.tuya.builder.EnchantedDeviceV2"
}

```

</details>

### Logs

```python

[Paste the logs here]

```

</details>

### Additional information

_No response_

MegaGio94:
## Proposed change

This PR fixes issue #4192 where the TH01Z quirk (_TZE200_qrztc3ev TS0601) generates errors when setting negative temperature alarm values.

The device supports temperature range from -20°C to 60°C according to datasheet, but the quirk had three main issues:

1. **Incorrect enum values** for `TuyaNousTempHumiAlarm` 
2. **Wrong data type** (`t.uint16_t` instead of `t.int16s`) preventing negative temperature support for alarm settings
3. **No negative temperature testing** - none of TH01Z device was tested with negative values

### Changes made:
-  **Fixed enum values**: `Canceled=0x00`, `LowerAlarm=0x01`, `UpperAlarm=0x02` (was 0x02, 0x00, 0x01)
-  **Changed data types**: `t.uint16_t` → `t.int16s` for alarm temperature parameters to support negative values
-  **Added negative temperature support**: Custom converter for TH01Z quirk
-  **Enabled comprehensive testing**: Updated test parameters for all TH01Z devices to include negative temperature testing

## Additional information

**Devices affected**: All devices using TH01Z quirk (Temperature & humidity sensor with clock):
- `_TZE200_qrztc3ev` (primary device from issue)
- `_TZE200_lve3dvpy`, `_TZE200_c7emyjom`, `_TZE200_locansqn`
- `_TZE200_snloy4rw`, `_TZE200_eanjj2pa`, `_TZE200_ydrdfkim`
- `_TZE284_locansqn`, `_TZE200_vvmbj46n`

**Testing**: All existing tests pass (22/22) including new negative temperature tests for all TH01Z devices.

**No breaking changes**: The modifications are backward compatible and only fix existing functionality.

Fixes #4192

## Checklist

### DIFF
--- a/zhaquirks/tuya/tuya_sensor.py
+++ b/zhaquirks/tuya/tuya_sensor.py
@@ -27,9 +27,9 @@ class TuyaTempUnitConvert(t.enum8):
 class TuyaNousTempHumiAlarm(t.enum8):
     """Tuya temperature and humidity alarm enum."""
 
-    LowerAlarm = 0x00
-    UpperAlarm = 0x01
-    Canceled = 0x02
+    Canceled = 0x00
+    LowerAlarm = 0x01
+    UpperAlarm = 0x02
 
 
 class NoManufTimeTuyaMCUCluster(TuyaMCUCluster):
@@ -116,7 +116,14 @@ class NoManufTimeTuyaMCUCluster(TuyaMCUCluster):
     .applies_to("_TZE200_w6n8jeuu", "TS0601")
     .applies_to("_TZE200_vvmbj46n", "TS0601")
     .applies_to("_TZE284_vvmbj46n", "TS0601")
-    .tuya_temperature(dp_id=1, scale=10)
+    # Not using tuya_temperature because device reports negative values incorrectly
+    .tuya_dp(
+        dp_id=1,
+        ep_attribute=TuyaTemperatureMeasurement.ep_attribute,
+        attribute_name=TuyaTemperatureMeasurement.AttributeDefs.measured_value.name,
+        converter=lambda x: ((x - 0xFFFF if x > 0x2000 else x) * 10),
+    )
+    .adds(TuyaTemperatureMeasurement)
     .tuya_humidity(dp_id=2)
     .tuya_battery(dp_id=4)
     .tuya_number(
@@ -165,7 +172,7 @@ class NoManufTimeTuyaMCUCluster(TuyaMCUCluster):
     .tuya_number(
         dp_id=10,
         attribute_name="alarm_temperature_max",
-        type=t.uint16_t,
+        type=t.int16s,
         unit=UnitOfTemperature.CELSIUS,
         min_value=-20,
         max_value=60,
@@ -178,7 +185,7 @@ class NoManufTimeTuyaMCUCluster(TuyaMCUCluster):
     .tuya_number(
         dp_id=11,
         attribute_name="alarm_temperature_min",
-        type=t.uint16_t,
+        type=t.int16s,
         unit=UnitOfTemperature.CELSIUS,
         min_value=-20,
         max_value=60,
--- a/tests/test_tuya_sensor.py
+++ b/tests/test_tuya_sensor.py
@@ -36,15 +36,15 @@ zhaquirks.setup()
         ("_TZE200_bq5c8xfe", "TS0601", 100, 10, True),
         ("_TZE200_vs0skpuc", "TS0601", 100, 10, True),
         ("_TZE200_44af8vyi", "TS0601", 100, 10, True),
-        ("_TZE200_lve3dvpy", "TS0601", 100, 10, False),  # TH01Z - Temp & humid w/ clock
-        ("_TZE200_c7emyjom", "TS0601", 100, 10, False),
-        ("_TZE200_locansqn", "TS0601", 100, 10, False),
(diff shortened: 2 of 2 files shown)


zhaquirks/tuya/tuya_sensor.py
```diff
@@ -116,7 +116,14 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
     .applies_to("_TZE200_w6n8jeuu", "TS0601")
     .applies_to("_TZE200_vvmbj46n", "TS0601")
     .applies_to("_TZE284_vvmbj46n", "TS0601")
-    .tuya_temperature(dp_id=1, scale=10)
+    # Not using tuya_temperature because device reports negative values incorrectly
+    .tuya_dp(
+        dp_id=1,
+        ep_attribute=TuyaTemperatureMeasurement.ep_attribute,
+        attribute_name=TuyaTemperatureMeasurement.AttributeDefs.measured_value.name,
+        converter=lambda x: ((x - 0xFFFF if x > 0x2000 else x) * 10),
```
puddly, line 124: `TuyaTemperatureMeasurement.AttributeDefs.measured_value` is already a `int16s` so this converter I think isn't necessary. Unrelated, but I think the converter logic I think is also invalid: `2000`, for example, would be turned into `-63535`, which is both outside the range of a signed 16-bit integer and also invalid.

puddly:
@TheJulianJES I've actually seeing `x - 0xFFFF if x > 0x2000 else x` in quite a few spots in the Tuya quirks. To me, this looks like a bad unsigned -> signed conversion. The actual effect of this discrepancy is *really* minuscule due to ZCL attribute division, since it is effectively an off-by-one error for all negative numbers.

I think the Tuya datapoint's data type should be a signed 16 bit integer, no?

zhaquirks/tuya/tuya_sensor.py
MegaGio94: You're absolutely right on both points. I used the same converter pattern that appears in several other Tuya quirks in the codebase (I can see it in at least 3-4 different device definitions), assuming it was the correct approach since it was so widespread.
But you're correct that:
If TuyaTemperatureMeasurement.AttributeDefs.measured_value is already int16s, then no conversion should be needed.
The logic is indeed flawed - as you pointed out, 2000 would become -63535, which is invalid
Should I:
Remove the converter entirely from this quirk since the attribute is already the correct type?
Open a separate issue/PR to fix this pattern in the other existing quirks where it appears?

TheJulianJES:
I'll need to have a closer look at this in the next few days or so. I do remember that this was already adjusted in some quirks when porting them from v1 to v2 (most of what prairiesnpr did in January). I think we already removed a bunch of that explicit conversion then(?)
 
It seems to be ported from Z2M generally. It's also used in other places there, but for the Tuya air sensors, see:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/b8cd9aa10bb6c0e8f21d9721ced34c398a4ab6d3/src/lib/legacy.ts#L2097

> I think the Tuya datapoint's data type should be a signed 16 bit integer, no?

We never assign data types to the Tuya datapoints IIRC. I think that's done *automagically* by some old Tuya code. That might be an issue though...
`tuya_temperature` just remaps the Tuya DP to the a fake/local `TemperatureMeasurement` ZCL cluster, where the data type for the `measured_value` attribute is already a signed int. But yeah, the issue might already be in the Tuya part.

PR #4239, opened 2025-08-06
Silence swapped direction warning for Aqara devices

puddly:
## Proposed change

Reported here: https://github.com/zigpy/zigpy/issues/1640#issuecomment-3160504392

Currently, Aqara devices emit the "swapped direction" warning because we try to use `_find_zcl_cluster` instead of  `_find_zcl_cluster_strict` when implementing the swapping logic, triggering the zigpy warning.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ readme = "README.md"
 license = {text = "Apache-2.0"}
 requires-python = ">=3.12"
 dependencies = [
-    "zigpy>=0.82.0",
+    "zigpy>=0.82.2",
 ]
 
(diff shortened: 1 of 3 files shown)


TheJulianJES (commented):
We may also wanna bump zigpy to [0.82.2](https://github.com/zigpy/zigpy/releases/tag/0.82.2) in quirks to make sure tests run with that version

TheJulianJES (approved):

puddly:
Ah, good point. I was testing locally but I have editable installs so it's always the latest for me.

PR #4241, opened 2025-08-07
Manage dependencies with uv

carlolars:
## Proposed change

Use `uv` for dependency management and lock packet versions to have a reproducable environment.

## Checklist


README.md
```diff
@@ -560,6 +552,38 @@ def test_ts0121_signature(assert_signature_matches_quirk):
     assert_signature_matches_quirk(zhaquirks.tuya.ts0121_plug.Plug, signature)
 ```
 
+## Managing dependencies
+
+Running `uv sync` will install all packets using the locked version in the `uv.lock` file. The setup script will do this for you.
```
puddly: I think the word `packet` should be replaced with `package`, there are quite a few occurrences in this section.

carlolars: You are right.

I asked CoPilot to review the branch in my fork :smile:
bild

I've fixed the wording and did some other improvements after the feedback from CoPilot.

script/setup
```diff
@@ -17,18 +17,17 @@ fi
...
 if [ ! -n "$VIRTUAL_ENV" ]; then
   if [ -x "$(command -v uv)" ]; then
-    uv venv venv
+    uv venv
   else
-    python3 -m venv venv
+    python3 -m venv .venv
   fi
-  source venv/bin/activate
+  source .venv/bin/activate
```
TheJulianJES, line 24: Changing the venv folder from `venv` to `.venv` is not ideal, as this kind of breaks the setup script for existing setups. Do we wanna change it back?

PR #4244, opened 2025-08-08
Add `NimlyPRO24` model variant for Nimly Touch Pro

andersgenberg:
## Proposed change

I bought a new Nimly Touch Pro lock and paired it with ZHA. The pairing worked, but the only entity that appeared was the lock entity.

I did some digging and found out that the standard quirk for Nimly Touch Pro handled the model named "NimlyPRO", "NimlyCode", "NimlyTouch" and "NimlyIn" where as the model my lock have is "NimlyPRO24".

## Additional information

I made a custom quirk, a copy of the standard quirk, and added the missing model name and the missing entities appeared in my unit. 

## Checklist

### DIFF
--- a/zhaquirks/nimly/lock.py
+++ b/zhaquirks/nimly/lock.py
@@ -148,7 +148,8 @@ def last_action_user_converter(value: int) -> int:
 
 
 (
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
Thanks!

PR #4255, opened 2025-08-13
Move Tuya remote commands to correct cluster

puddly:
## Proposed change

Fix command definitions for `_TYZB01_1xktopx6` `TS0041A`. From https://github.com/home-assistant/core/issues/150528, it appears that the definition we have right now is incorrect:

```python
2025-08-13 00:02:10.780 DEBUG (MainThread) [zigpy.zcl] [0xBA95:1:0x0006] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl<0x01>(frame_type=<FrameType.CLUSTER_COMMAND: 1>, is_manufacturer_specific=0, direction=<Direction.Client_to_Server: 0>, disable_default_response=0, reserved=0, *is_cluster=True, *is_general=False), tsn=100, command_id=253, *direction=<Direction.Client_to_Server: 0>)
2025-08-13 00:02:10.785 DEBUG (MainThread) [zigpy.zcl] [0xBA95:1:0x0006] Unknown cluster command 253 b'\x00'
2025-08-13 00:02:10.785 WARNING (MainThread) [zigpy.listeners] Matcher checkin() and command ZCLHeader(frame_control=FrameControl<0x01>(frame_type=<FrameType.CLUSTER_COMMAND: 1>, is_manufacturer_specific=0, direction=<Direction.Client_to_Server: 0>, disable_default_response=0, reserved=0, *is_cluster=True, *is_general=False), tsn=100, command_id=253, *direction=<Direction.Client_to_Server: 0>) b'\x00' are incompatible
2025-08-13 00:02:10.789 DEBUG (MainThread) [zigpy.zcl] [0xBA95:1:0x0006] Received command 0xFD (TSN 100): b'\x00'
2025-08-13 00:02:10.790 DEBUG (MainThread) [zigpy.zcl] [0xBA95:1:0x0006] No explicit handler for cluster command 0xfd: b'\x00'
```

Whether this change is valid for other devices is debatable. I believe it is, it would make sense for these to be client commands.

## Device diagnostics

[zha-9b493435cd2e247b918e1d13f0096c87-_TYZB01_1xktopx6.TS0041A-98ff68bf6f977b05335622602f29deb8.json](https://github.com/user-attachments/files/21746256/zha-9b493435cd2e247b918e1d13f0096c87-_TYZB01_1xktopx6.TS0041A-98ff68bf6f977b05335622602f29deb8.json)

## Checklist

### DIFF
--- a/zhaquirks/tuya/__init__.py
+++ b/zhaquirks/tuya/__init__.py
@@ -1065,7 +1065,7 @@ class TuyaSmartRemoteOnOffCluster(OnOff, EventableCluster):
         self.last_tsn = -1
         super().__init__(*args, **kwargs)
 
-    class ServerCommandDefs(OnOff.ServerCommandDefs):
+    class ClientCommandDefs(OnOff.ClientCommandDefs):
         """Server command definitions."""
 
         rotate_type = foundation.ZCLCommandDef(


zhaquirks/tuya/__init__.py
```diff
@@ -1065,7 +1065,7 @@ def __init__(self, *args, **kwargs):
         self.last_tsn = -1
         super().__init__(*args, **kwargs)
 
-    class ServerCommandDefs(OnOff.ServerCommandDefs):
+    class ClientCommandDefs(OnOff.ClientCommandDefs):
         """Server command definitions."""
```
TheJulianJES, line 1069: nit: we should also change the docstring if we adjust this
```suggestion
        """Client command definitions."""
```

TheJulianJES (commented):
It's late, so I might think about this the wrong way (and the previously swapped `Direction` enum doesn't help my memories), but here we go:

> Whether this change is valid for other devices is debatable. I believe it is, it would make sense for these to be client commands.

In some quirks, including [the one](https://github.com/zigpy/zha-device-handlers/blob/48191cc6ca52b7ad9746d433251638f6966004ac/zhaquirks/tuya/ts0041.py#L189-L233) in the HA issue, `TuyaSmartRemoteOnOffCluster` is used as an input/server cluster, where the client command definitions would make sense, since the remote (server) generates them to be received by ZHA (client).

However, in other quirks, it's used as a output/client cluster. There, the cluster on the device assumes the role of client and ZHA the one of server. If the device now wants to send a command to ZHA, it should be a server command (definition), since the device/client generates it, in order for that to be received by the server/ZHA.
(See `OnOff` as an output/client cluster for binding remotes to lights/plugs as an example. Those are all server cmd defs.)

Looking at the direction from the logs you included, it's `Client_to_Server`. This means the commands come from an output/client cluster. If that's the case, they should be server command definitions, since they're intended to be received by the server. (Same `OnOff` output/client binding example again to make this clear.)

However, this PR changes the command definitions to be client ones, meaning they're supposed to be received by a client, which again, seems correct, since the quirk uses an input/server remote cluster. But, the direction is still `Client_to_Server`, so it seems to me that it should not be an input/server cluster in the quirk and also not client command definitions, but server command definitions.

**So, is the only issue maybe just that some of the quirks are wrong where the Tuya remote cluster is used as a input/server cluster, instead of a output/client one?**

PR #4271, opened 2025-08-17
Add Legrand devices: Connected Outlet, Micromodule Switch, Shutter Switch, Remote Shutter Switch, Switch and Dimmer

rapgenic:
## Proposed change

Add some custom Legrand clusters:
* `LegrandIdentify` which extends the `Identify` cluster to support the custom `identify` command needed for Legrand devices, which requires sending a `trigger_effect` command before.
* `LegrandShutterCluster` which extends the `Shutter` cluster adding a `calibration_mode` enum attribute

Support the new following Legrand devices:
* Connected Outlet (I own [this](https://catalogo.bticino.it/prodotto/soluzioni-per-edifici-residenziali-e-nel-terziario/livinglight/livinglight-smart/BTI-N4531C-IT) equivalent by BTicino)
* Micromodule Switch (I own [this](https://catalogo.bticino.it/prodotto/soluzioni-per-edifici-residenziali-e-nel-terziario/livinglight/livinglight-smart/BTI-3584C-IT) equivalent by BTicino)
* Shutter Switches (I own [this](https://catalogo.bticino.it/prodotto/soluzioni-per-edifici-residenziali-e-nel-terziario/livinglight/livinglight-smart/BTI-N4027C-IT) equivalent by BTicino)
* Remote Shutter Switches (I own [this](https://catalogo.bticino.it/prodotto/soluzioni-per-edifici-residenziali-e-nel-terziario/livinglight/livinglight-smart/comandi-wireless-per-gestione-luci-e-tapparelle/BTI-LN4027CWI-IT) equivalent by BTicino)

Convert to quirk V2 the following Legrand devices:
* Switch (I own [this](https://catalogo.bticino.it/prodotto/soluzioni-per-edifici-residenziali-e-nel-terziario/livinglight/livinglight-smart/BTI-N4003C-IT) equivalent by BTcino)
* Dimmer with Neutral (I own [this](https://catalogo.bticino.it/prodotto/soluzioni-per-edifici-residenziali-e-nel-terziario/livinglight/livinglight-smart/BTI-N4411C-IT) equivalent by BTcino)

## Additional information

* The `LegrandCluster` manufacturer specific cluster has been slightly refactored renaming the `dimmer` attribute to `device_mode` and adding some values to be consistent with the Z2M implementation (see [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/8e9404b28c6b4ee032bb29d6528695e3d260d38f/src/lib/legrand.ts#L205)) 
* I removed from all devices I implemented, the binary input entity (which is non functional)
* I removed from all devices the "opening" entity associated with the OnOff cluster (which is also non functional)
* I added toggle entities to control the LED behavior (led_on_when_on and led_on_when_off, which turn on the integrated LED respectively when the device is in the on or off state)

Information has been derived from the [Legrand developer portal](https://developer.legrand.com/local-interoperability/#Products%20cluster%20list) and the Z2M implementation.

Commits have been separated by domain for easier review, but only tested as a whole (single commits might not be functional)

## Device diagnostics

* Connected outlet: [zha-01K1JQT83XMY0X6WTRD4FQQ602- Legrand  Connected outlet-529ebe0364e3213c61a16076c44cc0c9.json](https://github.com/user-attachments/files/21822953/zha-01K1JQT83XMY0X6WTRD4FQQ602-.Legrand.Connected.outlet-529ebe0364e3213c61a16076c44cc0c9.json)
* Micromodule switch: [zha-01K1JQT83XMY0X6WTRD4FQQ602- Legrand  Micromodule switch-088bf7986cfa77914a6928421d9cbdb6.json](https://github.com/user-attachments/files/21822963/zha-01K1JQT83XMY0X6WTRD4FQQ602-.Legrand.Micromodule.switch-088bf7986cfa77914a6928421d9cbdb6.json)
* Shutter switch: [zha-01K1JQT83XMY0X6WTRD4FQQ602- Legrand  Shutter SW with level control-2069c9719e8762c86d12a52448a28351.json](https://github.com/user-attachments/files/21822966/zha-01K1JQT83XMY0X6WTRD4FQQ602-.Legrand.Shutter.SW.with.level.control-2069c9719e8762c86d12a52448a28351.json)
* Remote shutter switch: [zha-01K1JQT83XMY0X6WTRD4FQQ602- Legrand  Shutters central remote switch-5b1504d6d6f62e29083617f5f6db65ea.json](https://github.com/user-attachments/files/21822969/zha-01K1JQT83XMY0X6WTRD4FQQ602-.Legrand.Shutters.central.remote.switch-5b1504d6d6f62e29083617f5f6db65ea.json)
* Switch: [zha-01K1JQT83XMY0X6WTRD4FQQ602- Legrand  Light switch with neutral-b82c9674f6a570bf286e0c4a803b2328.json](https://github.com/user-attachments/files/21822972/zha-01K1JQT83XMY0X6WTRD4FQQ602-.Legrand.Light.switch.with.neutral-b82c9674f6a570bf286e0c4a803b2328.json)
* Dimmer: [zha-01K1JQT83XMY0X6WTRD4FQQ602- Legrand  Dimmer switch with neutral-1957f2f675e97488d01ee2abd9d9a068 (2).json](https://github.com/user-attachments/files/21822973/zha-01K1JQT83XMY0X6WTRD4FQQ602-.Legrand.Dimmer.switch.with.neutral-1957f2f675e97488d01ee2abd9d9a068.2.json)

## Checklist

- [x] The changes are tested and work correctly (**NOTE**: I could not test _exactly_ the files uploaded here, for example i cannot put a `__init__.py` file in the custom_zha_quirks folder so I had to modify them slightly. Any help on how to test the entire quirks package for example by replacing it is appreciated - consider I am using a HA Green HW) 
- [x] Tests have been added to verify that the new code works (**NOTE**: More than happy to write some, I just need some guidance on what to test and how)

### DIFF
--- a/zhaquirks/legrand/__init__.py
+++ b/zhaquirks/legrand/__init__.py
@@ -62,29 +62,28 @@ class LegrandIdentify(CustomCluster, Identify):
         self,
         command_id: GeneralCommand | int | t.uint8_t,
         *args: Any,
-        manufacturer: int | t.uint16_t | None = None,
-        expect_reply: bool = True,
-        tsn: int | t.uint8_t | None = None,
         **kwargs: Any,
     ) -> Any:
         """Override the command method to customize the identify command."""
 
         if command_id == Identify.ServerCommandDefs.identify.id:
+            # Remove identify command specific arguments
+            identify_time = kwargs.pop("identify_time", None)
+
             await super().command(
                 command_id=Identify.ServerCommandDefs.trigger_effect.id,
                 effect_id=EffectIdentifier.Blink,
                 effect_variant=EffectVariant.Default,
-                manufacturer=manufacturer,
-                expect_reply=expect_reply,
-                tsn=tsn,
+                **kwargs,
             )
+
+            # Restore identify command specific arguments
+            if identify_time is not None:
+                kwargs["identify_time"] = identify_time
 
         return await super().command(
             command_id,
             *args,
-            manufacturer=manufacturer,
-            expect_reply=expect_reply,
-            tsn=tsn,
             **kwargs,
         )
 
--- a/zhaquirks/legrand/__init__.py
+++ b/zhaquirks/legrand/__init__.py
@@ -1,13 +1,36 @@
 """Module for Legrand devices."""
 
+from typing import Any
+
 from zigpy.quirks import CustomCluster
 import zigpy.types as t
-from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+from zigpy.zcl.clusters.closures import WindowCovering
+from zigpy.zcl.clusters.general import EffectIdentifier, EffectVariant, Identify
+from zigpy.zcl.foundation import (
+    BaseAttributeDefs,
+    DataTypeId,
+    GeneralCommand,
+    ZCLAttributeDef,
+)
 
 from zhaquirks import PowerConfigurationCluster
 
 LEGRAND = "Legrand"
 MANUFACTURER_SPECIFIC_CLUSTER_ID = 0xFC01  # decimal = 64513
+
+
+class DeviceMode(t.enum16):
+    """Device modes for some Legrand devices. To be used in custom cluster.
+
+    Note: some values are taken from the Z2M code.
+    """
+
+    Pilot_Off = 0x0001
+    Pilot_On = 0x0002
+    Switch = 0x0003
+    Auto = 0x0004
+    Dimmer_Off = 0x0100
+    Dimmer_On = 0x0101
 
 
 class LegrandCluster(CustomCluster):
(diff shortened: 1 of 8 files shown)


zhaquirks/legrand/__init__.py
```diff
@@ -20,15 +43,75 @@ class LegrandCluster(CustomCluster):
...
+        """Override the command method to customize the identify command."""
+
+        if command_id == Identify.ServerCommandDefs.identify.id:
+            await super().command(
+                command_id=Identify.ServerCommandDefs.trigger_effect.id,
+                effect_id=EffectIdentifier.Blink,
+                effect_variant=EffectVariant.Default,
+                manufacturer=manufacturer,
+                expect_reply=expect_reply,
+                tsn=tsn,
```
puddly: We should pass through the remaining `**kwargs` as well. You can `kwargs.pop("identify_time", None)` beforehand to make sure that any kwargs for `identify` aren't passed through erroneously.

zhaquirks/legrand/__init__.py
```diff
@@ -20,15 +43,75 @@ class LegrandCluster(CustomCluster):
...
+class LegrandIdentify(CustomCluster, Identify):
+    """Custom Identify cluster for Legrand devices. Replaces the identify command."""
+
+    async def command(
+        self,
+        command_id: GeneralCommand | int | t.uint8_t,
+        *args: Any,
+        manufacturer: int | t.uint16_t | None = None,
+        expect_reply: bool = True,
+        tsn: int | t.uint8_t | None = None,
```
puddly: I would also just get rid of `manufacturer`, `expect_reply`, and `tsn` in their entirety, since you only really need access to `command_id`.

puddly:
This looks great! I'll get the diagnostic JSON into the ZHA integration next week so that we can test to see if there will be any entity regressions for existing users.

rapgenic:
@puddly Thank you for the _extremely_ fast review!

I pushed a commit (I also tested it right now) that should solve the issues you raised in the comments, please tell me if it is what you had in mind.

I also don't understand why now the codecov is not passing whereas it was before...

puddly:
> I also don't understand why now the codecov is not passing whereas it was before...

There's now runtime-executed code (as opposed to import-time, like most quirks) so `await device.endpoints[1].identify.identify(identify_time=1234)` needs to be tested. It may not have been before.

rapgenic:
I added a new test for the identify command. I hardcoded the expected values, as the test should raise an error if someone were to change ids or parameter values unexpectedly, I think.

zhaquirks/legrand/connected_outled.py
```diff
@@ -0,0 +1,31 @@
...
+
+(
+    QuirkBuilder(f" {LEGRAND}", " Connected outlet")
+    .replaces(LegrandCluster)
+    .replaces(LegrandIdentify)
+    .prevent_default_entity_creation(endpoint_id=1, cluster_id=BinaryInput.cluster_id)
+    .prevent_default_entity_creation(
+        endpoint_id=1,
+        cluster_id=OnOff.cluster_id,
+        function=lambda entity: entity.device_class == "opening",
```
TheJulianJES, line 16: This is *fine*, but can we use `cluster_type=ClusterType.Client` instead of the `function`/device class stuff here?
That should also only match the ZHA opening binary sensor (for the Client `OnOff` cluster).
That would be a bit cleaner, since we don't need to access some "unknown ZHA stuff" (which could change and we might notice the breakage a bit late).

(If this works, it's also applicable to all other places where this was done.)

zhaquirks/legrand/__init__.py
```diff
@@ -20,15 +43,74 @@ class LegrandCluster(CustomCluster):
     class AttributeDefs(BaseAttributeDefs):
         """Cluster attributes."""
 
-        dimmer = ZCLAttributeDef(
-            id=0x0000, type=t.data16, is_manufacturer_specific=True
+        device_mode = ZCLAttributeDef(
+            id=0x0000,
+            type=DeviceMode,
```
TheJulianJES, line 48: This could be left like this, but do all devices actually accept all `DeviceMode` options?
Alternatively, we would only have different custom clusters per device (type) with matching enum options only.

zhaquirks/legrand/shutter.py
```diff
@@ -0,0 +1,45 @@
...
+        cluster_id=LegrandCluster.cluster_id,
+        translation_key="turn_on_led_when_on",
+        fallback_name="Turn on LED when on",
+    )
+    .enum(
+        attribute_name=LegrandShutterCluster.AttributeDefs.calibration_mode.name,
+        cluster_id=LegrandShutterCluster.cluster_id,
+        translation_key="calibration_mode",
+        fallback_name="Calibration mode",
+        enum_class=ShutterCalibrationMode,
```
TheJulianJES: Let's keep `translation_key` and then `fallback_name` at the bottom for all quirks v2 entities, e.g. here:
```suggestion
        enum_class=ShutterCalibrationMode,
        translation_key="calibration_mode",
        fallback_name="Calibration mode",
```

TheJulianJES:
> Any help on how to test the entire quirks package for example by replacing it is appreciated - consider I am using a HA Green HW)

A simple way is to just replace the quirks package at `/usr/local/lib/python3.13/site-packages/zhaquirks/`. You can copy over the modified package from `/config` for example, where you can put it using the Samba addon.
You can also use some (third-party) Portainer addon to access the shell of the Core `homeassistant` container, though you may need to unhide it if you don't see it.

Reinstalling some Core update will reset the whole container, so will revert `zhaquirks` to its original state.
But this PR is tested well enough that you likely don't need to do this here.

zhaquirks/legrand/connected_outled.py
rapgenic: That was my initial approach, however I cannot make it work... It will not make a difference whether I write `cluster_type=ClusterType.Client` or `cluster_type=ClusterType.Server`, it seems as if it was always interpreted as both, and in fact the main light/switch entity disappears _along with_ the opening entity.

TheJulianJES: Ok, we can leave it like this then. Created an issue to look at this later: https://github.com/zigpy/zha-device-handlers/issues/4284

TheJulianJES (approved):
Should be good to go

PR #4272, opened 2025-08-17
Fix Tuya Hiking power meter "momentary power" type

ofcaah:
Power can go into negative, as these meters are bi-directional. Switch uint16_t to int32s.

### DIFF
--- a/zhaquirks/tuya/ts0601_din_power.py
+++ b/zhaquirks/tuya/ts0601_din_power.py
@@ -182,7 +182,7 @@ class HikingManufClusterDinPower(TuyaManufClusterAttributes):
             is_manufacturer_specific=True,
         )
         power: Final = ZCLAttributeDef(
-            id=HIKING_POWER_ATTR, type=t.uint16_t, is_manufacturer_specific=True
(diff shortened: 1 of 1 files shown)


TheJulianJES (commented):
The underlying `active_power` attribute (on the "local/fake" `ElectricalMeasurement` cluster we use) is only a 16 bit signed int. I think we also have to use that for the Tuya attribute then.
Reference:
https://github.com/zigpy/zigpy/blob/88c3ba6fd84c3932b89996f04673edcdaf15ddcc/zigpy/zcl/clusters/homeautomation.py#L309

ofcaah:
I changed it to int32s in my local quirk and it behaves correctly, however max that this meter ever sees here is ~18kW, so I can't confirm what happens if 32.7kW is exceeded. That being said: 32.7kW is already over 80 amps even if spread between 3 phases, so I suppose non-zigbee, one-time, non-recoverable fully mechanical audio-visual auto alert isn't far  beyond int16's max/min, and I'm not sure if anyone is brave enough to use this device with such currents. I, for one, certainly am not ;)

TheJulianJES (approved):
Ok, let's merge this then.

PR #4276, opened 2025-08-20
Add Aqara S1 air quality monitor

MindFreeze:
## Proposed change

Model lumi.airm.fhac01
image

- fix CO2 reading
- fix device temp reading
- add temp & humidity

## Checklist


TheJulianJES (approved):
Thanks! Made some small changes and this is good to go.
I've removed the `adds(TemperatureMeasurement)` and `adds(RelativeHumidity)`, as we likely don't need to add them as they should already be advertised by the device. If they're not (for some reason), they can just be re-added to the quirk.

Please also attach device diagnostics (to check that) and so we can add them to our diagnostics database in the ZHA repo.

PR #4256, opened 2025-08-13
Add Tuya valve SGW08 `_TZE284_fhvpaltk` variant

the-dreamer:
## Proposed change

should fix #4082

## Checklist

- [x] The changes are tested and work correctly
- [ ] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works
- [x] Device diagnostics data has been attached

PR #4281, opened 2025-08-22
Add Tuya NovaDigital ZVL-DUAL valve `_TZE284_iilebqoo`

fdaneluzzi:
This PR adds support for the Tuya TS0601 dual valve, sold in Brazil under the "NovaDigital ZVL-DUAL" brand.

The device is a whitelabeled version of the standard Tuya TS0601 dual valve already supported by this quirk. 
It uses the same datapoints and behavior, only the manufacturer string differs:

- Manufacturer: _TZE284_iilebqoo
WhatsApp Image 2025-08-22 at 14 31 48

- Model: TS0601

Adding `.applies_to("_TZE284_iilebqoo", "TS0601")` makes ZHA correctly apply the dual-valve quirk and exposes two switch entities (Valve 1, Valve 2) plus error reporting.

This will help users in Brazil, where the device is distributed under the NovaDigital brand, but is otherwise identical to other Tuya-branded versions.

### DIFF
--- a/zhaquirks/tuya/tuya_valve.py
+++ b/zhaquirks/tuya/tuya_valve.py
@@ -363,6 +363,7 @@ class GiexIrrigationStatus(t.enum8):
 (
     TuyaQuirkBuilder("_TZE284_8zizsafo", "TS0601")  # Giex GX04
     .applies_to("_TZE284_eaet5qt5", "TS0601")  # Insoma SGW08W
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
Thanks! Looks good

Note for myself/this quirk:
`_TZE284_8zizsafo` and `_TZE284_iilebqoo` use the same converter in Z2M, though `_TZE284_eaet5qt5` seems to be a different one.
Also see/related: https://github.com/zigpy/zha-device-handlers/pull/4256

PR #4287, opened 2025-08-24
Move XBee replacement endpoints to ZHA profile

Shulyaka:
## Proposed change

ZHA now ignores all endpoints which are neither with ZLL nor ZHA profile.
This PR moves XBee replacement endpoints to ZHA profile.

## Checklist

### DIFF
--- a/zhaquirks/xbee/xbee3_io.py
+++ b/zhaquirks/xbee/xbee3_io.py
@@ -29,105 +29,105 @@ class XBee3Sensor(XBeeCommon):
             0xD0: {
                 # AD0/DIO0/Commissioning
                 DEVICE_TYPE: zha.DeviceType.LEVEL_CONTROL_SWITCH,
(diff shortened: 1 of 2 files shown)


TheJulianJES:
Ideally, we'd also have some diagnostics data in the ZHA repo for XBee devices. This would have been caught in advance then. Change done in ZHA PR:
- https://github.com/zigpy/zha/pull/413

Current ZHA code: https://github.com/zigpy/zha/blob/65644001a0ba19a04cfdd7c8cddc0a8e4710fb00/zha/zigbee/endpoint.py#L141-L146

We could also add an exception for XBee devices there? But for now, I think changing this via quirks is fine as well.

TheJulianJES (approved):
thanks!

Shulyaka:
How can I submit the diagnostic data?

Shulyaka:
I think I found it...

Issue #4036, opened 2025-04-30
Review PRs


TheJulianJES:
This is an issue, mostly for me, for tracking a few PRs. No action needs to be taken from any linked PRs.

### PRs

- https://github.com/zigpy/zha-device-handlers/pull/4019
  - Initial PR (https://github.com/zigpy/zha-device-handlers/pull/4016) had the device in a different spot.
  - Z2M: Tuya vs Nous files, this `_TZE200_w6n8jeuu` is odd one out.
  - Do all the config entities even work?
  - Potentially added in wrong base quirk?
  - Local quirk might have been different to PR'd version?
  - Note `patch-2` in branch name, so web editor used, potentially leading to wrong result, as there are multiple base quirks in same file.
  - Awaiting feedback in linked PR.
  - Update: Might be correct? Was confused by other Tuya temp PR below.

- https://github.com/zigpy/zha-device-handlers/pull/4006
  - Z2M: Different converter, apparently supposed to have way more config options.
  - Potentially added in wrong base quirk?
  - Potential fix PR: https://github.com/zigpy/zha-device-handlers/pull/4037
  - Awaiting feedback in linked PR.

- https://github.com/zigpy/zha-device-handlers/pull/3936
  - Battery reporting apparently not working, though different quirk was mentioned with which it did, but caused cover to wrongfully invert.
  - Potentially new quirk (v1 or v2?) needed for combining working battery logic and correct inversion.
  - Z2M: Not in Z2M?
  - Awaiting feedback in linked PR.

TheJulianJES:
- https://github.com/zigpy/zha-device-handlers/pull/4603
  - Different model in Z2M but likely fine
  - Some models are defined around here: https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa47d703ad8f36de724a616f1c0785294f8d013d/src/devices/tuya.ts#L3502-L3531

- https://github.com/zigpy/zha-device-handlers/pull/4311
  - Entirely different model in Z2M
  - These datapoints do not match existing models: https://github.com/Koenkk/zigbee-herdsman-converters/blob/aa47d703ad8f36de724a616f1c0785294f8d013d/src/devices/zemismart.ts#L88-L122
  - There's already an issue with multiple models being incorrectly merged at the moment

- https://github.com/zigpy/zha-device-handlers/pull/4639
  - Different model in Z2M
[...]

PR #4311, opened 2025-09-05
Add Tuya Zemismart cover `_TZE200_sq6affpe` variant

bdodoiu:
## Proposed change
Add quirk for Tuya _TZE200_sq6affpe (TS0601).

## Additional information
Only added signature to existing class.
This was tested with Zemismart ZM25R3

## Checklist


TheJulianJES (commented):
This is a different model looking at Z2M, so I'm not sure we should just add it to this quirk.
I've noted this down for review later:
- https://github.com/zigpy/zha-device-handlers/issues/4036

PR #4344, opened 2025-09-14
Improve Bosch Room Thermostat II, add heating/cooling toggle

Sopsy:
## Proposed change

1. Fixes the missing cooling mode this thermostat has, but is not easily available to be toggled due to ZHA misintepreting "HEATING_ONLY" as something that cannot be toggled to "COOLING_ONLY". Bosch probably means that it does not automatically change to cooling when it's heating only.
It uses the undocumented 0x06 "centralite specific" value, which works perfectly for this thermostat as well.

2. Adds Imperial/Metric config option for the temperature reading on the thermostat itself.

3. Moves Operating mode from Diagnostic to Config, so it is easier to change from the UI. In reality Manual is usually the only mode people would like to use with HA, but this change makes it easier to change to that (default on the thermostat is Schedule).

4. Adds back "Heating demand" that was removed with the fix of https://github.com/home-assistant/core/issues/151763. It is useful information, and I would not want to remove it completely, rather rename it so ZHA does not get confused.

5. Adds an entity for outdoor temperature sensor, useful for automations. This is shown in the top right corner of the screen along with a tree icon indicating "outdoors". The default is indeed -327.68C, but as long as the value is never changed, nothing is displayed on the screen. Valid range is -99.9 - 99.9 C. -100C and below is displayed as Lo, and 100C and higher is not displayed at all, so it can be used as a way to remove the number from the screen.
To get rid of the temperature in the corner, the thermostat can be rebooted or the value set to anything over 10 000. After a reboot it's gone, but comes back after the first write to the value.

6. Adds the NC/NO setting for valves.

7. Adds the Min/Max cool setpoint limit config.

8. Adds the external sensor (usually a floor sensor) connection setting.

9. Adds the Valve status LED setting.

10. Adds the heater type (underfloor, boiler, radiator, central heating) setting.
According to Bosch documentation and recent research, the meaning is following:
Underfloor heating: 230V relay output.
Boiler: Potential free relay output. (Removed from the Bosch Smart Home app, replaced with Central heating)
Radiator: Radio only.
Central heating: Relay disconnected (not used), everything is controlled by the controller and not the thermostat. Thermostat only reports the heating demand to the controller.

Especially 6, 8 and 10 are crucial for some setups, previously completely preventing using this thermostat without the Bosch Smart Home Controller because they are not changeable from the thermostat itself when it is connected to a Zigbee network.

## Additional information

3 visual issues persist that this does not fix. I did not spend too much time trying to figure out how to override the attributes (seemingly not very easy) as they do work, but are just confusing in the UI.

1. The "Switch" entity in controls is read-only by the thermostat so it should be moved to Sensors and renamed to something like "Relay output" On/Off or "Valve" Open/Closed.
2. Keypad lockout should be a boolean 0x00/0x01 Off/On, and not an enum. The thermostat does not support the other states.
3. Min heat setpoint has a weird step of 0.349999999... instead of 0.5

If someone knows how to fix these with just the quirk, I'd be happy to implement.

## Checklist


Sopsy:
Not related to the pull request, but I figured this is the best place to write about my findings.

There are some more values in the thermostat cluster that are not yet figured out. They seem to be read-only, writing to any of them has no effect.

The values are 0x4022, 0x4023, 0x4024, 0x4025, 0x4050, 0x4052, 0x405B, 0x4060, 0x4061, 0x4062 and 0x4063.

Sopsy:
Added some more missing features.

Sopsy:
Added some more missing features.

Sopsy:
Fixed conflicts and added the missing features researched with the help of @DerDreschner over at the Z2M side (thanks!).

principis:
Do you need help merging this?

Sopsy:
This should be ready to merge. I have been using it myself without any issues.

PacoCT:
Hi,
first of all thank you very much for the work on supporting the Bosch Room Thermostat II family — it already works very well in ZHA.

I’m using the battery-powered model:

Bosch RBSH-RTH0-BAT-ZB-EU
Firmware: 0x02096a90

Everything pairs fine and, for the limited use I’m making of the thermostat, the climate/setpoint functions work perfectly.
(My setup uses the Bosch thermostat mainly as a physical interface to an existing Home Assistant “Generic Thermostat” which actually controls my heating on a per-room basis.)

However, the battery entity always shows unavailable, which is a problem for me.
Since this thermostat is the only way for guests (friends and relatives who do not use the HA app) to adjust the temperature in each room, keeping the device powered at all times is very important. Battery replacement notifications are therefore essential in my setup.

To my (limited) knowledge, this seems to be because the device only exposes battery_voltage, while ZHA currently creates a battery percentage sensor, which the thermostat does not report.

From ZHA diagnostics (Power Configuration 0x0001):

`{
  "battery_alarm_state": 0,
  "battery_voltage": 62,
  "battery_percentage_remaining": { "unsupported": true },
  "battery_size": { "unsupported": true },
  "battery_quantity": { "unsupported": true }
}
`
Partial signature (cleaned):
`{
  "manufacturer": "Bosch",
  "model": "RBSH-RTH0-BAT-ZB-EU",
  "manufacturer_code": 4617,
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0020",
        "0x0201",
        "0x0204",
        "0x0405",
        "0x0B05"
      ],
      "out_clusters": ["0x000A", "0x0019"]
    }
  }
}
`
In my opinion (and I may easily be mistaken), ZHA sees only battery_voltage, but since battery_percentage_remaining is unsupported, the generated battery sensor ends up as “%” with a null state, which then becomes permanently unavailable.

If possible, it would be great if the quirk for this thermostat could:

- Expose battery_voltage directly as the battery sensor,
- Or map voltage → percentage (as, I think,  ZHA does for some other devices),
- Or find a percentage-based battery entity .

I’m happy to provide any other information required (if you provide some guidance).

Thanks again for all the work on this device family — it’s already extremely useful, and proper battery handling would make it complete, especially for setups where this thermostat is the only physical way guests can control the heating.

Cheers,
Paco

TheJulianJES:
The changes look fine to me. @mrrstux Mind having a look at this, since you worked on most of the Bosch TRV quirks?

zhaquirks/bosch/rbsh_rth0_zb_eu.py
```diff
@@ -1,41 +1,85 @@
...
+ACTUATOR_TYPE_ATTR_ID = 0x4060
+
+# External sensor connection config.
+SENSOR_CONNECTION_ATTR_ID = 0x4062
+
+# Heater type (UFH, boiler, radiator, central heating).
+HEATER_TYPE_ATTR_ID = 0x4063
+
+# Error codes (these do not match the E**-codes displayed on the screen).
+ERROR_CODE_ATTR_ID = 0x5000
```
TheJulianJES, line 46: The attributes ID could be defined directly inside the attribute definitions instead, with the comments moved there as well. They should generally be accessed with `cluster.AttributeDefs.attr.id` if the id is needed then. But it's fine to keep it like this now.

zhaquirks/bosch/rbsh_rth0_zb_eu.py
```diff
@@ -107,23 +255,136 @@ class AttributeDefs(UserInterface.AttributeDefs):
...
+        BoschThermostatCluster.AttributeDefs.heating_demand.name,
+        BoschThermostatCluster.cluster_id,
+        entity_type=EntityType.DIAGNOSTIC,
+        state_class=SensorStateClass.MEASUREMENT,
+        unit=PERCENTAGE,
+        translation_key="heating_demand",
+        fallback_name="Heating demand",
+        reporting_config=ReportingConfig(
+            min_interval=1, max_interval=900, reportable_change=1
+        ),
```
TheJulianJES: Let's keep `translation_key` and `fallback_name` at the bottom, similar to all other v2 quirks.
I've moved them with [`ffd1f4f` (#4344)](https://github.com/zigpy/zha-device-handlers/pull/4344/commits/ffd1f4f4fd1041102fe3aac09540a083a49adb66) now.

TheJulianJES (approved):
Yeah, I think this all looks good. Thanks!

zhaquirks/bosch/rbsh_rth0_zb_eu.py
Sopsy: I actually thought that it would look cleaner, but as the old quirk had them defined like so and as the code style guidelines are quite strict (and rightfully so), I didn't want to move them there thinking it wouldn't fly.

I can do a separate pull request for that if you wish!

zhaquirks/bosch/rbsh_rth0_zb_eu.py
Sopsy: Good point. Thank you!

metallmanu:
@Sopsy @TheJulianJES yesterday I get an update to Firmware 0x03096a90 for my bosch.rbsh_rth0_zb_eu. There is a new switch which I don´t know what it is.
Bosch Firmware support page says:
Raum-Thermostat II 230V (0.3.9): Fehlerbehebung: Der Entkalkungsschutz ist jetzt auch im Modus Heizpause/Aus aktiv.

PR #4352, opened 2025-09-20
Add Third Reality button settings

3reality-support:
## Proposed change

This PR has added button quirk v2 and some basic attributes, automation, and private clusters
1.cancel_bouble_click----cancel button double-click function

## Additional information

This PR fixes the issue where buttons cannot be automated in zha，and do I need to remove the previously submitted button.py？

## Device diagnostics

[zha-01JEN6GWZMMHTF70X8QNKV91VP-Third Reality, Inc 3RSB22BZ-0d377385e48c0e61cd04ea86b316bf17.json](https://github.com/user-attachments/files/22437632/zha-01JEN6GWZMMHTF70X8QNKV91VP-Third.Reality.Inc.3RSB22BZ-0d377385e48c0e61cd04ea86b316bf17.json)

## Checklist


3reality-support:
@TheJulianJES codecov fail , please advise on how to fix it

3reality-support:
@TheJulianJES

zhaquirks/thirdreality/button_v2.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class MultistateInputCluster(CustomCluster, MultistateInput):
+    """Multistate input cluster."""
+
+    def __init__(self, *args, **kwargs):
+        """Init."""
+        self._current_state = {}
+        super().__init__(*args, **kwargs)
+
```
TheJulianJES: We don't really need  `self._current_state`.

```suggestion
```

zhaquirks/thirdreality/button_v2.py
```diff
@@ -0,0 +1,92 @@
...
+    def _update_attribute(self, attrid, value):
+        super()._update_attribute(attrid, value)
+        if attrid == 0x0055:
+            self._current_state[0x0055] = action = MOVEMENT_TYPE.get(value)
```
TheJulianJES: We also don't need to set `self._current_state[0x0055]` here.
```suggestion
            action = MOVEMENT_TYPE.get(value)
```

zhaquirks/thirdreality/button_v2.py
```diff
@@ -0,0 +1,92 @@
...
+            event_args = {VALUE: value}
+            if action is not None:
+                self.listener_event(ZHA_SEND_EVENT, action, event_args)
+
+            # show something in the sensor in HA
+            super()._update_attribute(0, action)
```
TheJulianJES: This also doesn't do anything, as there's no sensor in HA. I know this is copied a lot from some old quirks, but those should be updated as well.
```suggestion
```

zhaquirks/thirdreality/button_v2.py
```diff
@@ -0,0 +1,92 @@
...
+    .number(
+        attribute_name=ThirdRealityButtonCluster.AttributeDefs.cancel_bouble_click.name,
+        cluster_id=ThirdRealityButtonCluster.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        step=1,
+        translation_key="cancel_bouble_click",
+        fallback_name="Cancel double click",
+    )
```
TheJulianJES: What is the unit for this? If it's seconds/time, the number device class should ideally also be set to `DURATION`.

tests/test_third_button.py
```diff
@@ -0,0 +1,76 @@
...
+    # Find the MultistateInputCluster
+    multistate_cluster = next(
+        (
+            cluster
+            for cluster in device.endpoints[1].in_clusters.values()
+            if isinstance(cluster, MultistateInputCluster)
+        ),
+        None,
+    )
+    assert multistate_cluster is not None, "MultistateInputCluster not found"
```
TheJulianJES: We can also get it via `MultistateInput.cluster_id` here.

tests/test_third_button.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def zha_send_event(self, action, event_args):
+        """Record ZHA events.
+
+        Args:
+            action (str): The type of action for the event.
+            event_args (dict): Relevant parameters of the event.
+
+        """
+        self.zha_send_events.append((action, event_args))
```
TheJulianJES: We might be able to just use a `MagicMock` for this, similar to other tests using `zha_send_event`.
I'll take a look at this.

TheJulianJES (commented):
> and do I need to remove the previously submitted button.py？

Oh, I think I understood this wrong in some other PR. If you have a v2 quirk for the same device as an old v1 quirk, the old v1 quirk should be removed.

--- 

I'll take a look at addressing the comments in this PR now.

zhaquirks/thirdreality/button_v2.py
TheJulianJES: Can you explain what this does exactly?

TheJulianJES (commented):
The old v1 quirk also used a custom `PowerConfiguration` cluster to change the battery percentage reported in HA based on the device voltage, instead of the percentage reported by the device. Is this no longer needed?
https://github.com/zigpy/zha-device-handlers/blob/203b56f8b6b9430fab1276c4cada5fac20dd0a2b/zhaquirks/thirdreality/button.py#L31-L35

zhaquirks/thirdreality/button_v2.py
```diff
@@ -0,0 +1,92 @@
...
+class ThirdRealityButtonCluster(CustomCluster):
+    """Third Reality's button private cluster."""
+
+    cluster_id = 0xFF01
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Define the attributes of a private cluster."""
+
+        # cancel double click
+        cancel_bouble_click: Final = ZCLAttributeDef(
```
TheJulianJES: Typo + the comment really isn't helpful here.
```suggestion
        cancel_double_click: Final = ZCLAttributeDef(
```

zhaquirks/thirdreality/button_v2.py
```diff
@@ -0,0 +1,92 @@
...
+            is_manufacturer_specific=True,
+        )
+
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RSB22BZ")
+    .replaces(ThirdRealityButtonCluster)
+    .replaces(MultistateInputCluster)
+    .number(
+        attribute_name=ThirdRealityButtonCluster.AttributeDefs.cancel_bouble_click.name,
```
TheJulianJES: typo:
```suggestion
        attribute_name=ThirdRealityButtonCluster.AttributeDefs.cancel_double_click.name,
```

zhaquirks/thirdreality/button_v2.py
```diff
@@ -0,0 +1,92 @@
...
+        cluster_id=ThirdRealityButtonCluster.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        step=1,
+        translation_key="cancel_bouble_click",
```
TheJulianJES: typo:
```suggestion
        translation_key="cancel_double_click",
```

zhaquirks/thirdreality/button.py
```diff
@@ -46,67 +33,48 @@ class CustomPowerConfigurationCluster(PowerConfigurationCluster):
...
+        fallback_name="Cancel double click",
+    )
+    .device_automation_triggers(
+        {
+            (DOUBLE_PRESS, DOUBLE_PRESS): {COMMAND: COMMAND_DOUBLE},
+            (SHORT_PRESS, SHORT_PRESS): {COMMAND: COMMAND_SINGLE},
+            (LONG_PRESS, LONG_PRESS): {COMMAND: COMMAND_HOLD},
+            (LONG_RELEASE, LONG_RELEASE): {COMMAND: COMMAND_RELEASE},
+        }
+    )
```
TheJulianJES, line 74: These are unchanged, but I don't think `(DOUBLE_PRESS, DOUBLE_PRESS)` is really nice. I think this is (press_type, button) normally. Device triggers will be replaced at some point anyways.

We'll have to keep this to not break existing automations though.

zhaquirks/thirdreality/button_v2.py
3reality-support: This product has double clicking and single clicking, both of which will send different commands. Some customers do not want double clicking, so we use a private cluster to remove double clicking

3reality-support: I feel that using .switch is more suitable

3reality-support:
> The old v1 quirk also used a custom `PowerConfiguration` cluster to change the battery percentage reported in HA based on the device voltage, instead of the percentage reported by the device. Is this no longer needed?
> 
> https://github.com/zigpy/zha-device-handlers/blob/203b56f8b6b9430fab1276c4cada5fac20dd0a2b/zhaquirks/thirdreality/button.py#L31-L35

Yes, it's no longer needed

zhaquirks/thirdreality/button_v2.py
TheJulianJES: Ok, but this is a number entity. So, what is the difference between setting it to 0, 10, and 1000 for example?

3reality-support: Actually, only 0 and 1 have an effect. 0 represents opening and double clicking, and 1 represents shutting down and double clicking

TheJulianJES: So I can just change this to a switch entity? I can do that and this should be good to go.

3reality-support: ok， I will resubmit later and modify it to a switch entity

3reality-support:
@TheJulianJES If there are no issues, please help merge as soon as possible

tests/test_quirks.py
```diff
@@ -661,12 +661,6 @@ def test_migrated_lighting_automation_triggers(quirk: CustomDevice) -> None:
             (const.LONG_RELEASE, const.BUTTON_4),
         ],
     ],
-    zhaquirks.thirdreality.button.Button: [
-        [
-            (const.LONG_PRESS, const.LONG_PRESS),
-            (const.LONG_RELEASE, const.LONG_RELEASE),
-        ]
-    ],
```
TheJulianJES, line 669: Just a note for us (no action needed): This is technically an issue with the new quirk as well, but this test only checks v1 quirks. We also can't change these triggers now, since they might already be in use.

zhaquirks/thirdreality/button.py
```diff
@@ -46,67 +33,44 @@ class CustomPowerConfigurationCluster(PowerConfigurationCluster):
...
+(
+    QuirkBuilder("Third Reality, Inc", "3RSB22BZ")
+    .replaces(ThirdRealityButtonCluster)
+    .replaces(MultistateInputCluster)
+    .switch(
+        attribute_name=ThirdRealityButtonCluster.AttributeDefs.cancel_double_click.name,
+        cluster_id=ThirdRealityButtonCluster.cluster_id,
+        translation_key="cancel_double_click",
+        fallback_name="Cancel double click",
+    )
```
TheJulianJES, line 66: I think "disabling" makes more sense here than "canceling", so I'll just change the `translation_key` and `fallback_name`:
```suggestion
    .switch(
        attribute_name=ThirdRealityButtonCluster.AttributeDefs.cancel_double_click.name,
        cluster_id=ThirdRealityButtonCluster.cluster_id,
        translation_key="disable_double_click",
        fallback_name="Disable double click",
    )
```
("Clicking" might be slightly more correct than "Click" as well, but "Click" is shorter.)

zhaquirks/thirdreality/button.py
```diff
@@ -46,67 +33,44 @@ class CustomPowerConfigurationCluster(PowerConfigurationCluster):
...
+
+class ThirdRealityButtonCluster(CustomCluster):
+    """Third Reality's button private cluster."""
+
+    cluster_id = 0xFF01
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Define the attributes of a private cluster."""
+
+        cancel_double_click: Final = ZCLAttributeDef(
```
TheJulianJES: For consistency, I'll also adjust the `attribute_name` from "cancel" to "disable"
```suggestion
        disable_double_click: Final = ZCLAttributeDef(
```

zhaquirks/thirdreality/button.py
```diff
@@ -46,67 +33,44 @@ class CustomPowerConfigurationCluster(PowerConfigurationCluster):
...
+            is_manufacturer_specific=True,
+        )
+
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RSB22BZ")
+    .replaces(ThirdRealityButtonCluster)
+    .replaces(MultistateInputCluster)
+    .switch(
+        attribute_name=ThirdRealityButtonCluster.AttributeDefs.cancel_double_click.name,
```
TheJulianJES: and here as well:
```suggestion
        attribute_name=ThirdRealityButtonCluster.AttributeDefs.disable_double_click.name,
```

TheJulianJES (approved):
Looks good. Thanks! I'll just apply a couple of small changes and it'll be merged.

zhaquirks/thirdreality/button.py
```diff
@@ -46,67 +33,44 @@ class CustomPowerConfigurationCluster(PowerConfigurationCluster):
...
```
TheJulianJES: Apparently this didn't change correctly, so let's reapply the suggestion:
```suggestion
        attribute_name=ThirdRealityButtonCluster.AttributeDefs.disable_double_click.name,
```

TheJulianJES (approved):

zhaquirks/thirdreality/button.py
```diff
@@ -46,67 +33,44 @@ class CustomPowerConfigurationCluster(PowerConfigurationCluster):
...
-                OUTPUT_CLUSTERS: [
-                    OnOff.cluster_id,
-                    LevelControl.cluster_id,
-                    Ota.cluster_id,
-                ],
-            }
-        },
-    }
-    replacement = {
-        SKIP_CONFIGURATION: True,
```
TheJulianJES, line 89: Note: `SKIP_CONFIGURATION` was included in the v1 quirk, which means that the clusters aren't bound and no attribute reporting is attempted to be set up. With the v2 quirk, that's done now (even if the device doesn't need this to happen), but we also need to read the `disable_double_click` attribute during pairing.
If we `.skip_configuration()` in the v2 quirk as well, we'd not read the attribute during pairing, which isn't ideal.

So, I think we can leave this, unless the device doesn't like extra configuration.

TheJulianJES:
> This PR fixes the issue where buttons cannot be automated in zha，and do I need to remove the previously submitted button.py？

One last question: Are there multiple variants of this device that the v1 quirk didn't match previously, or why does this fix an issue where the device triggers didn't show up? They are unchanged between v1 and v2 quirks, so I guess the v2 quirk must match another device variant that the v1 quirk did not? Or am I missing something?

3reality-support:
> > This PR fixes the issue where buttons cannot be automated in zha，and do I need to remove the previously submitted button.py？
> 
> One last question: Are there multiple variants of this device that the v1 quirk didn't match previously, or why does this fix an issue where the device triggers didn't show up? They are unchanged between v1 and v2 quirks, so I guess the v2 quirk must match another device variant that the v1 quirk did not? Or am I missing something?

This is because the product itself registers a private cluster 0xFF01, but there is no corresponding adaptation script for it.

PR #4373, opened 2025-10-07
Add Tuya dimmable LED controller `_TZ3210_syh4kuef` variant

Niek:
## Proposed change
This is a one-line change that adds support for the `_TZ3210_syh4kuef` device. This is listed on e.g. AliExpress as:

[Zigbee 3.0 12 Watts 12V Dimmable LED Driver Dimming LED Power Supply Voice APP Smart Control Suitable for downlight spotlight
](https://www.aliexpress.com/item/1005006630186296.html)

The device works fine, but lists a color dimmer instead of a regular one. This quirk fixes it.


TheJulianJES (approved):
Thanks!

PR #4378, opened 2025-10-09
Add Third Reality curtain gen1 settings

3reality-support:
## Proposed change

This PR added adaptation code for curtain_gen1 and added 2 additional private clusters

1. enable_disable_pir_remote: turn on or off the infrared remote control control
2. allow_bind: open or close bind

## Additional information

f1935114fd9588d5145f5a53d1dbbf9f

## Checklist


zhaquirks/thirdreality/curtain_gen1.py
```diff
@@ -0,0 +1,50 @@
...
+        translation_key="enable_pir_mode",
+        fallback_name="Enable PIR remote",
+    )
+    .write_attr_button(
+        attribute_name=ThirdRealityCurtainCluster.AttributeDefs.allow_bind.name,
+        attribute_value=0x01,
+        cluster_id=ThirdRealityCurtainCluster.cluster_id,
+        translation_key="allow_bind",
+        fallback_name="Allow bind",
+    )
```
TheJulianJES, line 48: What does "Allow bind" do exactly? Is this related to cluster binding or to curtains being bound?
Does the device reset the value from `1` to `0` itself after this button is pressed?

3reality-support: image
For example, sending a bind request to the button, opening the bind, the curtain can send a request to establish contact, triggering an allow bind once, which can last for three minutes

3reality-support: After writing 1, wait for 3 minutes until the end, and this value will become 0

TheJulianJES: Sorry, I still don't understand what this button does exactly. The screenshot also doesn't seem to be related?
A couple of questions that might help me understand this:
1. Is this related to "Zigbee device/group binding" at all? Or is this another "binding"?
2. Is this to bind a remote to the curtain? And the remote then has to be bound in the three minutes?

---

And to clarify what I don't understand from your example:
> sending a bind request to the button,

3. Is this referencing the "Allow bind" button entity in Home Assistant? Or is this about a physical button/remote to which the "bind request" is sent?

> opening the bind,

4. What is opened exactly?

> the curtain can send a request to establish contact,

5. The curtain sends a request to what? A different button/remote device?

> triggering an allow bind once, which can last for three minutes

6. So pairing for a different button/remote device is allowed (for three minutes)?

TheJulianJES (commented):
This generally looks good as well, but please help me understand the "Allow bind" entity in the comment above.
For example, is this about binding a different remote to the device? Does this use standard Zigbee binding?

I've looked at the Zigbee2MQTT converters and the attribute is not exposed an entity there, so that wasn't much help for me.

Also, I tried looking at the user instructions for the "Smart Blind" on your website here: https://thirdreality.com/user-manuals/. But clicking on "Smart Blind" opens the manual for the "Smart Button".

If the "Allow bind" entity is clarified, this can be merged. We just need to somewhat know what it does, as this will be translated into other languages and those people might ask us.

zhaquirks/thirdreality/curtain_gen1.py
3reality-support: 1.This is related to binding with Zigbee device.
2.It mainly involves devices with the On/Off cluster (output), and the binding must be completed within three minutes.
3.It refers to the physical button/remote control.
4/5/6. The curtain acts as the device being bound. The binding process is as follows: the remote control sends an Identify Query, and the curtain responds with an Identify Query Response upon receiving it, thus establishing the binding. Normally, the curtain does not process received Identify Query messages; it will only respond and perform the binding operation when you enable "Allow Bind".

zhaquirks/thirdreality/curtain_gen1.py
```diff
@@ -0,0 +1,50 @@
...
+        force_inverted=True,
+        translation_key="enable_pir_mode",
+        fallback_name="Enable PIR remote",
+    )
+    .write_attr_button(
+        attribute_name=ThirdRealityCurtainCluster.AttributeDefs.allow_bind.name,
+        attribute_value=0x01,
+        cluster_id=ThirdRealityCurtainCluster.cluster_id,
+        translation_key="allow_bind",
+        fallback_name="Allow bind",
```
TheJulianJES: With the explanation given above, I'll rename this to "Allow remote binding". I think that's slightly clearer.

TheJulianJES:
```suggestion
        translation_key="allow_remote_binding",
        fallback_name="Allow remote binding",
```

zhaquirks/thirdreality/curtain_gen1.py
TheJulianJES: Ok, thanks for the explanation. That helps a ton!

I've renamed the button to "Allow remote binding" now. I think that should be slightly clearer for some users, especially after the entity names have been translated into other languages.

It does seem a bit weird to kind of misuse the `Identify` cluster for binding, like this, but ok. If it works 😅.

TheJulianJES (approved):
This looks good. Thanks!

PR #4379, opened 2025-10-09
Add Third Reality door sensor settings

3reality-support:
## Proposed change

This PR added adaptation code for door_sensor and added 1 additional private cluster

1. open_delay_time: set for 5 seconds, after triggering, display on after 5 seconds

## Additional information

2f8a7a35c0428774d368d5c20bbb7729

## Checklist

### DIFF
--- a/zhaquirks/thirdreality/door_sensor_v2.py
+++ b/zhaquirks/thirdreality/door_sensor_v2.py
@@ -1,0 +1,40 @@
+"""Third Reality door sensor devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ThirdRealityDoorCluster(CustomCluster):
+    """Third Reality's door sensor private cluster."""
+
+    cluster_id = 0xFF01
+
+    class AttributeDefs(BaseAttributeDefs):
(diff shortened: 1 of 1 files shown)


zhaquirks/thirdreality/door_sensor_v2.py
```diff
@@ -0,0 +1,42 @@
...
+(
+    QuirkBuilder("Third Reality, Inc", "3RDS17BZ")
+    .replaces(ThirdRealityDoorCluster)
+    .number(
+        attribute_name=ThirdRealityDoorCluster.AttributeDefs.open_delay_time.name,
+        min_value=0,
+        max_value=3600,
+        unit=UnitOfTime.SECONDS,
+        device_class=NumberDeviceClass.DURATION,
+        cluster_id=ThirdRealityDoorCluster.cluster_id,
```
TheJulianJES: Just moving the `cluster_id` to another place here. Doesn't really matter, but I think we generally have attribute name, cluster id, and endpoint id as the first three options.
```suggestion
        attribute_name=ThirdRealityDoorCluster.AttributeDefs.open_delay_time.name,
        cluster_id=ThirdRealityDoorCluster.cluster_id,
        min_value=0,
        max_value=3600,
        unit=UnitOfTime.SECONDS,
        device_class=NumberDeviceClass.DURATION,
```
I'll apply this patch and merge the PR.

TheJulianJES (approved):
This looks good. Thanks!

For others looking at this PR: Having the "open delay time" be device specific might actually be useful if binding the `OnOff` cluster to a light, for example.
Since this device provides the option, we should allow for configuration.

TheJulianJES (approved):
Thanks!

PR #4380, opened 2025-10-09
Add Third Reality soil sensor settings

3reality-support:
## Proposed change

This PR added adaptation code for soil sensor and added 3 additional private clusters

1. temperature_correction_fahrenheit: Calibrate Fahrenheit
2. temperature_correction_celsius: Calibrate Celsius
3. humidity_correction: Calibrate humidity
4. The two types of temperature calibration are separated, making it convenient for users to calibrate according to their own needs

## Additional information

dd1c7c38-0aad-4679-8668-65288966085e

## Checklist


zhaquirks/thirdreality/soil_moisture_sensor.py
```diff
@@ -0,0 +1,78 @@
...
+            id=0x0033,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
+
+        temperature_correction_celsius: Final = ZCLAttributeDef(
+            id=0x0031,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
```
TheJulianJES: Does this device (also) have a display or why is the temperature correction for Celsius and Fahrenheit separate?

TheJulianJES:
If we add the temperature device class two the two temperature offset entities, I believe HA switches the entities to be in the preferred unit for the user.

1. Can you explain the difference for temperature C° and F° correction for all devices?
2. How/when is which option applied? Which one is used for ZCL vs which one is used for display? Or are both used for the display?
3. Can we just remove the "Fahrenheit" option and only keep the "Celsius" option (as it's similar to how temperature is represented within ZCL)? If we name it "Temperature offset" and set the device class, HA should show the preferred unit for the user.

Also see:
- https://github.com/zigpy/zha-device-handlers/pull/4442

zhaquirks/thirdreality/soil_moisture_sensor.py
3reality-support: There is no display screen, and the reason for separating Celsius and Fahrenheit is that some users use Celsius and Fahrenheit, so they can calibrate freely according to the unit without the need for additional conversion

3reality-support:
> If we add the temperature device class two the two temperature offset entities, I believe HA switches the entities to be in the preferred unit for the user.
> 
> 1. Can you explain the difference for temperature C° and F° correction for all devices?
> 2. How/when is which option applied? Which one is used for ZCL vs which one is used for display? Or are both used for the display?
> 3. Can we just remove the "Fahrenheit" option and only keep the "Celsius" option (as it's similar to how temperature is represented within ZCL)? If we name it "Temperature offset" and set the device class, HA should show the preferred unit for the user.
> 
> Also see:
> 
> * [Remove Fahrenheit temperature offset for Third Reality sensors #4442](https://github.com/zigpy/zha-device-handlers/pull/4442)

There is no difference between temperature C ° and F °. Setting one of them will convert the other into the corresponding value according to the formula, just for the convenience of the user. If it is possible to achieve the same calibration unit as the temperature unit set by the user, it would be even better

zhaquirks/thirdreality/soil_moisture_sensor.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RSM0147Z")
+    .applies_to("Third Reality, Inc", "3RSM0347Z")
+    .replaces(ThirdRealitySoilMoistureCluster)
+    .replaces(SoilMoisture)
```
TheJulianJES: `SoilMoisture` is a standard ZCL cluster. Does the device not advertise that cluster during the interview? If so, we should just do `.adds(SoilMoisture)`. It'll have the same effect then, just a bit more clear.

If it does advertise the `SoilMoisture` cluster on its own (so the soil moisture entity is already present without the quirk, this line can just be removed):

```suggestion
```

TheJulianJES (commented):
One question, but this looks good otherwise. Thanks!

(`temperature_offset_fahrenheit` is intentionally not exposed as an offset, as the "Temperature offset" entity should change the user-facing unit in HA due to it having `NumberDeviceClass.TEMPERATURE`. That essentially achieves the same, except we don't need to show two temperature offset entities for all users.)

zhaquirks/thirdreality/soil_moisture_sensor.py
```diff
@@ -0,0 +1,68 @@
...
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RSM0147Z")
+    .applies_to("Third Reality, Inc", "3RSM0347Z")
+    .replaces(ThirdRealitySoilMoistureCluster)
+    .number(
+        attribute_name=ThirdRealitySoilMoistureCluster.AttributeDefs.temperature_offset_celsius.name,
+        cluster_id=ThirdRealitySoilMoistureCluster.cluster_id,
+        min_value=-10000,
+        max_value=10000,
```
TheJulianJES, line 47: Allowing an offset of +/- 100°C does seem a bit much, but this should show as a box entity anyway, so if the device supports it, this isn't critical to change. Btw., to force the box mode display in HA, you can always add `mode="box"` here. Otherwise, it's depending on how min + max values + step count on whether a box or slider is shown.

If you think the min/max values should be tone down a bit for both temperature and humidity, please do so in another PR (and add `mode="box"` if needed).

zhaquirks/thirdreality/soil_moisture_sensor.py
```diff
@@ -0,0 +1,68 @@
...
+        device_class=NumberDeviceClass.TEMPERATURE,
+        unit=UnitOfTemperature.CELSIUS,
+        translation_key="temperature_offset",
+        fallback_name="Temperature offset",
+    )
+    .number(
+        attribute_name=ThirdRealitySoilMoistureCluster.AttributeDefs.humidity_offset.name,
+        cluster_id=ThirdRealitySoilMoistureCluster.cluster_id,
+        min_value=-10000,
+        max_value=10000,
```
TheJulianJES, line 59: Same comment as above. The humidity offset also allows for a huge range.

TheJulianJES (approved):
This looks good, thanks!

PR #4381, opened 2025-10-09
Add Third Reality curtain gen2 settings

3reality-support:
## Proposed change

This PR added adaptation code for curtain gen2 and added 2 additional private clusters

1. enable_disable_pir_remote: turn on or off the infrared remote control control
2. compensation_speed: There are 2 motors, one on each side, and a separate motor position can be set
3. limit_position: Set the lowest position

## Additional information

d26850a40e2fb710f34ef7afc870013a

## Checklist


zhaquirks/thirdreality/curtain_gen2.py
```diff
@@ -0,0 +1,69 @@
...
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RSB02015Z")
+    .replaces(THIRD_REALITY_Blind_Gen2_CLUSTER, endpoint_id=1)
+    .switch(
+        attribute_name=THIRD_REALITY_Blind_Gen2_CLUSTER.AttributeDefs.enable_disable_pir_remote.name,
+        cluster_id=THIRD_REALITY_Blind_Gen2_CLUSTER.cluster_id,
+        force_inverted=True,
+        translation_key="enable_disable_pir_mode",
+        fallback_name="Enable/Disable PIR Remote",
```
TheJulianJES: When this switch is on, the remote is enabled, right? Can we just name it something like "Enable PIR remote" then? That follows more what HA generally does.
```suggestion
        translation_key="enable_pir_mode",
        fallback_name="Enable PIR remote",
```
Also, we need to lower-case "remote", since we generally use sentence casing for the entity names, so something like this: "This is a text", and not: "This Is A Text".

zhaquirks/thirdreality/curtain_gen2.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+    .number(
+        attribute_name=THIRD_REALITY_Blind_Gen2_CLUSTER.AttributeDefs.compensation_speed.name,
+        cluster_id=THIRD_REALITY_Blind_Gen2_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=-100,
+        max_value=100,
+        step=1,
+        translation_key="compensation_speed",
+        fallback_name="compensation_speed",
```
TheJulianJES: We'll use the `fallback_name` for the (US) English translation in HA later, so we can make this something readable, like:
```suggestion
        fallback_name="Compensation speed",
```

zhaquirks/thirdreality/curtain_gen2.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+    .number(
+        attribute_name=THIRD_REALITY_Blind_Gen2_CLUSTER.AttributeDefs.limit_position.name,
+        cluster_id=THIRD_REALITY_Blind_Gen2_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=3800,
+        step=1,
+        translation_key="limit_position",
+        fallback_name="limit_position",
```
TheJulianJES: Same here:
```suggestion
        fallback_name="Limit position",
```

zhaquirks/thirdreality/curtain_gen2.py
```diff
@@ -0,0 +1,69 @@
...
+        translation_key="enable_disable_pir_mode",
+        fallback_name="Enable/Disable PIR Remote",
+    )
+    .number(
+        attribute_name=THIRD_REALITY_Blind_Gen2_CLUSTER.AttributeDefs.compensation_speed.name,
+        cluster_id=THIRD_REALITY_Blind_Gen2_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=-100,
+        max_value=100,
+        step=1,
```
TheJulianJES, line 54: For `number` entities, try adding `mode="box"` and see if you prefer how that makes the entity show up in HA:
```suggestion
        step=1,
        mode="box",
```
You won't get a slider then, but a "box" to input the number into instead. If you prefer this for these entities, feel free to also add it to the other number entity in this PR (or other PRs).

TheJulianJES (commented):
Few small things, otherwise good already!

3reality-support:
@TheJulianJES Thank you for your reply. The modifications have been made. Everything else is fine

zhaquirks/thirdreality/curtain_gen2.py
```diff
@@ -0,0 +1,70 @@
...
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class THIRD_REALITY_Blind_Gen2_CLUSTER(CustomCluster):
```
TheJulianJES, line 11: This class name is a bit long, but leave it like this for now. I'll go through some of the quirks after all are merged and see if they can be cleaned up slightly or combined.

TheJulianJES (approved):
Looks good, thanks!

PR #4383, opened 2025-10-10
Add Third Reality switch settings

3reality-support:
## Proposed change

This PR added adaptation code for dual_plug and added 2 additional private clusters

1. on_to_off_delay---off delay, when you set the attribute to 10 and the current status is on, it will change to off after 10 seconds
2. off_to_on_delay---on delay, similar to the above

## Additional information

9cebdaa5-95ab-46be-8b48-5dccfc5d02fe

## Checklist

### DIFF
--- a/zhaquirks/thirdreality/switch_v2.py
+++ b/zhaquirks/thirdreality/switch_v2.py
@@ -1,0 +1,61 @@
+"""Third Reality Blind Gen2 devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+from zigpy.quirks.v2.homeassistant import UnitOfTime
+
+
+
+class THIRD_REALITY_Switch_CLUSTER(CustomCluster):
+    """Third Reality's Switch private cluster."""
+
+    cluster_id = 0xFF02
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Define the attributes of a private cluster."""
+		
+        
+        on_to_off_delay: Final = ZCLAttributeDef(
+            id=0x0001,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+          
+        off_to_on_delay: Final = ZCLAttributeDef(
+            id=0x0002,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+
+
+(
+    QuirkBuilder("Third Reality, Inc", "3RSS009Z")
+    .replaces(THIRD_REALITY_Switch_CLUSTER)
+    .number(
+        attribute_name=THIRD_REALITY_Switch_CLUSTER.AttributeDefs.on_to_off_delay.name,
+        cluster_id=THIRD_REALITY_Switch_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        unit=UnitOfTime.SECONDS,
+        step=1,
+        translation_key="on_to_off_delay",
+        fallback_name="on_to_off_delay",
+    )
+    .number(
+        attribute_name=THIRD_REALITY_Switch_CLUSTER.AttributeDefs.off_to_on_delay.name,
+        cluster_id=THIRD_REALITY_Switch_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        unit=UnitOfTime.SECONDS,
+        step=1,
+        translation_key="off_to_on_delay",
+        fallback_name="off_to_on_delay",
+    )
+    .add_to_registry()
+)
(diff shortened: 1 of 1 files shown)


zhaquirks/thirdreality/switch_v2.py
```diff
@@ -0,0 +1,59 @@
...
+    .number(
+        attribute_name=THIRD_REALITY_Switch_CLUSTER.AttributeDefs.on_to_off_delay.name,
+        cluster_id=THIRD_REALITY_Switch_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        unit=UnitOfTime.SECONDS,
+        step=1,
+        translation_key="on_to_off_delay",
+        fallback_name="on_to_off_delay",
```
TheJulianJES: We'll use the `fallback_name` for the (US) English translation in HA later, so we can make this something readable, like:
```suggestion
        fallback_name="On to off delay",
```

zhaquirks/thirdreality/switch_v2.py
```diff
@@ -0,0 +1,59 @@
...
+    .number(
+        attribute_name=THIRD_REALITY_Switch_CLUSTER.AttributeDefs.off_to_on_delay.name,
+        cluster_id=THIRD_REALITY_Switch_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        unit=UnitOfTime.SECONDS,
+        step=1,
+        translation_key="off_to_on_delay",
+        fallback_name="off_to_on_delay",
```
TheJulianJES: and same here:
```suggestion
        fallback_name="Off to on delay",
```

zhaquirks/thirdreality/switch_v2.py
```diff
@@ -0,0 +1,59 @@
...
+(
+    QuirkBuilder("Third Reality, Inc", "3RSS009Z")
+    .replaces(THIRD_REALITY_Switch_CLUSTER)
+    .number(
+        attribute_name=THIRD_REALITY_Switch_CLUSTER.AttributeDefs.on_to_off_delay.name,
+        cluster_id=THIRD_REALITY_Switch_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        unit=UnitOfTime.SECONDS,
```
TheJulianJES, line 43: We can also add the duration device class here, since that's what it is:
```suggestion
        unit=UnitOfTime.SECONDS,
        device_class=NumberDeviceClass.DURATION,
```
It'll also need to be imported, so add this to the top as well:
```python
```

zhaquirks/thirdreality/switch_v2.py
```diff
@@ -0,0 +1,59 @@
...
+        translation_key="on_to_off_delay",
+        fallback_name="on_to_off_delay",
+    )
+    .number(
+        attribute_name=THIRD_REALITY_Switch_CLUSTER.AttributeDefs.off_to_on_delay.name,
+        cluster_id=THIRD_REALITY_Switch_CLUSTER.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        unit=UnitOfTime.SECONDS,
```
TheJulianJES, line 55: And we can also add the duration device class here:
```suggestion
        unit=UnitOfTime.SECONDS,
        device_class=NumberDeviceClass.DURATION,
```

zhaquirks/thirdreality/switch_v2.py
```diff
@@ -0,0 +1,59 @@
+"""Third Reality Blind Gen2 devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import UnitOfTime
```
TheJulianJES, line 7: To add the import for the device class:

```suggestion
from zigpy.quirks.v2.homeassistant import UnitOfTime
from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
```

(the pre-commit CI action should fix the order automatically or anything like that)

zhaquirks/thirdreality/switch_v2.py
```diff
@@ -0,0 +1,59 @@
+"""Third Reality Blind Gen2 devices."""
```
TheJulianJES: Is this correct? Or should it be something like this?
```suggestion
"""Third Reality Switch Gen2 devices."""
```

TheJulianJES (commented):
Awesome, only a few small things left, then we can merge this.

3reality-support:
Modified

3reality-support:
One more thing, the previous merge was for switch v1 version, and this time we are submitting v2 version with the same mod. Does it have any impact

TheJulianJES:
> One more thing, the previous merge was for switch v1 version, and this time we are submitting v2 version with the same mod. Does it have any impact

I'll go through all PRs now. It's possible we can combine some quirks later. I'll take a look at it then.

TheJulianJES (approved):
Thanks! You can also use `mode="box"` if you don't want a slider to appear for the number entities, but a box instead.
I think by default we have a slider for most number entities.

PR #4386, opened 2025-10-10
Add Third Reality temperature and humidity sensor settings

3reality-support:
## Proposed change

This PR added adaptation code for T_H sensor and added 3 additional private clusters

1. temperature_correction_fahrenheit: Calibrate Fahrenheit
2. temperature_correction_celsius: Calibrate Celsius
3. humidity_correction: Calibrate humidity

## Additional information

8d0fef7efb0f9e50279a95f0d18095a5

## Checklist

### DIFF
--- a/zhaquirks/thirdreality/temperature_and_humidity_sensor_v2.py
+++ b/zhaquirks/thirdreality/temperature_and_humidity_sensor_v2.py
@@ -3,7 +3,7 @@
 from typing import Final
 
 from zigpy.quirks import CustomCluster
-from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2 import NumberDeviceClass, QuirkBuilder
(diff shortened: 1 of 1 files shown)


TheJulianJES (commented):
Cleaned up similar to https://github.com/zigpy/zha-device-handlers/pull/4387.
I'll combine those quirks in a future PR, since they're very similar: only the `PollControl` should be removed for just the one sensor. I'll make sure to do that, but we can still use `clone()` on a base quirk.

TheJulianJES:
I might rename `"correction"` to `"offset"` in the attribute names as well then. That will change the unique ID, so if you might see new config entities appear in your dev environment for these devices when running the new versions later. That's expected, as you ran the initial version with the old names.

TheJulianJES:
The same question in https://github.com/zigpy/zha-device-handlers/pull/4380 also applies to this device/quirk.
Please answer in the linked PR (soil sensor).

PR #4390, opened 2025-10-11
Add Third Reality garage door lite sensor settings

3reality-support:
## Proposed change

This PR added adaptation code for garage door lite sensor and added 2 additional private cluster

1. open_delay_time： set for 5 seconds, after triggering, display on after 5 seconds
2. z_axis_calibration：When the device is placed flat, the theoretical z-axis angle is 90 degrees, but there is an error, so this attribute needs to be used for calibration to get as close to the theoretical value as possible

## Additional information

7c0d669c4c0f08f5940c5a052e1fc9d0

## Checklist

### DIFF
--- a/zhaquirks/thirdreality/garage_door_tilt_sensor.py
+++ b/zhaquirks/thirdreality/garage_door_tilt_sensor.py
@@ -18,7 +18,7 @@ class ThirdRealityGarageCluster(CustomCluster):
     class AttributeDefs(BaseAttributeDefs):
         """Define the attributes of a private cluster."""
 
-        undetected_to_detected_delay: Final = ZCLAttributeDef(
+        open_delay_time: Final = ZCLAttributeDef(
             id=0x0000,
             type=t.uint16_t,
             is_manufacturer_specific=True,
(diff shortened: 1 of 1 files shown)


zhaquirks/thirdreality/garage_door_tilt_sensor.py
```diff
@@ -0,0 +1,54 @@
+"""Third Reality garage door lite sensor devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+from zigpy.quirks.v2.homeassistant import NumberDeviceClass, UnitOfTime
```
TheJulianJES:
```suggestion
from zigpy.quirks.v2.homeassistant import UnitOfTime
from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
```

zhaquirks/thirdreality/garage_door_tilt_sensor.py
```diff
@@ -0,0 +1,55 @@
...
+        device_class=NumberDeviceClass.DURATION,
+        translation_key="open_delay_time",
+        fallback_name="Open delay time",
+    )
+    .write_attr_button(
+        attribute_name=ThirdRealityGarageCluster.AttributeDefs.z_axis_calibration.name,
+        cluster_id=ThirdRealityGarageCluster.cluster_id,
+        attribute_value=0x01,
+        translation_key="enable_z_axis_calibration",
+        fallback_name="Enable Z axis calibration",
```
TheJulianJES: What about this name?
```suggestion
        translation_key="calibrate_z_axis",
        fallback_name="Calibrate Z axis",
```

TheJulianJES (approved):
Thanks!

PR #4396, opened 2025-10-13
Fix Frient EMI Norwegian HAN divisor

TheJulianJES:
## Proposed change

This fixes the Metering divisor for the Frient EMI Norwegian HAN to a value of `1000` and ignores all attribute updates for that attribute, as the device apparently sends attribute reports with a value of `512` randomly.

TODO:
- add tests
- address TODOs in quirk
- get device diagnostics
- get invalid device report data (turn on debug logs for this)
 

## Additional information

See Z2M issue for reference:
- https://github.com/Koenkk/zigbee-herdsman-converters/issues/3066

## Testing

This is an experimental and untested quirk. We need device diagnostics to make sure this fixes the issue.

If you want to test this as a custom quirk, you can do the following:
1. Install an addon that allows you to modify files in Home Assistant's `/config` directory.
The "Samba share" addon or VSCode addon can be used for this.
2. Create a `zha_custom_quirks` folder in Home Assistant's `/config` directory. So, in the folder where HA's `configuration.yaml` is.
3. Use the download button on the [linked GitHub page](https://github.com/TheJulianJES/zha-device-handlers/blob/tjj/frient_emi_norw_han/zhaquirks/develco/emi_han.py) to get the custom quirk as a `.py` file.
4. Copy the custom quirk file you've just downloaded into the newly created `zha_custom_quirks`.
5. Now, add the following to HA's `configuration.yaml`:
```yaml
zha:
```
(Make sure you only have one `zha:` entry in your `configuration.yaml`. If you modified other ZHA YAML settings, you may need to merge your modifications with the one above.)

6. Restart Home Assistant.
7. Check if ZHA still works. if not, check the logs and post any errors.
8. Check if the custom quirk works as expected.

If you search the web, you should also be able to find some more info on installing custom quirks.

## Device diagnostics
We ideally need these before merging:
> You can find the diagnostics information by going to the device page, clicking the
  three dots, and then by clicking on "Download diagnostics". Drag-and-drop the
  downloaded file into this section.

## Checklist

### DIFF
--- a/zhaquirks/develco/emi_han.py
+++ b/zhaquirks/develco/emi_han.py
@@ -25,9 +25,7 @@ class FrientMetering(CustomCluster, Metering):
 
 
 (
-    # TODO: check which manufacturer name is used by the device
     QuirkBuilder("frient A/S", "EMIZB-132")
-    .applies_to("Develco Products A/S", "EMIZB-132")
-    .replaces(FrientMetering, endpoint_id=2)  # TODO: check endpoint
+    .replaces(FrientMetering, endpoint_id=2)
     .add_to_registry()
 )
--- a/zhaquirks/develco/emi_han.py
+++ b/zhaquirks/develco/emi_han.py
@@ -1,0 +1,33 @@
+"""Frient Electricity Meter Interface Norwegian HAN."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder
+import zigpy.types as t
+from zigpy.zcl.clusters.smartenergy import Metering
+
+
+class FrientMetering(CustomCluster, Metering):
+    """Frient EMI Norwegian HAN Metering cluster definition."""
+
+    # fix device issue
+    _CONSTANT_ATTRIBUTES = {Metering.AttributeDefs.divisor.id: 1000}
+
+    def _update_attribute(self, attrid: int | t.uint16_t, value: Any) -> None:
+        """Update attribute with value."""
+        # prevent attribute_updated events for divisor
+        if attrid == Metering.AttributeDefs.divisor.id:
+            return
+        super()._update_attribute(attrid, value)
+
+
+(
+    # TODO: check which manufacturer name is used by the device
+    QuirkBuilder("frient A/S", "EMIZB-132")
+    .applies_to("Develco Products A/S", "EMIZB-132")
+    .replaces(FrientMetering, endpoint_id=2)  # TODO: check endpoint
+    .add_to_registry()
+)
--- a/tests/test_develco.py
+++ b/tests/test_develco.py
@@ -141,24 +141,9 @@ async def test_frient_emi(zigpy_device_from_v2_quirk):
         assert result == (foundation.Status.SUCCESS, "done")
 
 
-def _get_packet_data(
-    command: foundation.GeneralCommand,
-    attr: foundation.Attribute | None = None,
-    dirc: foundation.Direction = foundation.Direction.Server_to_Client,
-) -> bytes:
-    hdr = foundation.ZCLHeader.general(1, command, 0, dirc).serialize()
-    if attr is not None:
-        cmd = foundation.GENERAL_COMMANDS[command].schema([attr]).serialize()
-    else:
-        cmd = b""
-    return t.SerializableBytes(hdr + cmd).serialize()
-
-
 async def test_mfg_cluster_events(zigpy_device_from_v2_quirk):
     """Test Frient EMI Norwegian HAN ignoring incorrect divisor attribute reports."""
     device = zigpy_device_from_v2_quirk("frient A/S", "EMIZB-132", endpoint_ids=[1, 2])
-
-    # TODO: maybe get data from real device
 
     metering_cluster = device.endpoints[2].smartenergy_metering
     metering_listener = ClusterListener(metering_cluster)
@@ -173,6 +158,7 @@ async def test_mfg_cluster_events(zigpy_device_from_v2_quirk):
             cluster_id=Metering.cluster_id,
             src_ep=2,
             dst_ep=2,
+            # XXX: get data from real device
             data=t.SerializableBytes(b'\x1c\x00\x00\x01\n\x02\x03"\x00\x02\x00'),
         )
     )
--- a/tests/test_develco.py
+++ b/tests/test_develco.py
@@ -2,9 +2,11 @@
 
 from unittest import mock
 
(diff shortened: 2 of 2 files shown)


uvNikita:
I was just going to submit a PR for this device, but you beat me to it :)

There is a zigbee spec doc for this device which explains issues with `divisor` attribute: [EMI Norwegian HAN - Onics Technical Manual - EMIZB-132.pdf](https://github.com/user-attachments/files/22903673/EMI.Norwegian.HAN.-.Onics.Technical.Manual.-.EMIZB-132.pdf)

image


So manufacture specific attibute 0x0302 conflicts with the standard `divisor` attribute.
The `divisor` attribute should indeed be fixed to 1000 as you've done in this PR which fixes the power readings for me.

However, ideally, it would be possible to write to this attribute to be able to change meter type using enum type like this:

```python
class InterfaceMode(t.enum16):
    """Interface Mode."""

    NorwegianHAN = 0x0200
    NorwegianHANExtraLoad = 0x0201
    NorwegianHANAidon = 0x0202
    NorwegianHANKaifa = 0x0203
    AutoDetect = 0x0204
```

I have no idea if this is something zigpy can handle (fixed attribute read with value 1000, but still possible to write).

If not, at least getting fixed `divisor` would be nice because this would at least fix the power readings.

zhaquirks/develco/emi_han.py
```diff
@@ -0,0 +1,33 @@
...
+    def _update_attribute(self, attrid: int | t.uint16_t, value: Any) -> None:
+        """Update attribute with value."""
+        # prevent attribute_updated events for divisor
+        if attrid == Metering.AttributeDefs.divisor.id:
+            return
+        super()._update_attribute(attrid, value)
+
+
+(
+    # TODO: check which manufacturer name is used by the device
```
uvNikita: I just bought this device and can confirm that manufacturer name is "frient A/S", model name is "EMIZB-132", endpoint is 2.

TheJulianJES: Thanks! I'll update the PR to only include that for now.

TheJulianJES:
@uvNikita Thanks for the comment! Can you provide both a diagnostics file for the device (ideally without the quirk, but shouldn't really matter) and a debug log of the "invalid divisor" attribute report? Would be great to have the actual messages the device sends in the tests.

uvNikita:
Here is signature and diagnostics files, let me know if you need something else:
[zha-frient A_S EMIZB-132-signature.json](https://github.com/user-attachments/files/22906141/zha-frient.A_S.EMIZB-132-signature.json)
[zha-frient A_S EMIZB-132-diagnostics.json](https://github.com/user-attachments/files/22906142/zha-frient.A_S.EMIZB-132-diagnostics.json)

As for the invalid divisor message, how can I get it? I enabled debug logging in zha integration and grepping my home assistant logs with:

```bash
$ sudo journalctl -u home-assistant --since today --grep 0x0302 -f
```

But nothing shows up for now, do I need to do anything else?

uvNikita:
It looks like solution proposed [here](https://github.com/zigpy/zha-device-handlers/pull/4421) should also work for this device (?)

TheJulianJES:
Yeah, I'll take another look at this device if the linked solution is generally ok. We may be able to ignore all "manufacturer specific" requests (e.g. attribute reports) on the `Metering` cluster for this Frient devices and then have another "fake" cluster that has all the manufacturer attributes.

Then we might not need to ignore specific attribute reports, like divisor, per ID on the `Metering` cluster.

TheJulianJES:
Can you send me all debug logs (via Discord or email)? Otherwise, try grepping logs for the nwk id.
For the divisor to change, it seems like your meter must be in auto-detect mode, so the attribute switches around.

zhaquirks/develco/emi_han.py
TheJulianJES: Removed the other manufacturer name with [`9d015c0` (#4396)](https://github.com/zigpy/zha-device-handlers/pull/4396/commits/9d015c0d01843c6e116497ebe3609dd53c3de6b5).

TheJulianJES:
I'll merge these changes for now. We can also use `ManufacturerDeviceV2` and a substitution cluster, similar to the other quirk for this device to change the interface mode.

We may also need to forward attribute updates from the normal `Metering` cluster (for 0x0302/divisor) to the manufacturer substitution cluster then, to make sure we parse those reports when the device switches modes itself.

TheJulianJES:
With https://github.com/zigpy/zigpy/pull/1719, we may no longer need this workaround from the PR.
Zigpy can now differentiate between attribute IDs depending on the manufacturer code. Now, I'm wondering if the attribute report for the manufacturer-specific attribute `0x0302` correctly has the manufacturer bit set as well. If so, we can drop this entire workaround and also implement the interface mode switch as normal.

It would be really nice to get debug logs for this device when such an attribute report comes in.
@uvNikita Is it possible for you to enable debug logs for a bit, remove the meter, maybe put it in the auto-detection mode, and send me the logs?
(And yes, I'll also get to your other PRs soon. We've just been busy with this migration.)

uvNikita:
Sure, I can try to do it 🙂

Should I look for something in the logs to know that I got what you need, e.g. something like 0x0302?

uvNikita:
Here is DEBUG log of messages related to this Frient device after I disconnected and connected it back:

[hass-frient.log](https://github.com/user-attachments/files/25219908/hass-frient.log)

Here are lines related to 0x0302 attribute:

```
350:Feb 10 21:41:54 server hass[37417]: 2026-02-10 21:41:54.015 DEBUG (MainThread) [zigpy.zcl] [0x879C:2:0x0702] Decoded ZCL frame: FrientMetering:Report_Attributes(attribute_reports=[Attribute(attrid=0x0302, value=TypeValue(type=enum16, value=<enum16.undefined_0x0200: 512>))])
351:Feb 10 21:41:54 server hass[37417]: 2026-02-10 21:41:54.015 DEBUG (MainThread) [zigpy.zcl] [0x879C:2:0x0702] Received command 0x0A (TSN 3): Report_Attributes(attribute_reports=[Attribute(attrid=0x0302, value=TypeValue(type=enum16, value=<enum16.undefined_0x0200: 512>))])
```


Let me know if you need something else.

TheJulianJES:
Awesome!! I was just about to send what you could look for. This is exactly it. Thanks for providing the full logs.

`is_manufacturer_specific=1` in the line above `FrientMetering:Report_Attributes` means the device is correctly setting the manufacturer bit. Theoretically, it means this workaround is no longer required and we can properly implement a sensor or select entity for the interface mode and so on, even though the ID is colliding with the ZCL attribute.

PR #4421, opened 2025-10-23
Add Frient EMI 2 LED pulse configuration

TheJulianJES:
## Proposed change

This adds the missing pulse configuration entity for the Frient Electricity Meter Interface 2 LED.

## Implementation detail

As the attribute ID for the manufacturer specific attribute conflicts with a ZCL defined attribute on the `Metering` cluster, this quirk adds some logic to move out all manufacturer related attributes on that cluster to a separate "fake" cluster.
This cluster just has a different ID compared to the `Metering` cluster with none of the ZCL attributes defined, allowing for the definition of all manufacturer specific attributes.

This solution can be more generalized and easily applied to other devices/quirks as well.
Luckily, there are very few devices where manufacturer specific attribute IDs conflict with ZCL ones (mostly Frient), but we now have an okay-ish solution.

Note: This also works around an issue in ZHA, where manufacturer specific attributes on ZCL clusters are not properly read during initialization/pairing: https://github.com/zigpy/backlog/issues/51. This still needs to be fixed at some point, but at least we're not affected by it with this PR now, as all manufacturer specific attributes are not on a ZCL cluster, but a manufacturer specific one.

### Specific code

This is essentially the entire code needed to make this work:

```python
class ManufacturerDevice(CustomDeviceV2):
    async def request(
        self,
        *args,
        **kwargs,
    ):
        endpoint_id = kwargs["src_ep"]
        cluster_id = kwargs["cluster"]

        if endpoint_id != 0 and cluster_id == 0xFD10:  # new custom cluster defined with that ID
            kwargs["cluster"] = t.ClusterId(0xFD10)

        return await super().request(
            *args,
            **kwargs,
        )
```

The PR implements a more generic solution, allowing for a shared `ManufacturerDevice` between quirks, so only the cluster has to be defined like usual, but with `SUBSTITUTION_FOR` added.

For the (very) few devices that use conflicting manufacturer specific attribute IDs, I think this is the best solution we can implement for now, and honestly, it's not as hacky as I originally thought this would be.

### First attempt

For example, this was my first attempt from a month ago (without having the device, purely relying on tests): https://github.com/zigpy/zha-device-handlers/commit/9d0c92eb40450390f3b6f56a92e1b9a52b456647
It didn't work correctly and it was also way worse than this implementation is.

### Future

If/when zigpy eventually reworks the implementation around attribute IDs per cluster, there will be no breaking changes for the user if we migrate this quirk to that solution, as the quirks v2 entity is device-based for the unique ID and only includes the endpoint ID, not the cluster one. So, we don't even need to change the UID then to keep matching the "old entity".

## Additional information

Fixes https://github.com/home-assistant/core/issues/152075

## Device diagnostics

[zha-9afa7813c02240642802443599f8f31b-frient A_S EMIZB-141-d66737e6b4b8d298aa4a1397d950a368.json](https://github.com/user-attachments/files/23198610/zha-9afa7813c02240642802443599f8f31b-frient.A_S.EMIZB-141-d66737e6b4b8d298aa4a1397d950a368.json) (with quirk)

## Checklist

- [x] Device diagnostics data has been attached (with quirk)


TheJulianJES:
Note: This also requires the following PR to correctly read the `pulse_config` attribute during pairing/initialization:
- https://github.com/zigpy/zha/pull/548

Otherwise, the number entity will have `unknown` state and you can't properly enter a number due to an implementation detail for HA's number platform.

zhaquirks/develco/emi.py
```diff
@@ -1,9 +1,86 @@
...
+
+    async def request(
+        self,
+        *args,
+        **kwargs,
+    ):
+        """Remap cluster IDs for clusters that substitute for others."""
+        # TODO: verify these are always kwargs
+        endpoint_id = kwargs["src_ep"]
+        cluster_id = kwargs["cluster"]
```
TheJulianJES: zigpy always calls `Device.request` with kwargs (from `Endpoint.request` and `Device.reply`). Only one test uses positional args: https://github.com/zigpy/zigpy/blob/1aba0f1c15af94120349c0ba18a58bd9ced8c484/tests/test_device.py#L117

I'd rather not copy the entire method signature: https://github.com/zigpy/zigpy/blob/1aba0f1c15af94120349c0ba18a58bd9ced8c484/zigpy/device.py#L516-L529
Considering the, uh, *improvable* typing for `Device.request`, I don't see much sense in having that duplicated "in the middle" here, especially since we only want to change one kwarg.
Or would you really want the full method signature here as well? It's doable of course.

Generally, I think this implementation is *fine* and will work without issues if new arguments are added. It is still somewhat of a hack after all, but similar to other quirks intercepting zigpy methods.
This will have full test coverage, so if this quirk breaks because of some zigpy changes, we'll know.

zhaquirks/develco/__init__.py
```diff
@@ -37,3 +38,30 @@ class ClientCommandDefs(IasZone.ClientCommandDefs):
...
+            cluster = self.endpoints[endpoint_id].in_clusters.get(cluster_id)
+            substitution_cluster = getattr(cluster, "SUBSTITUTION_FOR", None)
+
+            if substitution_cluster:
+                kwargs["cluster"] = t.ClusterId(substitution_cluster)
+
+        return await super().request(
+            *args,
+            **kwargs,
+        )
```
TheJulianJES, line 67: There'll be another Frient device to use this, so it's moved here for now.

TheJulianJES:
Tried to make the tests a bit more readable. Had some other variants, but this is the "best".
It's only temporary™.. 😄 

I think it's in a mergeable state. edit: now

PR #4425, opened 2025-10-24
Add IKEA Inspelning child lock and LED disable functionality

remmycat:
## Proposed change
The IKEA Inspelning smart plug at firmware 2.4.45 exposes functionality on its manufacturer-specific cluster for enabling a Child Lock (disable the hardware button) and disabling the status LED.

Important: I'm currently expecting that the Tretakt plug works the same, but I do not own a Tretakt device and cannot test this! I'm also happy with not touching the Tretakt device for now, if you prefer to merge the change for Inspelning only.

Proposed change:
- Split `ikea/plug.py` into `ikea/plug.py` for the old Tradfri plug and `ikea/btnplug.py` for the newer Inspelning and Tretakt plugs, which have a physical button and LED.
- Add `Child Lock` switch, that disables the functionality of the physical button. Translation key `child_lock` already exists.
- Add `Disable LED` switch, that disables the Status LED regardless of device state. See note in Additional Information.

## Additional information

I expected the functionality to exist, as a project integrating with the Dirigera hub found that these settings [are possible to configure](https://github.com/lpgera/dirigera/issues/10#issuecomment-2380893254).
After doing a device scan using ZHA tools I discovered the following:

`0xFC85` is a manufacturer-specific cluster and exposes two boolean attributes that can be set using the manufacturer ID (`4476`):

I tried them out on my Inspelning plugs and confirmed the following functionality:
`0x0000`: Boolean, Enable Child Lock, default False (`0`) (disables hardware button when on)
`0x0001`: Boolean, Enable Status LED, default True (`1`)

NOTE: The plug's logic is "Enable LED" - Changing the default `1` to a value of `0` deactivates the status LED. I mostly chose to flip the control ("Disable LED") because it matched existing settings in other devices, and because there is no `enable_led` translation as of now. I'm happy to change this to closer reflect the device's logic.

## Device diagnostics

[zha-6053e39fefe9a6b5b6b18e07810366d3-IKEA of Sweden INSPELNING Smart plug-427025dff2ec6fa8d4f70bcd9e6135e4.json](https://github.com/user-attachments/files/23124430/zha-6053e39fefe9a6b5b6b18e07810366d3-IKEA.of.Sweden.INSPELNING.Smart.plug-427025dff2ec6fa8d4f70bcd9e6135e4.json)

(At the time of diagnosis the switch is on, child lock is on, Status LED is enabled.)

## Checklist

### Steps Needed For Checking Everything

To make sure the changes are tested, someone needs to verify that this does indeed work for Tretakt plugs, and if the cluster/attribute IDs differ or not.
**Update**: I removed support for Tretakt, this can be added in the future if someone can confirm this.

Regarding tests I'm not sure what kind of test would be useful to the project in this case, I'm happy to implement a suggestion.

Also, I'd like someone to answer my open questions:
- Should we go forward with inverting the "Enable LED" logic as I proposed?
- If yes, is flipping the on/off values correct, or should I set `force_invert` or something else? I chose the values approach based on the same happening in other quirks.

## Further Notes

Reference to Z2M discussion I created in case someone wants to implement this for Z2M as well: https://github.com/Koenkk/zigbee2mqtt/discussions/29257


puddly:
Awesome! I can finally remove the piece of electrical tape I use for this.

puddly (approved):

puddly:
One minor suggestion: if you know the firmware version that exposes this feature, you can gate this quirk to not apply to older firmwares: https://github.com/zigpy/zigpy/blob/1aba0f1c15af94120349c0ba18a58bd9ced8c484/zigpy/quirks/v2/__init__.py#L683

You'll need to make a second one that removes the `LevelControl` cluster as a model-specific fallback for older versions.

remmycat:
Same, regarding the tape 😅 In addition to setting the startup behaviour to On and disabling the switch entity, I can almost convince myself that the sockets won't accidentally shut off…

Re version:
Unfortunately I do not know since when these attributes exist and I'm not sure how to find out. The firmware I stated in the PR is just the newest available one, for which I discovered the functionality.
If we want to be safe about it I could just set the current firmware as the minimum firmware version.

zhaquirks/ikea/btnplug.py
```diff
@@ -0,0 +1,50 @@
...
+        translation_key="child_lock",
+        fallback_name="Child lock",
+    )
+    .switch(
+        IkeaSmartPlugCluster.AttributeDefs.enable_led.name,
+        IkeaSmartPlugCluster.cluster_id,
+        translation_key="disable_led",
+        fallback_name="Disable LED",
+        off_value=1,
+        on_value=0,
```
TheJulianJES: minor thing: let's swap the order, so that `translation_key` and `fallback_name` are at the bottom. That's just how we have it in all other quirks:
```suggestion
        off_value=1,
        on_value=0,
        translation_key="disable_led",
        fallback_name="Disable LED",
```

TheJulianJES:
> Unfortunately I do not know since when these attributes exist and I'm not sure how to find out. The firmware I stated in the PR is just the newest available one, for which I discovered the functionality.
If we want to be safe about it I could just set the current firmware as the minimum firmware version.

You should be able to check the firmware entity, or read the `current_file_version` attribute on the `OTA` cluster using the Manage Zigbee device menu. It's on the device page -> three buttons on the left -> Manage Zigbee device -> Clusters tab (default) -> select OTA cluster -> select `current_file_version` attribute -> Press "Read Zigbee attribute".
Doing it with the dialog should freshly read the attribute again.

TheJulianJES:
Ah, you actually provided the diagnostics file! 😄 Looks to be `0x2040045` (`33816645` in decimal). Can you confirm using the dialog steps I mentioned? Sometimes, the latest version isn't cached properly.

remmycat:
Yes `0x02040045` is the firmware I tested this on (i only now noticed that the `2.4.45` version name, that is also returned by the `sw_build_id` attribute seems to be encoded as hex digits)

And I'm sorry if I wasn't being clear - I know the current firmware of the device. I just don't know if or when these attributes got added in the past. For all I know they might have been there since the first version that got shipped.

But I think setting the current version as minimum is a good idea then - If someone happens to ask for older version support, that can still be done in the future.

puddly:
`0x02040045` seems to have been the only firmware update for the device, published last September. There's no earlier one to download, just the firmware that ships with it.

TheJulianJES (commented):
The "manufacturer-model pair is unique" test needs to be improved to ignore firmware version filters, but I don't think we actually need a firmware filter in this case now, since all plugs should have these options?

(Also, no need to squash and force push for this repo. We squash all PRs before merging them anyways)

remmycat:
Right, it is also the version that is preinstalled on units sold in the Ikea store now (at least in Germany last week).

I had a look in https://github.com/zigpy/zha-device-handlers/issues/3374 and did not find the relevant attributes in the original debug info, so it might be an issue for these devices that haven't been updated yet.

- I applied the suggestion regarding code sorting
- For now I changed the PR to only update Inspelning because I have no data on Tretakt, but I left a comment in the code.

I also implemented the suggested firmware filter with a fallback, however now the tests fail, because both the fallback and the new quirk are for the same model.
Should I just take the fallback out, or change the tests? Or did I make a mistake in how I added the fallback?

remmycat:
I removed the firmware filter and fallback quirk and checked that the newest code version works in my home assistant.

Edit: And thanks for the note re squashing, I wasn't sure

TheJulianJES:
I think this is fine now 👍. It would be great if we can test this for a Tretakt plug as well, but we can always add it in the future. @puddly, you don't happen to have the Tretakt plugs as well? 😄 

One more thing: we should probably name the new quirk file `plug_inspelning.py` instead. It's a bit easier to recognize next to the more common `plug.py` quirk then, and doesn't require you to know IKEA's naming system.

(Side note: We may also wanna add a `require_attr_exists` argument to v2 quirk entity methods. This has come up in the past and would allow us to not create unsupported entities if the attribute doesn't exist (/is unsupported), without having to create multiple quirks.
For this instance, a firmware match filter would be completely fine though. I'll make sure to improve the "unique v2 quirks" test for future devices that need it. The test was added due to me missing some duplicate v2 quirks for Tuya devices.)

remmycat:
Alright, I renamed the file as suggested and fixed the comment in the `plug.py` file 🙂

puddly:
For `require_attr_exists`, I guess we could have ZHA try to read the attribute if it isn't explicitly unsupported (though we do have to clear those on OTA!). That sounds like it would work well.

TheJulianJES:
> For `require_attr_exists`, I guess we could have ZHA try to read the attribute if it isn't explicitly unsupported (though we do have to clear those on OTA!). That sounds like it would work well.

Yep! That would be similar to an option we have for (some) native ZHA entities. There's an issue or PR in *some* repo *somewhere* with more discussion on this for v2 quirks. I'll include that info when I get to doing the PRs for it.
The whole reinterview/clearing attribute cache task is probably the bigger issue. I'll put both of those in the backlog, so we have it listed somewhere.

TheJulianJES (approved):
LGTM, thanks!

MattWestb:
Firmware check for Inspelning is **not** needed the the first shipped firmware was not working OK for metering cluster and need being updated for working at all so if the user is not updating them they is not working at all only OnOff.
I have Tretackt  (3 Tact = on two tree jump) but dont have experience of Quirk V2 but if some can helping i can testing it with ZBT-_X_ in my test system.

MattWestb:
LED and Child lock is working on Tretack only adding `    .also_applies_to(IKEA, "TRETAKT Smart plug")` so now i dont need doing the _dirty_ "**Puddly** _hack_" on Tretackts !!!

MattWestb:
I think "plug_G3.py" is better then G4 is being thread (perhaps with Zigbee if they is implanting the dual / triple working mode as they have hinting of).
IKEA Gens: 
1 = Potted ZLL MG1 (all updated to ZB3)
2 = Unpotted ZB3 still MG1
3 = MG2X devices ZB3 only (MG21)
4 = MG2X with Zigbee / Thread (MG24+)

TheJulianJES:
I'll merge this as is for now.

@MattWestb If I understand you correctly, does this work on all TRETAKT firmware versions as well? If so, you can do a PR just adding the one model variant. If it only works on some firmware versions, we'll need to do some firmware version filtering.
Let me know what's right as soon as possible ideally, since the HA beta is releasing tomorrow. Please ping me!

MattWestb:
Its works on current IKEA production firmware by adding the device model string.
Pleas add the to the quirk so it getting the same functionality.

MattWestb:
Sorry was forgetting the PING @TheJulianJES  !!

TheJulianJES:
Thanks! @MattWestb. I'll add it to the same quirk then!

EDIT: See https://github.com/zigpy/zha-device-handlers/pull/4438

MattWestb:
Great and considering renaming it to some more general for the 3rd gen plugs as you is the maintainer.

MattWestb:
By the way i was looking on` ASKVADER on/off switch` (nice second gen plugin unit) and its only have the standard WWAH cluster that is not defined in ZHA (but is standard) and the level control cluster as it support level control with on/off as one good light domain device shall do..

PR #4438, opened 2025-10-28
Add IKEA Tretakt child lock and LED disable functionality

TheJulianJES:
## Proposed change

@MattWestb mentioned the IKEA TRETAKT plugs support disabling LED and child lock as well, so this PR adds support for it.

## Additional information

Follow-up to https://github.com/zigpy/zha-device-handlers/pull/4425 
Reference: https://github.com/zigpy/zha-device-handlers/pull/4425#issuecomment-3456436549

## Device diagnostics

Would be ideal if they can provided as well, so we can add them to our ZHA diagnostics tests.

## Checklist


zhaquirks/ikea/plug_g2.py
```diff
@@ -29,6 +29,7 @@ class AttributeDefs(BaseAttributeDefs):
 # remove LevelControl for plugs to not show config options in ZHA
 (
     QuirkBuilder(IKEA, "INSPELNING Smart plug")
+    .applies_to(IKEA, "TRETAKT Smart plug")
```
TheJulianJES, line 32: Note: `also_applies_to` is the same as `applies_to`. The different naming is due to quirk builders being allowed to be created without any manufacturer/model info now, so the "also" doesn't make sense in those case.

Doesn't really matter. They're both the same for now.

PR #4443, opened 2025-10-29
Clean up Third Reality switch quirk `3RSS009Z`

TheJulianJES:
## Proposed change

This cleans up the Third Reality switch `3RSS009Z` v2 quirk to be similar to the plug quirks:
- https://github.com/zigpy/zha-device-handlers/pull/4109

## Additional information

Follow-up to:
- https://github.com/zigpy/zha-device-handlers/pull/4383

## Checklist


TheJulianJES (commented):
This is essentially the same quirk as the ones in `plug`. We could also merge them. We'd have a difference between plug and switch in naming though.

PR #4449, opened 2025-10-30
Add Candeo LED strip controllers variants

dhc25:
Adds C-ZB-LC20v2 to the different modes of the zigbee LED strip controller


TheJulianJES (approved):
Thanks! Is this something important that should ideally still go into the current 2025.11.0 beta?

dhc25:
@TheJulianJES It would be great if you can. Thanks

PR #4458, opened 2025-11-02
Fix energy and voltage for Aqara T2 relay

abmantis:
## Proposed change

SSIA.

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/switch_acn047.py
+++ b/zhaquirks/xiaomi/aqara/switch_acn047.py
@@ -74,6 +74,15 @@ class MultistateInputCluster(CustomCluster, MultistateInput):
             super()._update_attribute(attrid, value)
 
 
+class T2MeteringCluster(MeteringCluster):
+    """T2 Metering cluster to fix divisor."""
+
+    def _update_attribute(self, attrid, value):
+        if attrid == self.CURRENT_SUMM_DELIVERED_ID and value > 0:
+            value = value / 1000
+        super()._update_attribute(attrid, value)
+
+
 class OppleCluster(XiaomiAqaraE1Cluster):
     """Opple cluster."""
 
@@ -197,7 +206,7 @@ class AqaraT2Relay(XiaomiCustomDevice):
                     OnOff.cluster_id,
                     MultistateInputCluster,
                     DeviceTemperature.cluster_id,
-                    MeteringCluster,
+                    T2MeteringCluster,
                     ElectricalMeasurementCluster,
                     OppleCluster,
                 ],
--- a/zhaquirks/xiaomi/aqara/switch_acn047.py
+++ b/zhaquirks/xiaomi/aqara/switch_acn047.py
@@ -77,10 +77,9 @@ class MultistateInputCluster(CustomCluster, MultistateInput):
 class T2MeteringCluster(MeteringCluster):
     """T2 Metering cluster to fix divisor."""
 
-    def _update_attribute(self, attrid, value):
-        if attrid == self.CURRENT_SUMM_DELIVERED_ID and value > 0:
-            value = value / 1000
-        super()._update_attribute(attrid, value)
+    _CONSTANT_ATTRIBUTES = {
+        MeteringCluster.AttributeDefs.divisor.id: 1000000,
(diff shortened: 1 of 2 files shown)


zhaquirks/xiaomi/aqara/switch_acn047.py
```diff
@@ -74,6 +74,15 @@ def _update_attribute(self, attrid, value):
             super()._update_attribute(attrid, value)
 
 
+class T2MeteringCluster(MeteringCluster):
+    """T2 Metering cluster to fix divisor."""
+
```
puddly, line 79: Would this work? It should avoid requiring any runtime-testable code:

```python
    _CONSTANT_ATTRIBUTES = {
        MeteringCluster.AttributeDefs.divisor.id: 1000,
    }
```

abmantis: Yes! But its 1000000 instead since 1000 is already what we have by default for xiaomi: https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/xiaomi/__init__.py#L712

puddly: Ah. In that case you'd need to do `_CONSTANT_ATTRIBUTES = {**MeteringCluster._CONSTANT_ATTRIBUTES, MeteringCluster.AttributeDefs.divisor.id: 1000000,}`.

Is the device really sending reports scaled this much? Or are we hiding a `* 1000` somewhere?

abmantis: > Ah. In that case you'd need to do `_CONSTANT_ATTRIBUTES = {**MeteringCluster._CONSTANT_ATTRIBUTES, MeteringCluster.AttributeDefs.divisor.id: 1000000,}`.

Oh, duh! This is a sign that I should probably go to sleep... :laughing: 

> Is the device really sending reports scaled this much? Or are we hiding a `* 1000` somewhere?

We scale it here: https://github.com/abmantis/zha-device-handlers/blob/xiaomi_t2_fix_energy/zhaquirks/xiaomi/__init__.py#L

zhaquirks/xiaomi/aqara/switch_acn047.py
```diff
@@ -74,6 +74,14 @@ def _update_attribute(self, attrid, value):
...
+    _CONSTANT_ATTRIBUTES = {
+        MeteringCluster.AttributeDefs.divisor.id: 1000000,
+    }
```
TheJulianJES, line 83: To not remove all other constants attribute, they'll need to be added again:
```suggestion
    _CONSTANT_ATTRIBUTES = {
        **MeteringCluster._CONSTANT_ATTRIBUTES,
        MeteringCluster.AttributeDefs.divisor.id: 1000000,
    }
```
(You might not notice anything breaking when not doing this, but that's because the other attributes would have been read during pairing, where they are cached.)

zhaquirks/xiaomi/aqara/switch_acn047.py
TheJulianJES: Z2M also has a special case for this device:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/ea780c92873ad254818f9bb7060f9417879893d6/src/lib/lumi.ts#L503-L507

The standard 1000 divisor is because of us needing to do the trip through zigpy cache/ZCL datatypes: https://github.com/abmantis/zha-device-handlers/blob/3916d9b9efa74e35a745c1c3d88cccb1645e86f2/zhaquirks/xiaomi/__init__.py#L329-L333
Z2M doesn't need to do this, but we'd just cut off all decimal points otherwise.

TheJulianJES (approved):
LGTM. Thanks!

TheJulianJES:
Energy and voltage just didn't display for this device before at all, right?

abmantis:
> Energy and voltage just didn't display for this device before at all, right?

Power was already displayed and updated more frequently than the custom xiaomi attribute updates.
IIRC energy and voltage where shown as 0. If you think it is relevant, I can unpair a device and pair it to my dev env to check it out with a clean database.

PR #4460, opened 2025-11-02
Unify quirks for frient SMSZB-120, HESZB-120, FLSZB-110

PrintingRat:
1. Hidden binary input entity.

2. Hidden unsupported siren configuration.

3. Added Max Duration attribute support.

4. Added IAS test support bit.

## Proposed change

Combined quirk for all three devices - SMSZB-120, HESZB-120, FLSZB-110.

SMSZB-120:
image

HESZB-120:
image

FLSZB-110:
image

## Device diagnostics

[zha-01K85V12X81M1HFSH4S6W7FT2D-frient A_S HESZB-120-03a0a77e7c1f77d316cd541a30d9e11b.json](https://github.com/user-attachments/files/23291208/zha-01K85V12X81M1HFSH4S6W7FT2D-frient.A_S.HESZB-120-03a0a77e7c1f77d316cd541a30d9e11b.json)
[zha-01K85V12X81M1HFSH4S6W7FT2D-frient A_S SMSZB-120-c7d68ee08423575724965700cfe639e7.json](https://github.com/user-attachments/files/23291209/zha-01K85V12X81M1HFSH4S6W7FT2D-frient.A_S.SMSZB-120-c7d68ee08423575724965700cfe639e7.json)
[zha-01K85V12X81M1HFSH4S6W7FT2D-frient A_S FLSZB-110-3cbb7bb473dc8c7c151d1329514fc75e.json](https://github.com/user-attachments/files/23291210/zha-01K85V12X81M1HFSH4S6W7FT2D-frient.A_S.FLSZB-110-3cbb7bb473dc8c7c151d1329514fc75e.json)

## Checklist


aubreyz:
This explains a bunch of stuff. How long do you think it might take to be merged?
Can you explain stepwise how to install the quirk for some beta testing prior to that?

Side comment: The "Works with Home Assistant" badge and statements by manufacturer should really not be used when it really is not set up to work in any way. That just discredits the whole enterprise..

aubreyz:
> This explains a bunch of stuff. How long do you think it might take to be merged? Can you explain stepwise how to install the quirk for some beta testing prior to that?
> 
> Side comment: The "Works with Home Assistant" badge and statements by manufacturer should really not be used when it really is not set up to work in any way. That just discredits the whole enterprise..

OK I installed the quirk and can confirm that it works but with some issues. I tested the IAS test (a critical thing that was previously missing) and that works. The incorrect entities are removed. The max_duration does not have any default value (not sure if that is expected) - if I switched the alarm on without changing anything the alarm stops after about 8 seconds. If I change the default nul value to 3 seconds then it alarms for 3 seconds. 

However if I change the duration to 40 seconds and trigger the siren then it stops after about 8 seconds so something is clearly not right there. In fact nothing seems to increase it beyond the 8 second apparent default but in can be decreased. 

The temperature seems to be reporting correctly. It is not clear what happens in an actual fire -- presumably both alarm switch will go on, and smoke will show as on -- but will the alarm go off after 8 seconds??

PrintingRat:
> > This explains a bunch of stuff. How long do you think it might take to be merged? Can you explain stepwise how to install the quirk for some beta testing prior to that?
> > Side comment: The "Works with Home Assistant" badge and statements by manufacturer should really not be used when it really is not set up to work in any way. That just discredits the whole enterprise..
> 
> OK I installed the quirk and can confirm that it works but with some issues. I tested the IAS test (a critical thing that was previously missing) and that works. The incorrect entities are removed. The max_duration does not have any default value (not sure if that is expected) - if I switched the alarm on without changing anything the alarm stops after about 8 seconds. If I change the default nul value to 3 seconds then it alarms for 3 seconds.
> 
> However if I change the duration to 40 seconds and trigger the siren then it stops after about 8 seconds so something is clearly not right there. In fact nothing seems to increase it beyond the 8 second apparent default but in can be decreased.
> 
> The temperature seems to be reporting correctly. It is not clear what happens in an actual fire -- presumably both alarm switch will go on, and smoke will show as on -- but will the alarm go off after 8 seconds??

Thanks for taking your time to try it out.

About your points:

1. The Max Duration has a default value of 240s (that's actual attribute on the device default value). If it doesn't show anything for you, it means it hasn't read it from the device. This makes me suggest you test the quirk in this order: load custom quirk --> restart your HA --> remove your sensor --> add it again - it should appear. If you still can't see it, share debug logs, otherwise it's impossible to find out what's wrong.

2. The default short siren duration (8 seconds you mention), you experience, has nothing to do with the device and it seems to be ZHA's default value for siren "on". You can change it in advanced settings of the siren entity or via automation. There is Max Duration attribute (device), and there is ZHA's siren entity duration setting - please don't mix them.

3. Regarding actual fire - there are two separate clusters IAS Zone and IAS WD. The siren you can control with the entity is IAS WD, and it allows you to simply turn the siren on and off, and also change the durations, etc.

In case of actual fire, it's IAS Zone that handles it and it is always in priority for the device, meaning, you cannot control it anyhow remotely. When smoke/fire/water is detected, device starts the siren and will continue to unless you physically interact with the device. Actual dangerous event has nothing to do with siren entity you can see.

aubreyz:
Thanks for that @PrintingRat and thanks for writing the Quirk. 

As I said the duration entity seems to be doing nothing at all beyond about 5 seconds (I mentioned 8 before but that was incorrect. However, I noticed I noticed there was an option in the gui to add a script to the Siren, the options are Warn or Sqawk where the only difference seems to be the length of the alarm. Squawk lasts 2 seconds, and Warn last 5 seconds (these are inbuilt settings and namings for the  IAS (Intruder Alarm System) Warning Device cluster in Zigbee, not the standard siren implementation.

I cannot find any way to trigger the siren to go for any longer duration than the Sqawk. 

Is it perhaps the case that the Quirk is exposing the Frient alarm is likely being exposed through the IAS WD (Warning Device) cluster (0x0502) rather than as a proper siren entity. These are hardcoded in the Zigbee specification itself - they're not configurable durations.

Prior to this Quirk it did go for a long duration before switching off.

PrintingRat:
This device doesn't support Volume, Warn or Squawk - this is visible because ZHA siren entity shows it always - this requires ZHA to create a custom siren entity just for devices (like this) that don't have this.

To trigger a custom duration, please use automation like this:

image

aubreyz:
Ahh - I just realised if I run this with any duration, it actually does run the siren for the duration specified in the duration entity (ie not the 50 seconds as in the below). This is highly confusing to me anyway. Just turning on the siren via the switch that shows up on the dashboard does it for the Squawk duration (5 seconds)

So experiment 1: This gives the Squawk duration (5 seconds) not the duration set in the duration entity

```
action: siren.turn_on
metadata: {}
target:
  entity_id: siren.fire_alarm_frient
```
  
experiment 2:  This gives the duration set in the entity exposed by the Quirk (not 50 seconds) in my case 23 seconds. So the setting of 50 actually does nothing apart from allowing the duration entity to work

```
action: siren.turn_on
metadata: {}
data:
  duration: "50"
target:
  entity_id: siren.fire_alarm_frient
```

experiment 3: Just flipping the switch on the Quirk dashboard
This again gives the standard Squawk duration before it switches off

PrintingRat:
"Just turning on the siren via the switch that shows up on the dashboard does it for the Squawk duration (5 seconds)"

Yes, correct, this also got me very confused :D

To overcome strange behavior I even made a version where I create my own "custom siren entity", but it's not really recognized as siren entity elsewhere, so I decided to go with the default ZHA one hoping it will get adjusted for this kind of devices soon.

aubreyz:
So what someone needs to do to get a siren which goes for 20 seconds is something like this where the duration entity is set first to 20, and the number 50 at the bottom is totally irrelevant in terms of its value but I guess needs to be longer than the first number and at least some value needs to be present. 

This seems to me to be a bug in Home Assistant, but anyway there is a working mechanism

```
actions:
  - device_id: 99e037c6ba6d498bf3f0add022d47e80
    domain: number
    entity_id: b84ed248b9468ee043733322b24665c6
    type: set_value
    value: 20
  - action: siren.turn_on
    metadata: {}
    target:
      entity_id: siren.fire_alarm_frient
    data:
      duration: "50"
```

zhaquirks/develco/smoke_heat_water_alarm.py
```diff
@@ -0,0 +1,77 @@
...
+    .replaces(DevelcoPowerConfiguration, endpoint_id=35)
+    .replaces(FrientSmokeHeatWaterIasZone, endpoint_id=35)
+    .binary_sensor(
+        attribute_name="test",
+        cluster_id=IasZone.cluster_id,
+        endpoint_id=35,
+        entity_type=EntityType.DIAGNOSTIC,
+        translation_key="test",
+        fallback_name="IAS test",
+    )
```
TheJulianJES, line 51: To avoid the virtual ZCL attribute, we can just point the quirks v2 binary sensor at the `zone_status` attribute directly and use an `attribute_converter`, similar to this:
https://github.com/zigpy/zha-device-handlers/blob/e2cafc6b26405a211df7daf222136aaeabe68638/zhaquirks/develco/smoke_alarm_us.py#L51-L59

We can also use the `translation_key="test"` and `fallback_name="Test"` already used by that Frient US smoke sensor quirk.

zhaquirks/develco/smoke_heat_water_alarm.py
```diff
@@ -0,0 +1,77 @@
...
+        cluster_id=IasWd.cluster_id,
+        endpoint_id=35,
+        min_value=0,
+        max_value=65535,
+        step=1,
+        unit=UnitOfTime.SECONDS,
+        translation_key="max_duration",
+        fallback_name="Max duration",
+        unique_id_suffix="max_duration",
+    )
```
TheJulianJES, line 63: I'd opt to leave this out for the initial version until we know what to do with it. Maybe we can just create that entity for all devices that support that attribute directly from ZHA.

zhaquirks/develco/smoke_heat_water_alarm.py
```diff
@@ -0,0 +1,77 @@
...
+        endpoint_id=35,
+        cluster_id=IasWd.cluster_id,
+        function=lambda entity: entity.translation_key
+        in (
+            "default_siren_tone",
+            "default_siren_level",
+            "default_strobe_level",
+            "default_strobe",
+        ),
+    )
```
TheJulianJES, line 74: The quirk can just do `.exposes_feature(SIREN_BASIC)` now to achieve the same thing (hide the siren select entities) AND only exposes a siren entity with the advanced options removed.

TheJulianJES (commented):
Just to add an update here: There were changes multiple HA versions ago, so that the unsupported siren-related entities are now no longer created for most of these devices:
https://github.com/zigpy/zha-device-handlers/blob/e2cafc6b26405a211df7daf222136aaeabe68638/zhaquirks/develco/smoke_alarm.py#L15-L16

However, it looks like the `heat_alarm` quirk (`HESZB-120`) was not adjusted. It's not WWHA certified, so I guess nobody of us had that device. That issue was fixed with https://github.com/zigpy/zha-device-handlers/pull/5156 but that didn't land in HA Core yet.

But in quirks `dev`, the (EU) smoke sensor, heat alarm, and water leak sensor quirks are basically identical – but the US smoke sensor quirk is a bit different.

So I'm not sure I really like the approach of unifying the (EU) smoke alarm + heat alarm + water leak sensors into one quirk, as it goes against what we do with like all other quirks. But on the other hand, they're also very similar devices in the first place, so it may be fine to unify them.

---

I think this would be good to go when the "IAS test" binary sensor is adjusted to use `attribute_converter` and the `fallback_name="Test"` (similar to the [Frient US smoke sensor quirk](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/develco/smoke_alarm_us.py)) + `max_duration` entity removed (for now – until we maybe move it directly into ZHA for all supporting devices) + remove no longer needed `prevent_default_entity_creation` for the select entities + the tests removed, as there's not really any code to test here then. The simple definitions are all checked by ZHA diagnostics snapshots later.

I can also make the changes here later.

---

One more thing: Please redownload the diagnostics for the three devices (with a newer HA Core version) and update the PR body. It looks like the `original_signature` was missing in them, as there was a bug for a few versions a while ago.

tests/test_develco.py
```diff
@@ -187,3 +192,56 @@ async def test_mfg_cluster_events(zigpy_device_from_v2_quirk):
     assert (
         metering_cluster.get(Metering.AttributeDefs.current_summ_delivered.id) == 1234
     )
+
+
+async def test_frient_smoke_heat_water_alarm_entities(zigpy_device_from_v2_quirk):
```
TheJulianJES, line 197: This test can also be removed then, as there won't be any actual code that's "run" in the quirk. It's only really the definitions, which will be tested in the ZHA repo with the diagnostics snapshots.

PR #4482, opened 2025-11-11
Add Third Reality curtain gen2 total cycle times

3reality-support:
## Proposed change

This PR added  added 2 additional private clusters
1. `total_cycle_times`: The large curtain descends from the top tightening point to the bottom, and then rises to the top tightening point in a loop. The total number of cycles is how many times this behavior can be cycled, and the battery will run out of power. The maximum power failure of 81 inch curtains after 220 cycles is 210 times, so this attribute value defaults to 210 times.
2. `last_remaining_battery_percentage`: The percentage of battery remaining from the last time the curtain was powered on until the power was cut off. Calculated based on the distance traveled by the curtains and the maximum distance they can travel.

## Checklist


zhaquirks/thirdreality/curtain_gen2.py
```diff
@@ -34,6 +34,18 @@ class AttributeDefs(BaseAttributeDefs):
...
+            id=0x0003,
+            type=t.uint16_t,
+            is_manufacturer_specific=True,
+        )
+
+        last_remaining_battery_percentage: Final = ZCLAttributeDef(
+            id=0x0004,
+            type=t.uint8_t,
+            is_manufacturer_specific=True,
+        )
```
TheJulianJES, line 47: Should there also be a sensor entity for this? Or is this not info that the user really needs?

TheJulianJES (approved):
Looks good. And thanks for the good description of what the attributes do. That's helpful!

zhaquirks/thirdreality/curtain_gen2.py
3reality-support: Yes, that's a great idea, we'll add it in the next PR.

PR #4484, opened 2025-11-12
Expose Frient EMI LED reset current summation button

TheJulianJES:
## Proposed change

This exposes the Frient manufacturer-specific `current_summation` attribute.
A write attribute button is also added to easily reset this to `0`.  This is what the user will see.

## Additional information

There will be no number entity to change the value, as the attribute is write-only and we'd duplicate some functionality we have in the sensor.

Advanced users can still set the `current_summation` to a value they want via the debug UI or a service action call.

## Checklist

### DIFF
--- a/zhaquirks/develco/emi_led.py
+++ b/zhaquirks/develco/emi_led.py
@@ -28,6 +28,7 @@ class ManufacturerMetering(CustomCluster):
         current_summation: Final = ZCLAttributeDef(
             id=0x0301,
             type=t.uint48_t,
+            access="r",
             is_manufacturer_specific=True,
         )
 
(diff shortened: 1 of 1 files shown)


TheJulianJES (commented):
Works as expected. As mentioned, users can manually write the `current_summation` field via the "Manage Zigbee device" UI or with service calls. Most users shouldn't need to do this, however. For them, there's the simple "Reset summation delivered" button.

zhaquirks/develco/emi_led.py
```diff
@@ -25,6 +25,12 @@ class AttributeDefs(BaseAttributeDefs):
             type=t.uint16_t,
             is_manufacturer_specific=True,
         )
+        current_summation: Final = ZCLAttributeDef(
+            id=0x0301,
+            type=t.uint48_t,
+            access="r",
```
TheJulianJES: But this should be write-only:
```suggestion
            access="w",
```

PR #4485, opened 2025-11-12
Fix Frient EMI LED divisor and multiplier

TheJulianJES:
For testing, see the [comment below](https://github.com/zigpy/zha-device-handlers/pull/4485#issuecomment-3519928906) this one.

## Proposed change

This PR sets the divisor and multiplier to fixed values to avoid possible conflicts with the `interface_mode` manufacturer specific attribute (`0x0302`, same as `divisor`

This PR is just for testing. On the HA forum, it was observed that the EMI LED shoots to random values for the `instantaneous_demand` about once a month. This is just a shot in the dark to see if this would fix that.

See:
https://community.home-assistant.io/t/frient-joins-works-with-home-assistant/926758/63?u=thejulianjes

## Additional information

This is similar to the implementation done for the Frient HAN EMI:
- https://github.com/zigpy/zha-device-handlers/pull/4396

Tests are expected to fail for now. Apparently just using the `CustomCluster` causes the existing EMI LED tests to fail.
I'll fix them before this PR is merged (if at all).

BASED ON (below should to be merged first):
- https://github.com/zigpy/zha-device-handlers/pull/4484

## Checklist


TheJulianJES:
Custom quirk download link:
https://github.com/TheJulianJES/zha-device-handlers/blob/tjj/frient_emi_led_fixed_divisor/zhaquirks/develco/emi_led.py
(use GitHub's download button or view RAW and then download using the browser's save page feature)

If you want to test this as a custom quirk, you can do the following:
1. Install an addon that allows you to modify files in Home Assistant's `/config` directory.
The "Samba share" addon or VSCode addon can be used for this.
2. Create a `zha_custom_quirks` folder in Home Assistant's `/config` directory. So, in the folder where HA's `configuration.yaml` is.
3. Use the download button on the linked GitHub page to get the custom quirk as a `.py` file.
4. Copy the custom quirk file you've just downloaded into the newly created `zha_custom_quirks`.
5. Now, add the following to HA's `configuration.yaml`:
```yaml
zha:
```
(Make sure you only have one `zha:` entry in your `configuration.yaml`. If you modified other ZHA YAML settings, you may need to merge your modifications with the one above.)

6. Restart Home Assistant.
7. Check if ZHA still works. if not, check the logs and post any errors.
8. Check if the custom quirk works as expected.

If you search the web, you should also be able to find some more info on installing custom quirks.

TheJulianJES:
This should no longer be needed with recent zigpy releases (in HA Core 2026.2.0) and later, as zigpy now won't parse attribute reports for the ZCL attributes if the manufacturer specific bit is set. That's a better fix than this PR anyway, as it now allows us to also define the Frient manufacturer specific attributes, even though they collide with built-in ZCL attribute IDs. Zigpy properly handles this.

TheJulianJES:
Actually, we need to add the attribute definitions for the manufacturer specific attributes to work properly here.
For reference, see a related comment here: https://github.com/zigpy/zigpy/pull/1760#discussion_r2800173787

PR #4488, opened 2025-11-13
Add dirty level and pressure for Third Reality filter sensor (3RAP0149BZ)

### LINKED ISSUE #4870: [Device Support]Third reality Air sensor
The dirty level of the device uses 0x000c cluster, but the interface displays temperature. 
Expected to remove the temperature entity, and the first entity name is dirty level

Image

3reality-support:
## Proposed change

This PR added device setting and display 2 entities
1. Dirty level
2. Pressure

## Additional information

77fdd49e-bc29-42ca-875d-90e5cd4dc74e

## Device diagnostics

[3RAP0149BZ.json](https://github.com/user-attachments/files/29999509/3RAP0149BZ.json)

## Checklist

Closes #4870


zhaquirks/thirdreality/air_pressure_sensro.py
```diff
@@ -0,0 +1,62 @@
...
+        cluster_id=ThirdRealityAirPressureSensorCluster.cluster_id,
+        endpoint_id=1,
+        min_value=0,
+        max_value=65535,
+        mode="box",
+        unit=UnitOfPressure.PA,
+        device_class=NumberDeviceClass.PRESSURE,
+        translation_key="send_command_up_threshold",
+        fallback_name="Send command up threshold",
+    )
```
TheJulianJES: Can you clarify what these entities do?
Like, does this configure the threshold of how much a value needs to change before it sends an attribute report? If so, why is there both an "up threshold" and "down threshold"?

3reality-support: send_command_up_threshold：After setting the threshold, if the value of this sampling is larger than the previous one and exceeds the threshold, send the "on" command
send_command_down_threshold：After setting the threshold, if the value of this sampling is smaller than the previous one and exceeds the threshold, send the "off" command

3reality-support:
Add level cluster, the current level attribute is used to determine whether the air conditioning filter needs to be clear

zhaquirks/thirdreality/air_pressure_sensro.py
```diff
@@ -0,0 +1,29 @@
...
+        ],
+    )
+    .sensor(
+        cluster_id=LevelControl.cluster_id,
+        attribute_name="current_level",
+        device_class=SensorDeviceClass.AQI,
+        state_class=SensorStateClass.MEASUREMENT,
+        unit=PERCENTAGE,
+        fallback_name="Air quality level",
+    )
```
TheJulianJES: Is there already a device available that exposes this information on the `LevelControl` cluster? (which is only meant for *controlling* devices.)

If there's already one device out to the public using `LevelControl`, we'll need to look into ZHA to make updates work properly for this cluster. If not, please update this quirk to use the (custom?) attribute you're using now.

3reality-support: We have already used the temperature cluster to replace the level, but forgot to update this configuration – I will modify it immediately.

puddly: You're exposing the air quality via a temperature cluster?

3reality-support: Yes, no need to add scripts anymore! It can be used directly once added to ZHA.

puddly: The temperature cluster is for exposing temperatures, not air quality. If you want to expose air quality, please use an AnalogInput cluster.

TheJulianJES: Alternatively, it could also make sense to just use a custom attribute for this (on a private cluster).

For reference, see this IKEA quirk which uses a custom entity (with the correct HA device and entity class) to display the "VOC level": https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/ikea/vindstyrka.py

3reality-support: I may have misled you with my description – it does not display air quality, but rather the dirtiness level of the filter screen: 0% means the filter is clean, and the higher the value, the dirtier the filter becomes.

3reality-support: In addition, the correct entity will be used to express it in the future

TheJulianJES (commented):
Just in case you haven't seen my comment in the other issue: https://github.com/zigpy/zha-device-handlers/issues/4870#issuecomment-4427520584

jintj:
image
According to the code you provided, the name of the pneumatic entity is wrong
We have a dial switch. When on, we can register the air pressure cluster. When off, we will not register
@zigpy-review-bot

PR #4491, opened 2025-11-14
Fix tamper sensor showing for wrong Frient entry sensors

TheJulianJES:
## Proposed change

This fixes an issue where the binary tamper sensor is shown for the Frient Entry Sensor 2 Pro and Entry Sensor (basic) which both do not support the tamper sensor according to specifications.
It's only supported by the 1st Gen Frient Entry Sensor Pro.

## Additional information

Follow-up to:
- https://github.com/zigpy/zha-device-handlers/pull/4246

## Device diagnostics

Should already be in ZHA diagnostics. ~~I don't have the devices anyway.~~ -> I have the Entry Sensor 2 Pro now.

## Checklist

### DIFF
--- a/zhaquirks/develco/open_close.py
+++ b/zhaquirks/develco/open_close.py
@@ -24,20 +24,20 @@ base_quirk = (
     .prevent_default_entity_creation(endpoint_id=35, cluster_id=BinaryInput.cluster_id)
 )
 
-# Entry Sensor 2 Pro + Entry Sensor (basic), no tamper
+# Entry Sensor 2 Pro, no tamper
 (
     base_quirk.clone()
     .applies_to("frient A/S", "WISZB-131")
-    .applies_to("Develco Products A/S", "WISZB-121")
-    .applies_to("frient A/S", "WISZB-121")
     .add_to_registry()
 )
 
-# Entry Sensor Pro, with tamper
+# Entry Sensor Pro + Entry Sensor (basic), with tamper
 (
     base_quirk.clone()
     .applies_to("Develco Products A/S", "WISZB-120")
     .applies_to("frient A/S", "WISZB-120")
+    .applies_to("Develco Products A/S", "WISZB-121")
+    .applies_to("frient A/S", "WISZB-121")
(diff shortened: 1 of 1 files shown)


zhaquirks/develco/open_close.py
```diff
@@ -16,16 +16,28 @@ class DevelcoPowerConfiguration(PowerConfigurationCluster):
...
     # The binary input cluster is a duplicate
     .prevent_default_entity_creation(endpoint_id=35, cluster_id=BinaryInput.cluster_id)
+)
+
+# Entry Sensor 2 Pro + Entry Sensor (basic), no tamper
+(
+    base_quirk.clone()
+    .applies_to("frient A/S", "WISZB-131")
+    .applies_to("Develco Products A/S", "WISZB-121")
+    .applies_to("frient A/S", "WISZB-121")
```
TheJulianJES: In Zigbee2MQTT, `WISZB-121` (basic Entry Sensor, 1st Gen) supports the tamper sensor, but according to specsheet, it doesn't: https://cdn.prod.website-files.com/67f4edb3e49fc9eedd4d58ff/68627bf514b30153d4133678_Entry%20sensor_frient_v1.0.pdf

TheJulianJES: deCONZ seems to support tamper for the basic sensor (1st Gen) as well: https://github.com/dresden-elektronik/deconz-rest-plugin/blob/41d9a4933f45b6d14fbc96b9df0f8524ffb66c4c/devices/frient/wiszb-121_open_close_sensor.json#L111-L114

I'll adjust this PR to only remove the tamper sensor for the 2nd Gen Pro sensor then.
It would be good to get confirmation on this though 😄

puddly: Let me check the device itself, I might have one. Every one I added explicitly should have tamper sensor support.

TheJulianJES:
Confirmed the "Entry Sensor 2 Pro" (WISZB-131) has no tamper support, but currently exposes that entity.

PR #4504, opened 2025-11-19
Add Aqara FP300 presence sensor quirk

### LINKED ISSUE #4487: [Device Support Request] Aqara FP300 (lumi.sensor_occupy.agl8)
### Problem description

Pairing Aqara FP300 in Zigbee 3.0 Mode with ZHA results in a limited and small set of entities:
- sensors: lux, humidity, temperature
- configuration: firmware
- diagnostics: batterie, identify

What's missing:
- pir/presense sensor entities
- presense detection options
- motion sensitivity options

### Solution description

Probably some quirks are required to get full functionality

### Screenshots/Video

Image

</details>

### Diagnostics information

[zha-01K5F5KH0S5V31DGARVANXYG50-Aqara lumi.sensor_occupy.agl8-4898f06b1056074f3a2b6cfae4479060.json](https://github.com/user-attachments/files/23510829/zha-01K5F5KH0S5V31DGARVANXYG50-Aqara.lumi.sensor_occupy.agl8-4898f06b1056074f3a2b6cfae4479060.json)

### Device signature

```json

{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4447,
    "maximum_buffer_size": 82,
    "server_mask": 11264,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0107",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0012",
        "0x0400",
        "0x0402",
        "0x0405",
        "0xfcc0"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "Aqara",
  "model": "lumi.sensor_occupy.agl8",
}

```

</details>

### Logs

```python

[Paste the logs here]

```

</details>

### Custom quirk

```python

```

</details>

### Additional information

_No response_

cemizm:
## Proposed change

This PR adds a custom quirk for the **Aqara FP300 presence sensor**  
(**manufacturer**: `LUMI`, **model**: `lumi.sensor_occupy.agl8`) to `xiaomi`.

Key points:

- Implements an `AqaraFP300ManuCluster` based on `XiaomiAqaraE1Cluster` for the manufacturer-specific cluster `0xFCC0`.
- Adds support for parsing Aqara TLV attribute `0x00F7` so that:
  - `0xff01-23` is mapped to `BATTERY_VOLTAGE_MV`
  - `0xff01-24` is mapped to `BATTERY_PERCENTAGE_REMAINING_ATTRIBUTE`
- Uses `FP300PowerConfigurationVoltage` (based on `XiaomiPowerConfiguration`) and derives battery percentage from voltage (2850–3000 mV curve), while ignoring directly reported battery percent for FP300.
- Implements `FP300DetectionRangeCluster` as `LocalDataCluster` for detection range handling via manufacturer attribute `0x019A` (`octet string`):
  - decodes/encodes the 5-byte payload (`prefix` + 24-bit mask, little-endian)
  - supports both a numeric mask and HA-friendly switches
- Exposes detection range in two ways:
  - `detection_range_mask` number entity (`0..0xFFFFFF`) for fine-grained control
  - six 1 m config switches for simple control
- Exposes the  FP300 configuration options as entities (e.g. absence delay, motion sensitivity, AI adaptive sensitivity, interference source self-identification, etc.) via the v2 QuirkBuilder API.
- Does **not** modify any existing device handlers and should be non-breaking.

This allows ZHA to support the FP300 with:
- presence detection
- voltage-based battery handling
- simple and advanced detection-range configuration (switches + numeric mask)

## Additional information

Fixes #4487 

Docs / install guide for using this quirk locally (custom_quirks_path), until it’s merged + released + picked up by Home Assistant:

https://meshstack.de/post/home-assistant/zha-custom-quirks/

## Device diagnostics

[zha-bb791876a15e4fd9b338fdb8a62f7a7c-Aqara lumi.sensor_occupy.agl8-c0ba404a3f7cc423c2650a543896d981.json](https://github.com/user-attachments/files/23634930/zha-bb791876a15e4fd9b338fdb8a62f7a7c-Aqara.lumi.sensor_occupy.agl8-c0ba404a3f7cc423c2650a543896d981.json)

## Checklist

- [x] The changes are tested and work correctly (see linked issue)


faizpuru:
I just tested it. It behaves well. One small remark: unlike the z2m version, the PIR detection state is missing (it triggers before the presence sensor is activated, which is handy for responsiveness).

cemizm:
> I just tested it. It behaves well. One small remark: unlike the z2m version, the PIR detection state is missing (it triggers before the presence sensor is activated, which is handy for responsiveness).

Thanks for the feedback! The PIR detection state is actually present but disabled by default. You should find it under the Disabled entities section of the diagnostics card. Once you enable it, you’ll start receiving the PIR detection state as well.

image

faizpuru:
> > I just tested it. It behaves well. One small remark: unlike the z2m version, the PIR detection state is missing (it triggers before the presence sensor is activated, which is handy for responsiveness).
> 
> Thanks for the feedback! The PIR detection state is actually present but disabled by default. You should find it under the Disabled entities section of the diagnostics card. Once you enable it, you’ll start receiving the PIR detection state as well.
> image

Indeed ! Strangely, I had to restart home assistant to see the disabled motion entity. But now it's working well ! Thank you :)

cemizm:
Hi @TheJulianJES 
I’ve just rebased this PR on the latest dev and all tests are passing.
Please let me know if there’s anything that’s still not quite right or needs further changes. Thanks!

adams79:
Hi, I don't see an option to "start spacial learning" that is available in the zigbee2mqtt integration...

cemizm:
Hi! The option should be there in ZHA as well:

image

If you still can’t find it, please make sure you installed the quirk exactly as described in the guide. In some cases it’s also necessary to remove the device from ZHA and pair it again, so the updated quirk is applied.

https://meshstack.de/post/home-assistant/zha-custom-quirks/

adams79:
Sorry my mistake! Many thanks for this quirk very useful!

PascalWillig:
@cemizm Thanks alot for this!
I seem to have some trouble though and wanted to know if any of you have the same issue.
If I try to change any value in the Config section, I do get the error:
Die Aktion number/set_value konnte nicht ausgeführt werden. Failed to send request: Failed to send request: <Status.NWK_NO_ROUTE: 205>

cemizm:
Hi @PascalWillig, the FP300 is battery powered, so it spends most of its time asleep (“sleepy” Zigbee device). If you try to change a setting while it’s sleeping, ZHA may not be able to deliver the command and you’ll see NWK_NO_ROUTE.

Quick fix: press the button on the FP300 to wake it up, then immediately try setting the parameter again.

PascalWillig:
> Hi @PascalWillig, the FP300 is battery powered, so it spends most of its time asleep (“sleepy” Zigbee device). If you try to change a setting while it’s sleeping, ZHA may not be able to deliver the command and you’ll see NWK_NO_ROUTE.
> 
> Quick fix: press the button on the FP300 to wake it up, then immediately try setting the parameter again.

Thanks a lot!

tomber71:
What's the holdup?

snellejelle99:
> Exposes the 24 × 0.25 m detection slices as six 1 m configuration switches in Home Assistant

@cemizm Why not use a slider like the quirk for the fp1e does?
This looks kinda messy.

Edit: Ah I see this is about the detection zones. Not the normal detection range.
The Z2M page however does also mention a [numeric detection range](https://www.zigbee2mqtt.io/devices/PS-S04D.html#detection-range-numeric).

cemizm:
> Edit: Ah I see this is about the detection zones. Not the normal detection range.
> The Z2M page however does also mention a [numeric detection range](https://www.zigbee2mqtt.io/devices/PS-S04D.html#detection-range-numeric).

Thanks for the suggestion. I did consider using a slider like the FP1E quirk, but in this case the main goal is configuring the detection zones, which are discrete rather than a single continuous range.

https://github.com/zigpy/zha-device-handlers/issues/4487#issuecomment-3543445441

To keep the UI usable without losing too much control, I reduced the granularity from 24 steps of 0.25 m to 6 steps of 1 m. That felt like the best trade-off compared to exposing 24 separate toggles on the device page.

The numeric raw value is still accessible if someone needs full granularity, for example via ZHA Toolkit, but Home Assistant does not really process or visualize that raw value out of the box, so I focused on what works immediately in the standard ZHA/HA UI.

macleodal:
> ## Proposed change
> This PR adds a custom quirk for the **Aqara FP300 presence sensor** (**manufacturer**: `LUMI`, **model**: `lumi.sensor_occupy.agl8`) to `xiaomi`.
> 
> Key points:
> 
> * Implements an `AqaraFP300ManuCluster` based on `XiaomiAqaraE1Cluster` for the manufacturer-specific cluster `0xFCC0`.
> * Adds support for parsing Aqara TLV attribute `0x00F7` so that:
>   
>   * `0xff01-23` is mapped to `BATTERY_VOLTAGE_MV`
>   * `0xff01-24` is mapped to `BATTERY_PERCENTAGE_REMAINING_ATTRIBUTE`
>   * Battery is then exposed via `XiaomiPowerConfigurationPercent`, reusing the existing Xiaomi battery logic.
> * Implements an `FP300DetectionRangeCluster` as a `LocalDataCluster` to handle the **detection range** for the FP300:
>   
>   * Reads and writes the manufacturer attribute `0x019A` (type `octet string`) on cluster `0xFCC0`
>   * Decodes/encodes the 5-byte payload (`prefix` + 24-bit range mask)
>   * Exposes the 24 × 0.25 m detection slices as **six 1 m configuration switches** in Home Assistant
> * Exposes the  FP300 configuration options as entities (e.g. absence delay, motion sensitivity, AI adaptive sensitivity, interference source self-identification, etc.) via the v2 QuirkBuilder API.
> * Does **not** modify any existing device handlers and should be non-breaking.
> 
> This allows ZHA to support the FP300 with:
> 
> * presence detection
> * proper battery reporting
> * configurable detection range in a HA-friendly way (6 config switches instead of 24 individual bits).
> 
> ## Additional information
> Fixes #4487
> 
> Docs / install guide for using this quirk locally (custom_quirks_path), until it’s merged + released + picked up by Home Assistant:
> 
> https://meshstack.de/post/home-assistant/zha-custom-quirks/
> 
> ## Device diagnostics
> [zha-bb791876a15e4fd9b338fdb8a62f7a7c-Aqara lumi.sensor_occupy.agl8-c0ba404a3f7cc423c2650a543896d981.json](https://github.com/user-attachments/files/23634930/zha-bb791876a15e4fd9b338fdb8a62f7a7c-Aqara.lumi.sensor_occupy.agl8-c0ba404a3f7cc423c2650a543896d981.json)
> 
> ## Checklist
> * [x]  The changes are tested and work correctly (see linked issue)
> * [x]  `pre-commit` checks pass / the code has been formatted using Black
> * [x]  Tests have been added to verify that the new code works
> * [x]  Device diagnostics data has been attached



> ## Proposed change
> This PR adds a custom quirk for the **Aqara FP300 presence sensor** (**manufacturer**: `LUMI`, **model**: `lumi.sensor_occupy.agl8`) to `xiaomi`.
> 
> Key points:
> 
> * Implements an `AqaraFP300ManuCluster` based on `XiaomiAqaraE1Cluster` for the manufacturer-specific cluster `0xFCC0`.
> * Adds support for parsing Aqara TLV attribute `0x00F7` so that:
>   
>   * `0xff01-23` is mapped to `BATTERY_VOLTAGE_MV`
>   * `0xff01-24` is mapped to `BATTERY_PERCENTAGE_REMAINING_ATTRIBUTE`
>   * Battery is then exposed via `XiaomiPowerConfigurationPercent`, reusing the existing Xiaomi battery logic.
> * Implements an `FP300DetectionRangeCluster` as a `LocalDataCluster` to handle the **detection range** for the FP300:
>   
>   * Reads and writes the manufacturer attribute `0x019A` (type `octet string`) on cluster `0xFCC0`
>   * Decodes/encodes the 5-byte payload (`prefix` + 24-bit range mask)
>   * Exposes the 24 × 0.25 m detection slices as **six 1 m configuration switches** in Home Assistant
> * Exposes the  FP300 configuration options as entities (e.g. absence delay, motion sensitivity, AI adaptive sensitivity, interference source self-identification, etc.) via the v2 QuirkBuilder API.
> * Does **not** modify any existing device handlers and should be non-breaking.
> 
> This allows ZHA to support the FP300 with:
> 
> * presence detection
> * proper battery reporting
> * configurable detection range in a HA-friendly way (6 config switches instead of 24 individual bits).
> 
> ## Additional information
> Fixes #4487
> 
> Docs / install guide for using this quirk locally (custom_quirks_path), until it’s merged + released + picked up by Home Assistant:
> 
> https://meshstack.de/post/home-assistant/zha-custom-quirks/
> 
> ## Device diagnostics
> [zha-bb791876a15e4fd9b338fdb8a62f7a7c-Aqara lumi.sensor_occupy.agl8-c0ba404a3f7cc423c2650a543896d981.json](https://github.com/user-attachments/files/23634930/zha-bb791876a15e4fd9b338fdb8a62f7a7c-Aqara.lumi.sensor_occupy.agl8-c0ba404a3f7cc423c2650a543896d981.json)
> 
> ## Checklist
> * [x]  The changes are tested and work correctly (see linked issue)
> * [x]  `pre-commit` checks pass / the code has been formatted using Black
> * [x]  Tests have been added to verify that the new code works
> * [x]  Device diagnostics data has been attached

Quick one for anyone trying to install the custom quirk on Home Assistant "out of the box" -- the `/config` directory is not visible in File Editor, the root appears as `/homeassistant/~` but on previous occasions when I've needed to drop a file or folder into` /config/~ `this has proven to be interchangeable -- to be clear you need to specify the path as `/config/custom_zha_quirks` (or whatever path you chose) in `configuration.yaml` even though the file path appears different in the editor.

Chances are most people on here already knew that but I'm new here and still figuring it out so hope that's useful to someone. 

UPDATE: Loaded quirk worked perfectly first time, very sophisticated quirk... nice work.

konstantinrudi:
@cemizm  / @TheJulianJES  when will this be merged? What are we waiting for currently :)?

TheJulianJES:
We recently worked on some major ZHA changes, which took some time. There are some things that still need to be changed in this PR, so hopefully I can leave a review and/or directly make the changes needed to get this merged then.
For now, you can already use this as a custom quirk.
(I've also been trying to get a sensor myself, but was unsuccessful, so far..)

zhaquirks/xiaomi/aqara/motion_agl8.py
```diff
@@ -0,0 +1,833 @@
...
+        endpoint_id=1,
+        device_class=BinarySensorDeviceClass.OCCUPANCY,
+        entity_type=EntityType.STANDARD,
+        reporting_config=ReportingConfig(
+            min_interval=1,
+            max_interval=300,
+            reportable_change=1,
+        ),
+        translation_key="presence",
+        fallback_name="Presence",
```
jeverley: I'd recommend using "occupancy" here to be consistent with the Home Assistant device class, there's a bit more detail on the binary sensors page for use of occupancy/presence:

https://www.home-assistant.io/integrations/binary_sensor/

> - occupancy: on means occupied (detected), off means not occupied (clear)
> - presence: on means home, off means away

I also found discussion covering the use of these classes on the forums:
https://community.home-assistant.io/t/what-is-the-difference-between-show-as-presence-motion-and-occupancy/910588/14

Otherwise, I've been using the quirk without issue for the last few months, great work!

cemizm: Thanks, good point. I’ve switched the translation key to occupancy for consistency with the device class.

cemizm:
> We recently worked on some major ZHA changes, which took some time. There are some things that still need to be changed in this PR, so hopefully I can leave a review and/or directly make the changes needed to get this merged then. For now, you can already use this as a custom quirk. (I've also been trying to get a sensor myself, but was unsuccessful, so far..)

Hi @TheJulianJES, thanks for the update.

I’ve brought this PR in line with the recent ZHA/zigpy changes (attribute event rework + `manufacturer_code` handling), although the FP300 appears to ignore manufacturer-specific codes. The branch is up to date and checks are still passing.

Since this PR has been open for about 4 months and I’ve kept it updated during that time, could you confirm whether this is still intended for mainline, and what concrete blockers remain?

I’m happy to address remaining items if merge is planned. If not, please let me know so I can stop maintaining this branch.

### What changed

- Added a proper `translation_key` for occupancy so entity naming/device-class mapping is consistent.
- Battery handling was switched to voltage-derived percentage (aligned with Z2M behavior):
  - Uses a 2850–3000 mV curve.
  - Ignores the device-reported battery percentage.
- Added manufacturer attribute `0x0203` (`led_disabled_night`) and exposed it as a config switch.
- Detection range was reworked:
  - Added a numeric `detection_range_mask` attribute/entity (`0 .. 0xFFFFFF`) for fine-grained control.
  - Kept the existing 1m switches, but made them operate on the same underlying mask.
  - Raw write flow is now safer: write raw first, update local cluster only on success.
  - Unknown/invalid write attrs are filtered safely.
  - Segment switch writes preserve untouched granular bits (no accidental widening to full 1m blocks).

zhaquirks/xiaomi/aqara/motion_agl8.py
```diff
@@ -0,0 +1,1064 @@
...
+          bridge because runtime updates are split across different event types
+          (written vs. reported/updated). Keeping the raw->derived transform inline
+          here keeps the behavior explicit and avoids hidden listener-coupling complexity.
+        """
+
+        if attrid == self.AttributeDefs.detection_range_raw.id:
+            dr_cluster = self.endpoint.in_clusters.get(
+                FP300DetectionRangeCluster.cluster_id
+            )
+            if dr_cluster is not None:
```
TheJulianJES: As the cluster is added by the quirk, this can never be `None`. The check should be removed.

zhaquirks/xiaomi/aqara/motion_agl8.py
```diff
@@ -0,0 +1,1064 @@
...
+        - Attribute*Event listeners are intentionally not used as the primary
+          bridge because runtime updates are split across different event types
+          (written vs. reported/updated). Keeping the raw->derived transform inline
+          here keeps the behavior explicit and avoids hidden listener-coupling complexity.
+        """
+
+        if attrid == self.AttributeDefs.detection_range_raw.id:
+            dr_cluster = self.endpoint.in_clusters.get(
+                FP300DetectionRangeCluster.cluster_id
+            )
```
TheJulianJES: Also, if `ep_attribute = "fp300_detection_range"` is added to `FP300DetectionRangeCluster`, you can just access that cluster with `self.endpoint.fp300_detection_range` from here. That can also be done for other clusters.

zhaquirks/xiaomi/aqara/motion_agl8.py
```diff
@@ -0,0 +1,1064 @@
...
+            manu = self.endpoint.in_clusters.get(AqaraFP300ManuCluster.cluster_id)
+            if manu is None:
+                return [
+                    [
+                        foundation.WriteAttributesStatusRecord(
+                            foundation.Status.FAILURE, attrid=attr_id
+                        )
+                        for attr_id in resolved_attrs
+                    ]
+                ]
```
TheJulianJES: This can never be `None`. The cluster is added by the quirk.
You can also use an `ep_attribute` to access that cluster via `self.endpoint.<ep-attribute>`

zhaquirks/xiaomi/aqara/motion_agl8.py
```diff
@@ -0,0 +1,1064 @@
...
+            type=MotionSensitivity,
+            zcl_type=DataTypeId.uint8,
+            access="rwp",
+            manufacturer_code=AQARA_MANUFACTURER_CODE,
+        )
+
+        absence_delay_timer: Final = ZCLAttributeDef(
+            id=0x0197,
+            type=t.uint32_t,
+            zcl_type=DataTypeId.uint32,
```
TheJulianJES: If `type` and `zcl_type` match, we don't really need to define the `zcl_type` again.
This is mostly only needed for attributes where we assign an enum `type`, but need to write it as an integer.

zhaquirks/xiaomi/aqara/motion_agl8.py
```diff
@@ -0,0 +1,1064 @@
...
+        This avoids optimistic cache drift when the sleepy-device write times out.
+        """
+
+        resolved_attrs: dict[int, Any] = {}
+        for attr, value in attributes.items():
+            try:
+                attrid = self.find_attribute(attr).id
+            except KeyError:
+                continue
+            resolved_attrs[attrid] = value
```
TheJulianJES: We don't need to catch raised `KeyError`s from `find_attribute`. ZHA will only ever call `write_attributes` for attributes that also have a corresponding definition. If some part of the quirk does this, that part should be fixed.
You can assume that a definition always exists at this point. If not, we should just fail entirely here.

zhaquirks/xiaomi/aqara/motion_agl8.py
```diff
@@ -0,0 +1,1064 @@
...
+            )
+
+            if not self._raw_write_succeeded(raw_result, self._RAW_ATTR_ID):
+                return raw_result
+
+            target_attrs.update(self._segment_attrs_from_mask(new_mask))
+
+            return await super().write_attributes(
+                target_attrs, manufacturer=manufacturer, **kwargs
+            )
```
TheJulianJES: When we call `write_attributes` on the real cluster for `detection_range_raw`, a successful write should also execute `update_attribute` on that cluster, so we can update the local cluster's individual attributes with that. That should call `_update_from_raw` already.

Then we never need to "write to the attributes" on the `LocalDataCluster` (that immediately update the attribute cache) optimistically. We just wait for the real write to re-populate the local attributes.

zhaquirks/xiaomi/aqara/motion_agl8.py
```diff
@@ -0,0 +1,1064 @@
...
+    _RAW_ATTR_ID: Final = AqaraFP300ManuCluster.AttributeDefs.detection_range_raw.id
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions for FP300 detection range cluster."""
+
+        prefix: Final = ZCLAttributeDef(
+            id=0x0000,
+            type=t.uint16_t,
+            zcl_type=DataTypeId.uint16,
+            access="rwp",
```
TheJulianJES: Defining `access` for local attributes doesn't really make sense. We can just drop it entirely. It's purely for information.

TheJulianJES:
This sensor does a lot of things that don't really fit regular HA entities, like the number mask entity for configuring different zones. We likely won't be allowed to officially expose that as a number entity in HA – maybe if we make it `initially_disabled` though? I'll try to check. Otherwise, you'd have to use the `zha.set_zigbee_cluster_attribute` service/action call or write the attribute via the "Manage Zigbee device" debug UI.


I had cleaned up an older variant of this quirk, but wasn't able to test it at the time. I didn't push it here, as this seems to be used as a custom quirk by a lot of people already. I finally got an FP300 now, so I'll try to find some time soon to re-apply the changes and test them. Then see what we're actually allowed to implement officially.

> I’m happy to address remaining items if merge is planned. If not, please let me know so I can stop maintaining this branch.

I have some notes below (and in the review), but you don't need to address them, or make any further changes to this branch if it's too much of a hassle. Just leave the PR open and maintainers can push changes and eventually merge some variant of this quirk to upstream. (and compare it to [your Gist](https://gist.github.com/cemizm/037f02827f5818baa77749876205943c) before)

Some initial notes: (again, no need to address them yourself. We can also push changes to this PR.)
- The `_parse_aqara_attributes` logic relying on internal IDs `0xff01-23` and `0xff01-24` being present is also not good. It should be moved to the main Aqara [`_parse_aqara_attributes`](https://github.com/zigpy/zha-device-handlers/blob/68ac36a1bb79d9333caf2241eb762bf2f150331b/zhaquirks/xiaomi/__init__.py#L379) method, like done in all other Aqara quirks. This will need to be done before merging, but it'll also break all existing custom quirks now, as those are internal IDs used for unknown attributes only.
- The whole write redirection logic with a `_write_mutex` is also something that no other quirks do, and very few HA integrations at all.
- The quirk also relies a lot on zigpy internals. `_raw_write_succeeded` expects a "global success" but this is already handled by zigpy. You get a status record for each attribute.
- Eventually, all the AI-generated docstrings can also be removed or at least shortened.

And thanks again for the PR!

Orange-GT3:
> We likely won't be allowed to officially expose that as a number entity in HA – maybe if we make it `initially_disabled` though? I'll try to check. Otherwise, you'd have to use the `zha.set_zigbee_cluster_attribute` service/action call or write the attribute via the "Manage Zigbee device" debug UI.

These sound like shortcomings of HA. Why should certain functionality of a device be crippled because HA can't match the entities? Surely, HA should be updated to match newer products and whatever entities are needed to support those devices.

It's really frustrating how ZHA gets so little attention from the devs. It is THE out of box experience for Zigbee and yet it plays second fiddle to pointless AI functionality.

puddly:
> Why should certain functionality of a device be crippled because HA can't match the entities?

Home Assistant doesn't have a platform or a UI component to select individual distance bands from 0-6m in steps of 25cm. That just doesn't exist and probably never will unless a bunch of other devices start requiring this type of feature. Exposing a text input box that shows a binary bitstring encoded as base-10 is not a user-friendly entity and not really something Home Assistant should expose.

That being said, you can use the ZHA cluster management UI to set the raw attribute value, same as you do now. Or you can add service/action to do this to ZHA. Or make a custom frontend card to make the above user friendly, like @cemizm did in the linked issue.

Not every feature can be exposed as a Home Assistant entity but that doesn't mean that the feature can't be used from Home Assistant.

---

That being said, you _could_ theoretically turn this bitmap...into 24 on/off switches, one for each distance range 😅.

cemizm:
> This sensor does a lot of things that don't really fit regular HA entities, like the number mask entity for configuring different zones. We likely won't be allowed to officially expose that as a number entity in HA – maybe if we make it `initially_disabled` though? I'll try to check. Otherwise, you'd have to use the `zha.set_zigbee_cluster_attribute` service/action call or write the attribute via the "Manage Zigbee device" debug UI.

I couldn’t really set this via “Manage Zigbee device”.  Reading the byte array worked, but writing it back failed because of payload formatting/validation. Using `zha_toolkit`, the same value could be written without issues.

> I have some notes below (and in the review), but you don't need to address them, or make any further changes to this branch if it's too much of a hassle. Just leave the PR open and maintainers can push changes and eventually merge some variant of this quirk to upstream.

Keeping the branch up to date is not a hassle for me; it would only be wasted effort if there is no real interest in merging this.

I’ll remove the defensive checks, clean up redundant type definitions, and trim/remove the long design-decision docstrings that were mainly meant to provide review context.

Regarding the mutex: I can leave that decision to you. As mentioned earlier, without serialization there is a race window where only the last pending change may win. On a sleepy device that window is not always small. It’s mostly a UX issue, and I’m personally fine with dropping it if that is preferred.

I still consider moving FP300 TLV mapping into the base parser suboptimal from a design perspective (custom quirks already show why), but I can align with your preferred direction.

I hope to get these updates in over the next few days. Since you have write access, feel free to push them directly as well if that is faster.

vvmcmxc:
Thanks for working on this quirk.
I used your current implementation as the starting point while testing on my device.
After some time testing the FP300, I ran a test suite for the current implementation.

Current results:
- 23 tests passing
- 100% coverage for this file

The tests cover device setup, detection range parsing/writes, LED schedule parsing/writes, battery handling, enum handling, bind behavior, and fallback/error cases.

Files:
- Test file: [tests/test_motion_agl8.py](https://github.com/vvmcmxc/zha-device-handlers/blob/FP300/tests/test_motion_agl8.py)
- Test results: [test_results_final.txt](https://github.com/vvmcmxc/zha-device-handlers/blob/FP300/test_results_final.txt)
- Quirk file used for testing: [motion_agl8.py](https://github.com/vvmcmxc/zha-device-handlers/blob/FP300/zhaquirks/xiaomi/aqara/motion_agl8.py)

Hope this is useful.

cemizm:
Quick PR update:

As announced earlier, I have now removed defensive checks, cleaned up redundant type definitions, and shortened/removed the long review-context docstrings. I also addressed the remaining review comments and rebased on current `dev`.

Big thanks to @vvmcmxc !!! ❤️ I integrated his functional updates into this branch.

I moved FP300 tests in a dedicated file (`tests/test_motion_agl8.py`), which was also introduced in @vvmcmxc’s variant. I know this differs from the current “per manufacturer” style, so I’m happy to move them back if that is preferred.

With that, all review points should now be addressed except two from @TheJulianJES:
- `_parse_aqara_attributes` is still used for battery-related TLV parsing.
- `_write_mutex` is still used to serialize parallel updates; @vvmcmxc’s LED schedule cluster also builds on top of this write-redirection path.

Both decisions can still be changed if you want strict alignment before merge.

By the way, I kept the full commit history for traceability, but I can happily squash the commits if preferred (unless that already happens automatically during merge).

zhaquirks/xiaomi/aqara/motion_agl8.py
cemizm: introduced named clusters for direct access.

zhaquirks/xiaomi/aqara/motion_agl8.py
cemizm: Cleaned up all redundant type definitions.

zhaquirks/xiaomi/aqara/motion_agl8.py
cemizm: Now uses a fast-fail path instead of null checks.

zhaquirks/xiaomi/aqara/motion_agl8.py
cemizm: I removed the optimistic local attribute writes and now rely only on the reported detection_range_raw update to repopulate the derived local attributes.

This path is not mutex-protected anymore, but the implementation is definitely leaner.

zhaquirks/xiaomi/aqara/motion_agl8.py
cemizm: Removed  access modifiers from the LocalDataCluster attributes.

zhaquirks/xiaomi/aqara/motion_agl8.py
cemizm: Removed the defensive check and introduced named clusters for direct access.

zhaquirks/xiaomi/aqara/motion_agl8.py
cemizm: Removed the defensive check and introduced named clusters for direct access.

macleodal:
Hi -- I've been using a previous version of the custom quirk, presence detection working well but I hadn't got into any of the more advanced features. I'm also just about hanging by my fingertips on this level of tech so please bear with me...

Just replaced a previous version of the FP300 quirk with the current version in the PR. Restarted HA and the sensor is still working, however I notice the battery is still showing 100% which (since it's been up for a couple of months) makes me think probably something still isn't right. 
- Is there any way for me to know if I am using the most recent version of the quirk with the updated logic for battery assessment? Tbh the thread lost me a bit but I believe the change was going to refocus on battery voltage and I see a reference to that in the .py file so I think I have the right version -- see below. 
- Is there any way for me to verify which quirk the sensor is using, from within ZHA?
- How (other than looking at the battery %) can I test that the quirk is doing everything it should be?
- Might the issue be that I need to reset and re-pair my two sensors? At the moment all I've done is replaced the Python file and restarted.

Thanks again for the work.

Alastair


`"""Quirk for Aqara lumi.sensor_occupy.agl8."""

\# ruff: noqa: D101, D102, D106, D107
import asyncio
from typing import Any, Final

from zigpy import types as t
from zigpy.quirks.v2 import QuirkBuilder, ReportingConfig
from zigpy.quirks.v2.homeassistant import (
    PERCENTAGE,
    EntityType,
    UnitOfElectricPotential,
    UnitOfLength,
    UnitOfTemperature,
    UnitOfTime,
)
from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
from zigpy.quirks.v2.homeassistant.sensor import SensorDeviceClass, SensorStateClass
from zigpy.typing import UNDEFINED, UndefinedType
from zigpy.zcl.foundation import BaseAttributeDefs, DataTypeId, ZCLAttributeDef

from zhaquirks import LocalDataCluster
from zhaquirks.xiaomi import (
    BATTERY_PERCENTAGE_REMAINING_ATTRIBUTE,
    BATTERY_VOLTAGE_MV,
    XiaomiAqaraE1Cluster,
    XiaomiPowerConfiguration,
)

\# Manufacturer-specific attribute keys present in the non-standard AQARA payloads
AQARA_MANUFACTURER_CODE: Final = 0x115F
MANU_ATTR_BATTERY_VOLTAGE: Final = "0xff01-23"
MANU_ATTR_BATTERY_PERCENT: Final = "0xff01-24"  # unused, keep for future
`

tomber71:
When will this be merged?

00lex:
I was also skeptical about the battery, so I decided to check it. Both were at 2.9V after 5 months. (This sensor was tested extensively during setup and is running in Zigbee mod)
anyway Home Assistant still shows 100% and a voltage of 30V (!!!)


EDIT: ok forget it! after installing latest quirk I get 79% battery and 3V voltage :)

jayrage887:
The only thing weird in this quirk is the temperature displayed, it seems to force it to exactly one degree steps (Celsius), any fix for that, that i am not seeing?

jayrage887:
I figured out that the temperature/humidity readings have currently issues and that this is already acknowledged by aqara: https://forum.aqara.com/t/fp300-zigbee-firmware-sends-duplicate-temperature-humidity-reports/203636/2

However, if someone wants a quick, dirty fix, add those classes and adapt the quirk builder
```
class FP300TemperatureCluster(CustomCluster, TemperatureMeasurement):
    """Ignore rounded duplicate temperature reports from FP300 firmware."""

    def _update_attribute(self, attrid: int, value: Any) -> None:
        if attrid == self.AttributeDefs.measured_value.id and value is not None:
            current = self.get(self.AttributeDefs.measured_value.id)
            if (
                current is not None
                and value % 100 == 0
                and current % 100 != 0
                and abs(value - current) <= 100
            ):
                return


class FP300HumidityCluster(CustomCluster, RelativeHumidity):
    """Ignore rounded duplicate humidity reports from FP300 firmware."""

    def _update_attribute(self, attrid: int, value: Any) -> None:
        if attrid == self.AttributeDefs.measured_value.id and value is not None:
            current = self.get(self.AttributeDefs.measured_value.id)
            if (
                current is not None
                and value % 100 == 0
                and current % 100 != 0
                and abs(value - current) <= 100
            ):
                return
        super()._update_attribute(attrid, value)```

.replaces(FP300TemperatureCluster)
.replaces(FP300HumidityCluster)

**Comment by cemizm** [comment]
What configuration are you using for the reporting thresholds, interval, and so on? I'm using the stock configuration and haven't had any issues with temperature reporting.

image

**Comment by jayrage887** [comment]
@cemizm i used it out of the box and noticed the same issues as described here: 
https://forum.aqara.com/t/fp300-zigbee-firmware-sends-duplicate-temperature-humidity-reports/203636/2

what settings do you use? did you change something? after having the issues i could not find a change that got rid of the behaviour, i tried intervall only, intervall and threshold as well as threshold only, additionally i played around with the intervall times etc. 

the only thing that helped was changing the quirks code

**Comment by jayrage887** [comment]
before the fix, the same issue:

Screenshot_20260421_173710_Home Assistant

compared to now:

Screenshot_20260421_183216_Home Assistant

**Comment by cemizm** [comment]
I'm using the following settings and I did not change anything in the quirks code:

- Temperature and humidity sampling: Low
- Temperature and humidity sampling period: 600 s
- Temperature reporting interval: 3600 s
- Temperature reporting mode: Threshold and interval
- Temperature reporting threshold: 0.5 °C

So in my case it's basically the stock/default configuration, and with these values I have not seen any issues with temperature reporting.

It may of course be the case that changing those values is what actually triggers the bug in the FP300. You could try switching the firmware to Thread in the Aqara app and then back to Zigbee, hoping that this resets the internal configuration.

**Comment by jayrage887** [comment]
Meh, it works now as intended, I'll just wait until aqara fixes it with a fw update, they acknowledged the issue after all, and for now the sensor just works, the quirk is amazing ;), i just wanted to supply a fix that worked for me for anyone installing this quirk and running into the same issue (as i said, it is not quirk related)

**Comment by Napetc** [comment]
When will target_distance X/Y coordinates be implemented to draw them on top of the room plan?

**Comment by macleodal** [comment]
I've been digging into the battery issue that I mentioned in my earlier comment (March 27). Still not working correctly for me (both FP300 sensors still show 100% all the time).

After pressing the physical button to wake the device and checking the raw state attributes in HA's developer tools, I can see:
`battery_voltage: 0.3
`

Working backwards through the quirk code — `battery_reported` stores `round(voltage_mv / 1000, 3)` as the display value — a stored value of 0.3 implies the raw incoming voltage from the TLV is 300, not 3000. That's off by a factor of 10, which would explain why the percentage calculation always resolves to 100% (300mV is below the MIN_VOLTS_MV floor of 2800).

So it looks like the TLV battery voltage attribute is being parsed and arriving at `battery_reported` as 300 rather than 3000mV. Possibly a units issue in the TLV parsing upstream?

Device: Den Aqara lumi.sensor_occupy.agl8, Firmware 0x00003a29

**Comment by sezissu** [comment]
I am excited to see this quirk added to zha fully, but Is there motion on merging this in?  The longer it lags behind head, the more of a chance something will silently break.

**Comment by 00lex** [comment]
> * Temperature reporting mode: Threshold and interval

BTW/OT: as i understand the z2m device page all settings are irrelevant as long as you don't use _custom_ sampling. I am team low too :)

**Comment by jayrage887** [comment]
> > * Temperature reporting mode: Threshold and interval
> 
> BTW/OT: as i understand the z2m device page all settings are irrelevant as long as you don't use _custom_ sampling. I am team low too :)

Well Aqara literally just released a FW update that, among other issues, fixes and achnowledges exactly that issue, so just because it works for you, it was an error and now after the update, it works flawlessly for me.

**Comment by tomber71** [comment]
When will this be merged? Why is this taking so long? Can anyone please explain?

**Comment by rich5741** [comment]
Reported a battery issue different from above here: https://github.com/zigpy/zha-device-handlers/issues/5116.  Battery jumps from zero to 100 and back again

**Comment by macleodal** [comment]
Have to admit I'm lost as to what's working, what's not, and for what's not what is still being worked on and what is parked as not gonna happen. I don't have the skills myself so I am just appreciative that people are trying but it's a nause to keep removing, resetting, updating and repairing my FP300's every time there's a new HA core so for now I'm accepting them as underperforming slightly disappointing devices with no functioning battery sensor and decided I'll try again in three months and see if the quirk is working by then.

On Tue Jun 23, 2026, 01:49 PM GMT, rich5741 ***@***.***> wrote:
[image]rich5741 left a comment (zigpy/zha-device-handlers#4504) <https://github.com/zigpy/zha-device-handlers/pull/4504#issuecomment-4779919547>

**Comment by rich5741** [comment]
@cemizm is it possible to exposure the mmwave sensor on its own (much the same way the PIR sensor is seen on its own)?

Also suggest adding a "version number" to your quirk, to track what is the latest until HA mergers this.

**Comment by frkd-dev** [comment]
The code in this PR wasn't updated since March 2026 (except merging dev branch) but others published quirks with improvements for this code, like in #4487. Those improved quirks bring reliability to the sensor, and I'm using them too as they did resolve the issue with sensors being stuck in the occupied state for hours.

I also saw battery reading issues (jumps between 85-100%), which I don't observe since I started using the quirk with improvements and stopped experimenting with sensors. Readings just became stable at 100% (my sensors are just two months old).

Could we bring those improvements into this PR?

**Comment by demianvann** [comment]
Wanted to add some diagnostic detail to the existing "stuck in occupied state for hours" and battery-instability reports on this thread, in case it's useful for isolating the cause.

**Setup:** Aqara FP300 (`lumi.sensor_occupy.agl8`, firmware `0x00003a29`) on Home Assistant 2026.2.3 / ZHA, using this quirk pulled in from the linked gist (same implementation as this PR — `_parse_aqara_attributes`, `_write_mutex` write-redirection, etc.).

**Occupancy attribute goes completely silent, not just delayed or stuck detecting**

I initially assumed "stuck occupied" meant the device kept (falsely) detecting motion. Digging into Home Assistant's recorder with duplicate-value reports included (not just state *changes*), that's not what's happening — the occupancy attribute stops sending reports of *any* kind, including repeated "still occupied" confirmations, for hours at a time. Two examples:

- 2026-07-20: one "occupied" report at 07:56:48, then **zero further reports on that attribute for over 8 hours** — no clear when the room was actually vacated, no re-affirm, nothing — until I pressed the restart-device write attribute, after which fresh reports resumed ~35 min later.
- 2026-07-21: same pattern, occupancy held at its last reported value from before 15:00 until a single fresh "clear" report at 19:13:18. Cross-checked against `person_home` tracking for the room's occupant, who was away from before 15:00 until 19:30 — confirming the room really was empty and the device just wasn't reporting, not a live false-positive.

Meanwhile the battery attribute kept reporting fine on its own cadence throughout both windows (including the known battery-percentage-jumps-to-0-then-recovers glitch also mentioned upthread) — so the device isn't dropping off the Zigbee network entirely, it's specifically the occupancy attribute's reporting that stalls.

**Re: @macleodal's battery_voltage off-by-10 report (Apr 26)** — I'm seeing the same thing on the current quirk revision. My device periodically reports `battery_voltage: 0.3` (should be ~3.0) right before the percentage drops to 0% and then self-corrects back to ~100% a few minutes later, on a roughly 3-hour cadence. So the March 22 voltage-derived-percentage fix doesn't seem to have fully resolved this for every device — mine hits it repeatedly, not just as a one-off.

**Working (temporary) remediation:** writing to the `restart_device` attribute (`0x00E8`) reliably gets occupancy reporting flowing again — confirmed by watching for a fresh state transition within ~30-40 min of the write. It's not durable though — the 07-21 case recurred less than 24 hours after a restart resolved the 07-20 case.

Happy to pull more diagnostics (raw Zigbee traffic, binding table state, etc.) if it'd help narrow down whether this is a reporting-config/binding issue vs. something in the device's own firmware. Thanks for the work on this quirk either way — it's been very usable day-to-day apart from this.

**Comment by macleodal** [comment]
@demianvann thanks for picking up on my point.

More broadly (and I feel I represent the less technical would-be users of the FP300 on Git / this thread) -- am I the only one who is feeling both a bit lost in all the back and forth, and a little disappointed in my FP300s after a promising start? I've got three deployed, I want to put them everywhere (design / size / functionality wise it's the perfect device if it worked) but I am obviously on pause currently.

With huge ongoing thanks and appreciation to everyone developing the quirk(s) - I do not have the skills even with AI in my hands - I have two questions if someone has a moment to respond please:
1. Currently, at time of writing, is there just _one_ or _more than one_ custom quirk which is usable and testable and somewhat functional? I don't know if we're down to a single thread here or where there are multiple threads / multiple forks / multiple custom quirks with different characteristics and performance on the problematic FP300 features. I am fine using custom quirks but I am struggling to know which one(s) or where it / they are now to be found.
2. I saw discussion about the PR being merged so I assume now, or soon, a version of the quirk will turn up in ZHA. But so far I see nothing on Blakaddr which is the closest to a canonical reference I guess, and since it's not clear if it's merged / when it will merge, I am between a rock and a hard place -- do I delete my custom quirk, restart, remove and repair all my FP300s in the hope that when I re-add them there's an official quirk inside ZHA waiting for me? 

Accepting this is the most basic question on Git this morning but if anyone can fill in the blanks for me I'd appreciate it. And thanks again to the devs, I am fully aware that without you guys we'd have nothing -- with hindsight wishing I'd gone Z2M rather than ZHA but way way way too late for that now! (I just don't hate myself enough to remove and re-pair my entire Zigbee network!).

Thanks all.

Alastair

**Comment by Orange-GT3** [comment]
Slightly related, ZHA is currently undergoing a fairly extensive overhaul in the quirk department. Several PRs have been merged and more work is ongoing. The end result is that it should be easier and faster for new products to be supported in ZHA.

Note that the devs have said that support for some products in Z2M is very 'hacky'. That approach wouldn't cut the mustard in ZHA because ZHA is the OOBE for new HA users. ZHA is also moving out of core and into a standalone app.

I'm not a dev, so please don't ask me any detailed questions. I just subscribe to certain issues/PRs/discussions on Github to keep tabs on what is happening.

**Comment by wiltodelta** [comment]
**Battery 0% to 100% toggling: measurements and a local fix**

I hit the same 0% to 100% alternation as #5116 on a single FP300, every report cycle (HA 2026.8.2, quirk from this PR). Reading the clusters directly over the ZHA WebSocket (`zha/devices/clusters/attributes/value`) shows two live writers fighting:

| Source | Attribute | Raw value |
|---|---|---|
| Standard PowerConfiguration (0x0001) | 0x0020 battery_voltage | 3 |
| Standard PowerConfiguration (0x0001) | 0x0021 battery_percentage_remaining | 200 (100%) |
| Aqara TLV 0x00F7, tag 0x17 | battery voltage | 320 |

Two defects in the current battery path:

1. Tag 0x17 is not millivolts, at least on my unit. 320 mV is impossible for a working radio. Read as units of 10 mV it is 3.2 V, consistent with a fresh cell and with the device's own 100% report. Fed to `battery_reported()` as mV, it clamps to `MIN_VOLTS_MV` and derives 0%. That is where the zeros come from.

2. `battery_reported()` writes `round(voltage_mv / 1000, 3)` into `BATTERY_VOLTAGE_ATTR` (0x0020), but that attribute is defined in units of 100 mV. The parent `XiaomiPowerConfiguration` writes `round(voltage_mv / 100, 1)`. The result is a voltage attribute showing 0.32 V on a full battery.

Together with the standard reports @nokside described in #5116, the entity flips between the device's 100% and the derived 0% every cycle.

This local variant has kept the entity stable since deployed today (rescales units of 10 mV, sanity checks the window, and drops the device's own standard battery reports):

```python
class FP300PowerConfigurationVoltage(XiaomiPowerConfiguration):
    """Battery level based on voltage."""

    MIN_VOLTS_MV = 2800
    MAX_VOLTS_MV = 3000

    def __init__(self, *args, **kwargs) -> None:
        """Init."""
        self._quirk_battery_update = False

    def _update_attribute(self, attrid: int, value: Any) -> None:
        """Drop the device's own contradictory battery reports.

        The FP300 pushes a stuck-at-full percentage through the standard power
        cluster while the TLV voltage needs rescaling, so only values this
        quirk derives from the TLV voltage may enter the attribute cache.
        """
        if (
            not self._quirk_battery_update
            and attrid
            in (self.BATTERY_VOLTAGE_ATTR, self.BATTERY_PERCENTAGE_REMAINING)
        ):
            return

    def battery_reported(self, voltage_mv: int) -> None:
        """Update voltage and derived battery percentage from an mV report."""
        if voltage_mv < 1000:
            voltage_mv *= 10
        if not 2000 <= voltage_mv <= 4000:
            return
        self._quirk_battery_update = True
        try:
            self._update_attribute(
                self.BATTERY_VOLTAGE_ATTR, round(voltage_mv / 100, 1)
            )
            self._update_battery_percentage(voltage_mv)
        finally:
            self._quirk_battery_update = False

    def battery_percent_reported(self, battery_percent: int) -> None:
        """Ignore buggy percentage reports from device."""
        pass
```

Caveat: one unit. If other units really send tag 0x17 in millivolts (2950-style values), the `voltage_mv < 1000` guard leaves them untouched, so the class is safe for both scales.

@cemizm happy to push this as a commit to the PR branch if you want it.

wiltodelta:
Opened the fix as a ready-to-merge PR against the PR branch, with tests updated to the corrected expectations: cemizm/zha-device-handlers#1

wiltodelta:
Since this branch has been idle since March and #4487 is still open, I opened a takeover PR to dev with the quirk as-is plus the battery fix commit from my comment above: #5259. @cemizm all quirk code is yours and credited as such in the commit history. If you would rather resume this PR, say the word and I close mine.

rich5741:
First, thanks @cemizm for this quirk — it's been running my four FP300s for weeks and is the only reason the device is usable on ZHA at all. And thanks @TheJulianJES for the ongoing review here, hopefully we will have an official ZHA quirk soon.

Did some troubleshooting with Claude - I had a few issues where device would drop or stick occupancy.  This seems to have fixed it.  My 0 to 100 swing was fixed by a fresh firmware pull on the FP 300 (reset, then switch to thread, switch back to zigbee) and a from scratch re-pair.  I noticed that this particular device was a bit sticky - removing it from ZHA many time did not actually remove it at all, after resetting the device it would automatically re-pair to ZHA, defeating the whole purpose of removing and repairing it.  Best method I found was to make sure it was awake with a few button presses while removing it from ZHA from within HA.

**Claude output:**

I've been running these on firmware 6542 and have found a reproducible defect plus a fix I've now validated across all four sensors. Posting here rather than opening a separate issue because the fix belongs in this PR — #5259 is already correctly stacked on top of this one for battery, but neither PR covers what follows.

**Environment:** Aqara FP300 (`lumi.sensor_occupy.agl8`), manufacturer code 4447, firmware `0x0000412a` (Aqara Zigbee build 6542, May 2026 — the newest available). Home Assistant 2026.9.0, HAOS, ZHA, 4 units. All were first installed *after* the May firmware, so the pre-6542 behaviour was never present here.

---

## Bug: presence/PIR stop reporting entirely

`presence` and `pir_detection` silently stop reporting. Captured 84 minutes with zero reports on either attribute while the device was demonstrably online (LQI 255, RSSI −17):

| Entity | State | Last report |
|---|---|---|
| Humidity | 41.16 | 20:31 |
| Temperature | 76.0 | 20:12 |
| Illuminance | 184 | 20:11 |
| Occupancy | on | 19:46:38 |
| PIR | on | 19:46:35 |

`last_reported == last_changed` on both — nothing arrived at all, not "reports saying still-on". Only cluster 0xFCC0 goes silent; the standard clusters keep flowing throughout.

**Steps to reproduce:** pair an FP300 on firmware 6542 with the current quirk, set `absence_delay_timer = 10s`, leave the room, watch occupancy. Reproduced on 3 of 4 sensors the same evening.

**Expected:** occupancy clears ~10 s after the room empties.

**Actual:** holds `on` for 40–84 minutes with no reports received.

Neither timer explains it — `absence_delay_timer=10s`, `pir_detection_interval=30s`. And the quirk declares `ReportingConfig(min_interval=0, max_interval=900, reportable_change=1)` on both attributes, which mandates a report every 15 minutes regardless of change. 84 minutes of silence means that configuration never reached the device.

## Root cause

`git log` on `zhaquirks/xiaomi/aqara/motion_agl8.py` shows **zero `configure_reporting` calls across all ten revisions**. `bind()` reads only `detection_range_raw` and `led_schedule_time_raw`. The quirk relies entirely on quirks-v2 metadata, which doesn't appear to reach this device — plausibly because it's a sleepy end device rarely awake when ZHA attempts it.

## Fix and validation

@0SkillAllLuck posted a variant on #4487 (timeline page 2) whose only functional difference is an explicit `configure_reporting_multiple()` for `presence` and `pir_detection` in `bind()`. Worth noting the "force uint8" element of that variant is already present here — both set `zcl_type=DataTypeId.uint8, access="rp"`, so the explicit call is the entire delta.

Confirmed on the wire:

```
Configure_Reporting(attrid=0x0142, min_interval=0, max_interval=300,
    reportable_change=1)  ->  Status.SUCCESS
Configure_Reporting(attrid=0x014D, min_interval=0, max_interval=300,
    reportable_change=1)  ->  Status.SUCCESS
```

After a successful bind, ~14 h of history showed clean cycles (Living Room 37 transitions, Office 24) with clears as short as 15 s and no multi-hour holds. Every longer hold was verified as genuine occupancy. Now ~12 days across all four sensors with no recurrence.

A fourth sensor whose bind failed became an unplanned control: it went unavailable exactly 3 hours after each transmission — ZHA's battery-device timeout — and recovered spontaneously when someone entered the room. The sensors with working reporting never went unavailable.

**Cross-stack precedent:** Zigbee2MQTT hit the same firmware-driven loss of autonomous reporting (zigbee2mqtt#32153) and resolved it by explicitly configuring reporting (zigbee-herdsman-converters#12383, merged 2026-06-09).

---

## Second finding: the Xiaomi lifeline parse is unreachable from the read path

Battery is populated only by the Aqara lifeline TLV (0x00F7). `battery_voltage` metadata has `reporting_config=None` and `bind()` never reads 0x00F7, so once the synthesized value is cleared (e.g. an HA restart) it's stranded — a button press and a Reconfigure both fail to restore it; only a full re-join does.

Adding a read shows the device answers fine:

```
Read_Attributes(attribute_ids=[247])
ReadAttributesResponse(attrid=247, status=Status.SUCCESS,
    value=LVBytes b'\x05!\x13\x00\n!\xb2\x1e\x0c \n\r#*A\x00\x00\x13 \x00\x17!\x9d\x0b\x18 d\x1c\x10\x00d \x00e \x02g \x00')
```

Decoded by tag: `0x0D` = 0x0000412a (firmware), `0x17` = 2973 (battery mV), `0x18` = 100 (device-reported percent).

Two gotchas:

**1. The lifeline must be declared on the cluster.** `AqaraFP300ManuCluster` overrides `XiaomiAqaraE1Cluster`'s `AttributeDefs`, which is empty, so 0x00F7 is unresolvable and `read_attributes` raises `KeyError(247)` before transmitting.

**2. Even when read, the TLV is discarded.** `XiaomiCluster` binds its parser to an event:

```python
self.on_event(AttributeUpdatedEvent.event_type, self._handle_attribute_event)
```

That fires only for *unsolicited* attribute reports. A `read_attributes()` response fills the attribute cache without emitting it, so the TLV is never parsed — and battery dispatch (`self.endpoint.power.battery_reported`) lives inside that handler. Observed exactly: the read returned SUCCESS with the payload above, no "Xiaomi attribute report" was logged, and battery stayed unknown. This looks like a general `zhaquirks/xiaomi` gap rather than an FP300-specific one — any Xiaomi lifeline read would hit it.

Working fix — declare the attribute, then read/parse/dispatch off the bind path:

```python
# in AqaraFP300ManuCluster.AttributeDefs
aqara_lifeline: Final = ZCLAttributeDef(
    id=0x00F7, type=t.LVBytes, access="rp",
    manufacturer_code=AQARA_MANUFACTURER_CODE,
)

# scheduled from bind() via asyncio.create_task(), NOT awaited — see caveat
async def _read_lifeline_battery(self, delay: float = 5.0) -> None:
    await asyncio.sleep(delay)
    try:
        result = await self.read_attributes(
            [self.AttributeDefs.aqara_lifeline.id],
            allow_cache=False, manufacturer=AQARA_MANUFACTURER_CODE,
        )
        raw = result[0].get(self.AttributeDefs.aqara_lifeline.id)
        if raw is None:
            return
        parsed = self._parse_aqara_attributes(raw)
        if BATTERY_VOLTAGE_MV in parsed:
            self.endpoint.power.battery_reported(parsed[BATTERY_VOLTAGE_MV])
    except Exception as exc:
        self.debug("Failed to read FP300 lifeline (battery): %r", exc)
```

Result: `battery_voltage_mV=2973` → battery entity 82.0 (2973 mV on the 2850–3000 curve). Verified on all four sensors.

**Caveat worth heeding if this is adopted:** I first awaited this read *inside* `bind()`. That lengthened an already-fragile operation on a sleepy device and correlated with partial reconfigure failures and device instability. Scheduling it as a background task avoids that — it should not extend the bind critical path.

---

## Notes for reproduction

Reconfigure on this device is unreliable — 6 attempts failed on one sensor and 2 on another before succeeding, logging:

```
[zigpy.device] Re-interview failed, keeping existing device
zigpy.exceptions.DeliveryError: Failed to deliver message:
    <sl_Status.ZIGBEE_DELIVERY_FAILED: 3074>
```

The ZHA dialog reports "successful" regardless; the absence of per-cluster green checkmarks is the real signal. Generating motion is **not** enough to hold this sleepy device awake — a physical button press repeated every 10–15 s throughout the operation is required.

Separately, a partially-failed reconfigure left devices in a **stuck-initializing** state that survived HA restarts and caused periodic unavailability, with all affected devices dropping within 1–3 seconds of each other on a shared retry timer. ZHA's add-device screen listed devices that were neither reset nor in pairing mode, which turned out to be the clearest diagnostic. Clearing it required a full reset: remove from HA → 10 short button presses → switch the device to Thread → switch back to Zigbee → re-pair. A plain factory reset was **not** sufficient; the device silently rejoined the network before the wipe completed, several times over.

Anonymized diagnostics for two units are attached to #5116. Happy to provide history exports, or to open a PR with tests for the lifeline fix if that would help.

PR #4510, opened 2025-11-20
Add HOBEIAN ZG-303Z

nschimme:
## Proposed change
This allows soil moisture and humidity to both been seen. Previously, only temperature, soil moisture, and battery were shown. Even worse, soil moisture was incorrectly displayed as humidity due to it being reported under cluster 0x405.

## Additional information
Other attempts to add support via the `TuyaQuirkBuilder` did not work because of how soil moisture would overwrite the humidity since they all shared the same endpoint. This quirk now uses endpoints 2 and 3 for Soil Moisture (DP 3) and Relative Humidity (DP 109), respectively. This prevents values from being overwritten when they are reported simultaneously by the device, ensuring stable readings for both attributes. We also drop the 0x405 cluster that was incorrectly reporting soil as humidity.

image

## Device diagnostics
[zha-01K8BQMJSKZFPA3YC43C2Q4TB0-HOBEIAN ZG-303Z-8ec34d4e404ba4a5f4ec9dcf4bc9e3c8-2.json](https://github.com/user-attachments/files/23664865/zha-01K8BQMJSKZFPA3YC43C2Q4TB0-HOBEIAN.ZG-303Z-8ec34d4e404ba4a5f4ec9dcf4bc9e3c8-2.json)

## Checklist

### DIFF
--- a/tests/test_tuya_soil.py
+++ b/tests/test_tuya_soil.py
@@ -1,0 +1,134 @@
+"""Test for HOBEIAN ZG-303Z soil sensor."""
+
+from unittest.mock import MagicMock
+
+import pytest
+from zigpy.zcl.clusters.general import PowerConfiguration
+from zigpy.zcl.clusters.measurement import (
+    RelativeHumidity,
+    SoilMoisture,
+    TemperatureMeasurement,
+)
+
+from zhaquirks.tuya import TuyaData, TuyaDatapointData, TuyaDPType
+from zhaquirks.tuya.ts0601_soil import (
+    HobeianMcuCluster,
+    HobeianRelativeHumidity,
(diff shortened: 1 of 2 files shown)


nschimme:
@TheJulianJES this is ready to go. I've looked at what is needed to enhance the V2 API to do the same thing and it doesn't look trivial. I suggest we do it in a second pass.

mjbogusz:
So far (few hours) this works way better than previous ones, though it would be nice to re-add the low moisture warning and configuration/calibration endpoints - it could be a followup though to get this merged sooner.

nschimme:
Closing in preference for #4536 which uses the V2 API

PR #4520, opened 2025-11-25
Add Tuya ZT08 weather station (_TZE284_hodyryli)

### LINKED ISSUE #4489: [Device Support Request] Tupa temp/humidity sensor with external temp probe TZE284_hodyryli
### Problem description

I ordered the following:  https://www.aliexpress.com/item/3256810055014641.html?spm=a2g0o.order_list.order_list_main.5.55b51802JqiQvK

It is not currently supported nor can I find any quirk for it.

### Solution description

likely needs a quirk.  I did get it to pair/detect.

### Screenshots/Video

</details>

### Diagnostics information

[zha-01JES93Y4NY89NXK3P04DYJNK0-_TZE284_hodyryli TS0601-0a2f9ff57075b050d6b4c6364ca08d3c.json](https://github.com/user-attachments/files/23537646/zha-01JES93Y4NY89NXK3P04DYJNK0-_TZE284_hodyryli.TS0601-0a2f9ff57075b050d6b4c6364ca08d3c.json)

### Device signature

```json
{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4417,
    "maximum_buffer_size": 66,
    "server_mask": 10752,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xed00",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE284_hodyryli",
  "model": "TS0601",
}

```

</details>

### Logs

```python

[Paste the logs here]

```

</details>

### Custom quirk

```python

```

</details>

### Additional information

_No response_

DesertRatUa:
## Checklist

Closes #4489


DesertRatUa:
https://github.com/zigpy/zha-device-handlers/issues/4489

seanmahrt:
It looks like it's sort of working, but it seems the temp probe scaling is off.  It may be a factor of 2 but I cannot tell for sure.  I'm going to put it in a different location and check.

[zha-01JES93Y4NY89NXK3P04DYJNK0-_TZE284_hodyryli TS0601-0a2f9ff57075b050d6b4c6364ca08d3c (1).json](https://github.com/user-attachments/files/23837567/zha-01JES93Y4NY89NXK3P04DYJNK0-_TZE284_hodyryli.TS0601-0a2f9ff57075b050d6b4c6364ca08d3c.1.json)

seanmahrt:
I put it out in the garage, and the two temperature lines still have the same shape, but it's not off by a factor of two.  next step is put the probe in the freezer to confirm it's reading two different temperatures.

DesertRatUa:
On my device, probe temperature and general temperature are different:
image
And device was able to receive the actual time

seanmahrt:
> On my device, probe temperature and general temperature are different: image And device was able to receive the actual time

yeah, time worked OK.   will know better here shortly with the two probes.

Would there be anything weird that I'm in Fahrenheit on the ZigBee device/home assistant?   I would assume under the hood everything would be metric.

image

current temps on device are 14.4F and 52.1F

seanmahrt:
found this:  https://github.com/Koenkk/zigbee2mqtt/issues/29627

looks like maybe the external probe temp needs some massaging?  

                if (dpValue.datatype === 2) { // VALUE type (4-byte integer)
                    if (dp === 38) {
                        // For external temperature, use signed 16-bit from last 2 bytes
                        // This supports both positive and negative temperatures
                        const uint16 = data.readUInt16BE(data.length - 2);
                        value = uint16 > 32767 ? uint16 - 65536 : uint16;

seanmahrt:
looking more into that other PR, there was some general tidbits that seems less specific to internal/external probe:

              let value;
                // temperature
                if (dpValue.datatype === 2) {
                    // Reads the data as an unsigned big-endian 32-bit integer (Negative numbers will be wrong, eg. 429496724.6)
                    // value = data.readUInt32BE(0);
                    // Reads the data as a signed big-endian 32-bit integer (Negative numbers will be correct)
                    value = data.readInt32BE(0);
                } else if (dpValue.datatype === 4) {
                    value = data[0];
                } else {
                    value = data[0];
                }
                
                switch (dp) {
                    // temperature
                    case 1:
                        result.temperature = value / 10;
                        break;
                    // humidity
                    case 2:
                        result.humidity = value;
                        break;
                    // battery_state
                    case 3:
                        result.battery_state = value;
                        result.battery = value === 0 ? 10 : (value === 1 ? 50 : 100);
                        result.battery_low = value === 0;
                        break;
                    // time_format
                    case 17:
                        result.time_format = value ? '12h' : '24h';
                        break;
                    // temperature_external
                    case 38:
                        result.temperature_external = value / 10;
                        break;

seanmahrt:
reading through your code, maybe you're already attempting to read it as 32bit signed.  do we need to confirm big endian? ('m a python/quirk noobie)

seanmahrt:
temps have stabilized:  
image

on the display:  8.6F for the probe, 52.1 for the sensor, 57% humidity

seanmahrt:
SOLVED:  sorry about the long posts.  change the /100 -> /10 for the external sensor and it looks good.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +355,34 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        attribute_name="battery_percentage_remaining",
+        converter=lambda x: {0: 10, 1: 50, 2: 100}[x],
+    )
+    .adds(TuyaPowerConfigurationCluster2AAA)
+    # External temperature sensor (DP 38, value needs to be divided by 100 to get correct temperature)
+    .tuya_sensor(
+        dp_id=38,
+        attribute_name="temperature_external",
+        type=t.int32s,
+        converter=lambda x: x / 100,  # Divide by 100 to get correct temperature
```
seanmahrt: if you set this to /10 it seems to work

DesertRatUa: For my device /10 is provided a temperature of 510.0 °C
What is very strange, cause in ZB2MQTT is also /10.
And the only visible reason is the Chinese random in the device firmware.

DesertRatUa:
> > On my device, probe temperature and general temperature are different: image And device was able to receive the actual time
> 
> yeah, time worked OK. will know better here shortly with the two probes.
> 
> Would there be anything weird that I'm in Fahrenheit on the ZigBee device/home assistant? I would assume under the hood everything would be metric.
> image
> 
> current temps on device are 14.4F and 52.1F

I was trying to switch my device to F, and got the same values. So looks like, under the hood, it's still metric and uses UI conversion.

DesertRatUa:
> result.time_format = value ? '12h' : '24h';

I'm also trying to get DP17 for the time format, but my device never sends it.

zhaquirks/tuya/tuya_sensor.py
seanmahrt: > For my device /10 is provided a temperature of 510.0 °C What is very strange, cause in ZB2MQTT is also /10. And the only visible reason is the Chinese random in the device firmware.

:)

DesertRatUa:
After reboot (change of batteries), my device started sending probe temperature in the same way as @seanmahrt, so I changed multiplayer to 10.

Bjoern3D:
Is the device now **officially** supported? Unfortunately, I couldn't find anything about it in the changelogs. Thank you in advance.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +355,34 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+    # Internal temperature (DP 1, scale=10)
+    .tuya_temperature(dp_id=1, scale=10)
+    # Internal humidity (DP 2)
+    .tuya_humidity(dp_id=2)
+    # Battery from battery_state (DP 3: 0=10%, 1=50%, 2=100%)
+    .tuya_dp(
+        dp_id=3,
+        ep_attribute=TuyaPowerConfigurationCluster2AAA.ep_attribute,
+        attribute_name="battery_percentage_remaining",
+        converter=lambda x: {0: 10, 1: 50, 2: 100}[x],
```
Aggrathon, line 371: Thank you for creating a quirk for this device. However, I think the `battery_percentage_remaining` scale should be doubled (just like other Tuya devices in this file), i.e. `lambda x: {0: 20, 1: 100, 2: 200}` instead of `lambda x: {0: 10, 1: 50, 2: 100}`. That is, at least, how my devices need to be configured.

lenucksi:
Also pointing to this issue here. It mentions this as working fix: https://github.com/zigpy/zha-device-handlers/issues/4766 
What's the state on getting this integrated? Any blockers?

zhaquirks/tuya/tuya_sensor.py
lenucksi: @Aggrathon Turned this into a commit suggestion if this is the last blocking element. Whoever has the rights to accept it, consider doing so.
```suggestion
        converter=lambda x: {0: 20, 1: 50, 2: 100}[x],
```

lenucksi:
I went around a bit and collated everything on this device to the attached gist. Following the integrated installation advice it did work well on my device. In case that helps anyone. Original PR is probably just as good but not merged yet, so this might be useful for the interim period.

https://gist.github.com/lenucksi/f448843ba3dfeeba9a45cd2b8c57710e

lenucksi (changes requested):
## Review: PR #4520 — `_TZE284_hodyryli`

Adds a quirk for a Tuya temperature/humidity sensor with external probe and LCD clock. The approach is sound, but several issues should be addressed before merging.

---

### 1. Battery percentage values are incorrect

```python
converter=lambda x: {0: 10, 1: 50, 2: 100}[x],
```

`battery_percentage_remaining` uses a 0–200 scale (200 = 100%). These values translate to 5 % / 25 % / 50 %, not 10 % / 50 % / 100 % as the comment suggests.

The Z2M converter for the identical device uses `{0: 10, 1: 50, 2: 100}` as *percentage* values directly (see Koenkk/zigbee2mqtt#29627). The equivalent on the 0–200 scale would be `{0: 20, 1: 100, 2: 200}`.

Additionally, the bare subscript `[x]` will raise `KeyError` if the device ever sends an unexpected battery state. Prefer `.get(x, 0)` for safety.

**Fix**: `converter=lambda x: {0: 20, 1: 100, 2: 200}.get(x, 0),`

---

### 2. Comment says /100, code does /10

```python
    # External temperature sensor (DP 38, value needs to be divided by 100 to get correct temperature)
    .tuya_sensor(
        ...
        converter=lambda x: x / 10,  # Divide by 10 to get correct temperature
```

Line 374 and line 379 contradict each other. Both @seanmahrt and @DesertRatUa confirmed on real hardware that `/10` is correct.

**Fix**: Remove or correct the comment on line 374.

---

### 3. `TuyaBatteryState` and `TuyaTimeFormat` are defined but unused

`tuya_sensor.py:35–47` defines these two enums, but neither is referenced in the quirk. Dead code should be removed unless there is a plan to use them in a follow-up (e.g. for a `tuya_enum` on DP 3 or DP 17).

---

### 4. Missing tests

Every other quirk in `tuya_sensor.py` has a corresponding test entry in `tests/test_tuya_sensor.py`. For this device we need test coverage for:

- DP 1: internal temperature (positive + negative values)
- DP 2: humidity
- DP 3: battery (enum → percentage mapping)
- DP 38: external probe temperature (including negative/freezer values)
- USP handling (unsupported DP)

---

### 5. DP 17 (time_format) is not handled (optional)

The device exposes DP 17 as `time_format` (0 = 24 h, 1 = 12 h). The Z2M side has already merged support for this in Koenkk/zigbee-herdsman-converters#12180. Consider adding it as a read-only sensor for consistency.

---

### Recommendation

**Blocking:**
1. Fix battery converter values
2. Add tests

**Should fix:**
3. Correct / remove misleading comment on line 374

**Nice to have:**
4. Remove unused enums
5. Consider adding DP 17 support

lenucksi (commented):
Probably sane, forgot about the caveats of implementing it for my use back when so probably will not be of much help anymore. 

There's a bunch of automated review comments in the tuya_sensor.py and apparently there's a migration thing going on as per the labels. If those are an easy mod I guess fixing them up would be good. 
Otherwise gettings this thing in for one and then iterating in a second follow-up PR is probably the sanest way? 🤷

lenucksi (commented):
The auto-PR questions are gone. Do we get a judgement from @TheJulianJES ?

lenucksi (approved):
Removing the blocker-marker.

PR #4523, opened 2025-11-26
Support for the Tuya PJ-1203 single channel clamp power meter _TZE204_cjbofhxw, _TZE284_cjbofhxw Model: TS0601

chrisf4lc0n:
This quirk adds support for the Tuya PJ-1203 single channel clamp power meter.
The device reports voltage, current, power and energy used. Apparent power and power factor are calculated from these values.

Manufacturer IDs: _TZE204_cjbofhxw, _TZE284_cjbofhxw
Model: TS0601
Purchase link: https://www.aliexpress.com/item/1005005994777032.html

Datapoints:
DP 18: Current (mA)
DP 19: Power (W × 10)
DP 20: Voltage (V × 10)
DP 101: Energy (Wh)

Exposed entities:
Voltage (V)
Current (A)
Power (W)
Apparent Power (VA) - calculated
Power Factor (%) - calculated
Energy (kWh)

Proposed change:
Add a new quirk for the Tuya PJ-1203 single channel clamp energy meter that maps Tuya datapoints to ZHA ElectricalMeasurement cluster attributes and calculates apparent power and power factor from the reported values.

Additional information:
This is a single channel variant; dual channel PJ-1203A models use different datapoints and would need a separate quirk
The device does not report AC frequency, so that entity will remain unavailable
Tested and working with ZHA on Home Assistant 2025.11.2

Device diagnostics:
[zha-01KAEDWDH9AYZDVE1HTD8MCW6X-_TZE284_cjbofhxw TS0601-6850c3f422189896ca2536b313d6e7fb.json](https://github.com/user-attachments/files/23652406/zha-01KAEDWDH9AYZDVE1HTD8MCW6X-_TZE284_cjbofhxw.TS0601-6850c3f422189896ca2536b313d6e7fb.json)

Checklist:
The changes are tested and work correctly
pre-commit checks pass / the code has been formatted using Black
Tests have been added to verify that the new code works
Device diagnostics data has been attached


chrisf4lc0n:
I am struggling to implement the reliable function to monitor the Consumed Energy total, it looks like the guys from Z2M have the same issue. It seems to reset the value each time each time it reconnects. I have gone through logs of power DP 19 and DP 101 to see if Claude could come up with something but all it did was an absolute mess out of the commit, lesson learnt I should have kept it separate before committing...
The easiest way would be to create a separate sensor in HA to calculate cumulative energy used over time instead of playing substandard data provided by DP 101.

chrisf4lc0n:
After pulling enough logs and comparing the energy used vs power I found a dip in cumulative energy consumption.
To alleviate that an implementation of energy offset restoration from ZHA's attribute cache to maintain
energy accumulation continuity across quirk reloads/restarts was introduced. When the device
counter is lower than the cached value, the offset is restored to prevent energy loss.

I will monitor the data for the next 24-48hrs and see if that has fixed the problem.
There are still edge cases when stale cache from crash, multiple resets during offline period or cache loss and a few more even more obscure cases can lead to some data loss and dips in energy used but it is still far better than the original implementation, which was all over the place.

PR #4524, opened 2025-11-26
Update Third Reality plug entity names + add entities

3reality-support:
## Proposed change

There are two types of dual plugs, one with left and right sockets, and the other with top and bottom sockets, with EP1 located below
Modify the property name of a private cluster

Add 3 private attributes：
1. `power_rise_threshold`: Reports sudden power changes. Power rise and fall alerts can be enabled separately. Threshold adjustable.
2. `power_drop_threshold`: Reports sudden power changes. Power rise and drop alerts can be enabled separately. Threshold adjustable.Threshold adjustable.
3. `metering_only_mode`: When enabled, the device enters metering-only mode and the relay is forced to stay ON.
## Checklist


3reality-support:
Follow up with this PR
https://github.com/zigpy/zha-device-handlers/pull/5024

PR #4526, opened 2025-11-26
Add Sonoff SNZB-02DR2 LCD temperature sensor settings

zmguko:
## Checklist


puddly:
Could you attach diagnostics JSON for this device please? We rely on diagnostics JSON for automated tests.

TheJulianJES (approved):
Z2M has these as separate models, but the only difference seems to be `batteryVoltage` support and that `batteryVoltageReport` is set up.

So, I think we should be good to have them merged in one quirk. But a diagnostics file would still be needed.

For comparasion:
- `SNZB-02D`: https://github.com/Koenkk/zigbee-herdsman-converters/blob/ce520abf31550c81220b57766cfa7fcda58d77be/src/devices/sonoff.ts#L1636-L1747
- `SNZB-02DR2`: https://github.com/Koenkk/zigbee-herdsman-converters/blob/ce520abf31550c81220b57766cfa7fcda58d77be/src/devices/sonoff.ts#L1845-L1956

DariukaB:
is it possible to add the attribute for how often to send/update temp and humidity records? would be useful to lower the default interval

PR #4535, opened 2025-11-29
Allow assigning Tuya datapoints to different endpoints

nschimme:
## Proposed change
TuyaQuirkBuilder did not have the ability to assign data to different endpoints which resulted in some devices not being easily supported as values from the same endpoint are overwritten when they are reported simultaneously by the device.

## Additional information
See https://github.com/zigpy/zha-device-handlers/pull/4510 for more details

## Device diagnostics
N/A

## Checklist

### DIFF
--- a/tests/test_tuya_builder.py
+++ b/tests/test_tuya_builder.py
@@ -561,3 +561,28 @@ async def test_tuya_override_mcu_command(
 
     assert tuya_listener.attribute_updates[0][0] == 0xEF0A
     assert tuya_listener.attribute_updates[0][1] == TestEnum.B
+
+
+async def test_tuya_quirk_builder_endpoint_id(device_mock):
+    """Test TuyaQuirkBuilder endpoint_id."""
+
+    registry = DeviceRegistry()
+
+    (
+        TuyaQuirkBuilder(device_mock.manufacturer, device_mock.model, registry=registry)
+        .adds_endpoint(2)
+        .tuya_humidity(dp_id=1, endpoint_id=2)
+        .tuya_soil_moisture(dp_id=2)
+        .skip_configuration()
+        .add_to_registry()
+    )
+
+    quirked = registry.get_device(device_mock)
+    assert isinstance(quirked, CustomDeviceV2)
+    assert quirked in registry
+
+    assert not hasattr(quirked.endpoints[1], "humidity")
+    assert hasattr(quirked.endpoints[2], "humidity")
+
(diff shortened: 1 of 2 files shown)


nschimme:
#4536 can be further simplified once this is merged

nschimme:
@TheJulianJES this PR should allow us to set humidity data to another endpoint to prevent conflicts seen in #4536

nschimme:
@puddly @TheJulianJES polite bump as this can simplify things in #4536

tests/test_tuya.py
```diff
@@ -2023,3 +2023,19 @@ async def test_ts601_door_sensor(
     attrs = await cluster.read_attributes(attributes=[attribute])
 
     assert attrs[0].get(attribute) == expected_value
+
+
+async def test_tuya_quirk_builder_endpoint_id(zigpy_device_from_v2_quirk):
+    """Test TuyaQuirkBuilder endpoint_id."""
+
+    (
+        zhaquirks.tuya.builder.TuyaQuirkBuilder("manufacturer", "model")
```
TheJulianJES: Please import `TuyaQuirkBuilder`.

tests/test_tuya.py
```diff
@@ -2023,3 +2023,19 @@ async def test_ts601_door_sensor(
...
+        .adds_endpoint(2)
+        .tuya_humidity(dp_id=1, endpoint_id=2)
+        .add_to_registry()
+    )
+
+    device: Device = zigpy_device_from_v2_quirk("manufacturer", "model")
+
+    assert not hasattr(device.endpoints[1], "humidity")
+    assert hasattr(device.endpoints[2], "humidity")
```
TheJulianJES: I think this test should be moved to the dedicated test file for the Tuya quirk builder.

TheJulianJES (commented):
Looks good, thanks! Sorry for the long wait. Two small comments.

In general, the way of how the datapoints need to be remapped to an individual `LocalDataCluster` is not great. It's not ideal that we have to teach the `TuyaQuirkBuilder` to use different endpoints here, but it should also be rarely needed.

In the future, we can hopefully change some of the internals, so this is all handled a bit cleaner behind the scenes – without affecting existing Tuya v2 quirks

nschimme:
@TheJulianJES Thanks for the review, please take another look. I'm also investigating how to do this automatically.

TheJulianJES (approved):

TheJulianJES:
Thanks!

(No need to force push on this repo btw. We squash when merging anyway)

PR #4160, opened 2025-07-02
Add HOBEIAN ZG-303Z / Tuya _TZE200_npj9bug3 soil sensor

makuser:
## Proposed change
Add Support for HOBEIAN (Tuya) soil sensor `ZG-303Z` and Tuya soil sensor `_TZE200_npj9bug3`

## Additional information

Fixes #4144, https://github.com/zigpy/zha-device-handlers/issues/4122#issuecomment-2983097722

## Checklist

- [ ] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

PR #4457, opened 2025-11-01
Add HOBEIAN ZG-204ZV mmWave presence sensor

zimme:
## Proposed change
Add support for HOBEIAN ZG-204ZV 10GHz mmWave presence sensor with temperature, humidity, and illuminance sensors.

## Additional information
Additional supported manufacturers:
- _TZE200_uli8wasj
- _TZE200_grgol3xp
- _TZE200_rhgsbacq
- _TZE200_y8jijhba

Configuration options:
- Motion detection sensitivity (0-19)
- Motion keep time / fading time (0-28800 seconds)
- Temperature offset calibration (-2 to 2°C, step 0.1)
- Humidity offset calibration (-30 to 30%)
- Illuminance interval (1-720 minutes)
- LED indicator (on/off)
- Temperature unit (Celsius/Fahrenheit)

Environmental sensors (temperature, humidity, illuminance) report via standard Zigbee clusters. Occupancy detection and configuration settings use Tuya datapoints.

Implementation matches Z2M converter for this device.

Fixes #4268

## Device diagnostics

<details>
<summary>Device diagnostics JSON</summary>

```json
{
  "home_assistant": {
    "installation_type": "Home Assistant Container",
    "version": "2025.11.0b1",
    "dev": false,
    "hassio": false,
    "virtualenv": false,
    "python_version": "3.13.9",
    "docker": true,
    "arch": "x86_64",
    "timezone": "Europe/Stockholm",
    "os_name": "Linux",
    "os_version": "6.8.0-87-generic",
    "container_arch": "amd64",
    "run_as_root": true
  },
  "custom_components": {
    "ui_lovelace_minimalist": {
      "documentation": "https://ui-lovelace-minimalist.github.io/UI/",
      "version": "v1.3.18",
      "requirements": [
        "aiofiles>=0.8.0",
        "aiogithubapi>=22.2.4"
      ]
    },
    "robovac": {
      "documentation": "https://github.com/codefoodpixels/robovac",
      "version": "1.0.0",
      "requirements": []
    },
    "config_editor": {
      "documentation": "https://github.com/junkfix/config-editor-card",
      "version": "5.0.0",
      "requirements": []
    },
    "hacs": {
      "documentation": "https://hacs.xyz/docs/use/",
      "version": "2.0.5",
      "requirements": [
        "aiogithubapi>=22.10.1"
      ]
    },
    "zha_toolkit": {
      "documentation": "https://github.com/mdeweerd/zha-toolkit",
      "version": "v1.1.33",
[...]

PR #4536, opened 2025-11-29
Add HOBEIAN ZG-303Z + Tuya soil sensor `_TZE200_npj9bug3`

nschimme:
## Proposed change
This allows soil moisture and humidity to both been seen. Previously, only temperature, soil moisture, and battery were shown. Even worse, soil moisture was incorrectly displayed as humidity due to it being reported under cluster 0x405.

## Additional information
image

This PR replaces #4510 and #4160

## Device diagnostics
[zha-01K8BQMJSKZFPA3YC43C2Q4TB0-HOBEIAN ZG-303Z-aee38a7af56551153037cf883271628f.json](https://github.com/user-attachments/files/23835168/zha-01K8BQMJSKZFPA3YC43C2Q4TB0-HOBEIAN.ZG-303Z-aee38a7af56551153037cf883271628f.json)

## Checklist


dmnagornyi:
This looks really cool. I hope it will be merged and included in the next update.

itsnotrisky:
LGTM as a custom quirk module 👍 Looking forward to be merged

nschimme:
Polite bump @TheJulianJES ; please review

nschimme:
polite bump @TheJulianJES

puddly (commented):
For v2 quirks that have no "code" and just use standard v2 APIs, you can delete the explicit unit test.

puddly:
Ah, I didn't see the `converter=lambda x: not bool(x),`.

I guess you'd need a very minimal test for that to ensure coverage 😄. If you could move the imports from your unit test to the top of the file and keep just `# 10. Test DP 106: Dry binary sensor` from your test, I think this looks good to go.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        attribute_name="soil_moisture_calibration",
+        type=t.uint16_t,
+        unit=PERCENTAGE,
+        min_value=-30,
+        max_value=30,
+        step=1,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.HUMIDITY,
+        translation_key="soil_moisture_calibration",
+        fallback_name="Soil moisture Calibration",
```
TheJulianJES: Due to HA naming conventions, we should use sentence case (though abbreviations like "LED" are fine). The `fallback_name` is the the default English translation in HA. So, this should be "Soil moisture calibration" (lower-case "c").

However, to be aligned with other quirks v2 entities, we should use offset for all these entities instead. So, this should be:

```suggestion
        translation_key="soil_moisture_offset",
        fallback_name="Soil moisture offset",
```

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        type=t.uint16_t,
+        unit=UnitOfTemperature.KELVIN,
+        min_value=-2,
+        max_value=2,
+        step=0.1,
+        multiplier=10,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.TEMPERATURE,
+        translation_key="temperature_calibration",
+        fallback_name="Temperature Calibration",
```
TheJulianJES:
```suggestion
        translation_key="temperature_offset",
        fallback_name="Temperature offset",
```

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        attribute_name="humidity_calibration",
+        type=t.uint16_t,
+        unit=PERCENTAGE,
+        min_value=-30,
+        max_value=30,
+        step=1,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.HUMIDITY,
+        translation_key="humidity_calibration",
+        fallback_name="Humidity Calibration",
```
TheJulianJES:
```suggestion
        translation_key="humidity_offset",
        fallback_name="Humidity offset",
```

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+    )
+    .tuya_dp_attribute(
+        dp_id=106,
+        attribute_name="dry",
+        converter=lambda x: not bool(x),
+    )
```
TheJulianJES, line 404: Is this "dry" after or before being inverted? The `attribute_name` should match whatever the state is when enabled after the conversion. So, if this attribute was originally "on" when dry, but "on" when wet after the conversion, we should use `wet` or `moisture` as the attribute name here.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+    .binary_sensor(
+        attribute_name="dry",
+        cluster_id=TuyaMCUCluster.cluster_id,
+        endpoint_id=1,
+        entity_type=EntityType.STANDARD,
+        device_class=BinarySensorDeviceClass.MOISTURE,
+        initially_disabled=False,
+        attribute_initialized_from_cache=False,
+        translation_key="dry",
```
TheJulianJES: We can use the name of the Home Assistant device class here. It'll be "Moisture" for the entity name.
We can do that by removing the `translation_key`:
```suggestion
```

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        fallback_name="dry",
```
TheJulianJES: This won't be used, but let's adjust it to fit with the HA entity name as well:
```suggestion
        fallback_name="Moisture",
```

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+    .tuya_sensor(
+        dp_id=109,
+        type=t.uint16_t,
+        attribute_name="humidity_value",
+        unit=PERCENTAGE,
+        device_class=SensorDeviceClass.HUMIDITY,
+        entity_type=EntityType.STANDARD,
+        translation_key="humidity",
+        fallback_name="Humidity",
+    )
```
TheJulianJES: Please use the [`tuya_humidity` method](https://github.com/zigpy/zha-device-handlers/blob/7a57309cb6aff76739bd0e6e9438320e33ec4afa/zhaquirks/tuya/builder/__init__.py#L420-L434).

(At the moment, it's hooked up to a `LocalDataCluster` emulating a ZCL-compliant cluster like the `tuya_temperature` you're already using, though this might change in the future. It'll mean you will see a new entity appear in Home Assistant though.)

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        attribute_name="temperature_sampling",
+        type=t.uint16_t,
+        unit=UnitOfTime.SECONDS,
+        min_value=5,
+        max_value=3600,
+        step=1,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.DURATION,
+        translation_key="temperature_sampling",
+        fallback_name="Temperature Sampling",
```
TheJulianJES: We have a [`temperature_report_interval` translation](https://github.com/home-assistant/core/blob/10d32b7f23b0a399e845ea9f047ed3eb99b23864/homeassistant/components/zha/strings.json#L899-L901) already. I think we can use that instead:

```suggestion
        translation_key="temperature_report_interval",
        fallback_name="Temperature report interval",
```

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,129 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        attribute_name="soil_moisture_sampling",
+        type=t.uint16_t,
+        unit=UnitOfTime.SECONDS,
+        min_value=5,
+        max_value=3600,
+        step=1,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.DURATION,
+        translation_key="soil_moisture_sampling",
+        fallback_name="Soil Moisture Sampling",
```
TheJulianJES:
```suggestion
        translation_key="soil_moisture_report_interval",
        fallback_name="Soil moisture report interval",
```

TheJulianJES:
I've left some feedback above. Thanks for the PR!

Per the other comments:
The test is indeed not really necessary. We want added logic in custom clusters to be fully covered. Just raw definitions don't have to be covered. They'll automatically be tested in ZHA later with the provided diagnostics files.

`converter=lambda x: not bool(x)` is technically some logic, though it won't affect code coverage (technical reasons due to `lambda` definition). It's nice to cover, but since it's so trivial, I'd say it's not really necessary. But you can keep that part of the test. I think the other stuff can be removed, as I don't see a huge benefit here. Or do you think it's good to keep?

(Also note that `pytest.mark.parametrize` can generally be used for repetitive tests like this. But if the other stuff is removed, that's not required here.)

zhaquirks/tuya/tuya_sensor.py
nschimme: When I was testing this I wasn't able to as soil moisture would overwrite the humidity since they all shared the same endpoint.

We would need https://github.com/zigpy/zha-device-handlers/pull/4535 to allow us to use endpoints 2 and 3 for Soil Moisture (DP 3) and Relative Humidity (DP 109), respectively. This prevents values from being overwritten when they are reported simultaneously by the device. Can you review that too?

zhaquirks/tuya/tuya_sensor.py
nschimme: It is "wet" but the boolean represents dry so we invert it.

nschimme:
> `converter=lambda x: not bool(x)` is technically some logic, though it won't affect code coverage (technical reasons due to `lambda` definition). It's nice to cover, but since it's so trivial, I'd say it's not really necessary. But you can keep that part of the test. I think the other stuff can be removed, as I don't see a huge benefit here. Or do you think it's good to keep?

We can simplify, I was just adding tests as the PR checklist required it.

zhaquirks/tuya/tuya_sensor.py
nschimme: Renamed to "moisture" as you suggested

zhaquirks/tuya/tuya_sensor.py
nschimme: I've addressed the other comments @TheJulianJES and confirmed they work. Should we do the remainder in #4535?

nschimme: @TheJulianJES polite bump

nschimme: @puddly could you assist?

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,128 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        enum_class=TuyaTempUnitConvert,
+        entity_type=EntityType.CONFIG,
+        translation_key="display_unit",
+        fallback_name="Display unit",
+    )
+    .tuya_battery(dp_id=15, battery_type=BatterySize.AAA)
+    .tuya_number(
+        dp_id=102,
+        attribute_name="soil_moisture_offset",
+        type=t.uint16_t,
```
Machalkas:
```suggestion
        type=t.int8s,
```
Offset value can be negative

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,128 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        step=1,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.HUMIDITY,
+        translation_key="soil_moisture_offset",
+        fallback_name="Soil moisture offset",
+    )
+    .tuya_number(
+        dp_id=104,
+        attribute_name="temperature_offset",
+        type=t.uint16_t,
```
Machalkas:
```suggestion
        type=t.int8s,
```

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,128 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        multiplier=10,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.TEMPERATURE,
+        translation_key="temperature_offset",
+        fallback_name="Temperature offset",
+    )
+    .tuya_number(
+        dp_id=105,
+        attribute_name="humidity_offset",
+        type=t.uint16_t,
```
Machalkas:
```suggestion
        type=t.int8s,
```

zhaquirks/tuya/tuya_sensor.py
nschimme: Polite bump @puddly @TheJulianJES

puddly: Sorry for the reviewing delay! We're currently in the last stage of a huge refactor and will be reviewing and merging a lot of the pending quirk PRs soon.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -333,3 +337,128 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        endpoint_id=1,
+        entity_type=EntityType.STANDARD,
+        device_class=BinarySensorDeviceClass.MOISTURE,
+        initially_disabled=False,
+        attribute_initialized_from_cache=False,
+        fallback_name="Moisture",
+    )
+    .removes(RelativeHumidity.cluster_id)
+    .tuya_sensor(
+        dp_id=109,
```
butzei:
```suggestion
        dp_id=3,
```

zhaquirks/tuya/tuya_sensor.py
nschimme: Resolved by using new adds_endpoint() and endpoint_id

nschimme:
@TheJulianJES this is ready now, it uses the new functionality `adds_endpoint(2)` and specifying `endpoint_id=2` for `tuya_humidity()` to avoid the conflict with `tuya_soil_moisture()`

slymas:
Hi,

First of all, thank you so much for working on this! 

I have been following this process for a while, and I think the "moisture min" is still not working correctly. At least "moisture" doesn't change between dry or wet if I change the "moisture min" below or above it. To me it looks like as if "moisture min" was hardcoded to 20% somewhere, as any moisture reading below 20% will report dry, while any above will report wet. I have 5 sensors from the same vendor, all ZG-303Z, and they exhibit the same behaviour.

Example from 2 of my sensors where 49% is wet with a "moisture min" of 50%, and 19% is dry with a "moisture min" of 12%:
2026-03-07 10_57_38-Climate – Home Assistant

aagit:
Hello,

The humidity/temp/soil offsets give an error about int16 if I touch them, I found a different version of this quirk that defines those 3 as int8s and the error goes away, but even with int8s I haven't noticed any other effect from those tuning sliders yet.

temp:
Tried it out and works great for my ZG-303Z sensor (soil moisture was missing before), but in home assistant the soil moisture value is added with device_class "humidity", not "moisture". And after a re-add I only get 0%, before it was the correct value 13%.

SuperSandro2000:
Could it be that the type of the sensor is not updated when adding the quirk afterwards? I have the same problem.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+    .tuya_number(
+        dp_id=102,
+        attribute_name="soil_moisture_offset",
+        type=t.uint16_t,
+        unit=PERCENTAGE,
+        min_value=-30,
+        max_value=30,
+        step=1,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.HUMIDITY,
```
SuperSandro2000:
```suggestion
        device_class=NumberDeviceClass.MOISTURE,
```
Shouldn't this be like that?

SuperSandro2000: @temp that actually fixes my problem

temp: But isn't this just the offset value? I'm having problems with the soil moisture value itself...

SuperSandro2000: I assume the offset overrides the type. Please try it out, it worked for me.

temp: Did you re-add the device? Just changing the quirk and restarting HA didn't work for me. It's still humidity.

SuperSandro2000: Before I tried editing the types in the developer tools which worked until I restarted where they then reverted back. Then I adjusted the quirk and restarted and they are correct now. Are quirks path dependent? I also changed that but I thought that doesn't really matter. I didn't pair the device new.

SuperSandro2000: Now it broke again :( I have no idea why...
It worked yesterday across multiple restarts...

SuperSandro2000: I just found https://github.com/Olen/homeassistant-plant/blob/main/TIPS.md#option-1-use-customizeyaml-simplest so this might be a generell issue not even specific to this device.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        fallback_name="Display unit",
+    )
+    .tuya_battery(dp_id=15, battery_type=BatterySize.AAA)
+    .tuya_number(
+        dp_id=102,
+        attribute_name="soil_moisture_offset",
+        type=t.uint16_t,
+        unit=PERCENTAGE,
+        min_value=-30,
+        max_value=30,
```
SuperSandro2000: I am not sure what the correct type here is, but an unsigned integer cannot be negative and that probably produces the errors people described and I observed myself that the offset cannot be negative.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        multiplier=10,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.TEMPERATURE,
+        translation_key="temperature_offset",
+        fallback_name="Temperature offset",
+    )
+    .tuya_number(
+        dp_id=105,
+        attribute_name="humidity_offset",
+        type=t.uint16_t,
```
SuperSandro2000: same here

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        max_value=30,
+        step=1,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.MOISTURE,
+        translation_key="soil_moisture_offset",
+        fallback_name="Soil moisture offset",
+    )
+    .tuya_number(
+        dp_id=104,
+        attribute_name="temperature_offset",
```
SuperSandro2000, line 375: and same here but HASS does not produce an error

zhaquirks/tuya/tuya_sensor.py
nschimme: aha, this was identified earlier by @Machalkas and I accidentally reverted it. Fixed!

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        type=t.int8s,
+        unit=UnitOfTemperature.KELVIN,
```
SuperSandro2000: Other temperature_offset I could find use CELSIUS and nowhere else in this repo KELVIN is used. Not sure if my observation is correct though that this should be CELSIUS.

nschimme: Z2M documentation seems to agree as well: https://www.zigbee2mqtt.io/devices/ZG-303Z.html

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        endpoint_id=1,
+        entity_type=EntityType.STANDARD,
+        device_class=BinarySensorDeviceClass.MOISTURE,
+        initially_disabled=False,
+        attribute_initialized_from_cache=False,
+        fallback_name="Moisture",
+    )
+    .removes(RelativeHumidity.cluster_id)
+    .adds_endpoint(2)
+    .tuya_humidity(dp_id=109, endpoint_id=2)
```
SuperSandro2000, line 417: I am a noob when writing quriks but do we really need to replace the standard humidity cluster? When enabling the quirk I need to manually delete the old humdiity entity from hass.

SuperSandro2000:
Seems to work well, definitely better than without it.

image

(I am doing a calibration test, so 100% moisture is expected)

@TheJulianJES @puddly can you take another look to move forward, please?

nschimme:
Polite bump @TheJulianJES and @puddly ; please review. Works for at least three folks now

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        device_class=NumberDeviceClass.HUMIDITY,
+        translation_key="humidity_offset",
+        fallback_name="Humidity offset",
+    )
+    .tuya_dp_attribute(
+        dp_id=106,
+        attribute_name="moisture",
+        converter=lambda x: not bool(x),
+    )
+    .binary_sensor(
```
TheJulianJES, line 444: note: Might make sense to introduce a Tuya "moisture binary sensor" builder method for this in the future.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        attribute_name="moisture",
+        cluster_id=TuyaMCUCluster.cluster_id,
+        endpoint_id=1,
+        entity_type=EntityType.STANDARD,
+        device_class=BinarySensorDeviceClass.MOISTURE,
+        initially_disabled=False,
```
TheJulianJES: By default, sensors are already enabled, so no need to explicitly specify that again.
```suggestion
```

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        attribute_initialized_from_cache=False,
```
TheJulianJES: This is to force-read the virtual attribute (thus DP) for the device on every startup? Does that actually work? (i.e. the device doesn't sleep fully), or why was this added?

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        fallback_name="Moisture",
+    )
+    .removes(RelativeHumidity.cluster_id)
+    .adds_endpoint(2)
```
TheJulianJES, line 453: So, I'm not a huge fan of this, but we kind of need to do it like this at the moment, as we remap the `tuya_humidity` to local ZCL clusters and so on.

... but see the other comment. I wonder if we actually need to have a second endpoint?

zhaquirks/tuya/tuya_sensor.py
TheJulianJES, line 454: But I do have to ask for this again: Why did we do this whole different endpoint ID thing?
`RelativeHumidity` and `SoilMoisture` are different clusters, with different IDs, so they shouldn't conflict. What issue did the quirk show when it's not explicitly specified as `endpoint_id=2` here? If removed, is there still an issue with the latest HA version?

I only see that the incorrect device class is used, but that's a ZHA thing (soil moisture just didn't exist back then in HA – tracking that in https://github.com/zigpy/zha/pull/604).

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+(
+    TuyaQuirkBuilder("HOBEIAN", "ZG-303Z")
+    .applies_to("_TZE200_npj9bug3", "TS0601")  # COOLO CS-201Z
+    .tuya_soil_moisture(dp_id=3)
+    .tuya_temperature(dp_id=5, scale=10)
+    .tuya_enum(
+        dp_id=9,
+        attribute_name="display_unit",
+        enum_class=TuyaTempUnitConvert,
+        entity_type=EntityType.CONFIG,
```
TheJulianJES: As v2 quirks are mostly for config entities, `entity_type=EntityType.CONFIG` is the default for `tuya_enum` and `tuya_numbers` (and most other builder methods), so I'd just remove this everywhere.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -337,3 +341,120 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+        dp_id=104,
+        attribute_name="temperature_offset",
+        type=t.int8s,
+        unit=UnitOfTemperature.CELSIUS,
+        min_value=-2,
+        max_value=2,
+        step=0.1,
+        multiplier=10,
+        entity_type=EntityType.CONFIG,
+        device_class=NumberDeviceClass.TEMPERATURE,
```
TheJulianJES: This needs to use the temperature delta device class to not cause issues when using °F in HA:
```suggestion
        device_class=NumberDeviceClass.TEMPERATURE_DELTA,
```

TheJulianJES (commented):
Thanks. This is mostly good. Can you just explain why we need a different endpoint for the virtual `SoilMoisture` cluster? It doesn't share the same ID as `RelativeHumidity`, so they should be fine living on the same endpoint..?

zhaquirks/tuya/tuya_sensor.py
nschimme: It was there in the original PR I took over. I can test by removing it.

zhaquirks/tuya/tuya_sensor.py
nschimme: Confirmed that a second endpoint still required on current ZHA. If we use the same endpoint humidity and soil moisture collide and share a single value for both on updates.

zhaquirks/tuya/tuya_sensor.py
nschimme: See below comment, this is still required to prevent a collision.

zhaquirks/tuya/tuya_sensor.py
nschimme: Works without, removing

nschimme:
@TheJulianJES thanks for the review. I've addressed the feedback and found we still need the endpoint workaround.

mikelock-1991:
Works beautifully here!

simonwiles:
~~This seems to work for me (`_TZE200_npj9bug3`) except "Soil moisture" is always at 0% (as seemed to affect @temp and @SuperSandro2000) -- am I doing something wrong?~~

Apologies, it just wasn't reporting properly, my mistake 🫠.  All good now.

nschimme:
@TheJulianJES can we please get this merged? I've addressed all the feedback. Appreciate it!

mikelock-1991:
image

One issue, I have set the moisture alarm to 0% as this plant doesn’t need much water. But it still reports Dry. Why?
Also, after a reload of the ZHA integration, the alarm resets to default 20%.

When I manually change it in the device via manage device, it seems to survive a ZHA reload.

For now I created an automation that resets the moisture alarm % and status when they are incorrect

pingorczinger:
Works for me.

aagit:
> One issue, I have set the moisture alarm to 0% as this plant doesn’t need much water. But it still reports Dry. Why? Also, after a reload of the ZHA integration, the alarm resets to default 20%.

I've also one of these stuck in Dry despite soil moisture is high, like 70% soil moisture 93% humidity... and temperature works fine too, just he dry/wet is stuck in Dry. I don't use that sensor so it's only aesthetical, but it looks odd.

porelli (commented):
Adding another white-label of this hardware, plus independent data on the endpoint question that's been holding this up.

### AOYAN AY-303Z

Z2M already covers it in the same definition as the ZG-303Z (`zigbeeModel: ["ZG-303Z", "AY-303Z", "AY-302Z"]` — COOLO CS-201Z plus white labels), and all 12 DPs in this quirk match its behaviour exactly. Signature as ZHA sees it on HA 2026.8.2:

```
manufacturer: 'AOYAN  '   # two trailing spaces
model:        'AY-303Z'
endpoint 1: profile=260 device_type=770
  in:  0x0000, 0x0001, 0x0003, 0x0402, 0x0405, 0xEF00
  out: 0x0003
```

I've left a committable suggestion for the matcher inline.

I'd suggest **not** blanket-adding `AY-302Z` here: per the z2m converter it's the same DP set minus humidity (`humidity` and `humidity_calibration` are excluded for that model), so it wants its own builder without DP 109/105. I don't have one to test against.

### Re: whether the second endpoint is still needed

Independent confirmation that it is, on HA 2026.8.2 and on a different manufacturer variant. I ran both revisions against the same physical device:

- DP 109 humidity via `.tuya_sensor(...)` on **endpoint 1** → soil moisture reported 16 %, humidity stayed `unknown` indefinitely.
- `.adds_endpoint(2)` + `.tuya_humidity(dp_id=109, endpoint_id=2)` → both report and keep updating: soil moisture 19 %, humidity 72 %.

So the collision still reproduces on current ZHA, and `adds_endpoint(2)` is what fixes it. Full parity with z2m on all 12 DPs once applied.

### Minor note for whoever merges

`.removes(RelativeHumidity.cluster_id)` leaves the pre-quirk humidity entity behind as `unavailable` (as @SuperSandro2000 noted above). It has to be deleted by hand, otherwise the replacement lands as `..._humidity_2`. Probably worth a line in the PR description.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -379,3 +383,111 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
     .skip_configuration()
     .add_to_registry()
 )
+
+
+(
+    TuyaQuirkBuilder("HOBEIAN", "ZG-303Z")
+    .applies_to("_TZE200_npj9bug3", "TS0601")  # COOLO CS-201Z
```
porelli, line 390: Same hardware, different manufacturer string — the AOYAN white-label of this sensor. Note the **two trailing spaces** in the manufacturer, which zigpy preserves verbatim, so the matcher needs them:

```suggestion
    .applies_to("_TZE200_npj9bug3", "TS0601")  # COOLO CS-201Z
    .applies_to("AOYAN  ", "AY-303Z")  # AOYAN AY-303Z
```

Z2M's fingerprint carries the same `"AOYAN  "`, so it's the device reporting it that way rather than a zigpy artefact.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -379,3 +383,111 @@ class ServerCommandDefs(TuyaMCUCluster.ServerCommandDefs):
...
+    )
+    .tuya_number(
+        dp_id=104,
+        attribute_name="temperature_offset",
+        type=t.int8s,
+        unit=UnitOfTemperature.CELSIUS,
+        min_value=-2,
+        max_value=2,
+        step=0.1,
+        multiplier=10,
```
porelli: This multiplier is inverted, and it makes the entity completely non-functional.

ZHA's `NumberConfigurationEntity` does `value * self._multiplier` for display and `int(value / self._multiplier)` on write, and `tuya_number` passes the raw DP straight through with no converter, so a decidegree DP needs 0.1 here.

With 10, `int(value / 10)` is 0 for every value in the -2..2 range, so any write silently becomes zero. On my AY-303Z, setting 1.0 returned HTTP 200 and the entity stayed at 0, still 0 after 25 seconds. Changing only this line to 0.1: setting 1.0 gives 1.0, setting 0.5 gives 0.5, and both persist.

Display is wrong in the same way, a device-reported raw 5 renders as 50.0 °C, outside the entity's own -2..2 range. For reference, #4457 uses 0.1 for the equivalent DP on the ZG-204ZV.

```suggestion
        multiplier=0.1,
```

porelli (commented):
Found one more thing while testing this on hardware, flagged inline. It's a live functional break rather than cosmetics, so probably worth catching before this merges.

tomerhayundev:
Another AOYAN AY-303Z data point, on a newer stack: HA 2026.9.1 / zha-quirks 2.2.1. Signature identical to the one @porelli posted (manufacturer `"AOYAN  "`, two trailing spaces).

While waiting for this PR I ran a local custom quirk with a **single-endpoint** layout: `.removes(RelativeHumidity.cluster_id)`, DP 109 as a `.tuya_sensor(...)` attribute on the Tuya cluster, and `.tuya_soil_moisture(dp_id=3)` on endpoint 1. Both keep reporting with no collision:

- DP 3 soil moisture: 0% in air → 100% in water → 32% in freshly watered soil
- DP 109 humidity: tracks ambient (77% → 68% after pulling the probe out)
- DP 5 temperature, DP 15 battery, DP 106 dry flag (on at 0%, off at 32%): correct
- DP 112 write (soil sampling 300 s): confirmed, device echoed it back

Not tested: writes to DP 102 / 110 / 111.

So on current ZHA the second endpoint doesn't seem strictly required if humidity lives on the Tuya cluster instead of a RelativeHumidity cluster. Happy to test a revision on this device if useful.

PR #4537, opened 2025-11-30
Merge Tuya OnOff clusters and restore OnOff input cluster on TS004F

elupus:
## Proposed change

- Restore OnOff cluster on TS004F:
  These attributes are not available on the output side, only on the input side.
  
  This fixes warnings about direction being wrong on the onoff cluster, since
  the onoff cluster was removed on the input side.

- Restore standard name on the TuyaSmartRemoteOnOffCluster cluster.
  This is still a standard On/Off cluster, it just have some extensions.

- Ensure TuyaSmartRemoteOnOffCluster is an Eventable subclass of the attribute cluster. This is just an extension to allow for events

- Keep TuyaSmartRemoteOnOffCluster on input side for _TZ3000_famkxci2 since that uses attribute updates instead of commands to indicate button presses.

## Additional information

I have tested reading of the switch_mode attribute from this, which now works as expected.

Depends on: https://github.com/zigpy/zha/pull/591 to ensure the OnOff cluster does not end
up adding a switch entity.

Depends on: https://github.com/zigpy/zha/pull/590 to make sure this is added as a select entity

## Checklist


elupus:
@MattWestb Do you remember why you added this attributes to the output side of this cluster? The rotation events seem to come from the output cluster, but the attribute data is on the input cluster.

Feel like we should remove those attributes from the output cluster (TuyaSmartRemoteOnOffCluster)

elupus:
This looks like it should not be a complete OnOff cluster on input here, since not all attributes are supported on input and we end up creating a switch entity.

elupus:
This will depend on https://github.com/zigpy/zha/pull/590

MattWestb:
Hej du glade !!

Im not one code worrier and more "copy and paste / try and fail" but we was starting with the dimmer switch that  is only having the short and long press commands but it was one long road getting there.
The device is only having one EP then being paired and cant initiating the other 3. After sniffing we was finding that tuya is sending some attribute reads and the device is being initiated as one scene controller (tuya magic) (the EP1-4 is not visibly then init the device but is working so the quirk looks very strange).
Happy we was adding the light commands we was seen and also implanting the attribute for switching working more (also possible do on the device).
The was not getting group binding working and finding that tuya is doing it ad the device is one light and adding one (only the last is added s working) the device is sending light commands to light OK. The large problem was in scene mode the device need default response in wrong direction that was not possible doing in ZHA at that time > the device can waiting for DfRs but it timing out before sending sending next scene command (i think this is only one EP 2-4 and was working OK on EP1).

Next was the rotating knob that we was decoding and only adding the new commands it was sending and it was looking working OK but no deep diving so more then likely it can being in wrong direction like tuya oft have done.

In the end the deleting the On/Of in cluster so THA is not doing one light in the GUI.
First deleting the group in cluster but was finding its need it then dont support Zigbee binding and must being treated as one light.

I have one dimmer switch and one rotating knob so can doing testing and sniffing if you need more info for debugging the problem.

MattWestb:
By the way if you like doing some more deep fixing it shall being possible implanting the matter switch functionality for the scene commands its sending like IKEA was doing on Symfonisk 2 and sommrig and is exposed from Dirigera (the real matter commands was implanted on one manufacture cluster then it  was not  on the current matters spec) as switch  entrees with last status:
Symfonisk1
I also think (have not looking in the latest  matter (Z)Cluster Library) Matter shall have implanted rotary sensor that can being nice for the knob. 
But we shall knowing in one week then Bilresa is start selling in some regions (5 December in Sweden Austria perhaps little later for me).

Edit: Correcting my Sw gramma:  **Not** Bilresa (Eng Car trip) then its more then one device IKEA is releasing so its shall being Bilresor !!! (Eng Car tripos)

Edit2:
Not found rotary in "Matter Device Library Specification Version 1.4.2-mve-1" but keep looking !!

elupus:
Anyway you can test reading writing of the switch mode on any of the devices you have? And see if you get any data from that attribute when it's on output endpoint? I cant get that to work on any of the ts004f.

(I would like to remove it from the output cluster if possible)

MattWestb:
Was reading all attributes on Clusters:
```TuyaSmartRemoteOnOffCluster (Endpoint id: 1, Id: 0x0006, Type: out)``` 
and only ```cluster_revision (id: 0xfffd)``` was returning 1 and all other ```none```.
Then changing the "swish mode" on the device ZHA is logging:
```TS004F ROK Attribute Updated event was fired``` so something looks being broken in current ZHA.

Edit: Writing 0 or 1 say OK but no working mod not changed on the device and no attribute changed in ZHA log.
 => 2 is giving error = OK

**Comment by elupus** [comment]
Good then it looks like it works same as I see it. If you also add the onoff cluster or the attribute cluster on input side it works. (The events still must be on output side)

Maybe there was some compatibility that switches direction of a cluster command that was removed (the warning that is logged).

**Comment by MattWestb** [comment]
The TS004X devices is having the same defect  (and some other tuya devices) so i think they is also have lost  some functionality from ZHA but still working on the device then its possible doing it.

File: zhaquirks/tuya/ts004f.py [hunk 1]
```diff
@@ -57,17 +57,21 @@
     TURN_OFF,
     TURN_ON,
 )
+from zhaquirks.quirk_ids import TUYA_PLUG_ONOFF
```
MattWestb: I dont understand the logic of this naming then the tuya scene commands is only being used of scene remotes and no plugs what i knowing.

elupus: This may not be needed anymore. I think i originally used it to match some cluster handler.

Forgot this was not i draft. Will remove that change.

elupus: Actually.. Some way to mark the cluster having the "switch_mode" attribute is needed. The name of the quirk is bad thou.

The OnOff cluster have extensions. It's still a standard OnOff cluster on both client and server side.

Also I don't think we need to have TuyaSmartRemoteOnOffCluster and TuyaZBOnOffAttributeCluster. They are really only the server and the client side of same cluster. All other clusters are combined to describe both sides.

MattWestb: Most tuya quirk is in the pipe for being converted to V2 but its great that you is digging it it so some is knowing how the device is working and can getting the code catching it.
Sadly tuya is doing there "zigbee" implantation that is mostly OK but have making very bad shortcuts like on this devices is not supporting binding in the normal way (binding cluster to devices and / or groups) instead adding ther as lights that  is out of all standards.
And thanks for your work done !!

MattWestb: Most logic is renaming the `TUYA_PLUG_ONOFF` if its being used for other device types then plugs (the first use of the old cluster was for the TS004F and then _extended_ for the plugs and adding the back light and children lock).

MattWestb:
All TS004F devices is needed this for understanding the "scene commands".
https://github.com/zigpy/zha-device-handlers/blob/04c61bb5d47dd358dc00df71e1fc145f4edba4d8/zhaquirks/tuya/__init__.py#L1048-L1060
Plus the now not working attribute you is fixing here.
Only look what is needed then i remember the device is not sending commands in the right direction but if you can getting it working without the warnings its being great.

The quirks naming was from the start from the tuya is naming its devices that can being found in there dev docks but many have doing new quirks with other names / meanings.

elupus:
How about this change? I merged the two OnOff cluster's since it was just client side and server side of same extensions and regenerated the test cases in zha and it only render very minor changes for the testcases there.

Ps. The scene stuff there are just standard server commands sent same as any other levelcontrol/onoff, just a manufacture specific extension to the OnOff. Ie they are not doing anything strange with those. Can easily be handled as event entities once we have those changes in.

elupus:
This was tagged as needing 2.0 migration. But this touches old quirks. So does not seem to make sense to migrate in this step. (migration should be easier after since there are less custom things)

zhaquirks/tuya/ts011f_plug.py
```diff
@@ -90,7 +90,7 @@ class Plug(EnchantedDevice):
                     Identify.cluster_id,
                     Groups.cluster_id,
                     Scenes.cluster_id,
-                    TuyaZBOnOffAttributeCluster,
+                    TuyaOnOffCluster,
```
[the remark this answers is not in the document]
elupus: These events have been removed from server side.

elupus:
Okey.. Need to ponder some things:

- _TZ3000_famkxci2 is funky. I exposes only input clusters for OnOff yet claim to be a remote. I wonder if we can move that cluster over to output side without breaking stuff.
- The device database information does not contain information on unsupported attributes. So the normal filter on input side for supported attributes can't filter out the switches in the testdata.

elupus:
I've dropped the change on _TZ3000_famkxci2 since that uses attribute update values to event instead of normal press events on the output side.

PR #4552, opened 2025-12-06
Add Inovelli VZM32 MMwave Presence Sensing Switch attributes

InovelliUSA:
## Checklist

### DIFF
--- a/zhaquirks/inovelli/__init__.py
+++ b/zhaquirks/inovelli/__init__.py
@@ -1277,6 +1277,105 @@ class InovelliVZM32SNCluster(InovelliCluster):
         )
 
 
+class InovelliVZM32SNMMWaveCluster(CustomCluster):
+    """Inovelli VZM32-SN MMWave custom cluster."""
+
+    cluster_id = 0xFC32
+    ep_attribute = "inovelli_vzm32snmmwave_cluster"
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        mmwave_z_min = ZCLAttributeDef(
+            id=0x0065,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
+        mmwave_z_max = ZCLAttributeDef(
+            id=0x0066,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
+        mmwave_x_min = ZCLAttributeDef(
+            id=0x0067,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
+        mmwave_x_max = ZCLAttributeDef(
+            id=0x0068,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
+        mmwave_y_min = ZCLAttributeDef(
+            id=0x0069,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
+        mmwave_y_max = ZCLAttributeDef(
+            id=0x006A,
+            type=t.int16s,
+            is_manufacturer_specific=True,
+        )
+        mmwave_target_info_report = ZCLAttributeDef(
+            id=0x006B,
+            type=t.uint8_t,
+            is_manufacturer_specific=True,
+        )
+        mmwave_stay_life = ZCLAttributeDef(
+            id=0x006C,
+            type=t.uint32_t,
+            is_manufacturer_specific=True,
+        )
+        mmwave_detect_sensitivity = ZCLAttributeDef(
+            id=0x0070,
+            type=t.uint8_t,
+            is_manufacturer_specific=True,
+        )
+        mmwave_detect_trigger = ZCLAttributeDef(
+            id=0x0071,
+            type=t.uint8_t,
+            is_manufacturer_specific=True,
+        )
+        mmwave_hold_time = ZCLAttributeDef(
+            id=0x0072,
+            type=t.uint32_t,
+            is_manufacturer_specific=True,
+        )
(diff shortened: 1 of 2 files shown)


InovelliUSA:
@TheJulianJES is there anything else I need to do to this PR to get it merged?

TheJulianJES:
I don't think so. I'll take a proper look at this and other PRs when I'm back. From a quick glance, this looks good.

kenspangler:
@TheJulianJES is there an estimated time this change will be merged and in the production build.  I am installing these switches now and will hold off installing the custom quirks if this will be soon.  Otherwise, I will continue with the custom quirks. Thanks.

zhaquirks/inovelli/__init__.py
```diff
@@ -1277,6 +1277,105 @@ class AttributeDefs(InovelliCluster.AttributeDefs):
...
+            Union[t.Addressing.Group, t.Addressing.IEEE, t.Addressing.NWK]
+        ] = None,
+    ):
+        """Handle a cluster request."""
+        _LOGGER.debug(
+            "%s: handle_cluster_request - Command: %s Data: %s",
+            self.name,
+            hdr.command_id,
+            args,
+        )
```
puddly: This can be removed, since this code does nothing:
```suggestion
```

zhaquirks/inovelli/__init__.py
```diff
@@ -1277,6 +1277,105 @@ class AttributeDefs(InovelliCluster.AttributeDefs):
...
+            is_manufacturer_specific=True,
+        )
+
+    class ServerCommandDefs(BaseCommandDefs):
+        """Server command definitions."""
+
+        mmwave_control_command = ZCLCommandDef(
+            id=0x00,
+            schema={
+                "control_id": t.uint8_t,
```
puddly: Can `control_id` be turn into a proper `t.enum8` subclass with named commands?

InovelliUSA: @puddly I have made the requested changes.

TheJulianJES: If the device expects an `uint8` when that attribute is written, you'll likely also need to add `zcl_type` to the attribute definition now. Otherwise, it'll use `enum8`. See this example:
https://github.com/zigpy/zha-device-handlers/blob/9b5774d6cd76e624b377fcba4200c6b4e39a4889/zhaquirks/nodon/pilot_wire.py#L41-L42

InovelliUSA: Hmmm, I would think that too, but the current implementation does work:

image

I have tested it and it seems to be working. The command to reset the mmwave to default flashes the LED red and the one to set the interference area flashes the LED orange so I can see that it is working. I can still change it though to be sure but will have to do it later.

TheJulianJES: Sorry, I didn't read properly. What I wrote doesn't apply to commands anyway, only attributes. So, ignore 😄

zhaquirks/inovelli/__init__.py
```diff
@@ -1277,6 +1277,97 @@ class AttributeDefs(InovelliCluster.AttributeDefs):
         )
 
 
+class MMWaveControlId(t.enum8):
+    """MMWave control command IDs."""
+
+    RESET_TO_FACTORY = 0x00
+    AUTO_GENERATE_INTERFERENCE_AREA = 0x01
+    OBTAIN_AREAS = 0x02
+    CLEAR_INTERFERENCE_AREA = 0x03
```
TheJulianJES: We're unfortunately not really consistent on this yet, but I think we want a different casing for this.
(@puddly? 😄 )

1.:
```suggestion
    Reset_to_factory = 0x00
    Auto_generate_interference_area = 0x01
    Obtain_areas = 0x02
    Clear_interference_area = 0x03
```
or maybe 2.:
```suggestion
    Reset_To_Factory = 0x00
    Auto_Generate_Interference_Area = 0x01
    Obtain_Areas = 0x02
    Clear_Interference_Area = 0x03
```

puddly: I think we use 1. in most places

InovelliUSA: I have updated the case to mach #1.

wingrunr21:
@TheJulianJES can this PR get merged? it's coming up on being open for 2 months now

TheJulianJES (approved):
Thanks. Sorry about the delay. I was busy with a lot of other things.
We'll still need to support this in ZHA though. That won't be in 2026.3.0 unfortunately.

Inovelli devices are a bit special because they still have custom entities defined in ZHA itself. They can be moved to quirks but the `unique_id_suffix` needs be correctly set to match old entities from ZHA. IIRC it's `clusterId-attributeId`.

We'll look at either merging the ZHA PR until we move Inovelli entities directly to quirks, or directly do that with the new entities.

InovelliUSA:
> Thanks. Sorry about the delay. I was busy with a lot of other things. We'll still need to support this in ZHA though. That won't be in 2026.3.0 unfortunately.
> 
> Inovelli devices are a bit special because they still have custom entities defined in ZHA itself. They can be moved to quirks but the `unique_id_suffix` needs be correctly set to match old entities from ZHA. IIRC it's `clusterId-attributeId`.
> 
> We'll look at either merging the ZHA PR until we move Inovelli entities directly to quirks, or directly do that with the new entities.

Thanks, I was hoping we could move the entities to the v2 quirk but hopefully the ZHA change can be merged before that work gets started. :)

PR #4568, opened 2025-12-13
Add iHseno TS0601 human presence sensor (_TZE284_debczeci, _TZE204_debczeci, _TZE284_1lvln0x6)

a-gal:
…zeci, _TZE284_1lvln0x6)

## Proposed change

This adds support for a battery powered iHseno Tuya Human Presence Sensor

It's a port of [this](https://github.com/Koenkk/zigbee-herdsman-converters/blob/069547d3c85190d4831b52b5460975041559b3c1/src/devices/tuya.ts#L6020) z2m converter.

## Device diagnostics

[zha-01JYEQ11XP51H3NKH2CQYGCT58-_TZE284_debczeci TS0601-8ca1e243d10bc471fa8384d5e4306923.json](https://github.com/user-attachments/files/24141913/zha-01JYEQ11XP51H3NKH2CQYGCT58-_TZE284_debczeci.TS0601-8ca1e243d10bc471fa8384d5e4306923.json)

## Checklist

### DIFF
--- a/zhaquirks/tuya/tuya_motion.py
+++ b/zhaquirks/tuya/tuya_motion.py
@@ -170,6 +170,14 @@ class TuyaMotionFadeTime(t.enum8):
     _120_seconds = 0x03
 
 
+class TuyaMotionFadeTime3Step(t.enum8):
+    """Tuya motion 3 step fade time enum."""
+
+    _15_seconds = 0x00
+    _30_seconds = 0x01
+    _60_seconds = 0x02
+
+
 class TuyaSensitivityMode(t.enum8):
     """Tuya sensitivity mode enum."""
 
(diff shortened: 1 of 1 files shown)


BCSugoi:
I can confirm this is working for a "_TZE284_debczeci", but with one modification:  the fading time has only three possible settings, 15s, 30s and 60s, rather than the 4 steps (10, 30, 60, 120) in the class TuyaMotionFadeTime.  So I added a new class with three steps as shown below, and changed the reference in the code to this one:  

`class TuyaMotionFadeTime3step(t.enum8):
    """Tuya motion fade time enum."""

    _15_seconds = 0x00
    _30_seconds = 0x01
    _60_seconds = 0x02`

Also, I'm not sure if we need the "battery_qty=2" entry?   Mine works the same whether I include that or not.

The device is working well, although the fade time seems to be a little unpredictable, eg. if it is set at 15 seconds it often takes 30 to 45 seconds.  This behavior was tested using a Tuya hub and was the same as in ZHA, so it seems to be the device, not the quirk

For reference, I bought my device on Amazon here:
https://www.amazon.com/IHSENO-Presence-Detector-Automation-Required/dp/B0FHBB4D84
And found it on the manufacturer's website here:
http://ihseno.com/product/html/?195.html

And here is a photo of the inside of the device
iHseno_TZE284_debczeci

a-gal:
> 15s, 30s and 60s, rather than the 4 steps (10, 30, 60, 120) in the class TuyaMotionFadeTime. So I added a new class with three steps as shown below, and changed the reference in the code to this one:

Good to know, I wasn't sure about fourth value. I'll fix this.

> Also, I'm not sure if we need the "battery_qty=2" entry? Mine works the same whether I include that or not.

It's just an informational value, AFAIK, it doesn't do anything but define "Battery quantity" displayed in properties.

Issue #3857, opened 2025-02-12
[BUG] Innr SP240 newest firmware not compatible anymore with the quirk


Skaars-fr:
### Bug description

It seems that newest firmware behavior changed

With the current zha quirk and the newest firmware the reported consumption of the plug is far too high
The 1.7 firmware was working perfectly but the 1.9 firmware introduced a change in the divisors and multipliers behaviour.

https://github.com/Koenkk/zigbee-herdsman-converters/pull/8715#issuecomment-2640979923

### Steps to reproduce

1 update Innr SP240 to 1.9.30 (or 1.9.28)
2 check the energy consumption

### Expected behavior

According to what was reported on z2m github space, the new value is 10 times higher than the expected value

https://github.com/Koenkk/zigbee-herdsman-converters/issues/8712

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Device signature

<details><summary>Device signature</summary>

```json

[{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4454,
    "maximum_buffer_size": 74,
    "maximum_incoming_transfer_size": 80,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 80,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x010a",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0702",
        "0x0b04",
        "0x1000",
        "0xe001"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "innr",
  "model": "SP 240",
  "class": "zhaquirks.innr.innr_sp240_plug.SP240"
}]

```

</details>

### Diagnostic information

[...]

Issue #2781, opened 2023-11-28
[BUG] INNR SP 240 "summation delivered" kWh value is 100x expected value


rrozema:
### Bug description

Upon inclusion of the INNR SP 240 in ZHA, among others a "Summation delivered" sensor gets created. The value returned for this sensor is however 100x larger than expected: I ran a filament dryer for 5 hours of the SP 240 and the device measured 17kWh instead of the expected 0,17 kWh as measured using a Fibaro wall plug V2.

When I requested help from INNR support they responded that ZHA probably has incorrectly interpreted the configuration their device returned. My translation of their response: "An attribute "SummationFormatting" on the Metering cluster has the value 0x2A, which designates: 2 digits to the right of the decimal point, 5 digits to the left of the decimal point, don't suppress leading zeroes". Looks like his system (red: my ZHA integration in HA)) doesn't interpret or incorrectly interprets the value of the "SummationFormatting" attribute." 

The full mail text I got from INNR support (in Dutch, sorry):

> Beste Richard,
> 
>  
> 
> (This message is in Dutch because I assume based on your name and e-mail address that you also speak Dutch.)
> 
> Ik heb de vraag even bij één van onze engineers gelegd, en dit was zijn reactie:
> 
> Wat betreft Watt meting die op 0 decimalen wordt afgerond:
> 
> De waarde wordt opgeslagen en gecommuniceerd in een “int16” type, wat een bereik heeft van -32768 tot 32767. Als we op 1 decimaal zouden afronden zou het maximum slechts 3276.7 W zijn, maar de plug kan 3680 W aan dus de keuze was simpel. De afronding dynamisch maken (meer decimalen bij lagere waarden) is geen optie want dat ondersteunen de meeste systemen niet (voor zover ik weet geen enkel systeem).
> 
>  
> 
> Wat betreft de kWh formatting: het “SummationFormatting” attribuut van het Metering cluster heeft waarde 0x2A, wat betekent: 2 digits to the right of the decimal point, 5 digits to the left of the decimal point, don’t suppress leading zeroes. Lijkt er dus op dat zijn systeem de waarde van het “SummationFormatting” attribuut van het Metering cluster niet of verkeerd interpreteert.
> 
>  
> 
> Hopelijk heb je iets aan deze uitleg.
> 
>  
> 
> Met vriendelijke groet,
> 
>  
> 
>  
> 
> **REDACTED**
> 
> Innr Servicedesk 
[...]

PR #4592, opened 2025-12-21
Fix custom divisor for Innr SP 240 on `0x191E3685`

TheJulianJES:
## Proposed change

This quirk removes changing the divisor for Innr SP 240 plugs on firmware `0x191E3685` and later.

As old divisor values from the quirk may have persisted into zigpy cache, a quirk for newer versions is also added to force the correct `divisor: 1000`. This avoids users needing to re-pair devices to fix the issue and also allows OTA upgrades without re-pairing the device (just needing a ZHA reload or HA restart).

The config entities for the `LevelControl` cluster are also hidden.

## Additional information

Tested with the SP 240 (on a much newer version than the latest OTA).

Fixes:
- Addresses https://github.com/zigpy/zha-device-handlers/issues/3857
- Addresses https://github.com/zigpy/zha-device-handlers/issues/2781

## Device diagnostics

Freshly paired without quirk:
[Innr SP 240 without quirk zha-01KAH6BZ8GBKBJXCPX8V4BB4KP-innr SP 240-40d6d753fef0dfeafea5ce24ed764473.json](https://github.com/user-attachments/files/24413444/Innr.SP.240.without.quirk.zha-01KAH6BZ8GBKBJXCPX8V4BB4KP-innr.SP.240-40d6d753fef0dfeafea5ce24ed764473.json)

## Checklist

### DIFF
--- a/zhaquirks/innr/__init__.py
+++ b/zhaquirks/innr/__init__.py
@@ -7,15 +7,29 @@ from zigpy.zcl.clusters.smartenergy import Metering
 INNR = "innr"
 
 
-class MeteringClusterInnr(CustomCluster, Metering):
-    """Provide constant multiplier and divisor.
+class MeteringClusterInnrOld(CustomCluster, Metering):
+    """Provide constant multiplier and divisor for old Innr plug firmware.
 
-    The device seems to supply the summation_formatting attribute correctly, but ZHA doesn't use it for kWh for now.
+    Old firmware provides incorrect values for the divisor, so we override them.
     """
 
     _CONSTANT_ATTRIBUTES = {
         Metering.AttributeDefs.multiplier.id: 1,
         Metering.AttributeDefs.divisor.id: 100,
+    }
+
+
+class MeteringClusterInnrNew(CustomCluster, Metering):
+    """Provide constant multiplier and divisor for new Innr plug firmware.
+
+    New firmware provides already provides correct value, but the old quirk will have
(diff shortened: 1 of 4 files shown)


TheJulianJES:
Hmm, this does not work. The quirk still always applies. Only if `allow_missing=False` is added will the quirk (never) apply. Will take a look.. 
I hope we just load everything early enough for the filter to even work at all.

TheJulianJES:
Yeah... It seems like we only load attributes **after** we do the software version check. So, it'll never work.

TheJulianJES:
This correctly works with https://github.com/zigpy/zigpy/pull/1735 now. The remaining related "issues" are:
- The other plugs should also be fixed. Waiting on feedback here: https://github.com/zigpy/zha-device-handlers/issues/3857#issuecomment-3700400786
- OTA `current_file_version` isn't read early enough when pairing, so we default to the new quirk now (divisor: 1000).
  If a device on an old version is paired, the divisor will fix itself after restarting. With https://github.com/zigpy/zigpy/pull/1631, this should be fixed entirely.
- We should eventually also allow for "device reconfiguration" to re-read a bunch of attributes only read on initial pairing.
- When updating the device, ZHA needs to be reloaded once (or HA restarted) in order to change to the new quirk.

Except for fixing the other Innr plug models, everything else needs to be addressed in zigpy/ZHA. This PR should be good to go. We can fix other the other Innr plug in another PR.

PR #4613, opened 2025-12-31
Add COMMAND_MOVE_TO_LEVEL constant

MattWestb:
COMMAND_MOVE_TO_LEVEL is not implemented and is needed for some strange (new IKEA controller). Then its in place we can letting users trying local quirk easier and if its like the limited functionality we do one PR for the controller.

## Checklist


MattWestb:
Hey @TheJulianJES I was merger to the updated branch and its looks it some strange with python 3.12.13 `Error: Failed to restore cache entry`.

Is it any that blocking adding this "cosmetic" but very needed constant for getting our device automatons working with local quirks and testing for adding then later ??

PR #4646, opened 2026-01-11
Add CentraLite 3400-G keypad quirk with battery fix and motion sensor

yougotborked:
## Proposed change

This PR adds a v2 quirk for the CentraLite 3400-G Zigbee keypad that fixes battery reporting and adds virtual motion sensor support.

**Battery Fix:**
- Sets correct voltage range (2.1V-3.0V) for CR123A batteries
- ZHA automatically calculates battery percentage from voltage readings using this range

**Virtual Motion Sensor:**
- The keypad has an internal motion detector but doesn't report motion via standard Zigbee clusters
- When motion is detected, the keypad wakes and sends an IAS ACE GetPanelStatus command (0x07)
- The quirk intercepts this command and triggers a virtual motion sensor on endpoint 2
- Motion automatically clears after 90 seconds
- This matches the behavior implemented in Hubitat's driver for this device

**Implementation:**
- Uses v2 quirk format with `CustomDeviceV2` and `QuirkBuilder`
- Custom device class initializes the Bus system for motion events
- Creates virtual endpoint 2 for the motion sensor
- Preserves IAS Zone on endpoint 1 for tamper detection

**Preserved Features:**
- IAS Zone on endpoint 1 for tamper detection (disabled by default)
- Temperature sensor
- All standard keypad functionality

## Additional information

This quirk has been tested on 3 CentraLite 3400-G devices and confirmed working:
- Battery reporting works correctly
- Motion sensor triggers when walking in front of keypad
- Temperature and tamper detection preserved

No breaking changes. Uses v2 quirk format as recommended, with `CustomDeviceV2` for custom device initialization.

## Device diagnostics
[zha-f8e0b2aa77fdfad9a390fd0edf3af072-CentraLite 3400-G-ec6168d145797b43829512cc58517ee8.json](https://github.com/user-attachments/files/24554822/zha-f8e0b2aa77fdfad9a390fd0edf3af072-CentraLite.3400-G-ec6168d145797b43829512cc58517ee8.json)

## Checklist

### DIFF
--- a/zhaquirks/centralite/cl_3400_G.py
+++ b/zhaquirks/centralite/cl_3400_G.py
@@ -1,0 +1,75 @@
+"""Device handler for CentraLite 3400-G."""
+
+from zigpy.profiles import zha
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import CustomDeviceV2, QuirkBuilder
+import zigpy.types as t
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.security import IasAce, IasZone
+
+from zhaquirks import Bus, LocalDataCluster, MotionOnEvent, PowerConfigurationCluster
+from zhaquirks.const import MOTION_EVENT
+
+CENTRALITE = "CentraLite"
+ZONE_TYPE = 0x0001
+
+
+class CustomPowerConfigurationCluster(PowerConfigurationCluster):
+    """Custom PowerConfigurationCluster for CR123A batteries."""
+
+    MIN_VOLTS = 2.1
+    MAX_VOLTS = 3.0
+
+
+class MotionCluster(LocalDataCluster, MotionOnEvent, IasZone):
+    """Virtual motion sensor cluster."""
+
+    # Force zone type to Motion Sensor
+    _CONSTANT_ATTRIBUTES = {ZONE_TYPE: IasZone.ZoneType.Motion_Sensor}
+
+    # Reset motion after 90 seconds
+    reset_s = 90
+
+    # Skip configuration for virtual cluster
+    SKIP_CONFIGURATION = True
+
+
+class CustomIasAce(CustomCluster, IasAce):
+    """Custom IAS ACE cluster to detect GetPanelStatus as motion trigger."""
+
+    def handle_cluster_request(
+        self,
+        hdr: foundation.ZCLHeader,
+        args: list,
+        *,
+        dst_addressing: t.Addressing.Group
+        | t.Addressing.IEEE
+        | t.Addressing.NWK
+        | None = None,
+    ):
+        """Handle cluster request - detect GetPanelStatus (0x07) as motion event."""
+        # GetPanelStatus command - keypad sends this when it wakes up from motion
+        if hdr.command_id == IasAce.ServerCommandDefs.get_panel_status.id:
+            self.endpoint.device.motion_bus.listener_event(MOTION_EVENT)
+
+        return super().handle_cluster_request(hdr, args, dst_addressing=dst_addressing)
+
+
+class CentraLiteDeviceV2(CustomDeviceV2):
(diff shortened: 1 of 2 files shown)


zhaquirks/centralite/cl_3400_G.py
```diff
@@ -0,0 +1,163 @@
...
+            for record in records
+            if record.attrid != 0x0021  # battery_percentage_remaining
+        ]
+
+        if filtered_records:
+            result = await super().configure_reporting(
+                filtered_records, *args, **kwargs
+            )
+        else:
+            result = foundation.ConfigureReportingResponse.deserialize(b"\x00")[0]
```
puddly: If the attribute is not supported, the device responds as such. Why do we need to manually filter it out with this `configure_reporting` override?

zhaquirks/centralite/cl_3400_G.py
```diff
@@ -0,0 +1,163 @@
...
+        args: list,
+        *,
+        dst_addressing: t.Addressing.Group
+        | t.Addressing.IEEE
+        | t.Addressing.NWK
+        | None = None,
+    ):
+        """Handle cluster request - detect GetPanelStatus (0x07) as motion event."""
+        # Command 0x07 is GetPanelStatus - keypad sends this when it wakes up from motion
+        if hdr.command_id == 0x07:
```
puddly: `0x07` should use be a constant: `IASZone.ServerCommandDefs.get_panel_status.id`

zhaquirks/centralite/cl_3400_G.py
```diff
@@ -0,0 +1,163 @@
...
+            self.endpoint.device.motion_bus.listener_event(MOTION_EVENT)
+
+        return super().handle_cluster_request(hdr, args, dst_addressing=dst_addressing)
+
+
+class CentraLite3400G(CustomDevice):
```
puddly: `CustomDevice` and v1 quirks are being deprecated in favor of v2 quirks, which use `CustomDeviceV2` when you need to modify the device object. You can find this as an example: https://github.com/zigpy/zha-device-handlers/blob/42c59828ad67457f90a8a2d8aa6c6f318f0f2db7/zhaquirks/develco/emi_led.py#L39

zhaquirks/centralite/cl_3400_G.py
yougotborked: I have removed this, was left during debugging

zhaquirks/centralite/cl_3400_G.py
yougotborked: done

zhaquirks/centralite/cl_3400_G.py
yougotborked: I have switched to the v2 system, Thanks @puddly for all your reviews

zhaquirks/centralite/cl_3400_G.py
```diff
@@ -0,0 +1,85 @@
...
+        return super().handle_cluster_request(hdr, args, dst_addressing=dst_addressing)
+
+
+class CentraLiteDeviceV2(CustomDeviceV2):
+    """Custom device class for CentraLite 3400-G with motion bus."""
+
+    def __init__(self, *args, **kwargs):
+        """Init device with motion bus."""
+        self.motion_bus = Bus()
+
```
puddly: Do you need to add this virtual cluster to a new endpoint? If not, just add it to endpoint 1. If you do need to add it, there is an existing Quirks V2 API to do this. You don't need to manually touch the device object.

puddly (changes requested):

zhaquirks/centralite/cl_3400_G.py
yougotborked: Done, I see that now, adds_endpoint

PR #4649, opened 2026-01-13
Fix zigpy typing issues with Tuya and Aqara

puddly:
## Proposed change

A few quirks used zigpy types in creative ways that are not actually valid. This PR corrects these errors in manner both backward- and forward- compatible with current zigpy types.

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/tvoc.py
+++ b/zhaquirks/xiaomi/aqara/tvoc.py
@@ -71,7 +71,7 @@ class EmulatedTVOCMeasurement(LocalDataCluster):
         return result
 
 
-class TVOCDisplayUnit(t.enum_factory(t.uint8_t)):
+class TVOCDisplayUnit(t.enum8):
     """Display values."""
 
     mgm3_celsius = 0x00
--- a/zhaquirks/xiaomi/aqara/tvoc.py
+++ b/zhaquirks/xiaomi/aqara/tvoc.py
@@ -8,7 +8,7 @@ import zigpy.types as t
 from zigpy.zcl.clusters.general import AnalogInput, Basic, Identify, Ota
 from zigpy.zcl.clusters.measurement import RelativeHumidity, TemperatureMeasurement
 from zigpy.zcl.clusters.security import IasZone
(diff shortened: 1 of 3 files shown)


zhaquirks/xiaomi/aqara/tvoc.py
```diff
@@ -71,7 +71,7 @@ async def bind(self):
         return result
 
 
-class TVOCDisplayUnit(t.enum_factory(t.uint8_t)):
+class TVOCDisplayUnit(t.enum8):
```
TheJulianJES, line 74: We'll need to adjust the `zcl_type` to `uint8` for this attribute then.
Using an enum type here causes issues, as the device expects an int type: https://github.com/zigpy/zha-device-handlers/pull/2408#discussion_r1237194131

puddly: Ah, good point. Fixed.

TheJulianJES (approved):
Looks good

puddly:
The one Tuya test broken by this change I think exposes an underlying issue. Two test cases fail due to parsing errors being silently swallowed via `Error calling listener` logs, this being one of them:

```diff
diff --git a/tests/conftest.py b/tests/conftest.py
index 52e94865..b1facd47 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -334,3 +334,25 @@ def raise_on_bad_log_formatting():
         yield
     finally:
         root.removeHandler(handler)
+
+
+class FailOnBadListenerHandler(logging.Handler):
+    def emit(self, record):
+        msg = record.msg % record.args
+
+        if msg.startswith("Error calling listener"):
+            pytest.fail(msg)
+
+
+@pytest.fixture(autouse=True)
+def raise_on_bad_listener():
+    handler = FailOnBadListenerHandler()
+
+    root = logging.getLogger()
+    root.addHandler(handler)
+    root.setLevel(logging.DEBUG)
+
+    try:
+        yield
+    finally:
+        root.removeHandler(handler)
\ No newline at end of file
```

We incremented the TSN before the parsing error so the slight reorganization caused the increment to happen later.

PR #4655, opened 2026-01-15
Add Third Reality 60g radar VOC level

3reality-support:
## Proposed change

This PR added adaptation code for 60g radar and added 1 additional private clusters
volatile_organic_compounds：Measure the value of VOC
## Checklist

### DIFF
--- a/zhaquirks/thirdreality/60g_radar.py
+++ b/zhaquirks/thirdreality/60g_radar.py
@@ -1,0 +1,40 @@
+"""Third Reality formaldehyde sensor devices."""
+
+from typing import Final
+
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import QuirkBuilder, SensorDeviceClass, SensorStateClass
+from zigpy.quirks.v2.homeassistant import CONCENTRATION_PARTS_PER_BILLION
+import zigpy.types as t
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+
+class ThirdRealityRadarCluster(CustomCluster):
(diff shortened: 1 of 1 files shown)


zhaquirks/thirdreality/60g_radar.py
```diff
@@ -0,0 +1,41 @@
...
+    QuirkBuilder("Third Reality, Inc", "3RPL01084Z")
+    .replaces(ThirdRealityRadarCluster)
+    .sensor(
+        endpoint_id=1,
+        attribute_name=ThirdRealityRadarCluster.AttributeDefs.volatile_organic_compounds.name,
+        cluster_id=ThirdRealityRadarCluster.cluster_id,
+        device_class=SensorDeviceClass.VOLATILE_ORGANIC_COMPOUNDS_PARTS,
+        state_class=SensorStateClass.MEASUREMENT,
+        unit=CONCENTRATION_PARTS_PER_BILLION,
+        fallback_name="Volatile organic compounds",
```
TheJulianJES: My change doesn't make any real difference at all here, as Home Assistant uses the device class for the entity name (as no `translation_key` is defined, which is good): I'll just change the `fallback_name` to match the HA device class name (which includes "parts" at the end). Source here [sensor/strings.json](https://github.com/home-assistant/core/blob/57c49d0c4898df8ae19e455790394e281babf5fe/homeassistant/components/sensor/strings.json#L303-L305).
```suggestion
        fallback_name="Volatile organic compounds parts",
```

TheJulianJES (approved):
This looks perfect. Thank you!

PR #4658, opened 2026-01-15
Migrate existing quirks to new zigpy attribute events and `manufacturer_code`

puddly:
## Proposed change

Implements https://github.com/zigpy/zigpy/pull/1719 for zha-device-handlers.

The above PR implements a few backwards compatibility tweaks that allow many existing quirks to continue working without any changes but we should migrate our quirks repo to use the correct subscription events.

The most notable change is the introduction of dedicated events for attribute change subscription (with different events for each type) and a modification of `_attr_cache`. The attribute cache now has dedicated interfaces for real/virtual attributes but has backwards compatible APIs to automatically keep working as before.

## Checklist

### DIFF
--- a/tests/test_zunzunbee_slateswitch.py
+++ b/tests/test_zunzunbee_slateswitch.py
@@ -6,6 +6,7 @@ import pytest
 import zigpy.types as t
 from zigpy.zcl.clusters.security import IasZone
 
+from tests.common import ClusterListener
 import zhaquirks
 
 zhaquirks.setup()
@@ -16,8 +17,9 @@ async def test_button_ias(zigpy_device_from_v2_quirk):
     device = zigpy_device_from_v2_quirk("zunzunbee", "SSWZ8T")
     ias_zone_status_attr_id = IasZone.AttributeDefs.zone_status.id
     cluster = device.endpoints[1].ias_zone
-    listener = mock.MagicMock()
-    cluster.add_listener(listener)
+    cluster_listener = ClusterListener(cluster)
+    zha_listener = mock.MagicMock()
+    cluster.add_listener(zha_listener)
 
     # Define button press values (hex + long press variants)
     button_values = [
@@ -39,13 +41,12 @@ async def test_button_ias(zigpy_device_from_v2_quirk):
         0x101,
     ]
 
-    for value in button_values:
+    for i, value in enumerate(button_values):
         cluster.update_attribute(ias_zone_status_attr_id, value)
-        assert listener.attribute_updated.call_args[0][0] == ias_zone_status_attr_id
-        assert listener.attribute_updated.call_args[0][1] == value
+        assert cluster_listener.attribute_updates[i] == (ias_zone_status_attr_id, value)
 
-    assert listener.attribute_updated.call_count == len(button_values)
-    assert listener.zha_send_event.call_count == len(button_values)
+    assert len(cluster_listener.attribute_updates) == len(button_values)
+    assert zha_listener.zha_send_event.call_count == len(button_values)
 
 
 @pytest.mark.parametrize(
@@ -161,14 +162,17 @@ async def test_discard_invalid_value(zigpy_device_from_v2_quirk):
     """Test that invalid values are discarded without triggering events."""
     device = zigpy_device_from_v2_quirk("zunzunbee", "SSWZ8T")
     cluster = device.endpoints[1].ias_zone
-    listener = mock.MagicMock()
-    cluster.add_listener(listener)
+    cluster_listener = ClusterListener(cluster)
+    zha_listener = mock.MagicMock()
+    cluster.add_listener(zha_listener)
 
     ias_zone_status_attr_id = IasZone.AttributeDefs.zone_status.id
     invalid_value = 6  # example invalid value
     cluster.update_attribute(ias_zone_status_attr_id, invalid_value)
 
-    assert listener.attribute_updated.call_args[0][0] == ias_zone_status_attr_id
-    assert listener.attribute_updated.call_args[0][1] == invalid_value
+    assert cluster_listener.attribute_updates[0] == (
+        ias_zone_status_attr_id,
+        invalid_value,
+    )
 
-    listener.zha_send_event.assert_not_called()
+    assert len(zha_listener.zha_send_event.mock_calls) == 0
--- a/zhaquirks/develco/emi_han.py
+++ b/zhaquirks/develco/emi_han.py
@@ -2,11 +2,10 @@
 
 from __future__ import annotations
 
-from typing import Any
-
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.v2 import QuirkBuilder
 import zigpy.types as t
+from zigpy.zcl import foundation
 from zigpy.zcl.clusters.smartenergy import Metering
 
 
@@ -16,12 +15,30 @@ class FrientMetering(CustomCluster, Metering):
     # fix device issue
     _CONSTANT_ATTRIBUTES = {Metering.AttributeDefs.divisor.id: 1000}
 
-    def _update_attribute(self, attrid: int | t.uint16_t, value: Any) -> None:
-        """Update attribute with value."""
-        # prevent attribute_updated events for divisor
-        if attrid == Metering.AttributeDefs.divisor.id:
-            return
-        super()._update_attribute(attrid, value)
+    def handle_cluster_general_request(
+        self,
+        hdr: foundation.ZCLHeader,
+        args: list,
+        *,
+        dst_addressing: t.Addressing.Group
+        | t.Addressing.IEEE
+        | t.Addressing.NWK
+        | None = None,
+    ) -> None:
+        """Filter out incorrect divisor attribute reports from device."""
+        if hdr.command_id == foundation.GeneralCommand.Report_Attributes:
+            # Filter out divisor attribute reports
+            args.attribute_reports = [
+                attr
+                for attr in args.attribute_reports
+                if attr.attrid != Metering.AttributeDefs.divisor.id
+            ]
+
+            # Don't process if no attributes remain
+            if not args.attribute_reports:
+                return
+
+        super().handle_cluster_general_request(hdr, args, dst_addressing=dst_addressing)
 
 
 (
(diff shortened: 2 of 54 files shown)


zhaquirks/xiaomi/aqara/feeder_acn001.py
```diff
@@ -82,62 +83,86 @@
...
-            self._parse_feeder_attribute(value)
+        # Set default values for attributes
+        self._update_attribute(ZCL_DISABLE_LED_INDICATOR, False)
+        self._update_attribute(ZCL_CHILD_LOCK, False)
+        self._update_attribute(ZCL_FEEDING_MODE, FeedingMode.Manual)
+        self._update_attribute(ZCL_SERVING_SIZE, 1)
+        self._update_attribute(ZCL_PORTION_WEIGHT, 8)
+        self._update_attribute(ZCL_ERROR_DETECTED, False)
+        self._update_attribute(ZCL_PORTIONS_DISPENSED, 0)
+        self._update_attribute(ZCL_WEIGHT_DISPENSED, 0)
```
TheJulianJES: This won't reset the attributes to these default values when restarting HA?

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
TheJulianJES: I'l need to look at the zigpy PR first, so this might be a dumb question, but we need to override both now?
Would listening to `AttributeReportedEvent` + `AttributeReadEvent` also work? (but probably not `AttributeUpdatedEvent`, since we produce that ourselves here by calling `update_attribute`)?

TheJulianJES (commented):
Two questions for my understanding 😄 
Will take a look at the zigpy PR.

zhaquirks/xiaomi/aqara/feeder_acn001.py
puddly: Good point. I guess the purpose of the direct `_attr_cache` access is to set default values for the created entities?

TheJulianJES: Yeah, I think so. This Aqara device is somewhat similar to Tuya devices, uses a virtual cluster and so on.

In other places, we use this to set default values: https://github.com/puddly/zha-device-handlers/blob/e154e667c6d9b0e8e29c4901a2ce21315dee47ed/zhaquirks/xiaomi/__init__.py#L697-L703

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
puddly: This I think was a rebasing issue, sorry about that. I'll take a look at this Aqara quirk, it seems to be the most problematic one and may be the best case study for some new APIs.

TheJulianJES: I think the inversion stuff here and doubling battery percentage is something we could easily deal with by introducing a "exposed feature/hint" in quirks and then have a separate ZHA entity that matches them, with it doing the inversion or doubling the battery percentage, no longer affecting the underlying ZCL cache.

puddly: Fixed.

1. `WriteAwareUpdateAttribute` has been removed. We subscribe just to `AttributeReadEvent`, which is not emitted when attributes are written.
2. The overriding logic in `read_attributes` didn't support passing attribute definition objects directly so I've updated it and consolidated it into the main class instead of keeping it a mixin.

zhaquirks/xiaomi/aqara/feeder_acn001.py
puddly: Fixed. I'm not a huge fan of `_update_attribute` being used in quirks but we can address it in a followup PR.

zhaquirks/ikea/vallhorn.py
```diff
@@ -21,11 +21,13 @@ class AttributeDefs(BaseAttributeDefs):
         on_only_when_dark: Final = ZCLAttributeDef(
             id=0x0000,
             type=t.Bool,
+            manufacturer_code=4476,
```
TheJulianJES, line 24: nit: Do we prefer hex for `manufacturer_code`s?

zhaquirks/develco/emi_han.py
```diff
@@ -16,12 +15,30 @@ class FrientMetering(CustomCluster, Metering):
...
+                attr
+                for attr in args.attribute_reports
+                if attr.attrid != Metering.AttributeDefs.divisor.id
+            ]
+
+            # Don't process if no attributes remain
+            if not args.attribute_reports:
+                return
+
+        super().handle_cluster_general_request(hdr, args, dst_addressing=dst_addressing)
```
TheJulianJES, line 41: side note: This might even be fixed (or properly fixable) now due to the manufacturer code changes.
I'll take a look at that in the future. Z2M also has this exact workaround, but the divisor attribute collides with a manufacturer-specific attribute (mode or something like that) that's sent sometimes. So, if that report has the manufacturer bit properly set, it should never touch the divisor? (if zigpy now checks for that..?)

tests/test_zunzunbee_slateswitch.py
```diff
@@ -39,13 +41,12 @@ async def test_button_ias(zigpy_device_from_v2_quirk):
...
-    for value in button_values:
+    for i, value in enumerate(button_values):
         cluster.update_attribute(ias_zone_status_attr_id, value)
-        assert listener.attribute_updated.call_args[0][0] == ias_zone_status_attr_id
-        assert listener.attribute_updated.call_args[0][1] == value
+        assert cluster_listener.attribute_updates[i] == (ias_zone_status_attr_id, value)
 
-    assert listener.attribute_updated.call_count == len(button_values)
-    assert listener.zha_send_event.call_count == len(button_values)
+    assert len(cluster_listener.attribute_updates) == len(button_values)
```
TheJulianJES, line 48: nit: Why do we prefer `len(.attribute_updates) over `.call_count`? I think HA Core does it the other way around.

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
```diff
@@ -200,28 +107,31 @@ class AnalogOutputRollerE1(WriteAwareUpdateAttribute, CustomCluster, AnalogOutpu
...
-        super()._update_attribute(attrid, value)
-        if attrid == self.AttributeDefs.present_value.id and not is_write:
+    def _handle_attribute_read(self, event: AttributeReadEvent) -> None:
+        """Handle attribute read event."""
+        if event.attribute_id == self.AttributeDefs.present_value.id:
             self.endpoint.window_covering.update_attribute(
                 WindowCovering.AttributeDefs.current_position_lift_percentage.id,
-                t.uint8_t(100 - value),
+                t.uint8_t(100 - event.value),
             )
```
TheJulianJES, line 121: This works properly? In general, you don't happen to have tested the Aqara Roller Curtain E1 yourself, right?
`_handle_attribute_read` -> `update_attribute`?

TheJulianJES (approved):
Looks good. I hope the Xiaomi devices still work correctly 😅

zhaquirks/ikea/vallhorn.py
puddly: We can update them to hex. We display them as decimal in the UI and practically everywhere else but I believe Wireshark uses hex.

zhaquirks/develco/emi_han.py
puddly: > So, if that report has the manufacturer bit properly set, it should never touch the divisor? 

Yeah, we utilize the manufacturer code (or lack of one) to look up the exact attribute.

tests/test_zunzunbee_slateswitch.py
puddly: I think I just got into the habit of using it after we started moving away from `.call_args_list[0][0][0]` and such. It's a few characters shorter but if you think `.call_count` is better, we can start using that, it makes no difference at the end of the day.

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
puddly: I went over this file a bunch of times when rewriting the attribute handling so it _should_ be fine. It's also somewhat unit tested. I will have to check again, we may need to subscribe to attribute reports in addition to reads.

tests/test_zunzunbee_slateswitch.py
TheJulianJES: No, I don't care here. Was just wondering if there's a deeper reason behind it 😄

zhaquirks/xiaomi/aqara/feeder_acn001.py
```diff
@@ -82,62 +83,95 @@
...
+        if ZCL_SERVING_SIZE not in self._attr_cache:
+            self._update_attribute(ZCL_SERVING_SIZE, 1)
+        if ZCL_PORTION_WEIGHT not in self._attr_cache:
+            self._update_attribute(ZCL_PORTION_WEIGHT, 8)
+        if ZCL_ERROR_DETECTED not in self._attr_cache:
+            self._update_attribute(ZCL_ERROR_DETECTED, False)
+        if ZCL_PORTIONS_DISPENSED not in self._attr_cache:
+            self._update_attribute(ZCL_PORTIONS_DISPENSED, 0)
+        if ZCL_WEIGHT_DISPENSED not in self._attr_cache:
+            self._update_attribute(ZCL_WEIGHT_DISPENSED, 0)
```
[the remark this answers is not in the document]
TheJulianJES: `_VALID_ATTIRUBTES` might work as well (and show the sensors as "unknown" until the first report comes in), but I think we can leave it like this.

zhaquirks/xiaomi/aqara/roller_curtain_e1.py
codyc1515: Sorry, but this doesn't work for Aqara Roller Curtain E1 (https://github.com/zigpy/zha-device-handlers/issues/4662).

puddly: @codyc1515 can you please open an issue and include more context?

PR #4673, opened 2026-01-20
Fix temperature calibration scaling factor

HarvsG:
## Proposed change

On my TS0601 by _TZE200_cwnjrr72 Zigbee TRV the value of the local temperature offset entity displayed in decidegress (i.e an offset of `-2.0` degrees showed as `-20`). Making this edit fixed that display value.

However it is possible that the scaling by a factor of 10 error is elsewhere in the stack as when I query the raw endpoints I can see that both the `temperature`  (which reports correctly in HA) and `temperature_calibration` are reported in decidegrees - so it seems odd that the scaling works well for one but not the other - I suspect that scaling is also happening elsewhere (incorrectly)

It is also not clear to me why, for my TRVs the min and max values for the calibration are +2.5 and -2.5. [This is why](https://github.com/zigpy/zha/blob/c6ac89e07dc72b7e7b920fcb4e92d60dd5b270a1/zha/application/platforms/number/__init__.py#L786)

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_trv.py
+++ b/zhaquirks/tuya/ts0601_trv.py
@@ -575,10 +575,7 @@ class MoesThermostat(TuyaThermostatCluster):
             MOES_MAX_TEMPERATURE_ATTR,
             lambda value: round(value / 100),
         ),
-        "local_temperature_calibration": (
-            MOES_TEMP_CALIBRATION_ATTR,
-            lambda value: round(value / 10),
-        ),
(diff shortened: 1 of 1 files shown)


HarvsG:
https://github.com/zigpy/zha/blob/c6ac89e07dc72b7e7b920fcb4e92d60dd5b270a1/zha/application/platforms/number/__init__.py#L789
Looks like the default multiplier here is what is causing the scaling to be double counted.

PR #4695, opened 2026-01-29
Bump zigpy to 0.91.3 and adjust quirks

TheJulianJES:
## Proposed change
This bumps zigpy to [0.91.3](https://github.com/zigpy/zigpy/releases/tag/0.91.3) and adjusts a couple of quirks for the recent change of `read_attributes` and `write_attributes` using `UNDEFINED` as the default for the manufacturer code (meaning: "automatically determine" whether to use or not). Previously, that was `None`.
Instead, `None` now means "override by not sending one".

So far, only quirks that were failing tests were updated. Others should be updated as well, though we should explicitly pass `manufacturer=UNDEFINED` from ZHA when reading and writing attributes to account for older (custom) quirks.

`TuyaClusterData` was also updated to expect `manufacturer=UNDEFINED`. As far as I can see, it's only used internally.

## Additional information
Required for:
- https://github.com/zigpy/zha/pull/637

## Checklist

### DIFF
--- a/zhaquirks/bosch/rbsh_trv0_zb_eu.py
+++ b/zhaquirks/bosch/rbsh_trv0_zb_eu.py
@@ -7,6 +7,7 @@ from zigpy.quirks.v2 import QuirkBuilder, ReportingConfig
 from zigpy.quirks.v2.homeassistant import EntityPlatform, EntityType
 from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
 import zigpy.types as t
+from zigpy.typing import UNDEFINED, UndefinedType
 from zigpy.zcl import foundation
 from zigpy.zcl.clusters.hvac import (
     ControlSequenceOfOperation,
@@ -189,8 +190,11 @@ class BoschThermostatCluster(CustomCluster, Thermostat):
         )
 
     async def write_attributes(
-        self, attributes: dict[str | int, Any], manufacturer: int | None = None
-    ) -> list:
+        self,
+        attributes: dict[str | int | foundation.ZCLAttributeDef, Any],
+        manufacturer: int | UndefinedType | None = UNDEFINED,
+        **kwargs,
+    ) -> list[list[foundation.WriteAttributesStatusRecord]]:
         """system_mode special handling.
 
         - turn off by setting operating_mode to Pause
@@ -303,11 +307,12 @@ class BoschThermostatCluster(CustomCluster, Thermostat):
 
     async def read_attributes(
         self,
-        attributes: list[int | str],
+        attributes: list[int | str | foundation.ZCLAttributeDef],
         allow_cache: bool = False,
         only_cache: bool = False,
-        manufacturer: int | t.uint16_t | None = None,
-    ):
+        manufacturer: int | UndefinedType | None = UNDEFINED,
+        **kwargs,
+    ) -> Any:
         """system_mode special handling.
 
         - read and convert operating_mode to system_mode.
--- a/zhaquirks/bosch/rbsh_trv0_zb_eu.py
+++ b/zhaquirks/bosch/rbsh_trv0_zb_eu.py
@@ -7,7 +7,6 @@ from zigpy.quirks.v2 import QuirkBuilder, ReportingConfig
 from zigpy.quirks.v2.homeassistant import EntityPlatform, EntityType
 from zigpy.quirks.v2.homeassistant.number import NumberDeviceClass
 import zigpy.types as t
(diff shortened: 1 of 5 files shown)


zhaquirks/bosch/rbsh_trv0_zb_eu.py
```diff
@@ -298,16 +300,16 @@ async def write_attributes(
...
-            result += await super().write_attributes(remaining_attributes, manufacturer)
+            result += await super().write_attributes(remaining_attributes, **kwargs)
         return result
 
     async def read_attributes(
         self,
-        attributes: list[int | str],
+        attributes: list[int | str | foundation.ZCLAttributeDef],
         allow_cache: bool = False,
         only_cache: bool = False,
```
TheJulianJES: @puddly Should we also just use `**kwargs` for `allow_cache` and `only_cache`?

puddly: Yeah, it's best to keep the overridden signature as tiny as possible.

puddly: But there are a ton of those to clean up. We can leave them for now and do it later.

zhaquirks/xiaomi/aqara/thermostat_agl001.py
```diff
@@ -132,14 +136,14 @@ async def write_attributes(
         if system_mode_value is not None:
             self.debug("Passing 'system_mode' write to Xiaomi cluster")
             result += await self.endpoint.opple_cluster.write_attributes(
-                {SYSTEM_MODE: min(int(system_mode_value), 1)}
+                {SYSTEM_MODE: min(int(system_mode_value), 1)}, **kwargs
             )
```
TheJulianJES, line 135: Note: This did not pass through `manufacturer` before. It's always supposed to be a manufacturer-specific write. Since this is always called with `manufacturer=UNDEFINED` anyway* ("so automatically determine"), we can just pass through `**kwargs` here as well, for consistency.
*: I guess except for if you're manually calling it through the debug UI. In that case, we may even want for that to be properly passed through.

Either way, it doesn't really matter for now. I'll keep it for consistency with the rest. This whole attribute read/write redirection has to be cleaned up at some point anyway.

zhaquirks/bosch/rbsh_trv0_zb_eu.py
TheJulianJES: Changed it for these two only now. Rest later.. From what I saw, the methods are always called with them being specified as keyword args, not just `read_attributes(.., True, True)`, so this should be good.

Or should we also pass through (unknown) `*args` in all overridden quirks signatures?

puddly: I think all future signature changes will always be in `**kwargs`, adding in `*args` is technically "more correct" but will just be noise IMO.

TheJulianJES:
I think this can be merged now. I'll update a few of the other `read_attributes` and `write_attributes` just from `manufacturer=None` to `manufacturer=UNDEFINED` in a follow-up PR after this one.
The method signatures (with `**kwargs` being passed through) can be updated in the future then.

puddly (approved):

zhaquirks/bosch/rbsh_trv0_zb_eu.py
```diff
@@ -252,7 +254,7 @@ async def write_attributes(
                     Thermostat.AttributeDefs.ctrl_sequence_of_oper
                 )
                 successful_r, failed_r = await super().read_attributes(
-                    [ctrl_sequence_of_oper_attr.name], True, False, manufacturer
+                    [ctrl_sequence_of_oper_attr.name], True, False, **kwargs
```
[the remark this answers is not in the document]
TheJulianJES: Will fix in a follow-up PR.

zhaquirks/bosch/rbsh_trv0_zb_eu.py
```diff
@@ -278,7 +280,7 @@ async def write_attributes(
                 )
                 if ctrl_sequence_of_oper_value is not None:
                     successful_r, failed_r = await super().read_attributes(
-                        [operating_mode_attr.name], True, False, manufacturer
+                        [operating_mode_attr.name], True, False, **kwargs
                     )
```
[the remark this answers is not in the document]
TheJulianJES: Yep. Will fix in a follow-up PR. Should have waited for the Copilot review to finish.. 😄

PR #4710, opened 2026-01-31
Configure PM2.5 reporting interval for IKEA VINDSTYRKA

kevlare:
This quirk allows to change the polling frequency of the PM25 values. The default ist up to 900 seconds.

## Proposed change
 The default for polling the PM25 sensor in zha is minimum 30 seconds and maximum 900 seconds. This cannot be changed in the interface, with this change it is possible to set the values in a custom quirk without coding everything. The new values are min: 5 seconds and max: 45 seconds

## Device diagnostics

[zha-01JBP5G37KTXPAFK8N2RFG9EZB-IKEA of Sweden VINDSTYRKA-7f16e5b631dcf47b5be65a32550ee3f7.json](https://github.com/user-attachments/files/24985990/zha-01JBP5G37KTXPAFK8N2RFG9EZB-IKEA.of.Sweden.VINDSTYRKA-7f16e5b631dcf47b5be65a32550ee3f7.json)

## Checklist

- [ x] The changes are tested and work correctly
- [ x] Device diagnostics data has been attached

### DIFF
--- a/zhaquirks/ikea/vindstyrka.py
+++ b/zhaquirks/ikea/vindstyrka.py
@@ -37,7 +37,7 @@ class VOCIndex(CustomCluster):
         )
 
 
-class HPM25(CustomCluster, PM25):
+class PM25(CustomCluster, PM25):
     """PM2.5 cluster forced to override the IKEA default."""
 
     cluster_id = 0x042A
@@ -59,7 +59,7 @@ class HPM25(CustomCluster, PM25):
     )
     .sensor(
         attribute_name="measured_value",
-        cluster_id=HPM25.cluster_id,
+        cluster_id=PM25.cluster_id,
         device_class=SensorDeviceClass.PM25,
         state_class=SensorStateClass.MEASUREMENT,
         reporting_config=ReportingConfig(
--- a/zhaquirks/ikea/vindstyrka.py
+++ b/zhaquirks/ikea/vindstyrka.py
@@ -10,9 +10,11 @@ from zigpy.quirks.v2 import (
     SensorStateClass,
 )
 import zigpy.types as t
-from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 from zigpy.zcl.clusters.measurement import PM25  # Import the PM2.5 cluster
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
 from zhaquirks.ikea import IKEA
+
 
 class VOCIndex(CustomCluster):
     """IKEA VOC index cluster."""
@@ -33,9 +35,13 @@ class VOCIndex(CustomCluster):
         max_measured_value: Final = ZCLAttributeDef(
             id=0x0002, type=t.Single, access="r", is_manufacturer_specific=True
         )
+
+
 class HPM25(CustomCluster, PM25):
     """PM2.5 cluster forced to override the IKEA default."""
+
     cluster_id = 0x042A
+
 
 (
     QuirkBuilder(IKEA, "VINDSTYRKA")
@@ -57,9 +63,9 @@ class HPM25(CustomCluster, PM25):
         device_class=SensorDeviceClass.PM25,
         state_class=SensorStateClass.MEASUREMENT,
         reporting_config=ReportingConfig(
-            min_interval=5,      # 5 seconds
-            max_interval=45,     # 45 seconds max
-            reportable_change=1
+            min_interval=5,  # 5 seconds
+            max_interval=45,  # 45 seconds max
+            reportable_change=1,
         ),
         fallback_name="Particulate Matter 2.5",
     )
--- a/zhaquirks/ikea/vindstyrka.py
+++ b/zhaquirks/ikea/vindstyrka.py
@@ -36,8 +36,7 @@ class VOCIndex(CustomCluster):
             id=0x0002, type=t.Single, access="r", is_manufacturer_specific=True
         )
 
-
-#class PM25(CustomCluster, PM25):
+    # class PM25(CustomCluster, PM25):
     """PM2.5 cluster forced to override the IKEA default."""
 
     cluster_id = 0x042A
@@ -57,7 +56,7 @@ class VOCIndex(CustomCluster):
         translation_key="voc_index",
         fallback_name="VOC index",
     )
-    .replaces(PM25) #instead of new custom cluster PM25
+    .replaces(PM25)  # instead of new custom cluster PM25
     .sensor(
         attribute_name="measured_value",
         cluster_id=PM25.cluster_id,
--- a/zhaquirks/ikea/vindstyrka.py
+++ b/zhaquirks/ikea/vindstyrka.py
@@ -36,15 +36,10 @@ class VOCIndex(CustomCluster):
             id=0x0002, type=t.Single, access="r", is_manufacturer_specific=True
         )
 
-    # class PM25(CustomCluster, PM25):
-    """PM2.5 cluster forced to override the IKEA default."""
-
-    cluster_id = 0x042A
-
-
 (
     QuirkBuilder(IKEA, "VINDSTYRKA")
     .replaces(VOCIndex)
+    .replaces(PM25)
     .sensor(
         VOCIndex.AttributeDefs.measured_value.name,
         VOCIndex.cluster_id,
@@ -56,7 +51,6 @@ class VOCIndex(CustomCluster):
         translation_key="voc_index",
         fallback_name="VOC index",
     )
-    .replaces(PM25)  # instead of new custom cluster PM25
     .sensor(
         attribute_name="measured_value",
         cluster_id=PM25.cluster_id,
--- a/zhaquirks/ikea/vindstyrka.py
+++ b/zhaquirks/ikea/vindstyrka.py
@@ -37,7 +37,7 @@ class VOCIndex(CustomCluster):
         )
 
 
-class PM25(CustomCluster, PM25):
+class HPM25(CustomCluster, PM25):
     """PM2.5 cluster forced to override the IKEA default."""
 
     cluster_id = 0x042A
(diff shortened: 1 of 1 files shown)


zhaquirks/ikea/vindstyrka.py
```diff
@@ -36,6 +37,12 @@ class AttributeDefs(BaseAttributeDefs):
         )
 
 
+class HPM25(CustomCluster, PM25):
```
puddly: Why do we need this to be a custom cluster? `0x042A` is already the cluster ID for `PM25`.

puddly:
The ZHA default is to ask sensors: "update no more often than once every 30 seconds, at least once every 15 minutes (even if nothing changes), and push an update whenever the sensor value changes by 0.1".

ZHA doesn't poll sensors. Sensors push updates when things change. Changing the maximum reporting interval to 45 seconds will mean the sensor spams the network 20x more often with duplicate readings.

Were you seeing poor reporting with the default config? If your PM2.5 readings fluctuate wildly, the sensor should be constantly reporting every 30 seconds with the default config. It it isn't, that means your PM2.5 readings don't change that much so changing the reporting interval will only add unnecessary updates.

kevlare:
Even if the sensor fluctuated I only got one reading every 15 minutes. I saw the changes on the device display but in HA there was only one new datapoint after the 15 minute interval. That is why I want to have the possibility to increase the update frequency. I found other persons with the same problem and the solution was to get zha_toolkit or switch to Z2M. This solution works within ZHA.I could revert to the default values, but this offers a simple solution to get your own quirk with your own reporting update interval. 
I tried with an automation but this did not work because the automation just read the current value and did not initiate an update of the value. I also tried changing the reporting interval in the device settings but there were no options there.

MattWestb:
Great to implanting the reporting interval !!
But if i remember from our hack of Vindriktning is the sensor active some seconds and then is being  turned off and its normally one or 2 cycles in one minute, Then we have the practical things of spamming the HA log with not exact and not important data. So it shall being possible using much larger time span for reporting.

But after reading Puddlys comment i was looking on what the device is reporting and and all measurements is displayed on the device very fast but its only reporting PM25 every 15 minutes so i think the device is not honor the reporting setting from ZHA.
It was the first (Zigbee) sensor IKEA was doing and also the temperature is only reported in 1° that is very bad (Alpstuga 0.01°C).
If like have good one i can recommending Alpstuga that is reporting very "intense".

So first verify that the device is honor new reporting and if its doing make the reporting range more realistic.

zhaquirks/ikea/vindstyrka.py
```diff
@@ -50,5 +57,17 @@ class AttributeDefs(BaseAttributeDefs):
...
         fallback_name="VOC index",
     )
+    .sensor(
+        attribute_name="measured_value",
+        cluster_id=HPM25.cluster_id,
+        device_class=SensorDeviceClass.PM25,
+        state_class=SensorStateClass.MEASUREMENT,
+        reporting_config=ReportingConfig(
+            min_interval=5,  # 5 seconds
+            max_interval=45,  # 45 seconds max
```
MattWestb: Both values is unrealistic pleas pot in somthing that is working in real life for must users.

kevlare: I changed to 

`             min_interval=20,  # 20 seconds`
`             max_interval=120,  # 120 seconds max`

is this ok?

MattWestb:
@puddly !!
Looks like Z2M is having the reporting working but was having problem with the manufacture cluster https://github.com/Koenkk/zigbee2mqtt/issues/21807 but its fixed.
Was sniffing one reconfigure and the device is replaying success after getting this:
```
ZigBee Network Layer Data, Dst: 0x8c01, Src: 0x0000
ZigBee Application Support Layer Data, Dst Endpt: 1, Src Endpt: 1
    Frame Control Field: Data (0x00)
    Destination Endpoint: 1
    Cluster: Unknown (0x042a)
    Profile: Home Automation (0x0104)
    Source Endpoint: 1
    Counter: 183
ZigBee Cluster Library Frame, Command: Configure Reporting, Seq: 117
    Frame Control Field: Profile-wide (0x00)
    Sequence Number: 117
    Command: Configure Reporting (0x06)
    Reporting Configuration Record, Float: 0,1
        Direction: Reported (0x00)
        Attribute: 0x0000
        Data Type: Single Precision Floating Point (0x39)
        Float: 0,1
    Minimum Interval: 30
    Maximum Interval: 900
```
but the device is still reporting every 15 minutes. 
Is somthing broken or have we missing implementing some IKEA twerks ???

MattWestb:
Its looks like IKEA have using one not standard data type for reporting on its cluster and the device is not honoring the command if not being correct formed:    
```
pm25Measurement: {
        ID: 0x042a,
        attributes: {
            measuredValue: {ID: 0x0000, type: DataType.uint16},
            measuredMinValue: {ID: 0x0001, type: DataType.uint16},
            measuredMaxValue: {ID: 0x0002, type: DataType.uint16},
            measuredTolerance: {ID: 0x0003, type: DataType.uint16},
        },
        commands: {},
        commandsResponse: {},
    },
```
and
```
            await ep.configureReporting('pm25Measurement', [{
                attribute: {ID: 0x0000, type: zigbeeHerdsman.Zcl.DataType.singlePrec},
                minimumReportInterval: 60, maximumReportInterval: 120, reportableChange: 2,
            }]);

**Comment by puddly** [comment]
We use this in zigpy for all of the concentration clusters, including `PM25`:

```python
class _ConcentrationMixin:
    """Mixin for the common attributes of the concentration measurement clusters"""

        measured_value: Final = ZCLAttributeDef(
            id=0x0000, type=t.Single, access="rp", mandatory=True
        )  # fraction of 1 (one)
```

So when you set up reporting, it should do so with the same settings you see above.

kevlare:
@puddly and @MattWestb Thank you for looking into this.
I'm new to github and I'm not sure what I should do next? Include the _ConcentrationMixin class in my code? would this correct the way Vindstyrka handles the updates?

puddly:
Not at all, it's all included already. I'm just trying to understand where the device is not working. Is it not respecting the pushed attribute reporting configuration?

MattWestb:
What i see shall the ZCL using **uint16** for the setting up reporting but IKEA was before ZCL was implanting the PM25 so they was doing there way and not copy Silabs standard code (that was not existing for PM25 then).
Z2M is using "attribute: {ID: 0x0000, type: zigbeeHerdsman.Zcl.DataType.**singlePrec**}," for setting up attribute reporting on the cluster and the normal **uint16** on the other ZCL standard cluster on the device.
Then the device Zigbee stack is getting one valid ZCL command its sending success but the device "spp" is not understanding the command so is not updating the reporting settings for the cluster.

kevlare:
@puddly yes the Vindstyrka is not respecting the reporting configuration and only using the max interval even if there are variations in the values.

With the original quirk I found the following messages in the raw logs, but still it was only using the max interval even though there were changes of 3-4 units sometimes even more.

2026-01-31 14:59:38.249 DEBUG (MainThread) [zigpy.zcl] [0x0B8A:1:0x042a] Sending request: Configure_Reporting(config_records=[AttributeReportingConfig(direction=0, attrid=0x0000, datatype=<DataTypeId.single: 57>, min_interval=30, max_interval=900, reportable_change=0.1)])

2026-01-31 14:59:38.249 DEBUG (MainThread) [zigpy.device] Device concurrency (2) reached, delaying request (1 enqueued)

2026-01-31 14:59:38.250 DEBUG (MainThread) [zigpy.device] Previously delayed device request is now running, delayed by 0.05s

kevlare:
@puddly 
here are two history screenshots, one with the original intervals and one with a higher interval. As you can see the interval used is always the max. Even when there is a peak there is no earlier data point than after 900 seconds for the original interval.
Screenshot 2026-02-10 222128
Screenshot 2026-02-10 222202

puddly:
Very strange. From your log, it's almost as if IKEA is internally using an integer for this value, so the minimum reportable change of `0.1` is converted to `0` by the firmware and ignored but `1` passes through?

zhaquirks/ikea/vindstyrka.py
puddly: I think you can drop the `HPM25` subclass and just use `PM25` directly

kevlare: I used PM25 directly and it works in HA, but the GitHub Code Checker was throwing an error because I'm using the same   name twice. I'll change it back to PM25

kevlare: Check failed as expected, can this be ignored?

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

zhaquirks/ikea/vindstyrka.py:40: error: Name "PM25" already defined (possibly by an import)  [no-redef]
Found 1 error in 1 file (checked 455 source files)
```

puddly: No, I think you can just delete the whole `class PM25`.

kevlare: I deleted 
`class` PM25(CustomCluster, PM25):

and added 
`replaces(PM25)`

this does not work in my HA instance.
TVOC has the same value as PM25 and the specified time interval is not used, but the HA default.

I have changed the code to what I used, but it is not working as it is. Did I use the wrong syntax?

kevlare: Fixed the problem with the wrong values, but now it won't use the new time values and uses the 900 second default

kevlare: When I use the `replaces(PM25)` command the default values remain. I guess will I have to use the HPM25 subclass again or do you have another idea?

kevlare: Tried it again and now the current version of code seems to work and also the min value works now.
image
The air quality was real good and not fluctuating while I tried it out and without a change in the value the last updated value can go to multiple minutes even past the max update time. That's what led me to believe it was not updating correctly. It seems as if zigbee is not updating even past the max update time (it should update after 120 seconds with my quirk) or it could just be HA not writing a new value because nothing changed?
image

zhaquirks/ikea/vindstyrka.py
```diff
@@ -39,6 +40,7 @@ class AttributeDefs(BaseAttributeDefs):
 (
     QuirkBuilder(IKEA, "VINDSTYRKA")
     .replaces(VOCIndex)
+    .replaces(PM25)
```
javicalle, line 43: Just guessing, is that needed?

Issue #4620, opened 2026-01-05
[Bug] _TZE204_ltwbm23f temperature calibration quirk misses multiplier [Fix inside]


GWRon:
Heya,

in PR #3769 support for _TZE204_qyr2m29i and _TZE204_ltwbm23f was added.

But using the quirk and trying to configure the temperature calibration a "+1" leads to a "+0.1" change. In other words, the multiplier is missing.

I only own the latter (_TZE204_ltwbm23f), so I won't come up with a PR for this (I can only verify it works on my machine).

Only change is to add multiplier=0.1,` to the corresponding tuya-number.

```python
# Moes TRV602Z and TRV801Z
(
    TuyaQuirkBuilder("_TZE204_qyr2m29i", "TS0601")
    .applies_to("_TZE204_ltwbm23f", "TS0601")
    .tuya_dp(
        dp_id=3,
        ep_attribute=TuyaThermostatV2.ep_attribute,
        attribute_name=TuyaThermostatV2.AttributeDefs.running_state.name,
        converter=lambda x: RunningState.Heat_State_On if x else RunningState.Idle,
    )
	#...
    .tuya_number(
        dp_id=47,
        attribute_name=TuyaThermostatV2NoSchedule.AttributeDefs.local_temperature_calibration.name,
        type=t.int32s,
        min_value=-6,
        max_value=6,
        unit=UnitOfTemperature.CELSIUS,
        step=1,
        multiplier=0.1,															# <----- HERE
        translation_key="local_temperature_calibration",
        fallback_name="Local temperature calibration",
    )
#...
)
```
(https://github.com/zigpy/zha-device-handlers/blob/ec82cb8e66e3fe59bd9d81cd0a3b9769d83ba727/zhaquirks/tuya/tuya_trv.py#L677)

Other tuya_numbers/DPs for these devices care for this already, so I think it was simply missed in the PR but as said ... only owning one device and not the other.

sroeper:
I just discovered the same, but the step should be also changed to 0.5 which works at my side.
My devices also have 3 AA batteries not 2, which can easily corrected with:

```
...
 .tuya_battery(dp_id=6, battery_qty=3)
...
```
Now I have just 2 annoying things left, I am not able to set the devices to ECO mode from HA and I need a template sensor for the temperature readings to make some calculations with them. Any suggestions?

GWRon:
[...]

PR #4715, opened 2026-02-02
Fix Tuya TRV _TZE204_ltwbm23f quirk with corrected running_state logic

ghost:
## Summary
- Separate `_TZE204_ltwbm23f` from the Moes TRV602Z/TRV801Z quirk as this device has different DP mappings
- Fix inverted running_state on dp_id=102 (device sends True=idle, False=heating)
- Expand temperature calibration range to -60 to 60 decidegrees (-6.0°C to +6.0°C in 0.1°C increments)
- Simplify system_mode to Heat/Off on dp_id=2
- Add battery reporting and valve_position sensor

## Test plan
- [ ] Verify TRV correctly reports idle vs heating state
- [ ] Test temperature calibration with finer increments
- [ ] Confirm battery percentage reporting works
- [ ] Verify valve position sensor updates

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/zhaquirks/tuya/tuya_trv.py
+++ b/zhaquirks/tuya/tuya_trv.py
@@ -654,7 +654,6 @@ class TuyaThermostatV2NoSchedule(TuyaThermostatV2):
 # Moes TRV602Z and TRV801Z
 (
     TuyaQuirkBuilder("_TZE204_qyr2m29i", "TS0601")
-    .applies_to("_TZE204_ltwbm23f", "TS0601")
     .tuya_dp(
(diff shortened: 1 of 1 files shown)


TheJulianJES (changes requested):
Please check your PRs before submitting them. The datapoints are all over the place.
The min/max range with the multiplier also makes no sense.

- https://github.com/zigpy/zha-device-handlers/issues/4620 seems to be the better fix

I'll close this for now. But thanks for trying.

PR #4727, opened 2026-02-07
Add Tuya SNZB-03 PIR motion sensor quirk

nschimme:
## Proposed change
Adds support for the Tuya SNZB-03 which needs this quirk to see the battery level and to not be stuck permanently in the detected state.

## Additional information

This device does not have a MCU endpoint so we're not able to use `TuyaMotionWithReset` and have to use `MotionWithReset` instead.

Other devices might be supported too, a related project was able to re-use a configuration used by other Tuya motion sensors: https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8267

## Device diagnostics

[zha-01K8BQMJSKZFPA3YC43C2Q4TB0-_TZ3000_bb6xaihh SNZB-03-5f418bf0fa2d56ef1f1b15939ad252b5.json](https://github.com/user-attachments/files/25142330/zha-01K8BQMJSKZFPA3YC43C2Q4TB0-_TZ3000_bb6xaihh.SNZB-03-5f418bf0fa2d56ef1f1b15939ad252b5.json)

## Checklist


TheJulianJES (approved):
Thanks! Does the sensor also NOT get reset when you're constantly moving? E.g. is the `MotionWithReset` timeout correct?

nschimme:
> Thanks! Does the sensor also NOT get reset when you're constantly moving? E.g. is the `MotionWithReset` timeout correct?

The timeout seemed accurate in my testing

PR #4648, opened 2026-01-13
update Sonoff MINI DUO (ZB2GS) 2-channel switch

1775939187:
## Proposed change
Quirk adding support for Sonoff MINI DUO (ZB2GS) 2-channel switch.
## Device diagnostics
[SONOFF MINI-ZB2GS.json](https://github.com/user-attachments/files/24577953/SONOFF.MINI-ZB2GS.json)

## Checklist
- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works
- [x] Device diagnostics data has been attached

PR #4733, opened 2026-02-07
Add Sonoff ZBMINIR2 network LED setting

uschindler:
## Proposed change
This allows user to disable the annoying network indicator LED of the Sonoff ZBMINIR2.
By default the attribute is ON for those devices.

## Additional information
Based on code from zigbee-herdsman (z2m) and a recent PR by Sonoff #4648 (now #4742)
This code already uses @puddly's new manufacturer code override. ~~#4648 has to be updated.~~

## Device diagnostics
Not needed, device is already known, it is just a new attribute.

## Checklist


uschindler:
This also removes the useless 1/0 for a boolean attribute.

uschindler:
Any chance to get this in? @TheJulianJES 

I am planning to later possibly refactor the Sonoff part so they share the same cluster code and we just need to enable per device. This would allow to easier add new devices without always redefining the Sonoff cluster with same attributes.

TheJulianJES (approved):
Thanks!

PR #4740, opened 2026-02-09
Add Sonoff TRVZB adaptive mode

1775939187:
## Proposed change
Adaptive Mode is a feature designed to minimize temperature fluctuations.The device dynamically adjusts the valve opening based on the temperature changes and the target temperature to maintain a more stable environment.Note:Enabling Adaptive Mode will automatically disableValve Opening Percentage and Temp. Accuracy.

## Device diagnostics
[SONOFF TRVZB.json](https://github.com/user-attachments/files/25175807/SONOFF.TRVZB.json)

## Checklist


Schm1tz1:
That looks pretty good! I am working with the manual quirk at the moment and just wanted to create the same PR here.

TheJulianJES (commented):
I don't understand why we need two fake attributes (`boost_mode` + `timer_mode`) here? We do not create binary sensors for them and only ever write to them, so why don't we use a different `attribute_value` in `write_attr_button` and write directly to `temporary_mode`?

TheJulianJES (approved):
Per my comment above, I've refactored the quirk a bit to completely remove the `write_attributes` and `update_attribute` redirects. Unless I completely missed something, they should not be required. We just use a different `attribute_value` in the `write_attr_button`.

**BUT** we also have to provide a `unique_id_suffix` then. By default, it's the same of the attribute. But if two entities reference the same attribute, we need a suffix (ideally for both). Otherwise, only one will appear.

I'll merge this now for the next HA beta, releasing later today (2026.3.0b0). If there are any bugs with this, please ping me and I will have a look.

TheJulianJES:
We'll also merge the OTA update then, likely tomorrow. At first, it'll be in the [zigpy-ota beta](https://github.com/zigpy/zigpy-ota#using-pre-release-channels) channel.
We'll push it to the stable channel when 2026.3.0 is out, so everyone can also immediately use the new features with the updated built-in quirk. (The release notes mention the new adaptive mode, so it's good if it's also available to use/test then.)

PR #4745, opened 2026-02-10
udpate s60zbtpf/tpg and mini-zbdim

lcheng33775823:


zhaquirks/sonoff/s60zbtpf.py
```diff
@@ -0,0 +1,227 @@
...
+        return await super()._read_attributes(
+            attribute_ids,
+            *args,
+            manufacturer=foundation.ZCLHeader.NO_MANUFACTURER_ID,
+            **kwargs,
+        )
+
+    @property
+    def _is_manuf_specific(self):
+        return False
```
puddly: You can remove all of this code and just add `manufacturer_code=None` to every attribute definition.

vukisz:
I added minizbdim.py quirk into custom_zha_quirks on Feb 05,2026. Power, current, Voltage was working for a few days, but on ~ feb 9 Current and power stopped reporting. Tried taking latest quirk from this PR, but still power and current is unknown. Voltage is correct.
@lcheng33775823 is this as expected?

shingle-smith:
I've uploaded my changes (fixing manufacturer_code errors) here, maybe worth a try for those with issues
https://github.com/zigpy/zha-device-handlers/issues/4890#issuecomment-4431657717

PR #4749, opened 2026-02-12
Fix Aqara FP1E using incorrect manufacturer code

TheJulianJES:
## Proposed change
This changes the attribute definition's manufacturer codes to `None` for the Aqara FP1E mm-wave sensor.
Otherwise, we aren't parsing the attribute reports from the device, since they don't have the manufacturer bit set.
For writes, it looks like the manufacturer code doesn't matter at all(??)

## Additional information
Should address https://github.com/home-assistant/core/issues/162484

## Device diagnostics
[zha-01KAH6BZ8GBKBJXCPX8V4BB4KP-Aqara Presence Sensor FP1E-73e318d90a5606426bd23584f7e84776.json](https://github.com/user-attachments/files/25267670/zha-01KAH6BZ8GBKBJXCPX8V4BB4KP-Aqara.Presence.Sensor.FP1E-73e318d90a5606426bd23584f7e84776.json)

## Checklist

### DIFF
--- a/zhaquirks/xiaomi/aqara/motion_agl1.py
+++ b/zhaquirks/xiaomi/aqara/motion_agl1.py
@@ -70,7 +70,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
             id=0x015B,
             type=types.uint32_t,
             access="rw",
-            is_manufacturer_specific=True,
+            manufacturer_code=None,
         )
 
         # Detected motion
@@ -79,7 +79,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
             type=AqaraMotion,
             zcl_type=DataTypeId.uint8,
             access="rp",
-            is_manufacturer_specific=True,
+            manufacturer_code=None,
         )
 
         # Distance to the detected motion in millimeters
@@ -87,7 +87,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
             id=0x015F,
             type=types.uint32_t,
             access="rp",
-            is_manufacturer_specific=True,
+            manufacturer_code=None,
         )
 
         # The configurable detection sensitivity
@@ -96,7 +96,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
             type=AqaraMotionSensitivity,
             zcl_type=DataTypeId.uint8,
             access="rw",
-            is_manufacturer_specific=True,
+            manufacturer_code=None,
         )
 
         # Detected occupancy
@@ -105,7 +105,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
             type=AqaraOccupancy,
             zcl_type=DataTypeId.uint8,
             access="rp",
-            is_manufacturer_specific=True,
+            manufacturer_code=None,
         )
 
         # Trigger AI spatial learning (write 1)
@@ -113,7 +113,7 @@ class OppleCluster(XiaomiAqaraE1Cluster):
             id=0x0157,
             type=types.uint8_t,
             access="w",
-            is_manufacturer_specific=True,
+            manufacturer_code=None,
         )
 
         # Trigger device restart (write 0)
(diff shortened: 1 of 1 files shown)


TheJulianJES:
Z2M ignores the manufacturer code for attribute reports. For writing, it's using it for this device for both the restart and spatial learning writes (`manufacturerOptions.lumi`): https://github.com/Koenkk/zigbee-herdsman-converters/blob/e410903af3516701e48d4a851cda7488f70f928d/src/lib/lumi.ts#L3198-L3225

However, using both buttons in ZHA seems to work fine without the manufacturer code/bit, so I think we're actually doing this more correct with this PR, though the sensor seems to ignore it anyway.

TheJulianJES:
Paired an FP1E on HA 2026.2.1, reproduced the issue, applied the fix from this PR, and could no longer reproduce it.
This should be good to go.

puddly (approved):

TheJulianJES:
Ok, so I wasn't sure *why* this actually broke now, as you mentioned we still somewhat process attribute reports if the manufacturer code doesn't match the definition, but the reason is kind of stupid if I'm correct:

From what I understand, we took this path (without the PR):
- [zigpy/zcl/__init__.py#L861-L864](https://github.com/zigpy/zigpy/blob/0ceeb47a324f911fda830e23312fd47a4162d136/zigpy/zcl/__init__.py#L861-L864)
`find_attribute` raises `KeyError` as report is mfr-specific but definition is not, so `attr_def` is `None`
- https://github.com/zigpy/zigpy/blob/0ceeb47a324f911fda830e23312fd47a4162d136/zigpy/zcl/__init__.py#L1184
  `_update_attribute` searches attribute by ID only, so finds the correct definition (with mf-code) and updates the "typed cache"
- [zigpy/zcl/__init__.py#L866-L880](https://github.com/zigpy/zigpy/blob/0ceeb47a324f911fda830e23312fd47a4162d136/zigpy/zcl/__init__.py#L866-L880)
`AttributeReportedEvent` is emitted from `handle_cluster_general_request`, with the unmodified value, but this doesn't matter as ZHA sensors only need the event and get the state from the attribute cache using `cluster.get`.
That only works with the "new typed cache", but that's where the attribute value lives due to the more loose matching in `_update_attribute`.
  - However, for the ZHA sensor to actually check for updated state and possibly emit an event, it checks the `attribute_name` of the `ClusterAttributeUpdatedEvent` (just [passed through](https://github.com/zigpy/zha/blob/99ea585c3456b16e22fa395648f0dea79169e996/zha/zigbee/cluster_handlers/__init__.py#L246-L270) from zigpy). The check is here: [zha/application/platforms/sensor/__init__.py#L336](https://github.com/zigpy/zha/blob/99ea585c3456b16e22fa395648f0dea79169e996/zha/application/platforms/sensor/__init__.py#L336)
  - So, the sensor `attribute_name` needs to match the event from the zigpy event. And the zigpy event is always fired with `attribute_name=None` if it can't find the definition in `handle_cluster_general_request`.
  See: [zigpy/zcl/__init__.py#L873](https://github.com/zigpy/zigpy/blob/0ceeb47a324f911fda830e23312fd47a4162d136/zigpy/zcl/__init__.py#L873)

This isn't great because we first match with manufacturer code and fail to find a definition, but `update_attribute` matches without it and then finds the definition (that has a manufacturer code definition), even though that's somewhat incorrect.
Later, the attribute update event is emitted without the `attribute_name` and with the originally reported value, even though we store the value is the new typed runtime cache and with mf-code in the DB.

That should be somewhat improved with https://github.com/zigpy/zigpy/pull/1760 and https://github.com/TheJulianJES/zigpy/compare/tjj/manufacturer_update_fix...TheJulianJES:zigpy:tjj/manufacturer_update_fix_with_mfr_attr_fix_zcl_cache. But I think it still needs to be looked at afterwards. The behavior will be more consistent with those PRs, but yeah. Maybe not fully perfect yet if we can't find a definition.

PR #4750, opened 2026-02-12
Remove Frient EMI Norwegian HAN `divisor` workaround

TheJulianJES:
## Proposed change

This removes the workaround for the Frient EMI Norwegian HAN, which previously blocked attribute reports for the `divisor` attribute, similar to Z2M [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/e410903af3516701e48d4a851cda7488f70f928d/src/devices/develco.ts#L35-L46). With HA Core 2026.2.0 and later, this should no longer be needed.

The Frient EMI HAN test was also changed to use a message from a real device (with the mfr-specific bit set).

## Additional information

This essentially reverts the following PR:
- https://github.com/zigpy/zha-device-handlers/pull/4396

There's more context in the comments of that PR. @uvNikita was able to grab such an attribute report for the manufacturer specific attribute that conflicts with the ZCL `divisor` attribute and we confirmed the `is_manufacturer_specific` bit is correctly set, meaning that zigpy no longer parses those attribute reports for the ZCL attribute. (Thanks!!)

- In a future PR, we can implement support for the Frient specific attributes.
- The "quirk" (temporarily without function) was kept, as we still have a test for this device and plan to add more functionality to it soon. 
- We should only merge this for HA Core 2026.3.x.

## Device diagnostics
See this comment: https://github.com/zigpy/zha-device-handlers/pull/4396#issuecomment-3401998359

## Checklist

### DIFF
--- a/tests/test_develco.py
+++ b/tests/test_develco.py
@@ -156,7 +156,7 @@ async def test_mfg_cluster_events(zigpy_device_from_v2_quirk):
     # TODO: mfr-specific report should not update the standard divisor attribute,
     #  but zigpy currently does not filter this. Fix in zigpy.
     assert len(metering_listener.attribute_updates) == 0
-    assert metering_cluster.get(Metering.AttributeDefs.divisor.id) is None
+    assert metering_cluster.get(Metering.AttributeDefs.divisor) is None
 
     # send real attribute report with current_summ_delivered, current_summ_received,
     # instantaneous_demand, and status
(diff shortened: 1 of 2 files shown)


TheJulianJES:
This also removes the `_CONSTANT_ATTRIBUTES` workaround to set the divisor to a fixed `1000`. We should make sure that when reading the attribute (with this modified empty quirk), the device actually returns `1000`.

TheJulianJES:
Hmm, looks like zigpy is still parsing the attribute report for the `divisor`, even though it has a manufacturer code set? Need to check this later.

puddly:
We will still emit an attribute report event if an attribute definition doesn't exist, for backwards compatibility.

TheJulianJES:
Ah, right. A bunch of things in zigpy/quirks still rely on the IDs anyway (`ClusterListener`, `_CONSTANT_ATTRIBUTES`). Those need to allow definition reference (or name) for this.

For reference, also left a comment about this here: https://github.com/zigpy/zigpy/pull/1760#discussion_r2800173787

PR #4763, opened 2026-02-16
Fix `EventableCluster` not listening to `AttributeUpdatedEvent`

TheJulianJES:
## Proposed change
This fixes an issue where the `EventableCluster` does not listen to `AttributeUpdatedEvent` events.

For example, if a Tuya datapoint is remapped to a "ZCL" attribute, which is updated using `update_attribute`, it'll never fire `AttributeReportedEvent`, only `AttributeUpdatedEvent`. So, the `zha_event` for the attribute update is never fired.

The test for the Tuya TS021 remote is also improved to make sure the `EventableCluster` also listens to `AttributeUpdatedEvent`. 

## Additional information
Should address:
- https://github.com/home-assistant/core/issues/163118

See related quirk:
https://github.com/zigpy/zha-device-handlers/blob/40b41de4c0abae1f5265b1d05cbf47861f253afc/zhaquirks/tuya/ts0021.py#L36-L53

It would probably make sense to add decent test coverage for `EventableCluster` in another PR.
Maybe even utilizing the Tuya datapoint remapping, or just have `update_attribute` be called there.

## Device diagnostics
Possibly in linked HA issue or already in ZHA diagnostics.

## Checklist

### DIFF
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -154,7 +154,7 @@ async def test_ts0021_remote(
         "attribute_updated",
         {
             "attribute_id": expected_attr_name,
-            "attribute_name": "Unknown",
+            "attribute_name": "Unknown",  # no ZCLAttributeDef for Tuya DP-mapped attr
             "value": expected_attr_value,
         },
     )
--- a/tests/test_tuya.py
+++ b/tests/test_tuya.py
@@ -120,64 +120,44 @@ async def test_singleswitch_state_report(zigpy_device_from_quirk, quirk):
 
 
 @pytest.mark.parametrize(
-    "quirk,raw_event,expected_attr_name,expected_attr_value",
+    "raw_event,expected_attr_name,expected_attr_value",
     (
-        (
-            zhaquirks.tuya.ts0021.TS0021,
-            ZCL_TUYA_BUTTON_1_SINGLE_PRESS,
-            "btn_1_pressed",
-            0x00,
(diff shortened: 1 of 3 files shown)


puddly (approved):

tests/test_tuya.py
```diff
@@ -120,65 +120,45 @@ async def test_singleswitch_state_report(zigpy_device_from_quirk, quirk):
...
     assert len(switch_listener.attribute_updates) == 1
-
     assert switch_listener.attribute_updates[0][0] == expected_attr_name
     assert switch_listener.attribute_updates[0][1] == expected_attr_value
 
+    zha_listener.zha_send_event.assert_called_once_with(
+        "attribute_updated",
+        {
+            "attribute_id": expected_attr_name,
+            "attribute_name": "Unknown",  # no ZCLAttributeDef for Tuya DP-mapped attr
```
TheJulianJES, line 157: This might be something to look into soon. The quirk doesn't have a `ZCLAttributeDef`, but only maps the Tuya DP to names, so `update_attribute` is called with the name (present in `attribute_id` here).

The `TuyaQuirkBuilder` uses (fake) ZCL attribute definitions for all mapped Tuya DPs, but doesn't really support remotes yet. We should move over Tuya remotes to that soon and/or adjust these quirks.

If we make any changes in zigpy that break this, we'll (now) know at least.

PR #4597, opened 2025-12-23
Add support for Nous E10 CO2, Temperature and Humidity Detector

bexelbie:
## Proposed change
Add quirk for the Nous E10 (_TZE284_xpvamyfz / TS0601) which provides CO2, temperature, and humidity measurements using the TuyaQuirkBuilder v2 API.

Sensors:
- CO2 concentration (DP 2, ppm)
- Temperature (DP 18, °C, signed for negative values)
- Humidity (DP 19, %)

## Additional information
- This is my first time trying to do this - feedback welcome and appreciated.
- Tests are minimal but it isn't clear to me what else to test - examples welcome
- This code is working in my HA instance with the device

## Device diagnostics
[zha-2b5da96179570b3038c5dfdb61f698d8-_TZE284_xpvamyfz TS0601-9eae19e47cbad4e60c376b24b5c0679b.json](https://github.com/user-attachments/files/24311493/zha-2b5da96179570b3038c5dfdb61f698d8-_TZE284_xpvamyfz.TS0601-9eae19e47cbad4e60c376b24b5c0679b.json)

## Checklist
- [X] The changes are tested and work correctly
- [X] `pre-commit` checks pass / the code has been formatted using Black
- [X] Tests have been added to verify that the new code works
- [X] Device diagnostics data has been attached

PR #4773, opened 2026-02-19
Add Tuya CO2/temp/humidity sensors (Nous E10)

awehrfritz:
## Proposed change

This adds a quirk for the Nous E10 CO2/temp/humidity sensor.

## Additional information

The quirk was developed based in the Zigbee2MQTT device definition in
[zigbee-herdsman-converters/src/devices/tuya.ts](https://github.com/Koenkk/zigbee-herdsman-converters/blob/7eb0b699e500bb945a45bd3533fe0b5a47c721e6/src/devices/tuya.ts#L17480)
From the Zigbee2MQTT commits, it appears that this not only applies to the Nous E10, but also the Zorro Alert ZR360CDB device.

This PR supersedes a previous PR (https://github.com/zigpy/zha-device-handlers/pull/4597), adding all data points from Zigbee2MQTT such as information on the battery status.

This has been tested locally on my HA system with a Nous E10 CO2/temp/humidity sensor for a few days now.

Fixes: https://github.com/home-assistant/core/issues/151069

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_co2.py
+++ b/zhaquirks/tuya/ts0601_co2.py
@@ -1,0 +1,84 @@
+"""Nous E10 CO2, Temperature, and Humidity Detector.
+
+The quirk applies to the Zorro Alert ZR360CDB as well, which appears to be
+identical to the Nous E10.
+
+The quirk was developed based on the Zigbee2MQTT device definition:
+    https://github.com/Koenkk/zigbee-herdsman-converters/blob/7eb0b699e500bb945a45bd3533fe0b5a47c721e6/src/devices/tuya.ts#L17480
+"""
+
+from zigpy.quirks.v2 import EntityPlatform, EntityType
+import zigpy.types as t
+
+from zhaquirks.tuya.builder import TuyaQuirkBuilder
+
+
+class AirQuality(t.enum8):
+    """Tuya air quality enum."""
+
+    excellent = 0x00
+    moderate = 0x01
+    poor = 0x02
+
+
+class AlarmRingtone(t.enum8):
+    """Tuya alarm ringtone enum."""
+
+    volume_low = 0x00
+    volume_high = 0x01
+    OFF = 0x02
+
+
+class BatteryState(t.enum8):
+    """Tuya battery state enum."""
+
+    low = 0x00
+    medium = 0x01
+    high = 0x02
+
+
+(
+    TuyaQuirkBuilder("_TZE200_pl31aqf5", "TS0601")
+    .applies_to("_TZE200_xpvamyfz", "TS0601")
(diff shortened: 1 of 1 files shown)


zhaquirks/tuya/ts0601_co2.py
```diff
@@ -0,0 +1,84 @@
...
+    moderate = 0x01
+    poor = 0x02
+
+
+class AlarmRingtone(t.enum8):
+    """Tuya alarm ringtone enum."""
+
+    volume_low = 0x00
+    volume_high = 0x01
+    OFF = 0x02
```
TheJulianJES: Let's lower-case that as well then:
```suggestion
    off = 0x02
```

Or `Volume_low`, `Volume_high`, `Off`. At some point, these won't be used for raw HA state anymore. Right now, they still are.

zhaquirks/tuya/ts0601_co2.py
```diff
@@ -0,0 +1,84 @@
...
+    .tuya_number(
+        dp_id=17,
+        attribute_name="backlight_mode",
+        type=t.uint16_t,
+        min_value=1,
+        max_value=3,
+        step=1,
+        translation_key="backlight_mode",
+        fallback_name="Backlight mode",
+    )
```
TheJulianJES, line 79: Should this be an enum (select entity) instead maybe?

TheJulianJES (commented):
Nice! Looking good.

Should this go in https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/tuya_co.py?

awehrfritz:
> Should this go in https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/tuya_co.py?

I honestly don't know. I cannot make sense of the file naming in the tuya directory. tuya_co.py seems to have air quality sensors, but the filename is misleading at best (CO is not really air quality, more a safety issue). (I brought this up in a discussion here:  https://github.com/zigpy/zha-device-handlers/discussions/4774)

Also, since I added a few custom classes, I thought it was better to "hide" them in an own file. I am fine either way.

zhaquirks/tuya/ts0601_co2.py
awehrfritz: I also don't know the answer to this. This works, but I was wondering about the same thing. See also my questions in discussion https://github.com/zigpy/zha-device-handlers/discussions/4774 (point 4).

This implementation mimics the one in Z2M, but I am happy to change this if enums are better for that.

awehrfritz:
@TheJulianJES any chance to get this merged some time soon? Your triage bot closed the related [issue](https://github.com/home-assistant/core/issues/151069) already since it apparently ignores this PR.

Maybe also worth to mention that this implementation has been working reliably for half a year now in my HA setup.

PR #4777, opened 2026-02-21
Fix parsing for broken Aqara E1 attribute report

puddly:
## Proposed change

Add one more parsing special case for bad Aqara attribute reports that our parsing logic isn't handling correctly.

## Additional information

Reported here: https://github.com/home-assistant/core/issues/163692.

This is I think the first "bad" report that I've seen in a very long time so it's a little strange:
```python
# attr   [type]   [len] [data]
  f700   41       37    03283b05210900092100010a219f580b20000c20010d23200e00001123010000006520416620806720236920026a21451e6b2000
```

The length (0x37) isn't off by one, it's off by three.

## Checklist


TheJulianJES (approved):
Hmm, it's really weird we're getting a new variation after years.
We should check if Z2M also has this.

PR #4782, opened 2026-02-22
Add ELECRAM Ramses ESP32-C6 quirk

IMMRMKW:
## Proposed change
For the project [ramses_cc](https://github.com/ramses-rf/ramses_cc) I am working on Zigbee communication for my [Ramses ESP](https://github.com/IMMRMKW/ramses_esp) board. Until now, only serial and MQTT communication was possible.

## Additional information
The device supports two custom clusters, in similar fashion as the MQTT communication. One cluster is for sending received messages to Ramses RF, the other for receiving messages from Ramses RF.

## Device diagnostics
[diagnostics.txt](https://github.com/user-attachments/files/25468743/diagnostics.txt)

## Checklist
[x] The changes are tested and work correctly
[x] pre-commit checks pass / the code has been formatted using Black
[x] Tests have been added to verify that the new code works
[x] Device diagnostics data has been attached

### DIFF
--- a/tests/test_elecram.py
+++ b/tests/test_elecram.py
@@ -1,7 +1,6 @@
 """Tests for the ELECRAM RAMSES ESP32-C6 quirk."""
 
 import zhaquirks
-import zhaquirks.elecram.ramses_esp
 from zhaquirks.elecram.ramses_esp import (
     RAMSES_RX_CLUSTER,
     RAMSES_TX_CLUSTER,
--- a/tests/test_elecram.py
+++ b/tests/test_elecram.py
@@ -1,0 +1,80 @@
+"""Tests for the ELECRAM RAMSES ESP32-C6 quirk."""
+
+import zhaquirks
+import zhaquirks.elecram.ramses_esp
+from zhaquirks.elecram.ramses_esp import (
+    RAMSES_RX_CLUSTER,
+    RAMSES_TX_CLUSTER,
+    RamsesESP,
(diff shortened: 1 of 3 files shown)


puddly:
A few questions: 
- This device communicates via character strings. Is the protocol ASCII or can it be condensed? Why not use dedicated Zigbee command IDs for device functionality instead of tunneling an entire protocol?
- How much traffic is exchanged? Zigbee is not a very high bandwidth protocol.
- I noticed mention of ACK strings in both directions. How are they used? Zigbee has an APS ACK bit that you can flip to ask for a delivery receipt. You can also use the default response to convey status codes.

IMMRMKW:
@puddly

1. Yes, the protocol is ASCII. I tried to stay as close as possible to the original protocol as possible which is via USB serial, and MQTT. The device is basically a gateway passing on commands that come via Serial/MQTT/Zigbee to 868 MHz communication. The huge variety of commands and devices that use this protocol and could potentially be communicated with make it impractical to do it another way. This is also how Ramses cc expects it.
2. This depends on the number of devices, but let's say 1 message every 1-2 minutes.
3. As far as I know, the maximum character string length is limited. I thought ZHA is limiting it to like 64 or so, and ESP32 C6 would allow it but the buffer would every once in a while crash. As I did not manage to use this flag you are mentioning in the state machine, I explicitly send an ACK back to manage the flow of the program and prevent the buffer from being filled before a previous message is sent.

PR #4792, opened 2026-02-27
Limit Sonoff TRVZB entities for installed firmware

TheJulianJES:
## Proposed change

This limits Sonoff TRVZB entities based on the installed firmware version.

## Additional information

Related PRs:
- https://github.com/zigpy/zha-device-handlers/pull/4740

## Device diagnostics

TODO

## Checklist

### DIFF
--- a/zhaquirks/sonoff/trvzb.py
+++ b/zhaquirks/sonoff/trvzb.py
@@ -107,18 +107,9 @@ class CustomSonoffCluster(CustomCluster):
         )
 
 
-(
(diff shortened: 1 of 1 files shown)


TheJulianJES:
I'll do this later and verify the entities are correctly capped then.

The TRVZB update will be shown for everyone on 2026.3.0, where this quirk first landed, so users can install that update and use the entities without having to restart Home Assistant to match another quirk and discover new entities.
We should really improve the reconfiguration/reinterviewing... 😅

PR #4793, opened 2026-02-27
Let Hue sensor cluster inherit from `OccupancySensing`

TheJulianJES:
## Proposed change
This removes the `ep_attribute` override for the Hue sensor occupancy cluster. The cluster now also inherits from `OccupancySensing`.

## Additional information
Recently, ZHA introduced support for the `pir_o_to_u_delay` attribute/entity with https://github.com/zigpy/zha/pull/669. This should be supported by the Hue sensors as well. However, the number entities aren't created as they require the `occupancy` `ep_attribute` (cluster handler name), yet the Hue sensors use `philips_occupancy`: [zha/application/platforms/binary_sensor/__init__.py#L199-L209](https://github.com/zigpy/zha/blob/9d03d639ee93f61135c2c02445cae23f8abd6723/zha/application/platforms/binary_sensor/__init__.py#L199-L209).

This was done as the custom Hue attributes conflicted with the ZCL attribute IDs for `physical_contact_u_to_o_delay` + `physical_contact_o_to_u_delay` attributes. As zigpy now properly supports ZCL attributes and manufacturer-specific ones using the same attribute ID, we no longer need this workaround.

## Device diagnostics
Should be in ZHA.

## Checklist

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -773,7 +773,6 @@ def test_attributes_updated_not_replaced(quirk: CustomDevice) -> None:
                 # A few are expected to fail and are handled by ZHA
                 if cluster not in (
                     zhaquirks.konke.KonkeOnOffCluster,
-                    zhaquirks.philips.PhilipsOccupancySensing,
                     zhaquirks.xiaomi.aqara.vibration_aq1.VibrationAQ1.MultistateInputCluster,
                 ):
                     pytest.fail(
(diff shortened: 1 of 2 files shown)


tests/test_quirks.py
```diff
@@ -805,7 +804,10 @@ def test_attributes_updated_not_replaced(quirk: CustomDevice) -> None:
             base_attr_names = {a.name for a in base_cluster.attributes.values()}
             quirk_attr_names = {a.name for a in cluster.attributes.values()}
 
-            if not base_attr_names <= quirk_attr_names:
+            if not base_attr_names <= quirk_attr_names and cluster not in (
+                # XXX: Test to be updated for mf-attributes with same ID as ZCL ones
+                zhaquirks.philips.PhilipsOccupancySensing,
```
TheJulianJES, line 809: This will be done later with:
- https://github.com/zigpy/zha-device-handlers/pull/4794

TheJulianJES:
Tested and working with this ZHA PR:
- https://github.com/zigpy/zha/pull/698

It's a small-ish change and confirmed working. I'll go ahead with this, so users can finally get the timeout entities on Hue sensors.

PR #4794, opened 2026-02-27
Fix `test_attributes_updated_not_replaced` test

TheJulianJES:
DRAFT. Testing.

## Proposed change
- Update `test_attributes_updated_not_replaced` to validate inherited attributes using `cluster._attributes_by_id` instead of flat `.attributes`, so valid manufacturer-specific definitions sharing a ZCL ID are handled correctly.
  - Fixes false positive for: https://github.com/zigpy/zha-device-handlers/pull/4793
- Add a targeted guard that fails quirk-only attributes defined with `manufacturer_code=None` only when the ID collides with a base ZCL attribute, while allowing non-colliding IDs.
- Keep explicit ignore entries for known Sinope cluster issues.

## Device diagnostics
None.

## Checklist

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -818,6 +818,32 @@ def test_attributes_updated_not_replaced(quirk: CustomDevice) -> None:
                                 f" extending it"
                             )
 
+            # Ensure quirk-only attributes on standard clusters are not silently
+            # defined as non-manufacturer-specific (False/None). This catches
+            # attributes that should have a manufacturer code but were defined
+            # with manufacturer_code=None.
+            for attr_id, by_mfr in cluster._attributes_by_id.items():
+                if attr_id not in base_cluster._attributes_by_id:
+                    continue
+                base_names_at_id = {
+                    attr_def.name
+                    for by_code in base_cluster._attributes_by_id.get(
+                        attr_id, {}
+                    ).values()
+                    for attr_def in by_code.values()
+                }
+                for is_mfr, by_code in by_mfr.items():
+                    for mfr_code, quirk_attr in by_code.items():
+                        if quirk_attr.name in base_names_at_id:
+                            continue
+                        if is_mfr is False and mfr_code is None:
+                            pytest.fail(
+                                f"Cluster {cluster} defines quirk-only attribute"
+                                f" {quirk_attr.name!r} (id=0x{attr_id:04X}) as"
+                                " non-manufacturer-specific with"
+                                " manufacturer_code=None"
+                            )
+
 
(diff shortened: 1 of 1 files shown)


tests/test_quirks.py
```diff
@@ -791,25 +787,57 @@ def test_attributes_updated_not_replaced(quirk: CustomDevice) -> None:
...
+                    for mfr_code, base_attr in by_code.items():
+                        quirk_attr = (
+                            cluster._attributes_by_id.get(attr_id, {})
+                            .get(is_mfr, {})
+                            .get(mfr_code)
+                        )
+                        if quirk_attr is None or quirk_attr.name != base_attr.name:
+                            if cluster in (
+                                zhaquirks.sinope.switch.SinopeTechnologiesBasicCluster,
+                                zhaquirks.sinope.switch.SinopeTechnologiesMeteringCluster,
```
TheJulianJES, line 805: The test didn't catch these before but they're really interesting:
Here, the default `power_source` attribute is overwritten (by name) with a manufacturer attribute and different attribute type: https://github.com/zigpy/zha-device-handlers/blob/becb06881a8a9e30a7b04547f071682f293f8589/zhaquirks/sinope/switch.py#L272-L282

And two ZCL attributes on the `Metering` cluster are overwritten here (by name as well) with a manufacturer attribute and different data type: https://github.com/zigpy/zha-device-handlers/blob/becb06881a8a9e30a7b04547f071682f293f8589/zhaquirks/sinope/switch.py#L294-L302

There's a bit more discussion on that here: https://github.com/zigpy/zha-device-handlers/pull/4042#discussion_r2851663679

Ultimately, this should not be allowed, right? Should it be possible for a quirk just to change the data type, e.g. do this without `manufacturer_code` or `is_manufacturer_specific`? Also no, right?

I guess we can just rename the colliding Sinope attributes and we're good?

TheJulianJES:
There is another case we should consider: https://github.com/zigpy/zha-device-handlers/blob/a37b6311ec70c3cd7026557b6e4870c7aca139f3/zhaquirks/bosch/rbsh_trv0_zb_eu.py#L156-L162

Here, a ZCL attribute is overwritten with a manufacturer-specific one with the same name and a different data type, similar to the Sinope examples above, **but the ID is also changed**.
This should "mostly work", as ZHA reads attributes by name but it is likely still unexpected for a ZCL attribute to be replaced like this. We should look into validating this either directly in zigpy (but allowing exceptions?) or by improving this test.

For now, I'll try to see if I can add an exception for the Hue sensor in:
- https://github.com/zigpy/zha-device-handlers/pull/4793 (and unblock that PR)

PR #4197, opened 2025-07-18
Add Tuya solar soil sensor `_TZE284_awepdiwi` variant

rwalker777:
## Checklist

- [x] The changes are tested and work correctly
- [ ] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works

Issue #4143, opened 2025-06-23
[Device Support Request] add soil moisture and temperature sensor Tuya _TZE284_awepdiwi


dethegeek:
### Problem description

I bought the following device : https://fr.aliexpress.com/item/1005008785131043.html

My HomeAssistant is up to date, I had to update ts0601_sensor.py to make it work.

See the following diff: https://github.com/zigpy/zha-device-handlers/compare/dev...dethegeek:zha-device-handlers:patch-1

![Image](https://github.com/user-attachments/assets/a0f8974c-d54b-4dda-af3b-03526bbc9454)

### Solution description

The following diff seems to work: https://github.com/zigpy/zha-device-handlers/compare/dev...dethegeek:zha-device-handlers:patch-1

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

![Image](https://github.com/user-attachments/assets/0b1fb6ca-1114-4f81-8635-20216842bb3b)

</details>

### Device signature

<details><summary>Device signature</summary>

```json

{
  "node_descriptor": {
    "logical_type": 2,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4417,
    "maximum_buffer_size": 66,
    "maximum_incoming_transfer_size": 66,
    "server_mask": 10752,
    "maximum_outgoing_transfer_size": 66,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0004",
        "0x0005",
        "0x0402",
        "0x0408",
        "0xed00",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE284_awepdiwi",
  "model": "TS0601",
  "class": "zigpy.quirks.v2.CustomDeviceV2"
}

```

</details>

### Logs

<details><summary>Logs</summary>

```python

[Paste the logs here]

```

</details>

### Custom quirk

<details><summary>Custom quirk</summary>

```python

(
    TuyaQuirkBuilder("_TZE284_aao3yzhs", "TS0601")
    .applies_to("_TZE284_sgabhwa6", "TS0601")
[...]

PR #4800, opened 2026-03-01
Fix `_TZE284_awepdiwi` soil moisture scale (3x too low)

dougrathbone:
## Proposed change

The `_TZE284_awepdiwi` solar soil moisture sensor reports raw soil moisture values that are approximately **3x lower** than the actual percentage. The device was added in #4197 as a variant of the `_TZE284_aao3yzhs` group, which uses the default `scale=100` for `tuya_soil_moisture`. However, this device needs `scale=300` to produce correct readings.

**Evidence:** With 5 of these sensors deployed alongside `_TZE200_myd45weu` battery-powered soil sensors in the same locations, the solar sensors consistently read ~3x lower (e.g. 18% vs 54%, 39% vs 87% on the battery sensor next to it during heavy rain).

This PR:
- Separates `_TZE284_awepdiwi` from the `_TZE284_aao3yzhs` group into its own quirk definition
- Sets `scale=300` for `tuya_soil_moisture(dp_id=3)` to correct the 3x underreading
- Adds `tuya_electrical_conductivity(dp_id=1)` which this device also supports (confirmed via device data analysis)

Ref: https://github.com/zigpy/zha-device-handlers/issues/4143#issuecomment-2555455775 (@emogenet's analysis confirming `scale=3.0 * 100.0`)

## Additional information

- DP map confirmed from device attribute cache and Zigbee2MQTT converters:
  - dp 1: soil electrical conductivity (µS/cm)
  - dp 3: soil moisture (raw ÷ 3 = %)
  - dp 5: temperature (raw ÷ 10 = °C)
  - dp 14: battery state (enum)
  - dp 15: battery percentage (%)
- Tested with 5 `_TZE284_awepdiwi` devices on Home Assistant 2026.2.3

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/zhaquirks/tuya/tuya_sensor.py
+++ b/zhaquirks/tuya/tuya_sensor.py
@@ -259,11 +259,21 @@ class NoManufTimeTuyaMCUCluster(TuyaMCUCluster):
     .applies_to("_TZE284_sgabhwa6", "TS0601")
     .applies_to("_TZE284_nhgdf6qr", "TS0601")  # Giex GX04
     .applies_to("_TZE284_ap9owrsa", "TS0601")  # Novadigital SG-ZB
-    .applies_to("_TZE284_awepdiwi", "TS0601")  # Solar powered
     .applies_to("_TZE284_33bwcga2", "TS0601")  # iHseno
     .tuya_temperature(dp_id=5, scale=10)
     .tuya_battery(dp_id=15)
     .tuya_soil_moisture(dp_id=3)
+    .skip_configuration()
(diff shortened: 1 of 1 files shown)


dougrathbone:
Closing this PR — empirical testing confirmed the default scale=100 is correct for _TZE284_awepdiwi. The sensor reports raw DP values in the range 0–100 (verified by submerging in water → raw 100, dry air → raw 0). The scale=300 change was an overcorrection.

PR #4802, opened 2026-03-01
Move Zemismart ZM25R3 _TZE200_sq6affpe to V2 cover quirk with correct DPs

bdodoiu:
## Proposed change

Move `_TZE200_sq6affpe` (Zemismart ZM25R3) from the V1 `TuyaZemismartSmartCover0601_3` class to the V2 `_TZE284_3mzb0sdz` builder via `applies_to`.

The V1 class used DPs 2/3 for position, but the actual device uses DPs 8/9 (confirmed via https://github.com/Koenkk/zigbee-herdsman-converters/issues/9803). This meant position control was non-functional. The V2 builder from #4779 already has the correct DP mapping, so this PR simply adds the `_TZE200_sq6affpe` manufacturer ID to it, gaining:

- Working position control (DP 9)
- Position state reporting (DP 8)
- Battery support (DP 13)
- Motor direction setting (DP 11)
- Border/limit buttons (DP 16)
- Click control (DP 20)

## Additional information

Tested on a physical Zemismart ZM25R3 device running on HA 2026.3.1.
Open/close/stop, position slider, position inversion, and battery reporting all confirmed working.

This is a fix for existing users of `_TZE200_sq6affpe` whose position control was broken under the V1 quirk.

## Checklist


TheJulianJES (commented):
From a quick look, it does seem like there are some differences in Z2M between these models:
- Previous model: https://github.com/Koenkk/zigbee-herdsman-converters/blob/40f9a4298695a289fda4f246c99778b7de70e452/src/devices/zemismart.ts#L966-L999
- New model (this PR): https://github.com/Koenkk/zigbee-herdsman-converters/blob/40f9a4298695a289fda4f246c99778b7de70e452/src/devices/zemismart.ts#L87-L122

But you can confirm that everything is fully working? Enums? Border stuff etc.?

bdodoiu:
> From a quick look, it does seem like there are some differences in Z2M between these models:
> 
> * Previous model: https://github.com/Koenkk/zigbee-herdsman-converters/blob/40f9a4298695a289fda4f246c99778b7de70e452/src/devices/zemismart.ts#L966-L999
> * New model (this PR): https://github.com/Koenkk/zigbee-herdsman-converters/blob/40f9a4298695a289fda4f246c99778b7de70e452/src/devices/zemismart.ts#L87-L122
> 
> But you can confirm that everything is fully working? Enums? Border stuff etc.?

Interesting find, but I believe that the issue is actually in Z2M.
ZM25R1 (mains powered) and ZM25R3 (battery powered) seem to have different DPs (ZM25R3 is closer to the ZM16B).
I validated that the following works:
* set/reset upper/lower limit
* motor direction
* click control (step up/step down)

I also tried to make the set/reset middle position work, but didn't get anywhere (it seems that the Tuya app is missing this functionality as well, it only works with the RF remote, so maybe it's not supported via Zigbee)

PR #4632, opened 2026-01-08
Add Tuya TS004F rotary knob variant (_TZ3000_402vrq2i)

jrfigura:
Added _TZ3000_402vrq2i suport to TuyaSmartRemote004FRotary as suggested in #3738

closes #3738

## Proposed change

Added support for _TZ3000_402vrq2i variant
What works: "Single press, double press, long press, rotation events confirmed working"

## Additional information

Tested on Home Assistant with ZHA integration. All button functions work: single press, double press, long press, and rotation in both directions.

## Device diagnostics

[zha-cd7c1fa6596ea299d3f70e212321a83d-_TZ3000_402vrq2i TS004F-dfd5a4b5bee3a42176ef5aa8b2787779.json](https://github.com/user-attachments/files/25335873/zha-cd7c1fa6596ea299d3f70e212321a83d-_TZ3000_402vrq2i.TS004F-dfd5a4b5bee3a42176ef5aa8b2787779.json)

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [x] Tests have been added to verify that the new code works
- [x] Device diagnostics data has been attached

PR #4820, opened 2026-03-11
Add Tuya smart knob _TZ3000_402vrq2i variant 

atomiclama:
Simple add of model support

First PR here, be gentle ;-)


jrfigura:
there is already PR open for this change: https://github.com/zigpy/zha-device-handlers/pull/4632

atomiclama:
> there is already PR open for this change: #4632

So there is, not sure how I missed that in my search

PR #4830, opened 2026-03-16
Add Third Reality 3RPL01084Z 60g radar settings

3reality-support:
## Proposed change

This PR added added 3 additional private clusters and modify attribute name
1. tvoc_sensor_calibration: Calibrate the sampling sensor
2. sensor_sensitivity:  Used to set the sensitivity of sensors
3. air_threshold: When the VOC measurement value exceeds this threshold, the device will flash a red light, indicating poor air quality and danger
## Checklist


zhaquirks/thirdreality/60g_radar.py
```diff
@@ -10,32 +10,77 @@
...
+        )
+
 
 (
     QuirkBuilder("Third Reality, Inc", "3RPL01084Z")
     .replaces(ThirdRealityRadarCluster)
     .sensor(
         endpoint_id=1,
-        attribute_name=ThirdRealityRadarCluster.AttributeDefs.volatile_organic_compounds.name,
+        attribute_name=ThirdRealityRadarCluster.AttributeDefs.total_volatile_organic_compounds.name,
```
TheJulianJES: When we adjust the `attribute_name`, the unique ID of the entity changes. So, the old entity would no longer be provided by ZHA, and a new one would be created. This is not good.
However, as the PR adding this sensor was only accepted and merged a few days ago (https://github.com/zigpy/zha-device-handlers/pull/4655) and is not yet part of a Home Assistant release, we could change it.

But I don't see any reason to add the `total_` prefix here. So, I would change the `attribute_name` (both here and in the definition) back to `volatile_organic_compounds`. Or is there a reason we need to include "total" here?

So when we also rename the attribute definition again, then we can change this like so:
```suggestion
        attribute_name=ThirdRealityRadarCluster.AttributeDefs.volatile_organic_compounds.name,
```

zhaquirks/thirdreality/60g_radar.py
```diff
@@ -10,32 +10,77 @@
...
+    """Third Reality's 60G radar private cluster."""
 
     cluster_id = 0x042E
 
     class AttributeDefs(BaseAttributeDefs):
         """Define the attributes of a private cluster."""
 
         # measure the value of voc
-        volatile_organic_compounds: Final = ZCLAttributeDef(
+        total_volatile_organic_compounds: Final = ZCLAttributeDef(
```
TheJulianJES: Per the other comment, I don't think we should rename the attribute here to include a new `total_` prefix. This change should be reverted IMO, so be like tthis:
```suggestion
        volatile_organic_compounds: Final = ZCLAttributeDef(
```

zhaquirks/thirdreality/60g_radar.py
```diff
@@ -10,32 +10,77 @@
...
         endpoint_id=1,
-        attribute_name=ThirdRealityRadarCluster.AttributeDefs.volatile_organic_compounds.name,
+        attribute_name=ThirdRealityRadarCluster.AttributeDefs.total_volatile_organic_compounds.name,
         cluster_id=ThirdRealityRadarCluster.cluster_id,
         device_class=SensorDeviceClass.VOLATILE_ORGANIC_COMPOUNDS_PARTS,
         state_class=SensorStateClass.MEASUREMENT,
         unit=CONCENTRATION_PARTS_PER_BILLION,
-        fallback_name="Volatile organic compounds parts",
+        translation_key="total_volatile_organic_compounds",
+        fallback_name="Total volatile organic compounds",
```
TheJulianJES, line 67: For the entity name, is there a reason we need to add "Total" here? (This is not directly related to the attribute name. They can differ, without any issues.)

As we use `device_class=SensorDeviceClass.VOLATILE_ORGANIC_COMPOUNDS_PARTS`, Home Assistant would use the default name of "Volatile organic compounds parts" (from [here](https://github.com/home-assistant/core/blob/a57c65f5121d121380128afe8295cc1433963855/homeassistant/components/sensor/strings.json#L303-L305)) for that entity if the `translation_key` is NOT provided. So, that's what happened before the change here.

Now, "Total" is added and a `translation_key` is provided. Is there any reason the user-facing entity name should contain "Total" and not "parts" at the end? For `ppb` units, HA generally uses "Volatile organic compounds parts".

If we can revert the change (and use "Volatile organic compounds parts" as the HA entity name again by using that name from the device class), the change would be this:

```suggestion
        fallback_name="Volatile organic compounds parts",
```
(The `fallback_name` would not actually be used by HA, as no `translation_key` is present, but a device class is present. So, this is just for informational purposes in the quirk then.)

zhaquirks/thirdreality/60g_radar.py
```diff
@@ -10,32 +10,77 @@
...
+    )
+    .number(
+        attribute_name=ThirdRealityRadarCluster.AttributeDefs.air_threshold.name,
+        min_value=3000,
+        max_value=15000,
+        step=1,
+        cluster_id=ThirdRealityRadarCluster.cluster_id,
+        translation_key="air_threshold",
+        fallback_name="Air threshold",
     )
```
TheJulianJES, line 95: What does this entity do? Is there some sort of light on the device that changes if this threshold is reached?

Also, should `ppb` be added as the unit? If so, we could also added the device class for VOC.
(I'd also recommend moving `cluster_id` further up, as in the example below, so the attribute order is always `attribute_name`, `cluster_id`, ..., `translation_key`, and `fallback_name`. That just makes it a bit easier to read)

Example of how the change would look:

```suggestion
    .number(
        attribute_name=ThirdRealityRadarCluster.AttributeDefs.air_threshold.name,
        cluster_id=ThirdRealityRadarCluster.cluster_id,
        min_value=3000,
        max_value=15000,
        step=1,
        device_class=NumberDeviceClass.VOLATILE_ORGANIC_COMPOUNDS_PARTS,
        unit=CONCENTRATION_PARTS_PER_BILLION,
        translation_key="air_threshold",
        fallback_name="Air threshold",
)
```
Do note that `from zigpy.quirks.v2 import NumberDeviceClass` also has to be imported at the top. I can't add it to the suggestion here.

zhaquirks/thirdreality/60g_radar.py
```diff
@@ -10,32 +10,77 @@
...
+        fallback_name="Total volatile organic compounds",
+    )
+    .write_attr_button(
+        cluster_id=ThirdRealityRadarCluster.cluster_id,
+        attribute_name=ThirdRealityRadarCluster.AttributeDefs.tvoc_sensor_calibration.name,
+        attribute_value=0x01,
+        endpoint_id=1,
+        translation_key="tvoc_sensor_calibration",
+        fallback_name="TVOC sensor calibration",
+    )
```
TheJulianJES, line 75: Hmm. In the other PR, I renamed the button entity to "Calibrate sensor", as that should describe more what pressing the button does. So, if we want to follow that, this would be "Calibrate TVOC sensor", I guess:
```suggestion
    .write_attr_button(
        cluster_id=ThirdRealityRadarCluster.cluster_id,
        attribute_name=ThirdRealityRadarCluster.AttributeDefs.tvoc_sensor_calibration.name,
        attribute_value=0x01,
        endpoint_id=1,
        translation_key="calibrate_tvoc_sensor",
        fallback_name="Calibrate TVOC sensor",
    )
```

TheJulianJES (commented):
I think this mostly looks good. Couple of notes below.

zhaquirks/thirdreality/60g_radar.py
3reality-support: Yes, when the set threshold is exceeded, a red light will turn on

zhaquirks/thirdreality/60g_radar.py
3reality-support: Because the sensor's manual clearly states it is TVOC, we had it wrong before.

3reality-support: Submit to modify device class

zhaquirks/thirdreality/60g_radar.py
```diff
@@ -32,10 +55,39 @@ class AttributeDefs(BaseAttributeDefs):
...
+        fallback_name="Calibrate TVOC sensor",
+    )
+    .number(
+        attribute_name=ThirdRealityRadarCluster.AttributeDefs.sensor_sensitivity.name,
+        min_value=1,
+        max_value=6,
+        step=1,
+        cluster_id=ThirdRealityRadarCluster.cluster_id,
+        translation_key="sensor_sensitivity",
+        fallback_name="Presence sensor sensitivity",
```
TheJulianJES: Ah, tests are currently failing because another quirk now uses the translation key `sensor_sensitivity` with `fallback_name` "Sensor sensitivity". But this quirk uses it with "Presence sensor sensitivity".
Translation key and `fallback_name` have a 1:1 mapping. This is just relevant for how we generate translations later.

We can do either of these changes, depending on which name you want to show in HA.
To show "Presence sensor sensitivity":
```suggestion
        translation_key="presence_sensor_sensitivity",
        fallback_name="Presence sensor sensitivity",
```
OR to show only "Sensor sensitivity":
```suggestion
        translation_key="sensor_sensitivity",
        fallback_name="Sensor sensitivity",
```

For reference, the failing test:
```python
FAILED tests/test_quirks_v2.py::test_translation_key_and_fallback_name_match - AssertionError: Translation key 'sensor_sensitivity' is shared by quirks with different fallback names: ('/home/runner/work/zha-device-handlers/zha-device-handlers/zhaquirks/thirdreality/24G_radar.py:33', '/home/runner/work/zha-device-handlers/zha-device-handlers/zhaquirks/thirdreality/60g_radar.py:52')
assert 2 == 1
 +  where 2 = len({'Presence sensor sensitivity', 'Sensor sensitivity'})
 +    where {'Presence sensor sensitivity', 'Sensor sensitivity'} = set(('Sensor sensitivity', 'Presence sensor sensitivity'))
```

zhaquirks/thirdreality/60g_radar.py
TheJulianJES: Perfect, thanks!

zhaquirks/thirdreality/60g_radar.py
TheJulianJES: Ah, I wasn't aware there was a difference between TVOC and VOC 😅.

From the [HA developers page](https://developers.home-assistant.io/docs/core/entity/sensor/#:~:text=SensorDeviceClass.VOLATILE_ORGANIC_COMPOUNDS):
`SensorDeviceClass.VOLATILE_ORGANIC_COMPOUNDS` accepts units: `µg/m³` + `mg/m³`
`SensorDeviceClass.VOLATILE_ORGANIC_COMPOUNDS_PARTS` accepts units: `ppm` + `ppb`

`CONCENTRATION_PARTS_PER_BILLION` is just `ppb`, so we can only use `SensorDeviceClass.VOLATILE_ORGANIC_COMPOUNDS_PARTS` if the unit is correct then. This is just how HA works.

So I guess if it's important to have the "Total" in the entity name, we could indeed add the `translation_key` and `fallback_name` for that:
```python
        translation_key="total_volatile_organic_compounds",
        fallback_name="Total volatile organic compounds",
```
Only two other HA integrations seem to do this though..

3reality-support: If I want to use fallback_name="Total volatile organic compounds" in ppb, how should I modify it

3reality-support: 7284cb67-0cc3-4c09-9cbf-63acbd34916f
be3d8668-61d5-44ec-9fa0-59af94a46a3b

puddly: Home Assistant doesn't itself differentiate between various VOCs (e.g. formaldehyde) so I think its VOC (ppb) sensor device class is correct here, since it means TVOC anyways.

3reality-support: It has been modified to tvoc and uses µ g/m ³ units. The device reports ppb and converts it into µ g/m ³ through a script

TheJulianJES: I've simplified this a bit by using an `attribute_converter` for the entity, but I'm not sure if this is really correct.
I think we should just show the raw ppb output. At most, we can change add a `translation_key` with a `fallback_name`, so the "Total" part becomes clear. But there's no differentiation on the HA side here, really.

It's also a bit weird because "Air threshold" is still in ppb now. That's not as easy to modify into another unit though. I argue we also shouldn't do it here.

Either way, I'll merge this for now, since the HA beta cut-off is tomorrow. We may be able to revert to ppb in the beta time if you think that's better.

TheJulianJES (approved):
I'll merge this for now but please see the comment here: https://github.com/zigpy/zha-device-handlers/pull/4830#discussion_r2984972601

I think we should just show ppb and at most change the name (like we already do now).

zhaquirks/thirdreality/60g_radar.py
3reality-support: d5a69e08-ce5c-4f02-b1b4-82b0ae1d2796
aa998a14-4b25-4a53-9ab9-2e654a1e16b2
Can it be used like this

3reality-support: @TheJulianJES

TheJulianJES: Yep, that works. `VOLATILE_ORGANIC_COMPOUNDS_PARTS` accepts `ppb`, so just the name is changed to include "Total". That's fine. Do you quickly want to submit a PR with that, or should I change it?

TheJulianJES: This would change it to `ppb`, but keep the "Total" in the name: https://github.com/zigpy/zha-device-handlers/pull/4862

3reality-support: Thanks for your help

PR #4832, opened 2026-03-16
Downgrade log message after loading custom quirks to INFO from WARNING

mgiuffrida:
## Proposed change

This one-line change downgrades the logging call when custom quirks are loaded down to `info` from `warning`

- Warnings appear in the condensed system logs, which novice users like myself use at the first sign of trouble. Noisy log entries unrelated to potential problems shouldn't appear here.
- The [`INFO` level](https://docs.python.org/3/library/logging.html#logging.INFO) is a better fit for this message. `logging.WARNING` is `An indication that something unexpected happened, or that a problem might occur in the near future`. The existence of a custom quirk isn't any more of a sign of future problems than the existence of user-written automations or YAML already are.

## Additional information

There's arguably some value in prompting users to contribute custom quirks to the project, but the system error/warning log doesn't seem like the appropriate place.

Most users are *consuming* these custom quirks, not writing them, and probably aren't likely to have enough experience to know how to "contribute them" to the repository. 

## Device diagnostics

N/A, not a quirk

## Checklist

### DIFF
--- a/zhaquirks/__init__.py
+++ b/zhaquirks/__init__.py
@@ -510,7 +510,7 @@ def setup(custom_quirks_path: str | None = None) -> None:
             loaded = True
 
     if loaded:
-        _LOGGER.warning(
+        _LOGGER.info(
             "Loaded custom quirks. Please contribute them to"
             " https://github.com/zigpy/zha-device-handlers"
         )


puddly:
Custom quirks are meant for development. They're not something you should be running as a normal user of ZHA and their presence is a constant source of device breakage, database issues, and general instability. The message is logged as a warning for this reason.

Home Assistant Core itself logs something similar if you are using a custom component for the same reason: https://github.com/home-assistant/core/blob/39b44445ecfc64f6cb8b18ec235df3cfff671005/homeassistant/loader.py#L148-L153

mgiuffrida:
Didn't realize that. I see now there's a difference between a quirk and a *custom* quirk. That's fair.

I don't see custom component logs for HACS or for custom components I've forked. (I'm getting confused trying to follow the logic of HA's `loader.py` as it seems to treat `overwrites_built_in` as mutually exclusive with `is_built_in` in some places but not others, but that's a tangent.)

PR #4833, opened 2026-03-18
Add Third Reality curtain gen2 Last remaining battery percentage entity

3reality-support:
## Proposed change

This PR added added 1 additional private clusters entity

## Checklist


TheJulianJES (commented):
I think this looks good, but I do have to ask:

What is the difference between the "Last remaining battery percentage" and the normal "Battery percentage" entity? The device provides both, right? Do they show different values (always/sometimes)?

3reality-support:
> I think this looks good, but I do have to ask:
> 
> What is the difference between the "Last remaining battery percentage" and the normal "Battery percentage" entity? The device provides both, right? Do they show different values (always/sometimes)?

The 'Last remaining battery percentage' indicates the remaining amount of electricity used during the last power on to power off process. As the battery is rechargeable, using this attribute makes it convenient for users to calculate how long it takes to charge

TheJulianJES (approved):
Thanks!

PR #4835, opened 2026-03-18
Add TS0601 _TZE284_5m4nchbm RCBO quirk

alvarogl:
## Proposed change

Add support for the Tuya RCBO variant `_TZE284_5m4nchbm` / `TS0601`.

This PR adds a new v2 quirk entry in `zhaquirks/tuya/ts0601_rcbo.py` that maps the datapoints from a working Zigbee2MQTT converter, including the composite raw datapoints `48` and `49` used for the alarm/breaker threshold payloads. The implementation preserves the reserved bytes in those payloads when writing updated thresholds back to the device.

The PR also adds tests to verify:
- datapoint report parsing for the new variant
- serialized write payloads for normal enum writes and composite raw writes
- no regressions in quirks v2 registration

## Additional information

This is additive only and should not be a breaking change.

The device diagnostics used for this work show the unsupported device before this quirk is applied:
- manufacturer: `_TZE284_5m4nchbm`
- model: `TS0601`
- power source: `Mains`
- device type: `Router`
- `quirk_applied: false`

Validation run locally:
- `uv run pytest tests/test_tuya_rcbo.py -q`
- `uv run pytest tests/test_quirks_v2.py -q`
- `uv run pre-commit run --all-files`

## Device diagnostics

[zha-01KKZ1NW5GS19MYYHNAPKR43FC-_TZE284_5m4nchbm TS0601-bb977f4b1014ef6e4cf4f7a3f15e139a.json](https://github.com/user-attachments/files/26101362/zha-01KKZ1NW5GS19MYYHNAPKR43FC-_TZE284_5m4nchbm.TS0601-bb977f4b1014ef6e4cf4f7a3f15e139a.json)

## Checklist

### DIFF
--- a/zhaquirks/tuya/ts0601_rcbo.py
+++ b/zhaquirks/tuya/ts0601_rcbo.py
@@ -39,7 +39,12 @@ from zhaquirks.const import (
     OUTPUT_CLUSTERS,
     PROFILE_ID,
 )
-from zhaquirks.tuya import TUYA_MCU_COMMAND, AttributeWithMask, PowerOnState
+from zhaquirks.tuya import (
+    TUYA_CLUSTER_ID,
+    TUYA_MCU_COMMAND,
+    AttributeWithMask,
+    PowerOnState,
+)
 from zhaquirks.tuya.builder import TuyaQuirkBuilder
 from zhaquirks.tuya.mcu import (
     DPToAttributeMapping,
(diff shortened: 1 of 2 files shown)


zhaquirks/tuya/ts0601_rcbo.py
```diff
@@ -556,3 +570,551 @@ class TuyaCircuitBreaker(CustomDevice):
...
+        under_voltage_threshold: Final = ZCLAttributeDef(
+            id=0xEF21, type=t.uint16_t, is_manufacturer_specific=True
+        )
+
+    dp_to_attribute: dict[int, list[DPToAttributeMapping]] = {
+        1: [DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "state")],
+        17: [DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "energy")],
+        18: [DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "current")],
+        19: [DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "power")],
+        20: [DPToAttributeMapping(TuyaMCUCluster.ep_attribute, "voltage")],
```
TheJulianJES: We also have methods, like `tuya_dp`, in the `TuyaQuirkBuilder` to both create the local attributes and the mappings. Is there a reason why you've done it like this instead?
For reference, see:
https://github.com/zigpy/zha-device-handlers/blob/c5634ab01c6e5e5f0828489eae3153a6fa416564/zhaquirks/tuya/builder/__init__.py#L641-L671
and: https://github.com/zigpy/zha-device-handlers/blob/dev/tuya.md#tuya_dp_attribute.

alvarogl: Refactored this to use the v2 `TuyaQuirkBuilder` DP helpers instead of the hand-written `AttributeDefs` / `dp_to_attribute` block.

Concretely:
- single-DP entities now use `tuya_switch`, `tuya_sensor`, and `tuya_enum`
- the composite DP48/DP49 payloads now use `tuya_attribute` + `tuya_dp_multi` with the same reserved-byte-preserving write converters as before
- the custom `TuyaRCBO2ManufCluster` class is gone, so the manufacturer-cluster attributes/mappings are now declared the way `tuya.md` and the builder API recommend

I reran:
- `uv run pytest tests/test_tuya_rcbo.py -q`
- `uv run pytest tests/test_quirks_v2.py -q`
- `uv run pre-commit run --files zhaquirks/tuya/ts0601_rcbo.py tests/test_tuya_rcbo.py`

Thanks, that is a better fit for this v2 quirk.

alvarogl:
Hello @TheJulianJES, 
Do you think we can include this integration in the repository?

Thanks,
Álvaro

PR #4838, opened 2026-03-20
Add Third Reality temperature sensor gen2 3RTHS0324Z settings

3reality-support:
## Proposed change

Add a new device and modify the manufacturing

## Checklist


zhaquirks/thirdreality/temperature_sensor.py
```diff
@@ -21,20 +21,20 @@ class AttributeDefs(BaseAttributeDefs):
         temperature_offset_celsius: Final = ZCLAttributeDef(
             id=0x0031,
             type=t.int16s,
-            manufacturer_code=0x1233,
```
TheJulianJES, line 24: Since a few months, it's preferred to explicitly provide the `manufacturer_code` for each attribute, instead of just setting `is_manufacturer_specific`.

zhaquirks/thirdreality/temperature_sensor.py
TheJulianJES, line 131: Why do we not just add the new device here?
```suggestion
(
    base_quirk.clone()
    .applies_to("Third Reality, Inc", "3RTHS24BZ")
    .applies_to("Third Reality, Inc", "3RTHS0324Z")
    .add_to_registry()
)  # fmt: skip
```
(I'm not sure if the GitHub web suggestion here works. You may need to apply this manually.)

zhaquirks/thirdreality/temperature_sensor.py
```diff
@@ -79,3 +79,5 @@ class AttributeDefs(BaseAttributeDefs):
     .removes(PollControl.cluster_id)
     .add_to_registry()
 )
+
+(base_quirk.clone().applies_to("Third Reality, Inc", "3RTHS0324Z").add_to_registry())
```
TheJulianJES: This could be removed entirely then:
```suggestion
```

zhaquirks/thirdreality/temperature_sensor.py
3reality-support: If the manufacture ID of the new product is 1407, do I need to rewrite a class?

zhaquirks/thirdreality/temperature_sensor.py
```diff
@@ -67,6 +95,35 @@ class AttributeDefs(BaseAttributeDefs):
...
+    QuirkBuilder()
+    .replaces(ThirdRealityClusterGen2)
+    .number(
+        attribute_name=ThirdRealityClusterGen2.AttributeDefs.temperature_offset_celsius.name,
+        cluster_id=ThirdRealityClusterGen2.cluster_id,
+        min_value=-10000,
+        max_value=10000,
+        multiplier=0.01,
+        step=0.1,
+        device_class=NumberDeviceClass.TEMPERATURE,
```
TheJulianJES: We can now use a new device class for these offset temperatures (to be correctly translated into °F):
```suggestion
        device_class=NumberDeviceClass.TEMPERATURE_DELTA,
```

TheJulianJES (approved):
Is there a reason why you have multiple manufacturer codes?

And a couple of notes: The base quirk stuff could be changed, so that the base quirk adds the offset entities and the other quirks just add `replaces` and `applies_to`. Then, the entity definitions aren't duplicated.

But it's not a huge deal. Just unexpected that multiple manufacturer codes exist.
Thanks!

PR #4862, opened 2026-03-25
Change unit of Third Reality 60G radar TVOC sensor

TheJulianJES:
## Proposed change

Change unit of Third Reality 60G radar TVOC sensor from `µg/m³` to `ppb`.
We do keep the "Total" in the name though. This will be shown to HA users.

## Additional information

For reference, see the discussion here:
- https://github.com/zigpy/zha-device-handlers/pull/4830#discussion_r2955923843

This also matches the "Air threshold" entity then, which is already in `ppb`.

## Device diagnostics

N/A

## Checklist


TheJulianJES:
I'll merge this now, as we need to get ready for the HA beta. Let's hope we still make it at all 😄

PR #4879, opened 2026-03-30
Improve battery calculation for frient/Develco Humidity Sensors (HMSZB-110/120)

marcintyminski:
## Proposed change

The quirk has been made to extend functionality of the following frient/Develco Humidity Sensors:
 - frient Smart Humidity Sensor with model no. HMSZB-110
 - frient Smart Humidity Sensor with model no. HMSZB-120
 - Develco Smart Humidity Sensor with model no. HMSZB-110
 - Develco Smart Humidity Sensor with model no. HMSZB-120
 
 - Add compatible models
 - Adjust the PowerConfiguration voltage bounds

## Additional information

image

## Device diagnostics

[zha-01KJYXWD9VV3FMNDJEFN115MC4-frient A_S HMSZB-120-0299b8a59a4f0e1417138f0bc7e06d06.json](https://github.com/user-attachments/files/26346967/zha-01KJYXWD9VV3FMNDJEFN115MC4-frient.A_S.HMSZB-120-0299b8a59a4f0e1417138f0bc7e06d06.json)
[zha-01KJYXWD9VV3FMNDJEFN115MC4-Develco Products A_S HMSZB-110-5d0873109636666c8558e677197f081a.json](https://github.com/user-attachments/files/26346968/zha-01KJYXWD9VV3FMNDJEFN115MC4-Develco.Products.A_S.HMSZB-110-5d0873109636666c8558e677197f081a.json)

## Checklist


zhaquirks/develco/humidity.py
```diff
@@ -1,11 +1,298 @@
...
+        cluster_id=RelativeHumidity.cluster_id,
+        endpoint_id=38,
+        min_value=-10,
+        max_value=10,
+        step=1,
+        unit=PERCENTAGE,
+        translation_key="humidity_offset",
+        fallback_name="Humidity offset",
+        unique_id_suffix="humidity_offset",
+    )
```
TheJulianJES: Unfortunately, we can't add virtual config entities to Home Assistant Core – only entities where we really change a setting on the device are allowed.

You can achieve the same behavior by using templates in Home Assistant. It's not as user-friendly but it's to avoid cluttering different integrations with a lot of config entities that aren't "real".

TheJulianJES (commented):
Thanks for all the PRs! I've yet to have a close look at them, but we'll need to remove virtual config entities per Home Assistant rules, like "Temperature offset" and "Humidity offset".

This is also partly why it's so awkward to create these using quirks. Quirks are generally intended to expose simple config entities based on a single attribute. It's possible to expose other things as well, and you sometimes need "virtual ZCL attributes" to split up real device functionality, but even that is not great.

I'll need to have a closer look at the keypad and siren PRs. These device types are somewhat rare and have been neglected a bit. We may want to move some of those features directly into ZHA (instead of zha-quirks), but I'll let you know if I need anything more from you.

If you don't have enough time to further work on these PRs, I'll try to make the necessary changes soon. For now, just removing the virtual offset entities would be good.

tests/test_develco.py
```diff
@@ -177,3 +180,52 @@ async def test_mfg_cluster_events(zigpy_device_from_v2_quirk):
...
+                TemperatureMeasurement.cluster_id: ClusterType.Server,
+                RelativeHumidity.cluster_id: ClusterType.Server,
+            }
+        },
+    )
+
+    endpoint = device.endpoints[38]
+    assert isinstance(endpoint.power, HumidityPowerConfiguration)
+    assert isinstance(endpoint.temperature, TemperatureMeasurement)
+    assert isinstance(endpoint.humidity, RelativeHumidity)
```
TheJulianJES: @zigpy-review-bot Remove the `test_hmszb_120_replaces_power_config_and_exposes_measurement_clusters` test. We don't need to test pure definitions like this, as this is/will already done in ZHA diagnostics snapshot tests.

The `test_hmszb_120_power_config_battery_percent_from_voltage` also isn't really required because it's just the generic quirked `PowerConfigurationCluster` cluster in the end but that test is fine to keep. But it's not ideal to call the internals: `_calculate_battery_percentage`. `update_attribute` is already called, so we should be able to check the attribute cache. Do what works best here and push the changes to this PR/branch.

TheJulianJES (approved):
Thanks!

PR #4902, opened 2026-04-04
Add Sonoff MINI DUO-L - no neutral (ZB2GS-L) quirk

### LINKED ISSUE #4667: [Device Support Request] Sonoff MINI-ZB2GS - Add support for External trigger / Turbo / Detach relay options
### Problem description

Hi ZHA Device Handlers team,

I’d like to request device support/enhancement for the following Sonoff models:
Sonoff MINI-ZB2GS
Sonoff MINI-ZB2GS-L

Problem / Missing features
In ZHA, these models are currently missing the following configuration options:
External trigger mode
Turbo mode
Detach relay

These options are available on the Sonoff ZBMINIR2 model, which already has a dedicated quirk implementing the relevant configuration cluster/exposes.

### Solution description

Reference implementation
The feature support appears to have been added for ZBMINIR2 here:
zhaquirks/sonoff/zbminir2.py
https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/sonoff/zbminir2.py

Request
Could similar functionality be added for MINI-ZB2GS and MINI-ZB2GS-L?

Given the existing implementation for ZBMINIR2, it looks like this could potentially be extended quite easily by adding the corresponding model identifiers / signature matching and reusing the same/custom cluster logic.

Expected outcome
Once supported, these device options should become available in ZHA (Home Assistant) for MINI-ZB2GS and MINI-ZB2GS-L in the same way as ZBMINIR2.

### Screenshots/Video

_No response_

### Diagnostics information

Well it's not available..

### Device signature

```json

```

</details>

### Logs

```python

[Paste the logs here]

```

</details>

### Custom quirk

```python

```

</details>

### Additional information

_No response_

uschindler:
This device is basically identical to the MINI ZB2GS (which has neutral connection), but has fewer features due to missing neutral connection:
- No network LED
- No turbo mode

This is a followup PR and reuses https://github.com/zigpy/zha-device-handlers/pull/4742

## Proposed change
Reuse the quirk that's already there for the Mini-ZB2GS. As the Mini-ZB2GS-L has less features, refactor code a bit to reuse common code by first creating the smaller quirk (with less features) for the ZB2GS-L (new device) and later clone it and add the missing config entities for the ZB2GS device. This keeps code small.

## Device diagnostics
https://github.com/user-attachments/files/26517518/zha-01KN1ZMJ05EV56PWVPE42Q7WVK-SONOFF.MINI-ZB2GS-L-308e967a8cfdd208d66b895cc6e64277.json

Provided by @Crocmagnon.

## Checklist

This closes #4667

### DIFF
--- a/zhaquirks/sonoff/mini_zb2gs.py
+++ b/zhaquirks/sonoff/mini_zb2gs.py
@@ -1,4 +1,4 @@
-"""Sonoff MINI-ZB2GS - Zigbee Switch."""
+"""Sonoff MINI-ZB2GS and MINI-ZB2GS-L - Zigbee Switches."""
 
 from zigpy import types
 from zigpy.quirks import CustomCluster
@@ -57,8 +57,8 @@ class SonoffCluster(CustomCluster):
         )
 
 
-(
-    QuirkBuilder("SONOFF", "MINI-ZB2GS")
+zb2gs_l_quirk = (
+    QuirkBuilder("SONOFF", "MINI-ZB2GS-L")
     .replaces(SonoffCluster, endpoint_id=1)
     .replaces(SonoffCluster, endpoint_id=2)
     .enum(
@@ -85,6 +85,12 @@ class SonoffCluster(CustomCluster):
         translation_key="external_trigger_mode",
         fallback_name="External trigger mode",
     )
+)
+zb2gs_l_quirk.add_to_registry()
+
+zb2gs_quirk = (
+    zb2gs_l_quirk.clone()
+    .applies_to("SONOFF", "MINI-ZB2GS")
     .switch(
         SonoffCluster.AttributeDefs.turbo_mode.name,
         SonoffCluster.cluster_id,
@@ -101,5 +107,5 @@ class SonoffCluster(CustomCluster):
         translation_key="network_led",
         fallback_name="Network LED",
     )
-    .add_to_registry()
 )
+zb2gs_quirk.add_to_registry()


uschindler:
@TheJulianJES: You may remember my work on the previous quirk #4742. I created this one to support and close the corresponding issue #4667, because the second no-neutral device requested there was missing.

The code should work, I don't own this device, but the person on the issue will test it. I can't supply a device diagnostics, I can ask the person. I am not sure if we need it here. It is basically the same device with a different name and a bit less features, that i removed from the device specific quirk.

I will report back once it is confirmed to work.

P.S.: I will provide a PR for ZBMINIR2, MINI ZB2GS, and MINI ZB2GS-L soon that hides the "opening" sensor and instead adds a device trigger. The opening sensor is always confusing people because the state just toggles (as the device only sends "toggle" events which only toggles the zigpy binary state so it never reflects realy opening state). So in detached relay mode we should really only provide a device trigger on the onoff toggle event!

uschindler:
One question: People complained that because the device has 2 endpoints (channels/buttons), the entities created have exact same name in Home Assistant.

Do you have a good suggestion to fix this? The "Trigger mode" setting could be renamed here in the PR, but the relay on/of switches and the opening sensor (as added by ZHA automatically) have the exact same name ("Switch", "Opening").

Would this be an issue for ZHA or Zigpy so it creates better entity names, if several endpoints of a device expose the same settings or sensors or switches? In short: When a device has a Switch for endpoint 1 and endpoint 2 or a setting "trigger mode" from a quirk also for bothe endpoints, it should add the enpoint ID at end of name. "Switch 1", "Switch 2", "Trigger Mode 1", "Trigger mode 2".

I found no way to rename the default created entities for the on/off cluster to have the endpoint ID in their name. The same of couse also applies for the "Trigger Mode" added in this quirk - 2 endpoints same entity name.

uschindler:
I already got a confirmation: https://github.com/zigpy/zha-device-handlers/issues/4667#issuecomment-4187578401

uschindler:
> P.S.: I will provide a PR for ZBMINIR2, MINI ZB2GS, and MINI ZB2GS-L soon that hides the "opening" sensor and instead adds a device trigger. The opening sensor is always confusing people because the state just toggles (as the device only sends "toggle" events which only toggles the zigpy binary state so it never reflects realy opening state). So in detached relay mode we should really only provide a device trigger on the onoff toggle event!

Here is the cleanup for the device triggers and the "incorrectly behaving" Opening Sensor: https://github.com/zigpy/zha-device-handlers/pull/4903

It does not conflict so both changes can be separately merged while the changes in4903 should also affect the new device here.

Crocmagnon:
I've tested this specific quirk version with a real device, it works fine. Thanks!

uschindler:
> I've tested this specific quirk version with a real device, it works fine. Thanks!

For completeness: could you attach the diagnostic data of your device. As I don't own it, it would be great to have it:

Screenshot_20260406-194533

Crocmagnon:
Sure

[zha-01KN1ZMJ05EV56PWVPE42Q7WVK-SONOFF MINI-ZB2GS-L-308e967a8cfdd208d66b895cc6e64277.json](https://github.com/user-attachments/files/26517518/zha-01KN1ZMJ05EV56PWVPE42Q7WVK-SONOFF.MINI-ZB2GS-L-308e967a8cfdd208d66b895cc6e64277.json)

uschindler:
I added it above.

TheJulianJES (approved):
Thank you! Looks good.

TheJulianJES:
> One question: People complained that because the device has 2 endpoints (channels/buttons), the entities created have exact same name in Home Assistant.

Yeah, there's an open PR in the ZHA and Core repo.. There were some edge cases with that, but it's something we'll look at. Now, Home Assistant also has support for sub-devices, but I'm not sure we can just map one endpoint to a sub-device. Still needs to be investigated further.

PR #4916, opened 2026-04-10
Add YOKIS remote device triggers

jpaquereau:
## Proposed change

Add quirk for YOKIS remote.

This quirks implements handling for automation. Now a trigger will be available on OnOff toggle command to control an automation. This functionnality could be used for all remotes button.

## Additional information

This is not intended to be a breaking change for existing devices. It was added to allow all YOKIS remote controls to trigger automations. 
An update will coming on devices to generate OnOff toggle command instead of WindowCovering open/stop/close command.

## Device diagnostics

[zha-01K6ZBZTBVRBKK87315CSX96A3-YOKIS TLM2-UP-9f888d2eee46e3d8d554dc1264d838d8.json](https://github.com/user-attachments/files/26629309/zha-01K6ZBZTBVRBKK87315CSX96A3-YOKIS.TLM2-UP-9f888d2eee46e3d8d554dc1264d838d8.json)
[zha-01K6ZBZTBVRBKK87315CSX96A3-YOKIS TLC1-UP-2e21946c69d84e028ec57afeb4607b25.json](https://github.com/user-attachments/files/26629311/zha-01K6ZBZTBVRBKK87315CSX96A3-YOKIS.TLC1-UP-2e21946c69d84e028ec57afeb4607b25.json)

## Checklist


TheJulianJES (dismissed):
Thanks for the PR! Some comments:

Please use v2 quirks. This should drastically reduce the size of all quirks. See https://github.com/zigpy/zha-device-handlers/blob/dev/.github/copilot-instructions.md#quirk-architecture. You can also search `QuirkBuilder` in the repo using GitHub search to see real examples.

If you just add the device triggers, these v2 quirks will be really small.

If there's no "actual code" and just the v2 quirks definition and the defined device triggers, no tests are needed.

zhaquirks/yokis/remote.py
```diff
@@ -0,0 +1,150 @@
...
+    CLUSTER_ID,
+    COMMAND,
+    COMMAND_TOGGLE,
+    ENDPOINT_ID,
+    SHORT_PRESS,
+)
+
+(
+    QuirkBuilder("YOKIS", "TLC1-UP")
+    .also_applies_to("YOKIS", "TLM1-UP")
```
TheJulianJES: Can you change `also_applies_to` to `applies_to` everywhere? It's the exact same method, just another name.

(It was changed because `QuirkBuilder` can also be created without manufacturer/model now.)

TheJulianJES (commented):
Thanks for making the changes! This is looking really good. One more small comment on changing from the older `also_applies_to` method to the newer (name): `applies_to`.

TheJulianJES (approved):
Thanks!

PR #4819, opened 2026-03-10
Add Tuya SG502Z soil moisture sensor `_TZE284_nt4pquef`

mtheli:
## Summary

- Add support for the Tuya SG502Z soil moisture sensor with model ID `_TZE284_nt4pquef` (TS0601)
- Exposes soil moisture (dp 3), temperature (dp 5), light level (dp 2), display unit config (dp 9), and battery (dp 15)
- Light level is exposed as a sensor entity with three levels: Low, Normal, High

## Device info

- **Manufacturer**: Tuya
- **Model**: SG502Z
- **Zigbee model ID**: `_TZE284_nt4pquef` / `TS0601`
- **Reference**: https://github.com/Koenkk/zigbee-herdsman-converters/issues/10315

Closes #4707

PR #4919, opened 2026-04-12
Fix soil light level enum for Tuya `_TZE284_nt4pquef`

mtheli:
## Summary

Follow-up to #4819 as noted by @TheJulianJES in [review](https://github.com/zigpy/zha-device-handlers/pull/4819#pullrequestreview-4001978917).

The `TuyaSoilLightLevel` enum only mapped 3 of 5 values (0, 2, 4), causing `ValueError` exceptions for values 1 and 3 that the device actually sends:

```
ValueError: <enum8.undefined_0x01: 1> is not a valid LightLevel
ValueError: <enum8.undefined_0x03: 3> is not a valid LightLevel
```

Corrected mapping based on [zigbee-herdsman-converters](https://github.com/Koenkk/zigbee-herdsman-converters/issues/10315):

| Value | Old | New |
|-------|-----|-----|
| 0 | Low | Low_minus |
| 1 | ❌ missing | Low |
| 2 | Normal | Normal |
| 3 | ❌ missing | High |
| 4 | High | High_plus |

Both missing values (0x01 and 0x03) confirmed in production logs.


zhaquirks/tuya/tuya_sensor.py
```diff
@@ -27,9 +27,11 @@ class TuyaTempUnitConvert(t.enum8):
 class TuyaSoilLightLevel(t.enum8):
     """Tuya soil sensor light level enum."""
 
-    Low = 0x00
+    Low_minus = 0x00
+    Low = 0x01
     Normal = 0x02
-    High = 0x04
+    High = 0x03
+    High_plus = 0x04
```
TheJulianJES: Would `Lowest` and `Highest` make more sense? Now, we'd also change the existing meanings.
I also feel like this should have been a (numeric) sensor, not an enum one 😅

mtheli: Good point, thanks for the review! Switched to `Lowest`/`Highest` ([caac6e6](https://github.com/zigpy/zha-device-handlers/pull/4919/commits/caac6e6)) — reads much cleaner, and the re-meaning of `Low`/`High` is unavoidable either way.

I'm running the updated enum locally against my `_TZE284_nt4pquef`; the quirk loads cleanly and the sensor currently reports `lowest` for value `0`. I can't claim the `ValueError` for values `1`/`3` is gone from observation — my device hasn't produced those states since the rename — but the enum now covers the full 0–4 range that the upstream issue documents, so the earlier exception path is no longer reachable.

On the sensor type: agreed, numeric would have been the better choice in hindsight — but that would be a breaking change for existing devices. I'd treat it as a separate task; happy to send a follow-up PR if desired.

zhaquirks/tuya/tuya_sensor.py
```diff
@@ -27,9 +27,11 @@ class TuyaTempUnitConvert(t.enum8):
 class TuyaSoilLightLevel(t.enum8):
     """Tuya soil sensor light level enum."""
 
-    Low = 0x00
+    Lowest = 0x00
+    Low = 0x01
     Normal = 0x02
-    High = 0x04
+    High = 0x03
+    Highest = 0x04
```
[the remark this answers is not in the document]
TheJulianJES: No, we just renamed them.

Lolerloli:
Maybe its also possible to provide the lux value approximations for the values 1-5. At least my device came with a description which lux ranges should map to the values:
0-1000 Lux LOW,
1000-2000 Lux LOW,
2000-5000 Lux NOR,
5000-6000 Lux HIGH,
6000-7000 Lux HIGH+

mtheli:
@TheJulianJES Friendly ping — is this good as-is, or would you like further changes before it can go in?

State after your review: the enum covers all five DP values, with `Lowest`/`Highest` for 0x00/0x04 as you suggested, and `Low`/`High` moved to their actual values 0x01/0x03.

You also mentioned this should probably have been a numeric sensor — happy to look at that separately, but the DP only reports these five steps, not a lux value.

One thing from the thread: @Lolerloli's device ships a manual mapping the steps to approximate lux ranges (0–1000 / 1000–2000 / 2000–5000 / 5000–6000 / 6000–7000). I can add that as a comment above the enum if you find it useful.

Issue #4340, opened 2025-09-11
[Device Support Request] Sonoff SWV Smart Water Valve - timer function


ausfas:
### Problem description

The Sonoff water valve has a timer function. I can access that one via following service, but I'm hoping it can be added to quirk so that it be used from there. The reason to have this functionality added as if HA crashes the timer parameters are already passed on Sonoff valve and it will close when the timer expired even when HA is no responsive or unavailable. 

```

### Solution description

One can issue the following command to pass on the timer parameter.
```
      - service: zha.issue_zigbee_cluster_command
        data:
          ieee: xxxxxxxx
          endpoint_id: 1
          cluster_type: in
          command_type: server
          cluster_id: 0x0006
          command: 0x42
          args: [0, 60, 1]
```

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>


### Diagnostics information

No debugging is need, its just missing a feature that the valve supports.
The quirk is located in here zhaquirks/sonoff/swv.py

### Device signature

<details><summary>Device signature</summary>
zhaquirks/sonoff/swv.py
```json



```

</details>

### Logs

<details><summary>Logs</summary>

```python

[Paste the logs here]

```

</details>

### Custom quirk

<details><summary>Custom quirk</summary>

```python

[Paste your custom quirk here]

```

</details>

Hedda:
Also be aware of this related Device Support Request in the zha-device-handlers repo which asks for other attributes to be exposed:

* https://github.com/zigpy/zha-device-handlers/issues/3298

As well as this other feature request asking for timer function to be added when using newer firmware:

* https://github.com/zigpy/zha-device-handlers/issues/4340

FYI, Sonoff have also released four additional models of they Zigbee Smart Water Valve and all looks to expose the same:

* Sonoff SWV (original model with two different fittings) -> https://www.zigbee2mqtt.io/devices/SWV.html
  * Note that SWV-BSP and SWV-NH is the same product shipping with different fittings for Europe and North America.
* Sonoff SWV-ZFE (Hydro Series with Flow Meter) -> https://www.zigbee2mqtt.io/devices/SWV-ZFE.html
* Sonoff SWV-ZFU (Hydro Series with Flow Meter) -> https://www.zigbee2mqtt.io/devices/SWV-ZFU.html
[...]

PR #4927, opened 2026-04-16
Add Sonoff SWV-ZFE smart water valve

FraserKillip:
## Proposed change

Adds a v2 quirk for the SONOFF SWV-ZFE (Smart Water Valve, Flow Edition), exposing all accessible attributes from the private `0xFC11` ewelink cluster.

Entities added:

- **Child lock** switch — prevents accidental physical operation of the valve
- **Valve working** binary sensor — indicates the valve is actively open and flowing
- **Water shortage auto-close** switch — automatically closes the valve when water shortage is detected
- **Water shortage** binary sensor (PROBLEM) — bit 0 of `valve_abnormal_state`  
- **Water leakage** binary sensor (MOISTURE) — bit 1 of `valve_abnormal_state`
- **Frost protection active** binary sensor — bit 2 of `valve_abnormal_state`
- **Fail safe** binary sensor (SAFETY) — bit 3 of `valve_abnormal_state`
- **Real-time irrigation duration** sensor (seconds, MEASUREMENT)
- **Real-time irrigation volume** sensor (litres, MEASUREMENT)
- **Daily irrigation volume** sensor (litres, TOTAL_INCREASING)
- **Daily irrigation duration** sensor (minutes, TOTAL_INCREASING)
- **Hourly irrigation volume** sensor (litres, MEASUREMENT)
- **Hourly irrigation duration** sensor (minutes, MEASUREMENT)
- **Longitude** number — device location for weather-based irrigation
- **Latitude** number — device location for weather-based irrigation

Note: the real-time duration and volume attributes are sent big-endian by the device firmware (contrary to ZCL little-endian framing), so an `attribute_converter` byte-swaps them before display.

image

## Additional information

The device uses the same cluster ID (`0xFC11`) as the existing SWV quirk but with a different, extended attribute set to support the flow metering hardware.

## Device diagnostics

<details><summary>Details</summary>
<p>

{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2026.3.3",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.14.2",
    "docker": true,
    "arch": "x86_64",
    "timezone": "Pacific/Auckland",
    "os_name": "Linux",
    "os_version": "6.12.67-haos",
    "container_arch": "amd64",
    "supervisor": "2026.04.0",
    "host_os": "Home Assistant OS 17.1",
    "docker_version": "29.1.3",
    "chassis": "vm",
    "run_as_root": true
  },
  "custom_components": {
    "alarmo": {
      "documentation": "https://github.com/nielsfaber/alarmo",
      "version": "1.10.16",
      "requirements": []
    },
    "hacs": {
      "documentation": "https://hacs.xyz/docs/use/",
      "version": "2.0.5",
      "requirements": [
        "aiogithubapi>=22.10.1"
      ]
    },
    "bodymiscale": {
      "documentation": "https://github.com/dckiller51/bodymiscale",
      "version": "2024.6.0",
      "requirements": [
        "cachetools==5.3.0"
      ]
    }
  },
  "integration_manifest": {
    "domain": "zha",
    "name": "Zigbee Home Automation",
    "after_dependencies": [
      "hassio",
      "onboarding",
      "usb"
    ],
    "codeowners": [
      "dmulcahey",
      "adminiuga",
      "puddly",
      "TheJulianJES"
    ],
    "config_flow": true,
    "dependencies": [
      "file_upload",
      "homeassistant_hardware"
    ],
    "documentation": "https://www.home-assistant.io/integrations/zha",
    "integration_type": "hub",
    "iot_class": "local_polling",
    "loggers": [
      "aiosqlite",
      "bellows",
      "crccheck",
      "pure_pcapy3",
      "zhaquirks",
      "zigpy",
      "zigpy_deconz",
      "zigpy_xbee",
      "zigpy_zigate",
      "zigpy_znp",
      "zha",
      "universal_silabs_flasher",
      "serialx"
    ],
    "requirements": [
      "zha==1.0.2",
      "serialx==0.6.2"
    ],
    "usb": [
      {
        "description": "*2652*",
        "known_devices": [
          "slae.sh cc2652rb stick"
        ],
        "pid": "EA60",
        "vid": "10C4"
      },
      {
        "description": "*slzb-07*",
        "known_devices": [
          "smlight slzb-07"
        ],
        "pid": "EA60",
        "vid": "10C4"
      },
      {
        "description": "*sonoff*plus*",
        "known_devices": [
          "sonoff zigbee dongle plus v2"
        ],
        "pid": "55D4",
        "vid": "1A86"
      },
      {
        "description": "*sonoff*plus*",
        "known_devices": [
          "sonoff zigbee dongle plus"
        ],
        "pid": "EA60",
        "vid": "10C4"
      },
      {
        "description": "*tubeszb*",
        "known_devices": [
          "TubesZB Coordinator"
        ],
        "pid": "EA60",
        "vid": "10C4"
      },
      {
        "description": "*tubeszb*",
        "known_devices": [
          "TubesZB Coordinator"
        ],
        "pid": "7523",
        "vid": "1A86"
      },
      {
        "description": "*zigstar*",
        "known_devices": [
          "ZigStar Coordinators"
        ],
        "pid": "7523",
        "vid": "1A86"
      },
      {
        "description": "*conbee*",
        "known_devices": [
          "Conbee II"
        ],
        "pid": "0030",
        "vid": "1CF1"
      },
      {
        "description": "*conbee*",
        "known_devices": [
          "Conbee III"
        ],
        "pid": "6015",
        "vid": "0403"
      },
      {
        "description": "*zigbee*",
        "known_devices": [
          "Nortek HUSBZB-1"
        ],
        "pid": "8A2A",
        "vid": "10C4"
      },
      {
        "description": "*zigate*",
        "known_devices": [
          "ZiGate+"
        ],
        "pid": "6015",
        "vid": "0403"
      },
      {
        "description": "*zigate*",
        "known_devices": [
          "ZiGate"
        ],
        "pid": "EA60",
        "vid": "10C4"
      },
      {
        "description": "*bv 2010/10*",
        "known_devices": [
          "Bitron Video AV2010/10"
        ],
        "pid": "8B34",
        "vid": "10C4"
      },
      {
        "description": "*sonoff*max*",
        "known_devices": [
          "SONOFF Dongle Max MG24"
        ],
        "pid": "EA60",
        "vid": "10C4"
      },
      {
        "description": "*sonoff*lite*mg21*",
        "known_devices": [
          "sonoff zigbee dongle lite mg21"
        ],
        "pid": "EA60",
        "vid": "10C4"
      }
    ],
    "zeroconf": [
      {
        "name": "tube*",
        "type": "_esphomelib._tcp.local."
      },
      {
        "name": "*zigate*",
        "type": "_zigate-zigbee-gateway._tcp.local."
      },
      {
        "name": "*zigstar*",
        "type": "_zigstar_gw._tcp.local."
      },
      {
        "name": "uzg-01*",
        "type": "_uzg-01._tcp.local."
      },
      {
        "name": "slzb-06*",
        "type": "_slzb-06._tcp.local."
      },
      {
        "name": "xzg*",
        "type": "_xzg._tcp.local."
      },
      {
        "name": "czc*",
        "type": "_czc._tcp.local."
      },
      {
        "name": "*",
        "type": "_zigbee-coordinator._tcp.local."
      }
    ],
    "is_built_in": true,
    "overwrites_built_in": false
  },
  "setup_times": {
    "null": {
      "setup": 4.869699478149414e-05
    },
    "886df563250ea5a2b1ec9e51fdec0d42": {
      "wait_import_platforms": -0.0513524249981856,
      "wait_base_component": -0.0006972509872866794,
      "config_entry_setup": 7.450143057998503
    }
  },
  "data": {
    "version": 1,
    "ieee": "**REDACTED**",
    "nwk": "0xB3CB",
    "manufacturer": "SONOFF",
    "model": "SWV-ZFE",
    "friendly_manufacturer": "SONOFF",
    "friendly_model": "SWV-ZFE",
    "name": "SONOFF SWV-ZFE",
    "quirk_applied": true,
    "quirk_class": "zigpy.quirks.v2.CustomDeviceV2",
    "exposes_features": [],
    "manufacturer_code": 4742,
    "power_source": "Battery or Unknown",
    "lqi": 196,
    "rssi": -62,
    "last_seen": "2026-04-16T09:14:13.443382+00:00",
    "available": true,
    "device_type": "EndDevice",
    "active_coordinator": false,
    "node_descriptor": {
      "logical_type": "EndDevice",
      "complex_descriptor_available": false,
      "user_descriptor_available": false,
      "reserved": 0,
      "aps_flags": 0,
      "frequency_band": 8,
      "mac_capability_flags": 128,
      "manufacturer_code": 4742,
      "maximum_buffer_size": 74,
      "maximum_incoming_transfer_size": 404,
      "server_mask": 10752,
      "maximum_outgoing_transfer_size": 404,
    },
    "endpoints": {
      "1": {
        "profile_id": 260,
        "device_type": {
          "name": "ON_OFF_OUTPUT",
          "id": 2
        },
        "in_clusters": [
          {
            "cluster_id": "0x0000",
            "endpoint_attribute": "basic",
            "attributes": [
              {
                "id": "0x0004",
                "name": "manufacturer",
                "zcl_type": "string",
                "value": "SONOFF"
              },
              {
                "id": "0x0005",
                "name": "model",
                "zcl_type": "string",
                "value": "SWV-ZFE"
              }
            ]
          },
          {
            "cluster_id": "0x0001",
            "endpoint_attribute": "power",
            "attributes": [
              {
                "id": "0x0021",
                "name": "battery_percentage_remaining",
                "zcl_type": "uint8",
                "value": 200
              },
              {
                "id": "0x0033",
                "name": "battery_quantity",
                "zcl_type": "uint8",
                "unsupported": true
              },
              {
                "id": "0x0031",
                "name": "battery_size",
                "zcl_type": "enum8",
                "unsupported": true
              },
              {
                "id": "0x0020",
                "name": "battery_voltage",
                "zcl_type": "uint8",
                "unsupported": true
              }
            ]
          },
          {
            "cluster_id": "0x0003",
            "endpoint_attribute": "identify",
            "attributes": []
          },
          {
            "cluster_id": "0x0006",
            "attributes": [
              {
                "id": "0xfffd",
                "name": "cluster_revision",
                "zcl_type": "uint16",
                "value": 2
              },
              {
                "id": "0x4002",
                "name": "off_wait_time",
                "zcl_type": "uint16",
                "unsupported": true
              },
              {
                "id": "0x0000",
                "name": "on_off",
                "zcl_type": "bool",
                "value": 1
              },
              {
                "id": "0x4001",
                "name": "on_time",
                "zcl_type": "uint16",
                "unsupported": true
              },
              {
                "id": "0xfffe",
                "name": "reporting_status",
                "zcl_type": "enum8",
                "unsupported": true
              },
              {
                "id": "0x4003",
                "name": "start_up_on_off",
                "zcl_type": "enum8",
                "unsupported": true
              }
            ]
          },
          {
            "cluster_id": "0x0020",
            "endpoint_attribute": "poll_control",
            "attributes": [
              {
                "id": "0x0000",
                "name": "checkin_interval",
                "zcl_type": "uint32",
                "value": 14400
              },
              {
                "id": "0x0003",
                "name": "fast_poll_timeout",
                "zcl_type": "uint16",
                "value": 120
              }
            ]
          },
          {
            "cluster_id": "0xfc11",
            "endpoint_attribute": null,
            "attributes": [
              {
                "id": "0x0000",
                "name": "child_lock",
                "zcl_type": "bool",
                "value": 0
              },
              {
                "id": "0x501a",
                "name": "daily_irrigation_duration",
                "zcl_type": "uint32",
                "value": 337
              },
              {
                "id": "0x500f",
                "name": "daily_irrigation_volume",
                "zcl_type": "uint32",
                "value": 0
              },
              {
                "id": "0x501c",
                "name": "hour_irrigation_duration",
                "zcl_type": "uint32",
                "value": 3
              },
              {
                "id": "0x501b",
                "name": "hour_irrigation_volume",
                "zcl_type": "uint32",
                "value": 0
              },
              {
                "id": "0x5006",
                "name": "real_time_irrigation_duration",
                "zcl_type": "uint32",
                "value": 67108864
              },
              {
                "id": "0x5007",
                "name": "real_time_irrigation_volume",
                "zcl_type": "uint32",
                "value": 50331648
              },
              {
                "id": "0x500c",
                "name": "valve_abnormal_state",
                "zcl_type": "uint8",
                "value": 0
              }
            ]
          },
          {
            "cluster_id": "0xfc57",
            "endpoint_attribute": "manufacturer_specific",
            "attributes": []
          }
        ],
        "out_clusters": [
          {
            "cluster_id": "0x0003",
            "endpoint_attribute": "identify",
            "attributes": []
          },
          {
            "cluster_id": "0x0019",
            "endpoint_attribute": "ota",
            "attributes": [
              {
                "id": "0x0002",
                "name": "current_file_version",
                "zcl_type": "uint32",
                "value": 4103
              }
            ]
          }
        ]
      }
    },
    "original_signature": {
      "manufacturer": "SONOFF",
      "model": "SWV-ZFE",
      "node_desc": {
        "logical_type": 2,
        "reserved": 0,
        "aps_flags": 0,
        "frequency_band": 8,
        "mac_capability_flags": 128,
        "manufacturer_code": 4742,
        "maximum_buffer_size": 74,
        "maximum_incoming_transfer_size": 404,
        "server_mask": 10752,
        "maximum_outgoing_transfer_size": 404,
      },
      "endpoints": {
        "1": {
          "profile_id": "0x0104",
          "device_type": "0x0002",
          "input_clusters": [
            "0x0000",
            "0x0001",
            "0x0003",
            "0x0006",
            "0x0020",
            "0xfc57",
            "0xfc11"
          ],
          "output_clusters": [
            "0x0003",
            "0x0019"
          ]
        }
      }
    },
    "zha_lib_entities": {
      "binary_sensor": [
        {
          "info_object": {
            "fallback_name": "Fail safe",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "binary_sensor",
            "class_name": "BinarySensor",
            "translation_key": null,
            "translation_placeholders": null,
            "device_class": "safety",
            "state_class": null,
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "valve_abnormal_state"
          },
          "state": {
            "class_name": "BinarySensor",
            "available": true,
            "state": false
          }
        },
        {
          "info_object": {
            "fallback_name": "Frost protection",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "binary_sensor",
            "class_name": "BinarySensor",
            "translation_key": "frost_protection",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "valve_abnormal_state"
          },
          "state": {
            "class_name": "BinarySensor",
            "available": true,
            "state": false
          }
        },
        {
          "info_object": {
            "fallback_name": "Valve working",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "binary_sensor",
            "class_name": "BinarySensor",
            "translation_key": "valve_work_state",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "valve_work_state"
          },
          "state": {
            "class_name": "BinarySensor",
            "available": true,
            "state": false
          }
        },
        {
          "info_object": {
            "fallback_name": "Water leakage",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "binary_sensor",
            "class_name": "BinarySensor",
            "translation_key": null,
            "translation_placeholders": null,
            "device_class": "moisture",
            "state_class": null,
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "valve_abnormal_state"
          },
          "state": {
            "class_name": "BinarySensor",
            "available": true,
            "state": false
          }
        },
        {
          "info_object": {
            "fallback_name": "Water shortage",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "binary_sensor",
            "class_name": "BinarySensor",
            "translation_key": null,
            "translation_placeholders": null,
            "device_class": "problem",
            "state_class": null,
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "valve_abnormal_state"
          },
          "state": {
            "class_name": "BinarySensor",
            "available": true,
            "state": false
          }
        }
      ],
      "button": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "button",
            "class_name": "IdentifyButton",
            "translation_key": null,
            "translation_placeholders": null,
            "device_class": "identify",
            "state_class": null,
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "IdentifyClusterHandler",
                "generic_id": "cluster_handler_0x0003",
                "endpoint_id": 1,
                "cluster": {
                  "id": 3,
                  "name": "Identify",
                  "type": "server"
                },
                "id": "1:0x0003",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "command": "identify",
            "args": [
              5
            ],
            "kwargs": {}
          },
          "state": {
            "class_name": "IdentifyButton",
            "available": true
          }
        }
      ],
      "number": [
        {
          "info_object": {
            "fallback_name": "Latitude",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "number",
            "class_name": "NumberConfigurationEntity",
            "translation_key": "latitude",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "mode": "box",
            "native_max_value": 90,
            "native_min_value": -90,
            "native_step": 1,
            "native_unit_of_measurement": null
          },
          "state": {
            "class_name": "NumberConfigurationEntity",
            "available": true,
            "state": null
          }
        },
        {
          "info_object": {
            "fallback_name": "Longitude",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "number",
            "class_name": "NumberConfigurationEntity",
            "translation_key": "longitude",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "mode": "box",
            "native_max_value": 180,
            "native_min_value": -180,
            "native_step": 1,
            "native_unit_of_measurement": null
          },
          "state": {
            "class_name": "NumberConfigurationEntity",
            "available": true,
            "state": null
          }
        }
      ],
      "sensor": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "LQISensor",
            "translation_key": "lqi",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": "measurement",
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": false,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "BasicClusterHandler",
                "generic_id": "cluster_handler_0x0000",
                "endpoint_id": 1,
                "cluster": {
                  "id": 0,
                  "name": "Basic",
                  "type": "server"
                },
                "id": "1:0x0000",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": null
          },
          "state": {
            "class_name": "LQISensor",
            "available": true,
            "state": 196
          }
        },
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "RSSISensor",
            "translation_key": "rssi",
            "translation_placeholders": null,
            "device_class": "signal_strength",
            "state_class": "measurement",
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": false,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "BasicClusterHandler",
                "generic_id": "cluster_handler_0x0000",
                "endpoint_id": 1,
                "cluster": {
                  "id": 0,
                  "name": "Basic",
                  "type": "server"
                },
                "id": "1:0x0000",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": "dBm"
          },
          "state": {
            "class_name": "RSSISensor",
            "available": true,
            "state": -62
          }
        },
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "Battery",
            "translation_key": null,
            "translation_placeholders": null,
            "device_class": "battery",
            "state_class": "measurement",
            "entity_category": "diagnostic",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "PowerConfigurationClusterHandler",
                "generic_id": "cluster_handler_0x0001",
                "endpoint_id": 1,
                "cluster": {
                  "id": 1,
                  "name": "Power Configuration",
                  "type": "server"
                },
                "id": "1:0x0001",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": "battery_voltage"
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": 0,
            "unit": "%"
          },
          "state": {
            "class_name": "Battery",
            "available": true,
            "state": 100.0
          },
          "extra_state_attributes": [
            "battery_quantity",
            "battery_size",
            "battery_voltage"
          ]
        },
        {
          "info_object": {
            "fallback_name": "Daily irrigation duration",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "Sensor",
            "translation_key": "daily_irrigation_duration",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": "total_increasing",
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": "min"
          },
          "state": {
            "class_name": "Sensor",
            "available": true,
            "state": 337
          }
        },
        {
          "info_object": {
            "fallback_name": "Daily irrigation volume",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "Sensor",
            "translation_key": "daily_irrigation_volume",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": "total_increasing",
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": "L"
          },
          "state": {
            "class_name": "Sensor",
            "available": true,
            "state": 0
          }
        },
        {
          "info_object": {
            "fallback_name": "Hourly irrigation duration",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "Sensor",
            "translation_key": "hour_irrigation_duration",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": "measurement",
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": "min"
          },
          "state": {
            "class_name": "Sensor",
            "available": true,
            "state": 3
          }
        },
        {
          "info_object": {
            "fallback_name": "Hourly irrigation volume",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "Sensor",
            "translation_key": "hour_irrigation_volume",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": "measurement",
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": "L"
          },
          "state": {
            "class_name": "Sensor",
            "available": true,
            "state": 0
          }
        },
        {
          "info_object": {
            "fallback_name": "Real-time irrigation duration",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "Sensor",
            "translation_key": "real_time_irrigation_duration",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": "measurement",
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": "s"
          },
          "state": {
            "class_name": "Sensor",
            "available": true,
            "state": 4
          }
        },
        {
          "info_object": {
            "fallback_name": "Real-time irrigation volume",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "sensor",
            "class_name": "Sensor",
            "translation_key": "real_time_irrigation_volume",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": "measurement",
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "suggested_display_precision": null,
            "unit": "L"
          },
          "state": {
            "class_name": "Sensor",
            "available": true,
            "state": 3
          }
        }
      ],
      "switch": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "switch",
            "class_name": "Switch",
            "translation_key": "switch",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": null,
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": true,
            "cluster_handlers": [
              {
                "class_name": "OnOffClusterHandler",
                "generic_id": "cluster_handler_0x0006",
                "endpoint_id": 1,
                "cluster": {
                  "id": 6,
                  "name": "On/Off",
                  "type": "server"
                },
                "id": "1:0x0006",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": "on_off"
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null
          },
          "state": {
            "class_name": "Switch",
            "state": 1,
            "available": true
          }
        },
        {
          "info_object": {
            "fallback_name": "Water shortage auto-close",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "switch",
            "class_name": "ConfigurableAttributeSwitch",
            "translation_key": "water_shortage_auto_close",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "auto_close_water_shortage",
            "invert_attribute_name": null,
            "force_inverted": false,
            "off_value": 0,
            "on_value": 30
          },
          "state": {
            "class_name": "ConfigurableAttributeSwitch",
            "available": true,
            "state": false,
            "inverted": false
          }
        },
        {
          "info_object": {
            "fallback_name": "Child lock",
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "switch",
            "class_name": "ConfigurableAttributeSwitch",
            "translation_key": "child_lock",
            "translation_placeholders": null,
            "device_class": null,
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "SonoffPresenceSenorClusterHandler",
                "generic_id": "cluster_handler_0xfc11",
                "endpoint_id": 1,
                "cluster": {
                  "id": 64529,
                  "name": "SWVZFECluster",
                  "type": "server"
                },
                "id": "1:0xfc11",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "attribute_name": "child_lock",
            "invert_attribute_name": null,
            "force_inverted": false,
            "off_value": 0,
            "on_value": 1
          },
          "state": {
            "class_name": "ConfigurableAttributeSwitch",
            "available": true,
            "state": false,
            "inverted": false
          }
        }
      ],
      "update": [
        {
          "info_object": {
            "fallback_name": null,
            "unique_id": "**REDACTED**",
            "migrate_unique_ids": [],
            "platform": "update",
            "class_name": "FirmwareUpdateEntity",
            "translation_key": null,
            "translation_placeholders": null,
            "device_class": "firmware",
            "state_class": null,
            "entity_category": "config",
            "entity_registry_enabled_default": true,
            "enabled": true,
            "primary": false,
            "cluster_handlers": [
              {
                "class_name": "OtaClientClusterHandler",
                "generic_id": "cluster_handler_0x0019_client",
                "endpoint_id": 1,
                "cluster": {
                  "id": 25,
                  "name": "Ota",
                  "type": "client"
                },
                "id": "1:0x0019_client",
                "unique_id": "**REDACTED**",
                "status": "INITIALIZED",
                "value_attribute": null
              }
            ],
            "device_ieee": "**REDACTED**",
            "endpoint_id": 1,
            "available": true,
            "group_id": null,
            "supported_features": 7
          },
          "state": {
            "class_name": "FirmwareUpdateEntity",
            "available": true,
            "installed_version": "0x00001007",
            "in_progress": false,
            "update_percentage": null,
            "latest_version": null,
            "release_summary": null,
            "release_notes": null,
            "release_url": null
          }
        }
      ]
    },
    "neighbors": [],
    "routes": []
  },
  "issues": []
}
</p>
</details> 

## Checklist

> [!TIP]
> Disclaimer: This PR is written using Copilot, but tested using a real device manually

### DIFF
--- a/tests/test_sonoff.py
+++ b/tests/test_sonoff.py
@@ -10,6 +10,14 @@ from tests.common import ClusterListener
 import zhaquirks
 from zhaquirks.const import COMMAND_DOUBLE, COMMAND_HOLD, COMMAND_SINGLE, COMMAND_TRIPLE
 from zhaquirks.sonoff.snzb01m import SonoffButtonCluster
+from zhaquirks.sonoff.swv_zfe import (
+    SWVZFECluster,
+    swvzfe_be_swap,
+    swvzfe_fail_safe,
+    swvzfe_frost_protection,
+    swvzfe_water_leakage,
+    swvzfe_water_shortage,
+)
 from zhaquirks.sonoff.zbm5 import (
     SonoffCluster,
     SonoffDetachedRelayMask,
@@ -272,3 +280,63 @@ async def test_snzb01m_non_button_attribute_update(zigpy_device_from_v2_quirk):
 
     cluster.update_attribute(0x0001, 1)
     assert listener.zha_send_event.call_count == 0
+
+
+# ---------------------------------------------------------------------------
+# SWV-ZFE tests
+# ---------------------------------------------------------------------------
+
+
+def test_swvzfe_be_swap() -> None:
+    """Test that swvzfe_be_swap correctly reverses the byte order of a uint32."""
+    # Device sends 300 as big-endian bytes; zigpy reads it as little-endian
+    assert swvzfe_be_swap(0x2C010000) == 300
+    # Round-trip: swapping twice should return the original value
+    assert swvzfe_be_swap(swvzfe_be_swap(0xDEADBEEF)) == 0xDEADBEEF
+    # Zero is invariant under byte-swap
+    assert swvzfe_be_swap(0) == 0
+    assert swvzfe_be_swap(None) is None
+
+
+@pytest.mark.parametrize(
+    ("value", "water_shortage", "water_leakage", "frost_protection", "fail_safe"),
+    [
+        (0x00, False, False, False, False),
+        (0x01, True, False, False, False),
+        (0x02, False, True, False, False),
+        (0x04, False, False, True, False),
+        (0x08, False, False, False, True),
+        (0x0F, True, True, True, True),
+        (0x05, True, False, True, False),
+    ],
+    ids=["none", "shortage", "leakage", "frost", "failsafe", "all", "shortage+frost"],
+)
+def test_swvzfe_valve_abnormal_state_converters(
+    value, water_shortage, water_leakage, frost_protection, fail_safe
+) -> None:
+    """Test all valve_abnormal_state bit-extraction converter functions."""
+    assert swvzfe_water_shortage(value) == water_shortage
+    assert swvzfe_water_leakage(value) == water_leakage
+    assert swvzfe_frost_protection(value) == frost_protection
+    assert swvzfe_fail_safe(value) == fail_safe
+
+
+def test_swvzfe_converters_handle_none() -> None:
+    """All converter functions must return None when given None."""
+    assert swvzfe_water_shortage(None) is None
+    assert swvzfe_water_leakage(None) is None
+    assert swvzfe_frost_protection(None) is None
+    assert swvzfe_fail_safe(None) is None
+
+
+def test_swvzfe_quirk_applies(zigpy_device_from_v2_quirk) -> None:
+    """Verify the SWV-ZFE quirk is registered and replaces cluster 0xFC11."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+    assert SWVZFECluster.cluster_id in device.endpoints[1].in_clusters
+    assert isinstance(
+        device.endpoints[1].in_clusters[SWVZFECluster.cluster_id], SWVZFECluster
+    )
--- a/tests/test_sonoff.py
+++ b/tests/test_sonoff.py
@@ -272,4 +272,3 @@ async def test_snzb01m_non_button_attribute_update(zigpy_device_from_v2_quirk):
 
     cluster.update_attribute(0x0001, 1)
     assert listener.zha_send_event.call_count == 0
-
--- a/tests/test_sonoff.py
+++ b/tests/test_sonoff.py
@@ -10,14 +10,6 @@ from tests.common import ClusterListener
 import zhaquirks
 from zhaquirks.const import COMMAND_DOUBLE, COMMAND_HOLD, COMMAND_SINGLE, COMMAND_TRIPLE
 from zhaquirks.sonoff.snzb01m import SonoffButtonCluster
-from zhaquirks.sonoff.swv_zfe import (
-    SWVZFECluster,
-    swvzfe_be_swap,
-    swvzfe_fail_safe,
-    swvzfe_frost_protection,
-    swvzfe_water_leakage,
-    swvzfe_water_shortage,
-)
 from zhaquirks.sonoff.zbm5 import (
     SonoffCluster,
     SonoffDetachedRelayMask,
@@ -281,62 +273,3 @@ async def test_snzb01m_non_button_attribute_update(zigpy_device_from_v2_quirk):
     cluster.update_attribute(0x0001, 1)
     assert listener.zha_send_event.call_count == 0
 
-
-# ---------------------------------------------------------------------------
-# SWV-ZFE tests
-# ---------------------------------------------------------------------------
-
-
-def test_swvzfe_be_swap() -> None:
-    """Test that swvzfe_be_swap correctly reverses the byte order of a uint32."""
-    # Device sends 300 as big-endian bytes; zigpy reads it as little-endian
-    assert swvzfe_be_swap(0x2C010000) == 300
-    # Round-trip: swapping twice should return the original value
-    assert swvzfe_be_swap(swvzfe_be_swap(0xDEADBEEF)) == 0xDEADBEEF
-    # Zero is invariant under byte-swap
-    assert swvzfe_be_swap(0) == 0
-    assert swvzfe_be_swap(None) is None
-
-
-@pytest.mark.parametrize(
-    ("value", "water_shortage", "water_leakage", "frost_protection", "fail_safe"),
-    [
-        (0x00, False, False, False, False),
-        (0x01, True, False, False, False),
-        (0x02, False, True, False, False),
-        (0x04, False, False, True, False),
-        (0x08, False, False, False, True),
-        (0x0F, True, True, True, True),
-        (0x05, True, False, True, False),
-    ],
-    ids=["none", "shortage", "leakage", "frost", "failsafe", "all", "shortage+frost"],
-)
-def test_swvzfe_valve_abnormal_state_converters(
-    value, water_shortage, water_leakage, frost_protection, fail_safe
-) -> None:
-    """Test all valve_abnormal_state bit-extraction converter functions."""
-    assert swvzfe_water_shortage(value) == water_shortage
-    assert swvzfe_water_leakage(value) == water_leakage
-    assert swvzfe_frost_protection(value) == frost_protection
-    assert swvzfe_fail_safe(value) == fail_safe
-
-
-def test_swvzfe_converters_handle_none() -> None:
-    """All converter functions must return None when given None."""
-    assert swvzfe_water_shortage(None) is None
-    assert swvzfe_water_leakage(None) is None
-    assert swvzfe_frost_protection(None) is None
-    assert swvzfe_fail_safe(None) is None
-
-
-def test_swvzfe_quirk_applies(zigpy_device_from_v2_quirk) -> None:
-    """Verify the SWV-ZFE quirk is registered and replaces cluster 0xFC11."""
-    device = zigpy_device_from_v2_quirk(
-        "SONOFF",
-        "SWV-ZFE",
-        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
-    )
-    assert SWVZFECluster.cluster_id in device.endpoints[1].in_clusters
-    assert isinstance(
-        device.endpoints[1].in_clusters[SWVZFECluster.cluster_id], SWVZFECluster
-    )
--- a/tests/test_sonoff_swv_zfe.py
+++ b/tests/test_sonoff_swv_zfe.py
@@ -3,7 +3,7 @@
 from unittest import mock
 
 import pytest
-from zigpy.zcl import ClusterType, foundation
+from zigpy.zcl import AttributeUnsupportedEvent, ClusterType, foundation
 
 from tests.common import ClusterListener
 import zhaquirks
@@ -99,6 +99,100 @@ def test_swvzfe_normalize_valve_alarm_settings_rejects_bad_length() -> None:
         swvzfe_normalize_valve_alarm_settings([1, 2, 3])
 
 
+def test_swvzfe_normalize_valve_alarm_settings_accepts_foundation_array() -> None:
+    """The helper should coerce decoded ZCL Array values into the payload type."""
+    payload = swvzfe_normalize_valve_alarm_settings(
+        foundation.Array(
+            type=foundation.DataTypeId.uint8,
+            value=[1, 2, 3, 4],
+        )
+    )
+
+    assert isinstance(payload, SWVZFEValveAlarmSettingsPayload)
+    assert list(payload) == [1, 2, 3, 4]
+
+
+def test_swvzfe_valve_alarm_settings_payload_accepts_foundation_array() -> None:
+    """The payload wrapper should also unwrap decoded ZCL Array values directly."""
+    payload = SWVZFEValveAlarmSettingsPayload(
+        foundation.Array(
+            type=foundation.DataTypeId.uint8,
+            value=[1, 2, 3, 4],
+        )
+    )
+
+    assert list(payload) == [1, 2, 3, 4]
+
+
+@pytest.mark.parametrize(
+    ("value", "match"),
+    [
+        (None, "cannot be None"),
+        (object(), "iterable of four bytes"),
+        ([1, "bad", 3, 4], "must be an integer"),
+        ([1, 2, 3, 256], "range 0..255"),
+    ],
+    ids=["none", "non-iterable", "non-integer-item", "out-of-range-item"],
+)
+def test_swvzfe_normalize_valve_alarm_settings_rejects_invalid_values(
+    value, match: str
+) -> None:
+    """The helper should reject malformed packed alarm payloads."""
+    with pytest.raises(ValueError, match=match):
+        swvzfe_normalize_valve_alarm_settings(value)
+
+
+@pytest.mark.parametrize(
+    ("kwargs", "match"),
+    [
+        (
+            {"extra_enable_bits": "bad"},
+            "extra_enable_bits must be an integer",
+        ),
+        (
+            {"alarm_water_shortage_duration": -1},
+            "alarm_water_shortage_duration must be in the range 0..255",
+        ),
+    ],
+    ids=["bad-extra-enable-bits", "bad-duration"],
+)
+def test_swvzfe_pack_valve_alarm_settings_rejects_invalid_values(
+    kwargs, match: str
+) -> None:
+    """Packing should enforce uint8 validation on all numeric fields."""
+    valid_kwargs = {
+        "enable_alarm_water_shortage": True,
+        "enable_alarm_water_leak": False,
+        "enable_frost_protection": True,
+        "enable_water_shortage_auto_close": False,
+        "enable_water_leak_auto_close": True,
+        "alarm_water_shortage_duration": 7,
+        "alarm_water_leak_duration": 2,
+        "set_frost_temperature": 3,
+        "extra_enable_bits": 0,
+    }
+    valid_kwargs.update(kwargs)
+
+    with pytest.raises(ValueError, match=match):
+        swvzfe_pack_valve_alarm_settings(**valid_kwargs)
+
+
+def test_swvzfe_pack_valve_alarm_settings_rejects_non_integer_object() -> None:
+    """Packing should reject values that cannot be coerced to integers at all."""
+    with pytest.raises(ValueError, match="extra_enable_bits must be an integer"):
+        swvzfe_pack_valve_alarm_settings(
+            enable_alarm_water_shortage=True,
+            enable_alarm_water_leak=False,
+            enable_frost_protection=True,
+            enable_water_shortage_auto_close=False,
+            enable_water_leak_auto_close=True,
+            alarm_water_shortage_duration=7,
+            alarm_water_leak_duration=2,
+            set_frost_temperature=3,
+            extra_enable_bits=object(),
+        )
+
+
 @pytest.mark.parametrize(
     "model",
     ["SWV-ZFE", "SWV-ZNE", "SWV-ZNU", "SWV-ZFU"],
@@ -149,6 +243,49 @@ async def test_swvzfe_valve_alarm_settings_propagation(zigpy_device_from_v2_quir
     assert local_cluster.get("alarm_water_shortage_duration") == 8
     assert local_cluster.get("alarm_water_leak_duration") == 3
     assert local_cluster.get("set_frost_temperature") == 4
+
+
+async def test_swvzfe_non_alarm_attribute_update_does_not_propagate(
+    zigpy_device_from_v2_quirk,
+):
+    """Only valve_alarm_settings updates should drive the local config cluster."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+    local_cluster = device.endpoints[1].swvzfe_valve_alarm_config
+    local_listener = ClusterListener(local_cluster)
+
+    swvzfe_cluster.update_attribute(SWVZFECluster.AttributeDefs.child_lock.id, True)
+
+    assert local_listener.attribute_updates == []
+    assert local_cluster.get("enable_alarm_water_shortage") is None
+
+
+async def test_swvzfe_invalid_alarm_payload_logs_warning(zigpy_device_from_v2_quirk):
+    """Invalid 0x5020 payloads should be ignored after logging a warning."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+    local_cluster = device.endpoints[1].swvzfe_valve_alarm_config
+    local_listener = ClusterListener(local_cluster)
+
+    with mock.patch.object(swvzfe_cluster, "warning") as warning:
+        swvzfe_cluster.update_attribute(
+            SWVZFECluster.AttributeDefs.valve_alarm_settings.id,
+            [1, 2, 3],
+        )
+
+    assert local_listener.attribute_updates == []
+    assert local_cluster.get("enable_alarm_water_shortage") is None
+    warning.assert_called_once()
 
 
 async def test_swvzfe_valve_alarm_settings_write_attributes_logic(
@@ -204,6 +341,68 @@ async def test_swvzfe_valve_alarm_settings_write_attributes_logic(
     assert local_cluster.get("alarm_water_shortage_duration") == 5
     assert local_cluster.get("alarm_water_leak_duration") == 2
     assert local_cluster.get("set_frost_temperature") == 1
+
+
+async def test_swvzfe_valve_alarm_settings_numeric_write_updates_payload(
+    zigpy_device_from_v2_quirk,
+):
+    """Numeric local writes should be repacked into the device payload."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+    local_cluster = device.endpoints[1].swvzfe_valve_alarm_config
+
+    swvzfe_cluster.update_attribute(
+        SWVZFECluster.AttributeDefs.valve_alarm_settings.id,
+        SWVZFEValveAlarmSettingsPayload([0x05, 5, 2, 1]),
+    )
+
+    write_response = [
+        [foundation.WriteAttributesStatusRecord(status=foundation.Status.SUCCESS)]
+    ]
+    with mock.patch.object(
+        swvzfe_cluster,
+        "write_attributes_raw",
+        mock.AsyncMock(return_value=write_response),
+    ) as mock_write:
+        await local_cluster.write_attributes(
+            {
+                SWVZFEValveAlarmConfigCluster.AttributeDefs.set_frost_temperature.name: 6
+            }
+        )
+
+    written = mock_write.call_args[0][0][0]
+    assert list(written.value.value.value) == [0x05, 5, 2, 6]
+
+
+async def test_swvzfe_valve_alarm_settings_numeric_write_rejects_invalid_value(
+    zigpy_device_from_v2_quirk,
+):
+    """Numeric local writes should enforce uint8 validation."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+    local_cluster = device.endpoints[1].swvzfe_valve_alarm_config
+
+    swvzfe_cluster.update_attribute(
+        SWVZFECluster.AttributeDefs.valve_alarm_settings.id,
+        SWVZFEValveAlarmSettingsPayload([0x05, 5, 2, 1]),
+    )
+
+    with pytest.raises(ValueError, match="set_frost_temperature must be in the range"):
+        await local_cluster.write_attributes(
+            {
+                SWVZFEValveAlarmConfigCluster.AttributeDefs.set_frost_temperature.name: 256
+            }
+        )
 
 
 async def test_swvzfe_valve_alarm_settings_failed_write_does_not_propagate(
@@ -309,6 +508,56 @@ async def test_swvzfe_cluster_apply_custom_configuration_ignores_read_failure(
         await swvzfe_cluster.apply_custom_configuration()
 
 
+def test_swvzfe_repair_valve_alarm_settings_read_response_handles_bad_header(
+    zigpy_device_from_v2_quirk,
+):
+    """Malformed frames should be ignored by the repair helper."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+
+    assert swvzfe_cluster._repair_valve_alarm_settings_read_response(b"\xff") is None
+
+
+def test_swvzfe_repair_valve_alarm_settings_read_response_ignores_other_commands(
+    zigpy_device_from_v2_quirk,
+):
+    """Only read attribute responses should be considered for repair."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+
+    assert swvzfe_cluster._repair_valve_alarm_settings_read_response(b"\x00\x01\x00") is None
+
+
+def test_swvzfe_repair_valve_alarm_settings_read_response_ignores_well_formed_payloads(
+    zigpy_device_from_v2_quirk,
+):
+    """The repair helper should no-op when the duplicated array marker is absent."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+
+    assert (
+        swvzfe_cluster._repair_valve_alarm_settings_read_response(
+            b"\x18\x01\x01\x00\x00\x00\x10\x00"
+        )
+        is None
+    )
+
+
 def test_swvzfe_cluster_deserialize_logs_raw_frame_on_parse_failure(
     zigpy_device_from_v2_quirk,
 ):
@@ -390,6 +639,125 @@ def test_swvzfe_cluster_deserialize_repairs_duplicate_array_type_in_multi_record
     )
     assert response.status_records[1].value.type == foundation.DataTypeId.uint8
     assert list(response.status_records[1].value.value) == [0, 0, 0, 0]
+
+
+async def test_swvzfe_cluster_write_attributes_supports_mixed_writes(
+    zigpy_device_from_v2_quirk,
+):
+    """Child-lock and packed alarm writes should both be forwarded in one call."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+    write_response = [
+        [foundation.Status.SUCCESS],
+        [[foundation.WriteAttributesStatusRecord(status=foundation.Status.SUCCESS)]],
+    ]
+
+    with mock.patch.object(
+        swvzfe_cluster,
+        "write_attributes_raw",
+        mock.AsyncMock(side_effect=write_response),
+    ) as mock_write:
+        await swvzfe_cluster.write_attributes(
+            {
+                SWVZFECluster.AttributeDefs.child_lock.name: True,
+                SWVZFECluster.AttributeDefs.valve_alarm_settings.name: [1, 2, 3, 4],
+            }
+        )
+
+    assert mock_write.await_count == 2
+    assert mock_write.await_args_list[0].args[0][0].attrid == SWVZFECluster.AttributeDefs.child_lock.id
+    assert (
+        mock_write.await_args_list[1].args[0][0].attrid
+        == SWVZFECluster.AttributeDefs.valve_alarm_settings.id
+    )
+
+
+async def test_swvzfe_cluster_write_attributes_handles_empty_status_list(
+    zigpy_device_from_v2_quirk,
+):
+    """An empty raw status list should be normalized to a success record."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+
+    with mock.patch.object(
+        swvzfe_cluster,
+        "write_attributes_raw",
+        mock.AsyncMock(return_value=[[]]),
+    ):
+        result = await swvzfe_cluster.write_attributes(
+            {SWVZFECluster.AttributeDefs.valve_alarm_settings.name: [1, 2, 3, 4]}
+        )
+
+    assert result[0][0].status == foundation.Status.SUCCESS
+    assert result[0][0].attrid == SWVZFECluster.AttributeDefs.valve_alarm_settings.id
+
+
+async def test_swvzfe_cluster_write_attributes_handles_scalar_status(
+    zigpy_device_from_v2_quirk,
+):
+    """A scalar raw status should be converted into a status record."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+
+    with mock.patch.object(
+        swvzfe_cluster,
+        "write_attributes_raw",
+        mock.AsyncMock(return_value=[foundation.Status.SUCCESS]),
+    ):
+        result = await swvzfe_cluster.write_attributes(
+            {SWVZFECluster.AttributeDefs.valve_alarm_settings.name: [1, 2, 3, 4]}
+        )
+
+    assert result[0][0].status == foundation.Status.SUCCESS
+    assert result[0][0].attrid == SWVZFECluster.AttributeDefs.valve_alarm_settings.id
+
+
+async def test_swvzfe_cluster_write_attributes_emits_unsupported_event(
+    zigpy_device_from_v2_quirk,
+):
+    """Unsupported attribute writes should emit the unsupported event."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    swvzfe_cluster = device.endpoints[1].swvzfe_cluster
+    unsupported = foundation.WriteAttributesStatusRecord(
+        status=foundation.Status.UNSUPPORTED_ATTRIBUTE,
+        attrid=SWVZFECluster.AttributeDefs.valve_alarm_settings.id,
+    )
+
+    with (
+        mock.patch.object(
+            swvzfe_cluster,
+            "write_attributes_raw",
+            mock.AsyncMock(return_value=[[unsupported]]),
+        ),
+        mock.patch.object(swvzfe_cluster, "emit", wraps=swvzfe_cluster.emit) as emit,
+    ):
+        await swvzfe_cluster.write_attributes(
+            {SWVZFECluster.AttributeDefs.valve_alarm_settings.name: [1, 2, 3, 4]}
+        )
+
+    assert any(
+        call.args[0] == AttributeUnsupportedEvent.event_type for call in emit.call_args_list
+    )
 
 
 async def test_swvzfe_valve_alarm_settings_write_attributes_lazy_initializes(
--- a/tests/test_sonoff_swv_zfe.py
+++ b/tests/test_sonoff_swv_zfe.py
@@ -4,11 +4,13 @@ from unittest import mock
 
 import pytest
 from zigpy.zcl import AttributeUnsupportedEvent, ClusterType, foundation
+from zigpy.zcl.clusters.general import OnOff
 
 from tests.common import ClusterListener
 import zhaquirks
 from zhaquirks.sonoff.swv_zfe import (
     SWVZFECluster,
+    SWVZFEOnOffCluster,
     SWVZFEValveAlarmConfigCluster,
     SWVZFEValveAlarmSettingsPayload,
     swvzfe_be_swap,
@@ -16,6 +18,7 @@ from zhaquirks.sonoff.swv_zfe import (
     swvzfe_fail_safe,
     swvzfe_frost_protection,
     swvzfe_normalize_valve_alarm_settings,
+    swvzfe_on_with_timed_off_from_ds,
     swvzfe_pack_valve_alarm_settings,
     swvzfe_water_leakage,
     swvzfe_water_shortage,
@@ -64,6 +67,15 @@ def test_swvzfe_converters_handle_none() -> None:
     assert swvzfe_water_leakage(None) is None
     assert swvzfe_frost_protection(None) is None
     assert swvzfe_fail_safe(None) is None
+
+
+@pytest.mark.parametrize(
+    ("value", "expected"),
+    [(0, 0), (9, 0), (10, 1), (59, 5), (600, 60)],
+)
+def test_swvzfe_on_with_timed_off_from_ds(value: int, expected: int) -> None:
+    """Timed-off values should be converted from deciseconds to seconds."""
+    assert swvzfe_on_with_timed_off_from_ds(value) == expected
 
 
 def test_swvzfe_valve_alarm_settings_pack_decode() -> None:
@@ -210,10 +222,47 @@ def test_swvzfe_quirk_applies(zigpy_device_from_v2_quirk, model: str) -> None:
         device.endpoints[1].swvzfe_cluster,
         SWVZFECluster,
     )
+    assert isinstance(device.endpoints[1].on_off, SWVZFEOnOffCluster)
     assert isinstance(
         device.endpoints[1].swvzfe_valve_alarm_config,
         SWVZFEValveAlarmConfigCluster,
     )
+
+
+async def test_swvzfe_onoff_cluster_scales_on_with_timed_off_command(
+    zigpy_device_from_v2_quirk,
+):
+    """SWV-ZF* expects on_with_timed_off timers in seconds, not deciseconds."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+
+    on_off = device.endpoints[1].on_off
+    command_response = foundation.GENERAL_COMMANDS[
+        foundation.GeneralCommand.Default_Response
+    ].schema(
+        command_id=OnOff.ServerCommandDefs.on_with_timed_off.id,
+        status=foundation.Status.SUCCESS,
+    )
+
+    with mock.patch(
+        "zhaquirks.sonoff.swv_zfe.CustomCluster.command",
+        mock.AsyncMock(return_value=command_response),
+    ) as mock_command:
+        await on_off.command(
+            OnOff.ServerCommandDefs.on_with_timed_off.id,
+            0,
+            600,
+            120,
+        )
+
+    assert mock_command.await_count == 1
+    assert (
+        mock_command.await_args.args[0] == OnOff.ServerCommandDefs.on_with_timed_off.id
+    )
+    assert mock_command.await_args.args[1:] == (0, 60, 12)
 
 
 async def test_swvzfe_valve_alarm_settings_propagation(zigpy_device_from_v2_quirk):
@@ -472,8 +521,14 @@ async def test_swvzfe_cluster_apply_custom_configuration(zigpy_device_from_v2_qu
         swvzfe_cluster,
         "_read_attributes",
         mock.AsyncMock(return_value=[[read_response]]),
-    ):
+    ) as mock_read_attributes:
         await swvzfe_cluster.apply_custom_configuration()
+
+    assert mock_read_attributes.await_count == 1
+    assert set(mock_read_attributes.await_args.args[0]) == {
+        SWVZFECluster.AttributeDefs.valve_alarm_settings.id,
+        SWVZFECluster.AttributeDefs.valve_abnormal_state.id,
+    }
 
     assert len(local_listener.attribute_updates) == 8
     assert local_cluster.get("enable_alarm_water_shortage") is True
--- a/tests/test_sonoff_swv_zfe.py
+++ b/tests/test_sonoff_swv_zfe.py
@@ -1,0 +1,71 @@
+"""Tests for Sonoff SWV-ZFE smart water valve quirk."""
+
+import pytest
+from zigpy.zcl import ClusterType
+
+import zhaquirks
+from zhaquirks.sonoff.swv_zfe import (
+    SWVZFECluster,
+    swvzfe_be_swap,
+    swvzfe_fail_safe,
+    swvzfe_frost_protection,
+    swvzfe_water_leakage,
+    swvzfe_water_shortage,
+)
+
+zhaquirks.setup()
+
+
+def test_swvzfe_be_swap() -> None:
+    """Test that swvzfe_be_swap correctly reverses the byte order of a uint32."""
+    # Device sends 300 as big-endian bytes; zigpy reads it as little-endian
+    assert swvzfe_be_swap(0x2C010000) == 300
+    # Round-trip: swapping twice should return the original value
+    assert swvzfe_be_swap(swvzfe_be_swap(0xDEADBEEF)) == 0xDEADBEEF
+    # Zero is invariant under byte-swap
+    assert swvzfe_be_swap(0) == 0
+    assert swvzfe_be_swap(None) is None
+
+
+@pytest.mark.parametrize(
+    ("value", "water_shortage", "water_leakage", "frost_protection", "fail_safe"),
+    [
+        (0x00, False, False, False, False),
+        (0x01, True, False, False, False),
+        (0x02, False, True, False, False),
+        (0x04, False, False, True, False),
+        (0x08, False, False, False, True),
+        (0x0F, True, True, True, True),
+        (0x05, True, False, True, False),
+    ],
+    ids=["none", "shortage", "leakage", "frost", "failsafe", "all", "shortage+frost"],
+)
+def test_swvzfe_valve_abnormal_state_converters(
+    value, water_shortage, water_leakage, frost_protection, fail_safe
+) -> None:
+    """Test all valve_abnormal_state bit-extraction converter functions."""
+    assert swvzfe_water_shortage(value) == water_shortage
+    assert swvzfe_water_leakage(value) == water_leakage
+    assert swvzfe_frost_protection(value) == frost_protection
+    assert swvzfe_fail_safe(value) == fail_safe
+
+
+def test_swvzfe_converters_handle_none() -> None:
+    """All converter functions must return None when given None."""
+    assert swvzfe_water_shortage(None) is None
+    assert swvzfe_water_leakage(None) is None
+    assert swvzfe_frost_protection(None) is None
+    assert swvzfe_fail_safe(None) is None
+
+
+def test_swvzfe_quirk_applies(zigpy_device_from_v2_quirk) -> None:
+    """Verify the SWV-ZFE quirk is registered and replaces cluster 0xFC11."""
+    device = zigpy_device_from_v2_quirk(
+        "SONOFF",
+        "SWV-ZFE",
+        cluster_ids={1: {SWVZFECluster.cluster_id: ClusterType.Server}},
+    )
+    assert SWVZFECluster.cluster_id in device.endpoints[1].in_clusters
+    assert isinstance(
+        device.endpoints[1].in_clusters[SWVZFECluster.cluster_id], SWVZFECluster
+    )
--- a/tests/test_sonoff_swv_zfe.py
+++ b/tests/test_sonoff_swv_zfe.py
@@ -380,7 +380,9 @@ def test_swvzfe_cluster_deserialize_repairs_duplicate_array_type_in_multi_record
 
     assert hdr.command_id == foundation.GeneralCommand.Read_Attributes_rsp
     assert len(response.status_records) == 2
-    assert response.status_records[0].attrid == SWVZFECluster.AttributeDefs.child_lock.id
+    assert (
+        response.status_records[0].attrid == SWVZFECluster.AttributeDefs.child_lock.id
+    )
     assert bool(response.status_records[0].value.value) is False
     assert (
         response.status_records[1].attrid
--- a/tests/test_sonoff_swv_zfe.py
+++ b/tests/test_sonoff_swv_zfe.py
@@ -370,9 +370,7 @@ async def test_swvzfe_valve_alarm_settings_numeric_write_updates_payload(
         mock.AsyncMock(return_value=write_response),
     ) as mock_write:
         await local_cluster.write_attributes(
-            {
-                SWVZFEValveAlarmConfigCluster.AttributeDefs.set_frost_temperature.name: 6
-            }
+            {SWVZFEValveAlarmConfigCluster.AttributeDefs.set_frost_temperature.name: 6}
         )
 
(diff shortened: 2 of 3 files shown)


nazjunaid:
My device auto turns off after 10mins, is this a setting chat can be changed via a quirk?

470mr0f1:
> My device auto turns off after 10mins, is this a setting chat can be changed via a quirk?

Same for me!

hcrohland:
I have added with claude's help the configuration for the manual switch configuration here: https://gist.github.com/hcrohland/0f2720a55df21742237b71553c1c0f6c

You can now set the modes to "duration", "duration with interval", or "capacity". When switching the valve on it will follow the selected mode:

- "duration" will run for "Total irrigation duration"
- "duration with interval" will irrigate for "Total irrigation duration" but will pause after every "Interval irrigation duration" for "Interval pause"
- "capacity" will stop after irrigating for "Capacity amount" in "Capacity Unit" or "Fail safe timeout" minutes, whichever comes first

feel free to integrate

Hedda:
@FraserKillip Have you also seen this pull request for Sonoff Hydro Series single-channel and dual-channel water valves?

* https://github.com/zigpy/zha-device-handlers/pull/4993

PS: For reference. there is by the way also a feature request discussion that indirectly relates to all Sonoff water valves:

X-Ryl669:
In this comment, they list additional attributes that aren't exposed in this PR: https://github.com/zigpy/zha-device-handlers/issues/3298#issuecomment-3340700286

For example, the 500D (irrigation start time) and 500E (end time) (although, the device must learn the current time for this to work) can be very useful if the device is smart enough to go to sleep longer while waiting for the next irrigation and thus save batteries.

There's also the file [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/6c86f296808c9988be9d623dd1d2e23130daae45/src/devices/sonoff.ts#L1115) for a description of the schedule attribute (if it's any help for lowering the energy consumption)

ausfas:
Currently I send the timer command to SONFF valve and is run locally on the value so even if HA is down or an issue, SONOFF is internally running a timer and turn off the valve when the timer expires. I can't send this via quirk but send via zha toolkit. Can this this added to the quirk. 
```
action: zha.issue_zigbee_cluster_command
data:
  ieee: xxxx
  endpoint_id: 1
  cluster_type: in
  command_type: server
  cluster_id: 0x0006
  command: 0x42
  args: [0, 45, 1]
```

X-Ryl669:
On time and Off time settings don't work: writing to them gives "unsupported attribute:134" errors
I think they should be  0x4001 and 0x4002 attributes of the "On/Off" cluster_id (and they are, but the device answers with:
```
2026-05-23 10:18:40.747 DEBUG (MainThread) [zigpy.zcl] [0x8D11:1:0x0006] Received ZCL frame: '18 12 04 86 01 40'
2026-05-23 10:18:40.747 DEBUG (MainThread) [zigpy.zcl] [0x8D11:1:0x0006] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl<0x18>(frame_type=<FrameType.GLOBAL_COMMAND: 0>, is_manufacturer_specific=0, direction=<Direction.Server_to_Client: 1>, disable_default_response=1, reserved=0, *is_cluster=False, *is_general=True), tsn=18, command_id=4, *direction=<Direction.Server_to_Client: 1>)
2026-05-23 10:18:40.748 DEBUG (MainThread) [zigpy.zcl] [0x8D11:1:0x0006] Decoded ZCL frame: SWVZFEOnOffCluster:WriteAttributesResponseSchema(status_records=[WriteAttributesStatusRecord(status=<Status.UNSUPPORTED_ATTRIBUTE: 134>, attrid=0x4001)])
```
So it seems like their implementation doesn't support setting a timer when switching the valve on with the standard values. It's likely required to use another proprietary format here for setting this time.

The valve switch on time auto switches off after 600s (10mn) and it's not possible to change that time, rendering the device useless for any long watering or drip irrigation.

@ausfa : Can you explain what your command does? If I understand correctly, you're sending a "On with timed off" command with arguments ("on/off control" = 0, "on time" = 45s, "offtime" = 1s), or is it ("on/off control" = 0, "on time" = 0x12D = 301s ?, no "offtime")
@hcrohland : Your quirks don't work, it fails with many errors ("Failed to send" request when changing the options, unsupported attribute:134 for the times"), maybe I did something wrong? I've replaced this PR quirks file with your gist and restart HA, new entities appeared, then selected "Reconfigure" in the device page.

Yardco:
@X-Ryl669, @hcrohland’s quirk works fine. Make sure to remove and add the device again, otherwise commands will fail.

X-Ryl669:
I've deleted the device, updated the quirks, restarted HA, rejoined the device and still no luck. I can't send the on time and off time (I'm sure I'm using @hcrohland version since the offtime is in minutes in his version, not seconds). It still gives me the same unsupported attribute error. Changing the values in the select element don't return a "Failed to send" error anymore so something changed, but still not the pesky 10mn auto off timer.

Maybe you're talking about the other options that were added by the quirks which I haven't tested them all or I'm missing something. Do you have updated the firmware of the device (and if so, how?), I haven't.

hcrohland:
@X-Ryl669 what about the other settings like capacity? For me all of them work fine.
Another one: Do you have the European or American version?

X-Ryl669:
It's hard for me to test the capacity setting, since I have a rather long drip irrigation line. I think I have the European version, but how can I know?

I'll disconnect the line and try the capacity mode with a bucket to make sure it fills it. But since the valve has a default 10mn timeout, that's a lot of water to capture to ensure it doesn't auto switch off after 10mn.

Did you try to set the "On time" and "Off time" on your version? Are the attribute accepted ?

hcrohland:
Ah, I think I've got it: You need to set the "Irrigation total duration", not the "Off wait time". That one doesn't work for me either.

ausfas:
I'm on 'Firmware: 0x00001004'
The reference I used was the Zigbee2MQTT SONOFF SWV device page. It documents the SWV switch as supporting **“On with timed off”**.

Zigbee2MQTT describes this as sending `state: "ON"` with an additional `on_time` property, where `on_time` is the time in seconds the state should remain ON. It also supports `off_wait_time` as the cooldown time in seconds. The docs note that support depends on firmware and that some devices may require both values.

In Zigbee2MQTT/MQTT form, the equivalent idea is:

```json
{
  "state": "ON",
  "on_time": 300,
  "off_wait_time": 1
}
```

That led me to test the equivalent standard Zigbee command directly in ZHA, which is `OnWithTimedOff` on the On/Off cluster:

* Endpoint: `1`
* Cluster: `0x0006`
* Command: `0x42`
* `on_off_control`: `0`
* `on_time`: duration value
* `off_wait_time`: `1`

Using native ZHA, this works for me:

```yaml
action: zha.issue_zigbee_cluster_command
data:
  ieee: 00:11:22:33:44:55:66:77
  endpoint_id: 1
  cluster_type: in
  command_type: server
  cluster_id: 0x0006
  command: 0x42
  params:
    on_off_control: 0
    on_time: 300
    off_wait_time: 1
```

Using ZHA Toolkit, the same command can be sent as:

```yaml
action: zha_toolkit.zcl_cmd
data:
  ieee: 00:11:22:33:44:55:66:77
  endpoint: 1
  cluster: 0x0006
  cmd: 0x42
  args:
    - 0
    - 300
    - 1
```

So this is not writing the quirk timer attributes. It is sending the standard On/Off cluster timed-on command, which is the same concept exposed by Zigbee2MQTT as `on_time` / `off_wait_time`.

hcrohland:
I've commented out the two number entities in my gist.

X-Ryl669:
Ok, it worked with setting the irrigation total duration value, many thanks!

ikruglov:
> I have added with claude's help the configuration for the manual switch configuration here: https://gist.github.com/hcrohland/0f2720a55df21742237b71553c1c0f6c

I'm running this version on my brand-new SWV-ZFU with 1.0.8 firmware. I see the following problems:
1. `Water shortage` /  `Water leakage` / `Fail safe` are displayed as "Problem" / "Moisture" / "Safety" (see screenshot below). The corresponding sensors are missing translation keys.

Screenshot 2026-05-30 at 19 23 33

2. `Frost Protection` and `Valve working` in `Diagnostics` never change value from off. My valve doesn't report them.  I have `Frost Protection` configuration enabled, but the value in `Diagnostics` remains the same.

nglessner:
**Tested on SWV-ZFU (fw `0x00001007`) — partial success, with a firmware attribute-layout difference from ZFE**

Environment: HA 2026.6.4, zigpy 1.5.1, zha 1.4.1, zha-quirks 1.2.0, ZHA. The quirk imports cleanly via `custom_quirks_path` and creates 30 entities. Thanks for this work — most of it functions on ZFU.

**Works correctly on ZFU:**
- On/off, battery, and the alarm/auto-close switches.
- `real_time_irrigation_volume` / `_duration` (`0x5007`/`0x5006`). The big-endian handling is **required and correct** here: over-the-air the device sends LE bytes `14 00 00 00` / `98 00 00 00`; read as the declared BE uint32 they'd be 335544320 / 2550136832, and the swap yields the sane **20** / **152**.
- `0x500C` abnormal-state, `0x501B` / `0x501C` hourly, `0x5020` alarm settings.

**Two ZFU-specific observations:**

1. **`real_time_irrigation_*` behave as cumulative totals, not per-session** — they climbed monotonically (vol `0→1→28→65→86→115 L`; dur `0→3→8→11→15`) across several on/off cycles without resetting. And the duration looks like **minutes, not the declared seconds**: ~15 over ~15–19 min of runtime is ~1:1 (seconds would be ~900).

2. **Several attributes were never reported and their entities stayed `unknown` / `off` for the entire 80-min capture:** `valve_work_state 0x5010` (→ `valve_working` stuck `off` even while flowing), `daily_irrigation_volume 0x500F`, `daily_irrigation_duration 0x501A`, `longitude 0x5016`, `latitude 0x5017`. I did not separately confirm whether an explicit `ReadAttributes` of these returns `UNSUPPORTED_ATTRIBUTE`, so the defensible statement is "this firmware doesn't report them in normal operation," not "they're absent from the firmware."

**Likely why — an undefined packed array `0x501F`:** the ZFU streams an undefined attribute `0x501F` (21-byte array) **every ~6 s while flowing** (442 reports in my capture). It isn't defined in the quirk, so every report logs a DEBUG `zigpy.appdb … Error binding parameter 8: type 'Array' is not supported`. Proposed decode of a sample `[2,0,1,0, 31 CB 65 AC, 31 CB 68 04, 31 CB 67 F7, 1,0,0,0, 4B]`:
- `0x31CB65AC` = session **start** (big-endian Zigbee-epoch / seconds-since-2000)
- `0x31CB6804` = **end** = start **+ 600 s** (a 10-minute watering window)
- `0x31CB67F7` = **current time**, advancing +1/s (its LSB moves +6 per 6-s frame)
- final byte = frame counter; bytes `[2,0,1,0]` / `[1,0,0,0]` unconfirmed (flags / state).

So on this firmware the real-time / session data appears multiplexed into `0x501F` rather than the discrete `0x5010` / `0x500F` / `0x501A` attributes.

Happy to run controlled tests (fixed-duration watering, explicit attribute reads, full `0x501F` capture across a session) or to test a patch — I have the ZFU on the bench.

nglessner:
**Follow-up: more RE + refinements to my last comment**

Another bench session with the SWV-ZFU (fw `0x00001007`) — full 600 s debug capture (101 `0x501F` frames, write probes, and explicit `ReadAttributes` for the questionable attrs). Refining two claims from my earlier comment, plus the keystone find for the missing `valve_working` source. Calling this a *working* decode from one session, not a definitive field spec.

**Refinement #1 — timestamps inside `0x501F`.** I previously called these BE Zigbee-epoch seconds-since-2000. The magnitude (~26.5 years from a 2000 epoch lands in mid-2026, matching the capture date) is consistent with that — but **the device's RTC is poorly synced and drifts**. Comparing two captures 76 min of wall clock apart, the device-timestamp delta was 4497 s vs wall delta 4568 s (**71 s drift, ~1.5 %**), and the absolute time decoded sits ~32 h off real wall clock — not a tidy TZ offset. **Practical takeaway:** treat the fields as a drifting Zigbee-epoch-ish RTC. The deltas (`end − start` = configured duration, `current` advances ~1/s, `current = end` at auto-close) are reliable; the absolute wall-clock interpretation isn't.

**Refinement #2 — the "never reported" attrs (now write- AND read-confirmed).** I write-probed each (`zha.set_zigbee_cluster_attribute`, value `0`) and then forced ZHA to issue `ReadAttributes` (via `homeassistant.update_entity` on the matching entity — that's the only practical scripted-read path: `zha.issue_zigbee_cluster_command` is hard-coded to cluster-specific commands and raises `KeyError: 0` for the general `Read_Attributes` opcode). Status records were captured from the device:

| attr | quirk role | write resp | read resp (type, value) |
|---|---|---|---|
| `0xFC11 / 0x5010` valve_work_state | feeds `valve_working` | **UNSUPPORTED (0x86)** | **UNSUPPORTED** |
| `0xFC11 / 0x500F` daily_irrigation_volume | daily volume sensor | `SUCCESS` | `SUCCESS`, `uint32_t = 0` |
| `0xFC11 / 0x501A` daily_irrigation_duration | daily duration sensor | `SUCCESS` | `SUCCESS`, `uint32_t = 0` |
| `0xFC11 / 0x5016` longitude | number | `SUCCESS` | `SUCCESS`, **`int32s`** = 0 |
| `0xFC11 / 0x5017` latitude | number | `SUCCESS` | `SUCCESS`, **`int32s`** = 0 |
| `0x0006 / 0x4001` on_time | on_time number | **UNSUPPORTED** | **UNSUPPORTED** |
| `0x0006 / 0x4002` off_wait_time | off_wait_time number | not tested | **UNSUPPORTED** |

The reads return per-attribute `ReadAttributeRecord` with explicit datatypes, ruling out a rubber-stamped-write interpretation — the device persists what you write and returns it on read. End-to-end propagation works too: after the writes the corresponding HA entities (`number.…_longitude` etc.) transitioned from `unknown` → `0` immediately, so a read-on-init in the quirk would populate them. Two patch-relevant details: **`0x5016`/`0x5017` are `int32s`** (signed — lat/long can be negative), and **`off_wait_time` is `UNSUPPORTED` on read** too.

For `0x5010` / `on_time` / `off_wait_time` — drop them for ZFU and source `valve_working` from `0x501F` (next section).

**Probable `valve_working` source — `byte[0]` of `0x501F`.** Across 101 frames I see three distinct values. Working interpretation, based on timing vs. the OnOff cluster confirms:

- `0` — **opening / preamble**: a 15-byte frame sent ~57 ms *before* OnOff confirms `on`.
- `2` — **open / running**: every ongoing 21-byte frame during flow (×99 in this session).
- `1` — **closing**: a single 21-byte frame at auto-close (`current = end`), ~31 ms *before* OnOff confirms `off`.

Working decode of the 21-byte variant (15-byte variant has bytes 0–11 identical, then 3 trailing bytes I haven't decoded):

| byte(s) | field |
|---|---|
| 0 | session-state byte (`0` opening / `2` open / `1` closing) |
| 1 | `0x00` (observed constant) |
| 2 | `0x01` (observed constant) |
| 3 | `0x00` (observed constant) |
| 4–7 | BE uint32 — session **start** (drifting Zigbee-epoch-ish, see #1) |
| 8–11 | BE uint32 — session **end** = `start + duration` (observed `end − start = 600` in every frame; matched auto-close to <100 ms) |
| 12–15 | BE uint32 — session **current**, advances ~1/s; only in 21-byte variant; `= end` in the final close frame |
| 16–19 | `01 00 00 00` (constant across this whole session — needs eWeLink-scheduled vs manual vs no-water comparisons to decode) |
| 20 | uint8 — session-local **tick counter** (observed range 0 → 76 across the 600 s session; ticks roughly every 7–8 s; 100 frames vs 77 unique values means ~23 emission duplicates) |

So a ZFU patch could derive `valve_working = (0x501F.byte[0] == 2)`, and (bonus) expose `session_remaining = end − current` plus the configured duration. **Physical-reality caveat:** a motorised valve takes a couple of seconds to actually open/close, so the binary sensor will flip slightly before/after literal water flow during the `0→2` and `2→1` transitions — probably fine for HA-side purposes.

**Duration unit.** `real_time_irrigation_duration` (declared `s` for `0x5006`) advanced by **8 over an ~8.4 min on-time window** — ~1:1 in **minutes**, not seconds. (`hourly_irrigation_duration` is already labelled `min`.)

Raw frame log + decoded TSV from this session in a local RE workspace. Happy to push a draft PR with the `0x501F` struct + state-byte → `valve_working` derivation + `ReadAttributes` for the four `SUCCESS`-on-read attrs (`int32s` for lat/long) + `on_time`/`off_wait_time`/`0x5010` removed for ZFU + duration-unit fix — let me know if that path is welcome or you'd prefer the captures as a gist first.

nglessner:
**Patch — proposed ZFU support for #4927**

Following up on the [findings](#issuecomment-4759992756) from yesterday — drafted and bench-tested a surgical ~75-line addition to `zhaquirks/sonoff/swv_zfe.py` that takes the SWV-ZFU from "in `.applies_to` only" to actually-working. I ran the draft through adversarial review by both Codex (GPT-5.5) and Gemini (3.1 Pro), applied their consensus corrections, then re-deployed and re-verified on the bench.

What the patch does, narrowly scoped:

1. **Synthesises `valve_work_state` from `0x501F.byte[0] == 2`** — gated by `device.model == "SWV-ZFU"`. The standard attribute `0x5010` is device-confirmed `UNSUPPORTED_ATTRIBUTE` on this firmware (both reads and writes); the device instead multiplexes the open/closed state into `byte[0]` of the packed array attribute it streams every ~6 s during a session.
2. **Silently consumes `0x501F` on ZFU only** — eliminates the per-emission `zigpy.appdb: type 'Array' is not supported` save-error. ZFE/ZNE/ZNU pass through the default attribute pipeline; we don't drop their data on the floor.
3. **Adds explicit `ReadAttributes` for `0x500F`/`0x501A`/`0x5016`/`0x5017`** at config time. These attributes return per-attribute `SUCCESS` with explicit types on read (`uint32_t`, `uint32_t`, `int32s`, `int32s`) and are persisted by the device, but ZFU never auto-reports them. The reads are split into two batches to keep response payloads comfortably under the ~80 B APS MTU these Sonoff devices have been observed to truncate at.

```diff
--- a/zhaquirks/sonoff/swv_zfe.py
+++ b/zhaquirks/sonoff/swv_zfe.py
@@ -412,20 +412,79 @@
                 exc,
             )
 
+    def _update_attribute(self, attrid, value):
+        """Intercept the SWV-ZFU 0x501F packed-state array to drive valve_work_state.
+
+        SWV-ZFU firmware never emits 0x5010 (valve_work_state — confirmed
+        UNSUPPORTED on both read and write); instead it multiplexes the
+        valve open/closed state into byte[0] of a packed uint8 array on
+        attribute 0x501F, emitted every ~6 s during a session. byte[0] is
+        0 in the session preamble, 2 while running, 1 in the closing frame.
+        We parse byte[0] here and update valve_work_state synthetically so
+        the binary sensor works on ZFU.
+
+        We also silently consume 0x501F on ZFU (no super chain for that
+        attrid) to suppress the per-emission `zigpy.appdb: type 'Array' is
+        not supported` save-error. The consume is scoped strictly to the
+        ZFU model so other variants we have not characterised on the
+        bench (ZFE/ZNE/ZNU) keep the default attribute pipeline.
+        """
+        if (
+            attrid == 0x501F
+            and getattr(self.endpoint.device, "model", None) == "SWV-ZFU"
+        ):
+            # zigpy delivers Array(value=AnonymousLVList(value=[uint8,…])).
+            # Unwrap one .value to the LVList (which is a list subclass).
+            payload = getattr(value, "value", value)
+            try:
+                state_byte = int(payload[0])
+            except (IndexError, TypeError, ValueError):
+                state_byte = None
+            if state_byte is not None:
+                super()._update_attribute(
+                    self.AttributeDefs.valve_work_state.id,
+                    state_byte == 2,
+                )
+            return
+        super()._update_attribute(attrid, value)
+
     async def apply_custom_configuration(self, *args, **kwargs):
-        """Read private attrs during pairing to populate entities."""
-        try:
-            await self.read_attributes(
-                [
-                    self.AttributeDefs.valve_alarm_settings.id,
-                    self.AttributeDefs.valve_abnormal_state.id,
-                ]
-            )
-        except Exception as exc:
-            self.warning(
-                "Unable to read SWV-ZF* private attrs during configuration: %s; continuing without initialization",
-                exc,
-            )
+        """Read private attrs during pairing to populate entities.
+
+        On SWV-ZFU firmware several attributes are supported but never
+        auto-reported: daily_irrigation_volume (0x500F),
+        daily_irrigation_duration (0x501A), longitude (0x5016) and
+        latitude (0x5017). They return real per-attr ReadAttributesResponse
+        records with explicit types when read, but the entities stay
+        'unknown' without an explicit init read. Other variants either
+        already auto-report these or will return UNSUPPORTED_ATTRIBUTE
+        here (harmless).
+
+        The reads are split into two batches so each response payload
+        stays comfortably under the typical ~80-byte APS MTU these Sonoff
+        devices have been observed to truncate at; a single 6-attr read
+        with the int32 lat/lon fields can otherwise be silently dropped.
+        """
+        for attrs in (
+            [
+                self.AttributeDefs.valve_alarm_settings.id,
+                self.AttributeDefs.valve_abnormal_state.id,
+            ],
+            [
+                self.AttributeDefs.daily_irrigation_volume.id,
+                self.AttributeDefs.daily_irrigation_duration.id,
+                self.AttributeDefs.longitude.id,
+                self.AttributeDefs.latitude.id,
+            ],
+        ):
+            try:
+                await self.read_attributes(attrs)
+            except Exception as exc:
+                self.warning(
+                    "Unable to read SWV-ZF* private attrs %r during configuration: %s; continuing",
+                    attrs,
+                    exc,
+                )
```

**Bench verification** on SWV-ZFU fw `0x00001007`, HA 2026.6.4 / zigpy 1.5.1 / zha 1.4.1 / zha-quirks 1.2.0, after deploy + restart, on the corrected (reviewer-approved) version:

| event | t (UTC) | Δ |
|---|---|---|
| `switch.turn_on` | `21:32:30.581` | — |
| `binary_sensor…valve_working` → `on` | `21:32:31.016` | **+0.43 s** |
| `switch.turn_off` | `21:32:52.169` | — |
| `binary_sensor…valve_working` → `off` | `21:32:52.163` | **−6 ms** (before the OnOff confirm) |

`0x501F` state-byte sequence across the session: `0` (15-byte preamble) → `2` × 4 (21-byte ongoing) → `1` (final closing frame, `current = end`). **Zero** `zigpy.appdb "type 'Array' is not supported"` errors during the run. Both open and close transitions were also separately verified on the pre-correction draft of the patch (~57 s session before this 22 s one).

**Review notes worth flagging**, since two of these are non-obvious:

- The **model gate** is exact (`model == "SWV-ZFU"`) rather than substring — protects a hypothetical future `SWV-ZFUXL` from accidentally matching.
- The **silent-consume `return` is strictly inside the ZFU gate** — ZFE/ZNE/ZNU keep their default `_update_attribute` path. We don't actually know whether any of those emit `0x501F`; if a ZFE owner finds the same packed-array pattern, the gate can widen with their evidence.
- The **`apply_custom_configuration` read split** is a Gemini call-out: a single 6-attribute read with the `int32` lat/lon fields can exceed the ~80 B APS payload these Sonoff devices have been observed to truncate at (no fragmentation). Splitting into two reads is cheap insurance.
- Overriding `_update_attribute` (rather than the existing event-listener pattern used for `0x5020`) is intentional: the listener pattern would still let `zigpy.appdb` see the `Array` and log the save-error. Override hijacks the pipeline before the cache/db layer.

Not currently covered by a unit test — happy to add one against the observed wire-format (`Array(value=AnonymousLVList(value=[2,0,1,0,…]))`) if that would help. And happy to turn this into a PR against your branch if that's easier to apply than `git am` against this comment.

Si-Hill:
I have the new SWV-ZF2, I have used hcrohland's quirk but had to add .applies_to("SONOFF", "SWV-ZF2") to get it working (locally only). Looks like the sensors are working, but none of the configurations, they all "Failed to send request" which means I cant change the 10min timer for example.

EDIT: the real-time irrigation number seems a little high...

image

image

HarlemSquirrel:
Yeah I had this issue initially with SWV-ZFU after loading this custom quirk which this is the first time I've done this. I had to unpair the device and pair it again and then it started working to change the config. I tested changing the local timer to a few different values and it worked well to turn off after the set time.

carlonluca:
Can this be of help? https://forum.ewelink.cc/t/hydro-one-series-water-valves-now-updated-with-enhanced-zha-script-for-more-powerful-features/208644

nglessner:
Following up on my earlier ZFU notes, and picking up the recent reports from
@Si-Hill and @HarlemSquirrel. Two things from the bench plus a heads-up about an
official quirk.

### 1. Verified bug: the real-time sensors are byte-swapped wrong on SWV-ZFU

@Si-Hill's "the real-time irrigation number seems a little high" reproduces here,
and it traces to `swvzfe_be_swap`.

Bench: **SWV-ZFU** fw `0x00001007`, HA `2026.6.4` / zigpy `1.5.1` / zha `1.4.1` /
zha-quirks `1.2.0`. During an active watering session the real-time entities read:

| entity | displayed | hex |
|---|---|---|
| `real_time_irrigation_volume` | `100663296` L | `0x06000000` |
| `real_time_irrigation_duration` | `16777216` s | `0x01000000` |

But the wire value is already correct little-endian — from the `zigpy.event`
debug log on the same read:

```
attribute_read … attribute_name='real_time_irrigation_volume',   attribute_id=20487, raw_value=6, value=6
attribute_read … attribute_name='real_time_irrigation_duration', attribute_id=20486, raw_value=1, value=1
```

zigpy decodes `6 L` / `1 min` (sane for the run length); `swvzfe_be_swap` then
swaps `6` → `0x06000000` = `100663296`. So on ZFU the converter is *introducing*
the corruption rather than fixing it. The non-swapped sensors on the same device
(`hour_irrigation_*`) display sane values, consistent with the firmware sending
these little-endian.

**One caveat before anyone just deletes the swap:** the encoding looks
session-state-dependent. After the valve closed, the same sensor settled to sane
little-endian values (`14`, then `6`) — but mid-session it showed the swapped
garbage. That mirrors the dual frame format I documented for `0x501F` (a
shorter "preamble" vs. the longer "running" report). I only had a ~2-minute
capture, which isn't enough to characterize it. If a maintainer/author with a
device can grab a full single-session debug capture of every `0x5006`/`0x5007`
frame (preamble → running → closing), that would settle whether the fix is
"drop the swap," "swap only the running-frame format," or "model-gate it." I'm
happy to run a longer capture on the ZFU if that helps — just constrained on
water right now.

(For what it's worth, this is independent of my `valve_working` synthesis from
`0x501F.byte[0]`, which continues to work: `on` at +0.46 s after open, `off` on
close.)

### 2. Sonoff has published an official ZHA quirk

@carlonluca's link points to one: eWeLink forum →
`help.sonoff.tech/docs/SWVZFU-SWVZFE` ships `Hydro_Series_ZHA_Script.zip`
(`swv1c&2c.py`, a v2 `QuirkBuilder`, applies to ZFU/ZFE/ZF2E/ZF2U/ZNU/ZNE). It's
much smaller than this PR and makes some different choices worth knowing about:

- It exposes water usage via **`0x501B`/`0x501C`**, not the `0x500F`/`0x501A`
  ("daily") attributes this PR uses — possibly relevant to the "number seems off"
  reports.
- It decodes **`0x500C`** as an alarm **bitfield** (`shortage=bit0`, `leak=bit1`,
  `frost=bit2`, `ch2 shortage=bit4`).
- It does **not** expose the real-time `0x5006`/`0x5007` sensors at all — which is
  how it sidesteps the byte-order issue in (1).
- It does **not** expose a valve open/running binary sensor on ZFU either.

I don't think the official script obsoletes this PR — it's thinner and drops the
running-state sensor — but its attribute choices are a useful cross-check for
where this one should land.

Glad to help however is most useful: longer ZFU captures, or turning the ZFU
fixes into a PR against this branch.

nglessner:
Another **SWV-ZFU** data point (fw `0x00001007`), plus one addition to offer.

### 1. @hcrohland's manual duration-config works on a second ZFU

Integrated the `0x501D` `manual_default_settings` handling from @hcrohland's gist
and it works end-to-end here. Read back `irrigation_mode=duration`,
`irrigation_total_duration=10` (the stock 10-min default everyone's hitting),
set it to 15 via the number entity, and the valve auto-closed on its own at
**15 min 01.9 s** (bench-timed open→close). So the packed-array write is solid on
ZFU.

Corroborating @HarlemSquirrel: the write only takes **after a re-interview**. A
plain quirk reload/restart creates the entities but leaves them `unknown` and the
device rejects writes — because the `0x501D` read + config path only runs during
the interview. Re-pairing over the top (permit-join + 5 s button, same IEEE, no
device removal) fixes it and preserves entity IDs.

### 2. Offering a `valve_working` (running-state) source for ZFU

The PR derives `valve_working` from `0x5010`, but **ZFU never emits `0x5010`** —
it's `UNSUPPORTED` on read and write here, so the binary sensor is dead on ZFU.
The running state is instead multiplexed into **`byte[0]` of the packed `0x501F`
array** (emitted ~every 6 s during a session): `0` = preamble, `2` = running,
`1` = closing frame. Deriving `valve_working = (0x501F.byte[0] == 2)` gives a
working sensor, fires ~0.46 s after `switch.turn_on` and tracks auto-close to
within ~76 ms. (Sonoff's official quirk doesn't expose running-state at all, so
this may be worth keeping.) Happy to send it as a commit if useful — full decode
and the merge against this PR's head are here:
https://gist.github.com/nglessner/45dc518dd30b98826e4fb98277a1192b

### 3. re #4340 — ZCL `on_with_timed_off` is inert on ZFU

For anyone trying the standard timer path: `on_with_timed_off` (cluster `0x0006`,
cmd `0x42`) is silently ignored on ZFU fw `0x00001007` — it doesn't even open the
valve. `on_time`/`off_wait_time` attributes are `UNSUPPORTED` too. The `0x501D`
duration is the only timer mechanism that works on this variant.

danoene:
Hi, I've spent the past week building and live-debugging a full quirk for the SONOFF SWV-ZFE (fw 0x00001007, three devices, ZHA on current HA / zigpy). Sharing the findings here since this PR targets the same device — feel free to adopt anything useful.

Complete working quirk:
[sonoff_swv_zfe.py Gist](https://gist.github.com/danoene/8b5a6ccd06be405f04db73ac07c5a3df)

## Key findings

**1. Attribute mapping on 0xFC11 (all without manufacturer code):**
- `0x501B` / `0x501C` are **hourly** counters (reset every hour on the device clock), *not* lifetime totals — this is why earlier quirks saw "water usage resets every cycle".
- Daily totals are `0x500F` (volume, litres) / `0x501A` (duration, minutes); they report once at midnight (device-local) and can be read on demand.
- Real-time session stats are `0x5006` (duration) / `0x5007` (volume): **uint32 sent big-endian** inside the little-endian ZCL frame, so a byte swap is required (`int.from_bytes(v.to_bytes(4,'little'),'big')`). On this firmware the real-time *duration* unit is **minutes**, not seconds.
- `0x500C` abnormal-state bitmap: bit0 water shortage, bit1 leak, bit2 frost protection, bit3 fail-safe.
- `0x5020` `valve_alarm_settings` is a packed 4-byte uint8 array: [enable bit flags, shortage duration (min), leak duration (min), frost temperature (°C)].
- `0x501D` `manualDefaultSettings` (12-byte array, big-endian u16 fields) governs **every** plain start — physical button *and* plain Zigbee ON: duration mode or volume mode with fail-safe timeout. Default is a 10-minute window.

**2. Device clock (this makes or breaks the daily/hourly counters):**
The valve reads genTime (cluster 0x000A) from the coordinator right after joining and roughly every 45 minutes (cycles are aligned to the device clock, all our valves query within the same 2 seconds). Without a Time server cluster in the quirk, zigpy drops the request, the clock is never set and the daily/hourly counters stay dead (reads return unsupported). Adding a bare `CustomCluster, Time` fixes the sync — clocks end up accurate to ~3 s, verified via hourly/midnight rollovers.

**3. Critical stability issue (probably the "random reboots / flapping" people see):**
The valve requests attribute_ids `[0, 7, 2, 3, 4, 5]`. zigpy's built-in Time server answers `dst_start`/`dst_end`/`dst_shift` with `UNSUPPORTED_ATTRIBUTE` status records. The firmware *does* set its clock from the SUCCESS records — and then **crashes and reboots** on the response. After a battery change this becomes a boot → time read → crash loop (constant availability flapping). This matches why Z2M needed the custom `swvznGenTimeCompatResponse` (Koenkk/zigbee2mqtt#29582).
Robust fix without hand-crafting frames — filter the request down to the attributes zigpy actually serves, then let zigpy's proven handler reply (all-SUCCESS response):

```python
class EwelinkTimeServerCluster(CustomCluster, Time):
    def handle_cluster_general_request(self, hdr, args, *, dst_addressing=None):
        if hdr.command_id == foundation.GeneralCommand.Read_Attributes:
            filtered = [a for a in args.attribute_ids if a in (0x0000, 0x0001, 0x0002, 0x0007)]
            if filtered:
                args.attribute_ids[:] = filtered
        return super().handle_cluster_general_request(hdr, args, dst_addressing=dst_addressing)
```

With this in place our three valves have been running stable through dozens of 45-minute sync cycles — before, they rebooted within minutes of every time read.

**4. Firmware bug in array read responses:**
Read responses for the array attributes `0x5020` and `0x501D` declare the array *element* type as `array` (0x48) instead of `uint8` (0x20) — raw example: `... 1d 50 00 48 48 0c 00 ...`. zigpy deserialization fails on this; the quirk repairs the frame by rewriting the malformed prefix before deserializing.

**5. Not supported on this firmware (fw 0x00001007):**
`0x5021` (amount unit) returns UNSUPPORTED; OnOff `on_time` / `off_wait_time` attributes are unsupported and the `on_with_timed_off` (0x42) command is accepted (APS/Default Response) but **ignored** — hardware-timed runs only work via `0x501D` manualDefaultSettings.

**6. Reporting:**
`ReportingConfig(min_interval=30, max_interval=900, reportable_change=1)` on the stats attributes works, but the binding only sticks when reconfiguration happens while the sleepy device is awake (e.g. right after battery insert). Bindings survive power cycles.

Happy to provide logs/raw frames for any of the above.

---
*Disclosure: all reverse-engineering, testing, and hardware debugging above is my own work from live devices; this write-up was drafted with the help of Claude (Anthropic) based on my findings and logs.*

hcrohland:
@danoene : Looks great! Should I add the volume based config parameters or are you in progress of doing that?

danoene:
> @danoene : Looks great! Should I add the volume based config parameters or are you in progress of doing that?

Thank you very much. 
I don't have any plans to do that right now, but feel free to add it.

hcrohland:
Aah, they are already there. Somehow missed them in my first try.
One question though: Why do you limit config changes to parameters for the current active Manual Irrigation Mode?
I find it convenient to set them independent of current mode and have them ready when switching.

nglessner:
Confirming @danoene's `EwelinkTimeServerCluster` on a **SWV-ZFU** (fw `0x00001007`, HA `2026.7.2` / zigpy `2.0.0` / zha `2.0.0` / zha-quirks `2.1.1`). It fixes silent-reboot behavior for us.

### Before (no Time filter)

Watching `sensor.<valve>_lqi` — the switch entity doesn't flip on reboots, but LQI transitions `int → unknown → int` on each rejoin. On 2026-07-16 in a 90 min window I recorded **3 rejoins**:

```
18:32:55Z  185 → unknown → 185 at 18:34:25Z  (~90 s recovery)
18:40:32Z  182 → unknown → 189 at 18:44:38Z  (~4 min recovery)
19:03:31Z  185 → unknown → 182 at 19:04:14Z  (~43 s recovery)
```

### After (with Time filter, re-interviewed)

Ported danoene's class verbatim into our local quirk, restarted HA, held the valve button ~5 s to re-interview. Over the following **206 min** (covering 4–5 45-min Time-sync cycles): **0 `unknown` transitions**, LQI steady 178–185.

So this fix is not just a ZFE thing — it lands the ZFU crash too, and the "availability flapping after battery pulls" description matches what I was seeing without a battery pull (regular Time-read cycles are enough to trigger it on this fw).

### Aside: one ZFU-specific difference vs the ZFE decode

`0x5010 valve_work_state` returns `UNSUPPORTED` on ZFU fw `0x00001007`. I've been synthesizing running-state from `byte[0]` of `0x501F` instead (`0` = preamble, `2` = running, `1` = closing frame — fires ~0.46 s after `switch.turn_on`, tracks auto-close to within ~76 ms). Doesn't matter for the Time fix, just flagging so it doesn't get dropped when this PR consolidates. Happy to send a small commit if useful. Details in my 2026-07-02 comment.

starkillerOG:
I just tried: https://github.com/zigpy/zha-device-handlers/pull/5114 and it works greath for the SWV-ZFE and also allows control over the manual irrigation duration and volume. It looks like that quirk was made by a Sonoff employee.

hcrohland:
After running in some problems I updated my gist to include @danoene's timecluster and some bug fixes.
You can find the latest version at the usual place: https://gist.github.com/hcrohland/0f2720a55df21742237b71553c1c0f6c

PR #4928, opened 2026-04-16
Add support for Develco/frient Smart Sirens (SIRZB-110/111/112)

marcintyminski:
## Proposed change

The quirk has been made to extend functionality of the following frient/Develco Smart Sirens:

 - frient Smart Siren with model no. SIRZB-110
 - frient Smart Siren with model no. SIRZB-111
 - frient Smart Siren with model no. SIRZB-112
 - Develco Smart Siren with model no. SIRZB-110
 - Develco Smart Siren with model no. SIRZB-111
 - Develco Smart Siren with model no. SIRZB-112

 - Delete strobe settings
 - Add maximum siren duration setting
 - Add squawk buttons
 - Add squawk volume level setting
 - Add power entity to track if the device is AC or battery powered
## Additional information

image

## Device diagnostics

[zha-01KJYXWD9VV3FMNDJEFN115MC4-frient A_S SIRZB-110-e7de2bb8b79f142fff44856f7cb425d9.json](https://github.com/user-attachments/files/26777913/zha-01KJYXWD9VV3FMNDJEFN115MC4-frient.A_S.SIRZB-110-e7de2bb8b79f142fff44856f7cb425d9.json)
[zha-01KJYXWD9VV3FMNDJEFN115MC4-frient A_S SIRZB-111-7a90ebaec843facbbdb23a97b55e45b1.json](https://github.com/user-attachments/files/26777918/zha-01KJYXWD9VV3FMNDJEFN115MC4-frient.A_S.SIRZB-111-7a90ebaec843facbbdb23a97b55e45b1.json)
[zha-01KJYXWD9VV3FMNDJEFN115MC4-frient A_S SIRZB-112-34a1c8ad5ab2867faf337cf8e4ce11f7.json](https://github.com/user-attachments/files/26777920/zha-01KJYXWD9VV3FMNDJEFN115MC4-frient.A_S.SIRZB-112-34a1c8ad5ab2867faf337cf8e4ce11f7.json)

## Checklist


TheJulianJES (commented):
Yeah, we do have a service/action call and a device action in HA Core for this: `zha.warning_device_squawk`. It's not as friendly as the UI options but virtual config entities aren't generally allowed in HA Core, though we already have some old ones for some siren models...

I think the current solution with overriding the `squawk` command with values from local attributes isn't ideal. Personally, I do like having config entities for this, but it's not consistent with other siren implementations then, and technically, there's already the HA service/action call.

For an initial PR here, we may want to just add the "AC power" binary sensor. (That also requires no test coverage because it's purely "a definition".)
I'm not sure about `max_duration` yet. That is an attribute that actually exists on the device, but I'll need to double check when it would actually have an effect. (Using the siren entity should auto turn-off after like 15 IIRC, so maybe only for the service/action call?)

marcintyminski:
max-duration always takes an effect. The value of it always describes maximal time the siren will be on, even if you set other value in the advanced control. The siren will appear to be on in the UI (for whatever period you set in advanced control), but it will not make any sound after the max_duration time. 

Should we drop the squawk from the quirk then?

TheJulianJES:
Yeah, let's remove the virtual "squawk level" attribute and all the logic for that here. HA Core isn't a fan of new virtual entities being added and the `zha.warning_device_squawk` service/action call already exists, which can also set the squawk level when issuing the command.

For "max duration", I'm wondering if we should just globally add that to ZHA for all supported devices. We can just add that entity if the attribute is supported by a device.

PR #4934, opened 2026-04-19
Ignore missing default reply for Leviton dimmer

stephenjust:
## Proposed change

Leviton Load Controller dimmer modules LU107, ZK700-D0W both don't respond to input commands on their LevelControl clusters.

Add a simple quirk using the NoReplyMixin to fix LevelControl behavior.

## Additional information

These are two generations of the same 0-10V dimmer load controller. I suspect the same quirk is needed by other LuminaRF or GreenConnect load controllers, but these are the two I have on-hand.

## Device diagnostics

[zha-fe3d142bb254218f191475085bd9eaa1-LEVITON ZK700-D0W-4b70c9f394b8093dd900b632d84c3888.json](https://github.com/user-attachments/files/26876334/zha-fe3d142bb254218f191475085bd9eaa1-LEVITON.ZK700-D0W-4b70c9f394b8093dd900b632d84c3888.json)
[zha-fe3d142bb254218f191475085bd9eaa1-Leviton Inc LU107-6c9cc0605f063dcd07a70bfc9d438488 (1).json](https://github.com/user-attachments/files/26876336/zha-fe3d142bb254218f191475085bd9eaa1-Leviton.Inc.LU107-6c9cc0605f063dcd07a70bfc9d438488.1.json)

## Checklist

### DIFF
--- a/tests/test_leviton.py
+++ b/tests/test_leviton.py
@@ -1,0 +1,52 @@
+"""Tests for Leviton dimmers."""
+
+from unittest import mock
+
+import pytest
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.general import LevelControl
+import zigpy.zdo.types as zdo_t
+
+from zhaquirks.leviton.dimmer import LevitonLevelControl
+
+Default_Response = foundation.GENERAL_COMMANDS[
+    foundation.GeneralCommand.Default_Response
+].schema
+
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


TheJulianJES (approved):
Thanks! Looks good.

This is one of the few cases where we probably wouldn't really have needed test coverage but it's fine to keep. We'll improve our Copilot instructions. Either way, thanks for adding tests.

PR #4950, opened 2026-04-27
Add input-output linking and reverse polarity to frient IOMZB-110

marcintyminski:
## Proposed change

The quirk has been made to extend functionality of the following frient/Develco IO Modules:

 - frient IO Module with model no. IOMZB-110
 - Develco IO Module with model no. IOMZB-110

 - Add output being controlled by input functionality
 - Add reverse polarity setting for each input
 - Add on and off time delay to each output

## Additional information

image
image

## Device diagnostics

[zha-01KJYXWD9VV3FMNDJEFN115MC4-frient A_S IOMZB-110-ce03dbffd3e84878b8e7b03edc08fe38.json](https://github.com/user-attachments/files/27119555/zha-01KJYXWD9VV3FMNDJEFN115MC4-frient.A_S.IOMZB-110-ce03dbffd3e84878b8e7b03edc08fe38.json)

## Checklist


TheJulianJES (commented):
Hmm, we don't have a good way to expose `on_with_timed_off` in HA Core. I think the most correct way would be with a custom service/action, but even that isn't great.

The local/remote attribute stuff is also a hacky (with creating a task after an attribute update). Without that, it's similar to how it's used elsewhere to work around some limitations that will be lifted soon-ish. But even then, I'm not sure we can have a number entity like this influence a switch entity's normal on/off action.

If that on/off delay is removed for now, we may be able to keep the "linked output" settings but simplify that code a lot, maybe have the local stuff similar to the Ubisys quirks.

Eventually, the entire local stuff should become unnecessary for quirks. (Technically, you could already do it without that directly in ZHA, but we should also stop adding manufacturer-specific stuff directly to ZHA at this point. Quirks can do that soon.)

TheJulianJES:
And for the strings, they were made to be similar to what's on the device, but I think it's fine to change them.

PR #4980, opened 2026-05-07
Add settable current summation for frient EMIZB-141

marcintyminski:
## Proposed change

The quirk has been made to extend functionality of a frient Electricity Meter Interface 2, LED with model no. EMIZB-141.
The change introduced in this pull request adds a setter for current summation delivered, so the user can align it with reading from physical meter.

## Additional information

2 LED after change

## Device diagnostics

[zha-01KJYXWD9VV3FMNDJEFN115MC4-frient A_S EMIZB-141-05c92b77056a6a986a950df6acb3b17c.json](https://github.com/user-attachments/files/27466927/zha-01KJYXWD9VV3FMNDJEFN115MC4-frient.A_S.EMIZB-141-05c92b77056a6a986a950df6acb3b17c.json)

## Checklist


TheJulianJES (commented):
Hmm, yeah. I think this may not be acceptable for Home Assistant Core to have a `number` entity and a `sensor` entity like this. Even before, you could already use the "Manage Zigbee device" UI on the device page, select the cluster, and then manually write the attribute there (or use the service/action call for setting an attribute).

Even if the number entity is fully synchronized with the sensor entity, which we could do on the quirks/ZHA side, I think HA Core still wouldn't accept to have both a read/write number entity, and a dedicated sensor entity. This needs some more discussion.

(Side note: The `ManufacturerMetering` with the different/fake ID could be migrated away to a proper cluster extending `Metering` now, with the manufacturer attributes using Frient's `manufacturer_code`. Zigpy didn't properly support overlapping attribute IDs with different manufacturer codes a while ago, hence this current workaround is in place.)

marcintyminski:
The intent of the number entity is calibration. It allows the user to set the starting counter value to match the physical meter reading. The sensor entity shows the live reading, while the number entity is a write to sync the device's counter with the real-world meter. They serve different purposes. If HA Core policy still doesn't allow this combination, let me know and I can get rid of it, but it would be preferable for us to leave it as it is.

TheJulianJES:
Even without this PR, it should already be possible to set the "current summation" value, though it's not very user-friendly. To do that, you can follow these steps:
1. Go to the device page in HA.
2. Click on the three dots on the left upper card.
3. Click on "Manage Zigbee device".
4. Click on "Change cluster", and select the `ManufacturerMetering` cluster.
5. In the dropdown for the attribute selection, select the `current_summation` attribute.
6. Then enter the value you want to set the `current_summation` to in the first text field.
7. The second text field (manufacturer code) should be left empty, as it should be automatically populated in the background.
8. Finally, press "write attribute"
9. Check that the sensor has the new value.

However, I'll need to discuss this PR with someone from the HA Core team. As far as I know, it may not be allowed to have a second entity for the same state that's also present in a sensor entity, just read-only once and read/write for the number entity. But I think we also need to keep the sensor entity for the energy dashboard and so on. So we have a bit of a problem here.

The second problem, that's likely not allowed in HA Core, is that this PR would provide a user-visible way to change the value of a `SensorStateClass.TOTAL_INCREASING` sensor (which is what "Current summation delivered" is).
HA Core will always log warnings about this, since a `TOTAL_INCREASING` sensor entity is not allowed to go back to a lower state, except to 0 every day/month/week/year ([HA dev docs](https://developers.home-assistant.io/docs/core/entity/sensor?_highlight=sensor#available-state-classes)). A warning should already be logged when you use the above way to write the value.

In general, from a HA Core point of view, you wouldn't even want to modify the "current summation delivered", since you mess up long-term statistics that way. A sudden jump in that sensor will show on the energy dashboard like you've consumed that energy, even though you probably just wanted to synchronize one number in HA to the one on your meter.

But yeah, I'll ask around about this again.

TheJulianJES:
It seems like adding a number entity for this is not something we want at the moment.

Whilst the way above allows you to sync this number via the UI (which may log a warning in some cases and produce unexpected long-term statistics), I'm aware it's not easily discoverable.

The Z-Wave integration has a dedicated page in HA, similar to our "Manage Zigbee device" page but with preconfigured "values" (similar to attributes) that can be changed.  I wonder if we could also do something like that for ZHA, for when we don't want to (or can't) certain "device internals" as fully-blown entities, but just as some (slightly) easier-to-discover options. But this isn't something that'll happen in the short-term.

But essentially, we could mark certain attributes (or non-ZCL "state/values") with some flag to more easily expose them on a dedicated tab/page for the user to tweak, possibly also with a description.

PR #4896, opened 2026-04-02
Add Tuya soil sensor `_TZE284_tgrzpqf4` variant

jacsamell:
## Proposed change

Add support for the `_TZE284_tgrzpqf4` Tuya TS0601 soil moisture sensor variant.

This device is physically identical to other sensors in this family (same packet as a working `_TZE284_aao3yzhs` unit) but reports a different manufacturer string. The existing quirk for `_TZE284_aao3yzhs` applies correctly — tested and confirmed working with soil moisture, temperature, and battery entities all reporting.

## Additional information

- Two physically identical sensors from the same batch — one identified as `_TZE284_aao3yzhs` (matched existing quirk), the other as `_TZE284_tgrzpqf4` (did not match).
- After adding this fingerprint as a local custom quirk, the previously unsupported sensor paired correctly with all expected entities.

## Device diagnostics

**Zigbee device signature (before quirk):**
```json
{
  "node_descriptor": {
    "logical_type": 2,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4417,
    "maximum_buffer_size": 66,
    "maximum_incoming_transfer_size": 66,
    "server_mask": 10752,
    "maximum_outgoing_transfer_size": 66,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xed00",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE284_tgrzpqf4",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

**After applying the quirk, the device exposes:**
- Soil moisture sensor
- Temperature sensor
- Battery sensor

Using the same datapoint mapping as the sibling variants (dp 3 = soil moisture, dp 5 = temperature, dp 15 = battery).

## Checklist

- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [ ] Tests have been added to verify that the new code works
- [x] Device diagnostics data has been attached

PR #4987, opened 2026-05-10
Add Tuya soil sensor _TZE284_0ints6wl variant

Hafpaf:
## Proposed change

Add support for model TS0601 by _TZE284_0ints6wl, as it does not provide temperature, soil moisture and battery information correctly in Home Assistant. 

Similar to Pull request #4896

## Additional information

This is sold as "Zigbee Soil Moisture Sensor Meter Temperature Humidity Light Tester For Tuya Smart Life Z2M Home Assistant Automation" from Aliexpress: https://www.aliexpress.com/item/1005010168958171.html where I choose the "4in1 zigbee Tuya" version

I have not managed to make the Illumination sensor work yet

## Device diagnostics

Zigbee device signature before quirk:
```
{
  "node_descriptor": {
    "logical_type": 2,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4417,
    "maximum_buffer_size": 66,
    "server_mask": 10752,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xed00",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    }
  },
  "manufacturer": "_TZE284_0ints6wl",
  "model": "TS0601",
}
```

## Checklist

### DIFF
--- a/zhaquirks/tuya/tuya_sensor.py
+++ b/zhaquirks/tuya/tuya_sensor.py
@@ -270,6 +270,7 @@ class NoManufTimeTuyaMCUCluster(TuyaMCUCluster):
     .applies_to("_TZE284_awepdiwi", "TS0601")  # Solar powered
     .applies_to("_TZE284_33bwcga2", "TS0601")  # iHseno
     .applies_to("_TZE284_tgrzpqf4", "TS0601")
+    .applies_to("_TZE284_0ints6wl", "TS0601")
     .tuya_temperature(dp_id=5, scale=10)
     .tuya_battery(dp_id=15)
     .tuya_soil_moisture(dp_id=3)


Swissola:
Hi @Hafpaf, thanks for identifying support for this model — I have the same `_TZE284_0ints6wl` variant and have been running a local custom quirk based on your DP mapping for a few weeks. Wanted to share two corrections/additions I found, since I don't think this variant is actually identical to the `_TZE284_aao3yzhs` group.

**Battery (dp=15) doesn't fire for this manufacturer ID.** After 3+ days with zero reports despite temperature/soil-moisture DPs firing hourly, I enabled `zhaquirks.tuya`/`zigpy.zcl`/`zha` debug logging and physically woke the device, capturing *every* DP it sent rather than just checking dp=15. Two previously-unmapped DPs turned up: **dp=14** (enum, raw value 1) and dp=111 (still unidentified, not battery-related). I cross-referenced dp=14 against the old Tuya-cloud integration's equivalent (now-`unavailable`) entity for this device, which still carries its original schema: `options: ['low','middle','high']`, `device_class: enum`. Raw value 1 maps exactly to index 1 = `middle` — an exact match. So for `_TZE284_0ints6wl`, battery is a 3-tier enum on dp=14, not a 0–100 percentage on dp=15.

To still get a normal HA battery icon rather than a plain enum tile, I mapped Low/Middle/High to display percentages **derived from real percentage bands** (not arbitrary round numbers) — 20/60/100% — via a lookup-table converter rather than `.tuya_battery()`'s built-in linear scale (which would map Low=0 → 0%, reading as "dead" rather than "low"):

```python
from zhaquirks.const import BatterySize
from zhaquirks.tuya import TuyaPowerConfigurationCluster
from zigpy.zcl.clusters.general import PowerConfiguration

# Low/Middle/High (raw 0/1/2) -> stored half-percent ZCL units -> displayed 20%/60%/100%
_BATTERY_STATE_TO_STORED_PERCENT = {0: 40, 1: 120, 2: 200}

.tuya_dp(
    dp_id=14,
    ep_attribute=TuyaPowerConfigurationCluster.ep_attribute,
    attribute_name=PowerConfiguration.AttributeDefs.battery_percentage_remaining.name,
    converter=lambda x: _BATTERY_STATE_TO_STORED_PERCENT.get(x, 0),
)
.adds(
    TuyaPowerConfigurationCluster,
    constant_attributes={
        PowerConfiguration.AttributeDefs.battery_size: BatterySize.AA,
        PowerConfiguration.AttributeDefs.battery_rated_voltage: 15,
        PowerConfiguration.AttributeDefs.battery_quantity: 2,
    },
)
```

**Illuminance (dp=102) does work** — you noted in the PR description you hadn't managed to get this going. Same debug-log technique: shone a torch on the sensor mid-wake and watched the raw value jump from a baseline of 277 up to 6377→8294 across successive wake packets, tracking the light directly. It maps cleanly with the standard convenience method:

```python
.tuya_illuminance(dp_id=102)
```

One more thing worth flagging: my device also needed `.tuya_enchantment(data_query_spell=True)` — without it, dp=14/dp=102 (and originally dp=15) never volunteered a report on their own; this spell actively sends a `TUYA_QUERY_DATA` request during device configuration instead of passively waiting.

Given the battery DP and illuminance support differ from the aao3yzhs group, `_TZE284_0ints6wl` needs its own quirk block rather than joining that group's `.applies_to()` list — so I've opened #5175 with the corrected mapping as its own block, including illuminance, tests, and full device diagnostics. That should fix it all for this manufacturer ID; happy to adjust it based on any feedback here or there.

Issue #4125, opened 2025-06-13
[Device Support Request] Shelly 1 (PM) Gen 4 Specific Clusters


h1ghrise:
### Problem description

With Shelly now supporting Zigbee in their new Shelly 1 (PM) Gen4 devices, there were two specific clusters introduced for handling Manufacturer specific requests.

General Overview of Shelly Devices and their Zigbee capabilities:
- https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/DeviceFeatures

Clusters:
- Shelly RPC Cluster ([0xFC01](https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/RPCCluster))
- Shelly WiFi Setup Cluster ([0xFC02](https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/WiFiSetupCluster)) 

### Solution description

Be able to use those two clusters within the ZHA integration.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

currently, these 2 clusters are not populated:

![Image](https://github.com/user-attachments/assets/bb488dbd-ed6b-4c16-8385-dac73a6466b4)

</details>

### Device signature

<details><summary>Device signature</summary>

```json

{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 5264,
    "maximum_buffer_size": 108,
    "maximum_incoming_transfer_size": 1613,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 1613,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0702",
        "0x0b04"
      ],
      "output_clusters": []
    },
    "239": {
      "profile_id": "0xc001",
      "device_type": "0x2001",
      "input_clusters": [
        "0xfc01",
        "0xfc02"
      ],
      "output_clusters": []
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
[...]

PR #4990, opened 2026-05-11
Add Shelly Gen4 WiFi setup cluster

LaurynasGr:
## Proposed change

Add a custom Shelly WiFi setup quirk for Shelly 1PM and 2PM devices.

This adds support for the manufacturer-specific WiFi setup cluster on endpoint 239 and maps it to a typed custom cluster instead of leaving it as an unknown manufacturer-specific cluster. The quirk also uses a custom device class to handle Shelly responses sent on custom profile `0xC001`, which allows ZHA to decode WiFi cluster attribute reads and writes correctly.

## Additional information

Shelly sends WiFi setup cluster traffic on endpoint 239 using a custom profile (`0xC001`). Without the custom device handling, zigpy ignores the replies before they reach normal ZCL decoding.

The `action` attribute is defined as `uint8_t` because the device rejects `enum8` writes with `INVALID_DATA_TYPE`.

This work directly addresses the Shelly WiFi Setup Cluster part of open issue #4125, [Device Support Request] Shelly 1 (PM) Gen 4 Specific Clusters, which explicitly calls out support for the manufacturer-specific Shelly clusters `0xFC01` and `0xFC02`. This PR adds typed support for the WiFi setup cluster (`0xFC02`) for Shelly 1PM and 2PM devices and handles the custom profile reply path that previously prevented those responses from being decoded.

## Device diagnostics

[zha-01K7SVMTH5C2G4KYQP1C177HE0-Shelly 1PM-d96b2a1622eecd723911b40997b30cb9.json](https://github.com/user-attachments/files/27603279/zha-01K7SVMTH5C2G4KYQP1C177HE0-Shelly.1PM-d96b2a1622eecd723911b40997b30cb9.json)

## Checklist

### DIFF
--- a/tests/test_shelly.py
+++ b/tests/test_shelly.py
@@ -12,7 +12,6 @@ from zigpy.zcl.foundation import ZCLAttributeAccess
 from zhaquirks.shelly.wifi import (
     SHELLY_MANUFACTURER_CODE,
     SHELLY_WIFI_SETUP_CLUSTER_ID,
-    SHELLY_WIFI_SETUP_DEVICE_TYPE,
     SHELLY_WIFI_SETUP_ENDPOINT_ID,
     SHELLY_WIFI_SETUP_PROFILE_ID,
     ShellyCustomProfileDevice,
@@ -63,14 +62,6 @@ def test_shelly_wifi_setup_cluster_replaced(zigpy_device_from_v2_quirk, model) -
     ]
 
     assert isinstance(quirked, ShellyCustomProfileDevice)
-    assert (
-        quirked.endpoints[SHELLY_WIFI_SETUP_ENDPOINT_ID].profile_id
-        == SHELLY_WIFI_SETUP_PROFILE_ID
-    )
-    assert (
-        quirked.endpoints[SHELLY_WIFI_SETUP_ENDPOINT_ID].device_type
-        == SHELLY_WIFI_SETUP_DEVICE_TYPE
-    )
     assert isinstance(cluster, ShellyWiFiSetupCluster)
     assert cluster.ep_attribute == "shelly_wifi_setup"
     assert cluster.find_attribute("status") == cluster.AttributeDefs.status
--- a/tests/test_shelly.py
+++ b/tests/test_shelly.py
@@ -3,8 +3,10 @@
 import asyncio
 
 import pytest
+from zigpy.profiles import zha
 import zigpy.types as t
 from zigpy.zcl import ClusterType, foundation
+from zigpy.zcl.clusters.general import Basic
 from zigpy.zcl.foundation import ZCLAttributeAccess
 
 from zhaquirks.shelly.wifi import (
@@ -122,3 +124,42 @@ def test_shelly_wifi_custom_profile_packet_processed(
     asyncio.run(deliver_packet())
 
     assert cluster.get("status") == "got ip"
+
+
+@pytest.mark.parametrize("model", ["1PM", "2PM"])
+def test_shelly_wifi_standard_profile_packet_delegated(
+    zigpy_device_from_v2_quirk, model
+) -> None:
+    """Ensure standard ZHA profile packets fall through to normal zigpy parsing."""
+
+    quirked = zigpy_device_from_v2_quirk(
+        "Shelly",
+        model,
+        endpoint_ids=[1, SHELLY_WIFI_SETUP_ENDPOINT_ID],
+        cluster_ids={
+            SHELLY_WIFI_SETUP_ENDPOINT_ID: {
+                SHELLY_WIFI_SETUP_CLUSTER_ID: ClusterType.Server,
+            }
+        },
+    )
+
+    # Standard ZHA profile packet should delegate to super()
+    zha_packet = t.ZigbeePacket(
+        profile_id=zha.PROFILE_ID,
+        cluster_id=Basic.cluster_id,
+        src_ep=1,
+        dst_ep=1,
+        data=t.SerializableBytes(
+            _attribute_report_data(
+                Basic.AttributeDefs.model,
+                t.CharacterString("1PM"),
+            )
+        ),
+    )
+
+    hdr, rsp_key = quirked._parse_packet_header(zha_packet)
+
+    assert hdr is not None
+    assert rsp_key is not None
+    assert rsp_key.endpoint_id == 1
+    assert rsp_key.cluster_id == Basic.cluster_id
--- a/tests/test_shelly.py
+++ b/tests/test_shelly.py
@@ -9,6 +9,7 @@ from zigpy.zcl import ClusterType, foundation
 from zigpy.zcl.clusters.general import Basic
 from zigpy.zcl.foundation import ZCLAttributeAccess
 
+import zhaquirks
 from zhaquirks.shelly.wifi import (
     SHELLY_MANUFACTURER_CODE,
     SHELLY_WIFI_SETUP_CLUSTER_ID,
@@ -17,6 +18,8 @@ from zhaquirks.shelly.wifi import (
     ShellyCustomProfileDevice,
     ShellyWiFiSetupCluster,
 )
+
+zhaquirks.setup()
 
 
 def _attribute_report_data(
--- a/tests/test_shelly.py
+++ b/tests/test_shelly.py
@@ -1,0 +1,124 @@
+"""Tests for Shelly quirks."""
+
+import asyncio
+
+import pytest
+import zigpy.types as t
+from zigpy.zcl import ClusterType, foundation
+from zigpy.zcl.foundation import ZCLAttributeAccess
+
+from zhaquirks.shelly.wifi import (
+    SHELLY_MANUFACTURER_CODE,
+    SHELLY_WIFI_SETUP_CLUSTER_ID,
+    SHELLY_WIFI_SETUP_DEVICE_TYPE,
+    SHELLY_WIFI_SETUP_ENDPOINT_ID,
+    SHELLY_WIFI_SETUP_PROFILE_ID,
+    ShellyCustomProfileDevice,
+    ShellyWiFiSetupCluster,
+)
+
+
+def _attribute_report_data(
+    attribute: foundation.ZCLAttributeDef, value: object
+) -> bytes:
+    """Build a typed ZCL attribute report payload for Shelly WiFi attributes."""
+
+    header = foundation.ZCLHeader.general(
+        tsn=1,
+        command_id=foundation.GeneralCommand.Report_Attributes,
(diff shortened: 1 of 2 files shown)


zhaquirks/shelly/wifi.py
```diff
@@ -0,0 +1,118 @@
...
+
+class ShellyCustomProfileDevice(CustomDeviceV2):
+    """Handle Shelly responses sent on their custom endpoint profile."""
+
+    def custom_profile_packet_received(self, packet: t.ZigbeePacket) -> None:
+        """Treat Shelly's custom profile packets as standard ZCL packets."""
+        if packet.profile_id != SHELLY_WIFI_SETUP_PROFILE_ID:
+            return super().custom_profile_packet_received(packet)
+
+        self.packet_received(packet.replace(profile_id=zha.PROFILE_ID))
```
puddly: I don't think it's a good idea to mutate packets within a quirk. We should instead hook into the existing zigpy matching logic.

LaurynasGr: The packet itself was essentially fine, but `_parse_packet_header` was ignoring it due to an unrecognised `profile_id`.

We could override `_parse_packet_header` to handle header parsing for the `SHELLY_WIFI_SETUP_PROFILE_ID` without modifying the packet.
```python
def _parse_packet_header(
    self, packet: t.ZigbeePacket
) -> tuple[foundation.ZCLHeader, ResponseKey] | tuple[None, None]:
    """Parse Shelly custom-profile packets as ZCL for normal zigpy matching."""
    if packet.profile_id != SHELLY_WIFI_SETUP_PROFILE_ID:
        return super()._parse_packet_header(packet)

    hdr, _ = foundation.ZCLHeader.deserialize(packet.data.serialize())
    rsp_key = ResponseKey(
        endpoint_id=packet.src_ep,
        cluster_id=packet.cluster_id,
        direction=hdr.frame_control.direction,
        tsn=hdr.tsn,
    )
    return hdr, rsp_key
```
This also works and seems less "hacky" than modifying the packet and sending it back through the pipeline by retriggering `packet_received` inside `custom_profile_packet_received`

zhaquirks/shelly/wifi.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+(
+    QuirkBuilder("Shelly", "1PM")
+    .applies_to("Shelly", "2PM")
+    .device_class(ShellyCustomProfileDevice)
```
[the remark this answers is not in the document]
LaurynasGr: I only have access to Shelly `1PM` and `2PM` devices, so I'd be unable to test the quirk on `Mini1PM`/`Mini2PM`.

I'd suggest someone with those devices on hand raise a separate PR after this one with the additional `.applies_to` lines and the corresponding device diagnostics attached.

zhaquirks/shelly/wifi.py
LaurynasGr: Oh @puddly I see you already suggested basically the same approach in https://github.com/zigpy/zigpy/pull/1812#discussion_r3203500380. Should I update it to fully match that suggestion or the current implementation will do?

Hedda:
For reference as related to https://github.com/zigpy/zha-device-handlers/issues/4125 FYI, also see this PR from @dw86uk to expose Shelly RPC transport on endpoint in zigpy:

* https://github.com/zigpy/zigpy/pull/1812

  * https://github.com/zigpy/zigpy/issues/1810

LaurynasGr:
> For reference as related to #4125 FYI, also see this PR from @dw86uk to expose Shelly RPC transport on endpoint in zigpy:
> 
> * [Allow response matching on known custom-profile ZCL endpoints zigpy#1812](https://github.com/zigpy/zigpy/pull/1812)
>   
>   * [Shelly Gen4 RPC replies on profile 0xC001 are dropped because zigpy only parses ZCL packets for ZHA/ZLL profiles zigpy#1810](https://github.com/zigpy/zigpy/issues/1810)

Yep, saw it as well. If / when that gets merged, it would allow me to remove the `_parse_packet_header` override from this PR. I had to override it to parse the Shelly packets that use the custom `profile_id`. The remaining changes would stay the same.

Thanks @Hedda for the note.

zhaquirks/shelly/wifi.py
puddly: I think this is good. As long as we have a test we can catch anything breaking if we change zigpy around 😄. My only concern with modifying the packet itself is that it could lead to further issues down the line that would be _very_ hard to debug.

psym88:
Hi @LaurynasGr  https://github.com/zigpy/zha-device-handlers/issues/5076#issuecomment-4724213885

Yes, i`m successfully using your quirk with these

    .applies_to("Shelly", "Mini1PM")
    .applies_to("Shelly", "Mini1")
    .applies_to("Shelly", "EM Mini")

You can ignore Mini2PM, to my knowledge there is no such Shelly device.

Thanks for your work.

LaurynasGr:
> Hi @LaurynasGr [#5076 (comment)](https://github.com/zigpy/zha-device-handlers/issues/5076#issuecomment-4724213885)
> 
> Yes, i`m successfully using your quirk with these
> 
> ```
> .applies_to("Shelly", "Mini1PM")
> .applies_to("Shelly", "Mini1")
> .applies_to("Shelly", "EM Mini")
> ```
> 
> You can ignore Mini2PM, to my knowledge there is no such Shelly device.
> 
> Thanks for your work.

Great - thanks! Will add these to the list then so they work out of the box once this lands in the next HA release.

Will address everything after work tomorrow, and hopefully we can merge this.

LaurynasGr:
> **One note on `.replaces_endpoint(239, profile_id=0xC001, device_type=0x2001)`:** on real hardware this is effectively a no-op. When the quirk is applied, `CustomDeviceV2` builds its base by copying the real endpoint's `profile_id`/`device_type` (zigpy `quirks/v2/__init__.py` `_replacement_from_replaces`), so EP 239 already inherits `0xC001`/`0x2001` before this call runs. The override also keys on `packet.profile_id` from the wire, not the configured endpoint profile, so it doesn't depend on it either.
> 
> The only place this call actually matters is the test: `zigpy_device_from_v2_quirk` doesn't set a per-endpoint profile/device type on the synthesized device, so `replaces_endpoint` is what makes the `assert ...profile_id == SHELLY_WIFI_SETUP_PROFILE_ID` pass. That's a bit circular — the test asserts a value the quirk only sets to compensate for the fixture.
> 
> Rather than carry a production no-op for the test's sake, it'd be cleaner to extend the `zigpy_device_from_v2_quirk` fixture to let a test specify a real endpoint's `profile_id`/`device_type` (similar to how `firmware_version` was recently added). The test could then build a faithful pre-quirk device (EP 239 @ `0xC001`) and verify the quirk handles it, and the redundant `replaces_endpoint` could be dropped. Not a blocker — happy to see it land as-is if you'd prefer to keep the explicit call.

The two `profile_id` & `device_type` asserts in my test add no benefit and don't actually test anything about the quirk After dropping `.replaces_endpoint()`, those values come from the real device and zigpy's CustomDeviceV2 preserves them, so the quirk never touches them. Modifying `zigpy_device_from_v2_quirk` to accept `profile_id/device_type` just to assert back what we passed in is circular too. So I'll just remove the two useless asserts and drop the `.replaces_endpoint()` no-op.

LaurynasGr:
Comments addressed. The quirk tested again locally. Should be ready to merge

tests/test_shelly.py
```diff
@@ -0,0 +1,159 @@
...
+    assert cluster.ep_attribute == "shelly_wifi_setup"
+    assert cluster.find_attribute("status") == cluster.AttributeDefs.status
+    assert cluster.find_attribute("ssid") == cluster.AttributeDefs.ssid
+    assert cluster.AttributeDefs.status.manufacturer_code == SHELLY_MANUFACTURER_CODE
+    assert cluster.AttributeDefs.dhcp.access == ZCLAttributeAccess.Read
+    assert cluster.AttributeDefs.enable.access == (
+        ZCLAttributeAccess.Read | ZCLAttributeAccess.Write
+    )
+    assert cluster.AttributeDefs.action.access == ZCLAttributeAccess.Write
+    assert cluster.AttributeDefs.action.type is t.uint8_t
```
TheJulianJES, line 78: nit: We don't really need to test definitions applying correctly. When we add the device diagnostics to the ZHA repo's diagnostics database, we'd notice if there are any regressions in the future.

But it doesn't really hurt now, so it's also fine to keep.

tests/test_shelly.py
```diff
@@ -0,0 +1,159 @@
...
+    quirked = zigpy_device_from_v2_quirk(
+        "Shelly",
+        model,
+        endpoint_ids=[1, SHELLY_WIFI_SETUP_ENDPOINT_ID],
+        cluster_ids={
+            SHELLY_WIFI_SETUP_ENDPOINT_ID: {
+                SHELLY_WIFI_SETUP_CLUSTER_ID: ClusterType.Server,
+            }
+        },
+    )
```
TheJulianJES, line 96: note: I think it's good that we removed the `replaces_endpoint` call in the quirk that just set the Shelly profile again, but technically, this device now has the incorrect profile in the test (just the "ZHA" profile).

But as the packet explicitly contains the profile ID (and the quirk doesn't technically care about the one set on the profile), it's fine to keep it like this now. Still, we may want to change `zigpy_device_from_v2_quirk` to allow for different profile IDs in the future.

zhaquirks/shelly/wifi.py
```diff
@@ -0,0 +1,125 @@
...
+from zigpy.quirks import CustomCluster
+from zigpy.quirks.v2 import CustomDeviceV2, QuirkBuilder
+import zigpy.types as t
+from zigpy.zcl import foundation
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+
+SHELLY_MANUFACTURER_CODE = 0x1490
+SHELLY_WIFI_SETUP_ENDPOINT_ID = 239
+SHELLY_WIFI_SETUP_PROFILE_ID = 0xC001
+SHELLY_WIFI_SETUP_CLUSTER_ID = 0xFC02
```
TheJulianJES, line 15: nit/note: I think a constant for `SHELLY_WIFI_SETUP_ENDPOINT_ID` makes the `QuirkBuilder` part of quirk less readable standalone – you need to glance up here. We generally just inline these numbers/endpoints.

They're also used in the test, where it would be fine to duplicate them into IMO, but I guess this is fine as well. The same basically applies to the other constants. It's a bit of a style preference and how we handle it elsewhere in the project though.

TheJulianJES (approved):
Thanks!

PR #4992, opened 2026-05-11
Add Aqara Wall Outlet H2 UK (lumi.plug.aeu002) quirk

jasonmadigan:
## Proposed change

Adds a quirk for the Aqara Wall Outlet H2 UK (WP-P09D / `lumi.plug.aeu002`), a dual-socket UK wall outlet with per-socket and USB power monitoring.

All three endpoints emit `active_power` via standard ZCL `ElectricalMeasurement` 10x lower than the actual draw. The quirk multiplies the reported value by 10 inside `_update_attribute`. Scaling validated against a calibrated reference plug across the 3W to 2500W range.

## Additional information

This is the first Aqara plug quirk to use V2 `QuirkBuilder` and the first to use plain ZCL `ElectricalMeasurement` rather than the existing `zhaquirks.xiaomi.ElectricalMeasurementCluster` (a `LocalDataCluster`). The standard Aqara plug pattern (`ac_power_divisor: 10` on a `LocalDataCluster`) is not used here because:

- That pattern only nets x1 because power normally arrives pre-scaled x10 via the Lumi proprietary cluster or `AnalogInput`. This device emits power purely via standard ZCL on all three endpoints (no Lumi/AnalogInput hand-off), so the `LocalDataCluster` would block the only path that carries readings.
- Setting `ac_power_multiplier`/`ac_power_divisor` via `_CONSTANT_ATTRIBUTES` on a plain `CustomCluster, ElectricalMeasurement` was tested and produced no scaling change. The device appears to adjust its reporting scale based on what ZHA reads back from those attributes during cluster initialisation, creating a feedback loop. The `_update_attribute` override is the only approach that produced correct values.

The endpoint mapping is unusual: switch and power endpoints do not align.

| Endpoint | Switch | Power |
|-|-|-|
| 1 | Socket 1 | Total |
| 2 | Socket 2 | Socket 1 + USB |
| 3 | USB | Socket 2 |

The `_update_attribute` override pattern follows the precedent at `zhaquirks/sonoff/s60zbtpf.py`.

Hardware investigation and validation by absent42: https://github.com/absent42/Aqara-Wall-Outlet-H2-UK

## Checklist

- [x] Device diagnostics data has been attached [diagnostics-trimmed.json](https://github.com/user-attachments/files/27612361/diagnostics-trimmed.json)

### DIFF
--- a/tests/test_xiaomi_aqara_plug_aeu002.py
+++ b/tests/test_xiaomi_aqara_plug_aeu002.py
@@ -1,0 +1,45 @@
+"""Tests for the Aqara Wall Outlet H2 UK (lumi.plug.aeu002)."""
+
+from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
+
+import zhaquirks
+import zhaquirks.xiaomi.aqara.plug_aeu002  # noqa: F401
+
+zhaquirks.setup()
+
+POWER_ID = ElectricalMeasurement.AttributeDefs.active_power.id
+VOLTAGE_ID = ElectricalMeasurement.AttributeDefs.rms_voltage.id
+
+
+async def test_active_power_scaled_x10_on_all_endpoints(zigpy_device_from_v2_quirk):
+    """active_power must be multiplied by 10 on EP1, EP2 and EP3."""
+    device = zigpy_device_from_v2_quirk(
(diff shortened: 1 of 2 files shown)


jasonmadigan:
Updated

PR #4993, opened 2026-05-12
Add support for SONOFF SWV1C and SWV2C water valves

xiaoliu0001:
## Proposed change

Add ZHA quirks for SONOFF SWV single-channel and dual-channel Zigbee water valves.

This PR adds support for the following models:

Single-channel water valves:
- SONOFF SWV-ZFU
- SONOFF SWV-ZFE
- SONOFF SWV-ZNU
- SONOFF SWV-ZNE

Dual-channel water valves:
- SONOFF SWV-ZF2
- SONOFF SWV-ZF2U
- SONOFF SWV-ZF2E

The quirk adds SONOFF manufacturer-specific attributes for water valve status, water usage duration, and water usage volume, and exposes water leak, water depletion, duration, and volume entities where supported by the device.

## Additional information

No breaking changes are expected.

The new models share the SONOFF manufacturer-specific cluster used by the SWV water valve family. The single-channel and dual-channel variants are registered separately because the dual-channel devices expose channel-specific water usage duration on endpoint 1 and endpoint 2.

## Checklist

- [√] I have performed a self-review of my code
- [√] I have commented my code in hard-to-understand areas
- [√] I have verified my changes by testing with a device or have communicated a plan for testing

### DIFF
--- a/zhaquirks/sonoff/swv1c_swv2c.py
+++ b/zhaquirks/sonoff/swv1c_swv2c.py
@@ -2,31 +2,33 @@
 
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.v2 import (
-    QuirkBuilder, 
+    QuirkBuilder,
     ReportingConfig,
     SensorDeviceClass,  # Sensor device class
-    SensorStateClass    # Sensor state class (for line charts)
+    SensorStateClass,  # Sensor state class (for line charts)
 )
-from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
+
 # Import unit constants (duration/volume)
 from zigpy.quirks.v2.homeassistant import UnitOfTime, UnitOfVolume
+from zigpy.quirks.v2.homeassistant.binary_sensor import BinarySensorDeviceClass
 import zigpy.types as t
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 
 class ValveState(t.enum8):
     """Water valve state (8-bit value, bit-defined)."""
+
     # Basic states (single bit)
-    Normal = 0                    # 000 (no abnormal condition)
-    Water_Shortage = 1 << 0       # 001 (bit0: water shortage)
-    Water_Leakage = 1 << 1        # 010 (bit1: water leakage)
-    Anti_Frost_Alarm = 1 << 2     # 100 (bit2: anti-frost alarm)
+    Normal = 0  # 000 (no abnormal condition)
+    Water_Shortage = 1 << 0  # 001 (bit0: water shortage)
+    Water_Leakage = 1 << 1  # 010 (bit1: water leakage)
+    Anti_Frost_Alarm = 1 << 2  # 100 (bit2: anti-frost alarm)
     Water_Shortage_Channel_2 = 1 << 4  # bit4: channel 2 water shortage
     # Combined states (multiple bits triggered at the same time)
     Water_Shortage_And_Leakage = Water_Shortage | Water_Leakage  # 011
     Water_Shortage_And_Frost = Water_Shortage | Anti_Frost_Alarm  # 101
-    Water_Leakage_And_Frost = Water_Leakage | Anti_Frost_Alarm    # 110
-    All_Alarms = Water_Shortage | Water_Leakage | Anti_Frost_Alarm # 111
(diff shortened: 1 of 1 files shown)


Hedda:
@xiaoliu0001 Have you also seen these pull requests for Sonoff SNZB-SWVxC and SWV-ZFx water valves?

* https://github.com/zigpy/zha-device-handlers/pull/4927

and

* https://github.com/zigpy/zha-device-handlers/pull/5114

PS: There is by the way also a feature request discussion that indirectly relates to all Sonoff water valves:

gonz01d:
Tested on a SONOFF SWV‑ZF2U (Hydro Duo, dual-channel, sw_version 0x00001007) on Home Assistant Core 
 2026.5.4 — confirming this quirk works for the dual-channel variant. Thanks for the contribution 🙏

 A few notes from the install in case they're useful:

 1. Reconfigure step is essential for existing paired devices. Dropping the file into /config/zha_quirks/ +
  zha: custom_quirks_path: in configuration.yaml + an HA restart wasn't enough on its own — the new 
 attribute reporting bindings didn't get pushed to the sleepy device until I clicked Reconfigure in the ZHA
  device card. Once that ran, the reporting config (30/900/1) showed up correctly under the device's 
 CustomSonoffCluster (visible in "Manage Zigbee Device") and values started flowing. Worth a one-line note 
 in the PR description for users hitting the same.

 2. End-to-end works. Volume increments correctly (~2 L observed for a ~60 s CH2 run on a low-pressure 
 outdoor tap), both per-channel duration sensors populate, leak + depletion both initialise off.

 3. Duration sensor semantics worth documenting. The per-channel duration is "current / last session 
 minutes for that channel" and resets to 0 when the other channel runs. The state_class: measurement is the
  right choice for this — just worth a line in the README so users don't expect monotonic per-day totals 
 from the device (those have to be derived in HA via template/utility_meter helpers).

 4. Volume is device-wide (single sensor, no per-channel split). Fine in practice because the hardware only
  allows one channel open at a time — per-zone totals can be derived in HA from the switch state — but 
 worth flagging in docs to set expectations.

 5. Minor nits:
 - The water_depletion binary_sensor uses translation_key="Water depletion" (capital + space) — looks like 
 it should be water_depletion (snake_case) to match the other keys and avoid translation fallback to the 
 English string.
 - ValveState.Anti_Frost_Alarm (bit 2) is defined in the enum but no binary_sensor is exposed for it — 
 could be a useful winter sensor to add.

 Happy to test any iterations if it'd help.

t4ir1:
This quirk worked for me, I have access to the new sensors and they are working properly with the 1 channel hydro one valve.
I tried to turn on the valve but to shut the flow and I get the water depletion alarm, really good!
One thing that is missing, at least for my model, is that I can't see the cluster attribute to set an irrigation timer, the valve still closes automatically after 10 minutes :(
Is there a way you could add this attribute so that I could manage this timer?

starkillerOG:
I just tried: https://github.com/zigpy/zha-device-handlers/pull/5114 and it works greath for the SWV-ZFE and also allows control over the manual irrigation duration and volume. It looks like that quirk was made by a Sonoff employee.

FylBeall:
t4ir1 wrote:

> This quirk worked for me, I have access to the new sensors and they are working properly with the 1 channel hydro one valve. I tried to turn on the valve but to shut the flow and I get the water depletion alarm, really good! One thing that is missing, at least for my model, is that I can't see the cluster attribute to set an irrigation timer, the valve still closes automatically after 10 minutes :( Is there a way you could add this attribute so that I could manage this timer?

Have you found a solution for this 10 minute cutoff problem? I'm having the same trouble.

illgreen:
> t4ir1 wrote:
> 
> > This quirk worked for me, I have access to the new sensors and they are working properly with the 1 channel hydro one valve. I tried to turn on the valve but to shut the flow and I get the water depletion alarm, really good! One thing that is missing, at least for my model, is that I can't see the cluster attribute to set an irrigation timer, the valve still closes automatically after 10 minutes :( Is there a way you could add this attribute so that I could manage this timer?
> 
> Have you found a solution for this 10 minute cutoff problem? I'm having the same trouble.

I got this working on my SWV-ZF2 on firmware 0x00001009
I made a custom quirk based on the existing ZF2 support and the 0x501D implementation from #5114
It fixes the 10 minute cutoff and also exposes duration/capacity settings
[swv_zf2.py](https://github.com/user-attachments/files/32383843/swv_zf2.py)

Issue #3668, opened 2025-01-03
[BUG] Philips Hue Wall Switch and Dimmer Switch emit event 400 ms too late


carlolars:
### Bug description

Back in October, or was it November, I noticed that my Hue switches (wall and dimmer) started having a very noticeable delay when pressing a button until something actually happened (light turned on or off). I trigger on the `short_release` event which now is delayed alot.

For the Dimmer Switch there actually are two other events, `on` and `off_with_effect`, that are emitted immediately when pressing the button, but for the Wall Switch there are no such events so there is no workaround.

Looking at the logs there are two notifications received for a button, first one with `press_type=0` and then 100ms later one with `press_type=2`, and then 300 ms later (`ButtonPressQueue _ms_threshold`?) the event is emitted. This cause a significant and very noticeable delay.

I would suggest that always emit an event for the very first command that is received.

### Steps to reproduce

Just press a button on a Hue Wall or Dimmer switch and watch the logs or notice the delay if using it to turn on a light.

### Expected behavior

The light should toggle instantly when pushing a button on the switch :)

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Device signature

<details><summary>Device signature</summary>

```json
[Paste the device signature here]
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>

```json
[Paste the diagnostic information here]
```

</details>

### Logs

<details><summary>Logs</summary>
Hue Dimmer Switch:

```python
2025-01-03 23:28:22.015 DEBUG (MainThread) [zigpy.zcl] [0xCF49:2:0xfc00] Received ZCL frame: b'\x1d\x0b\x10\x9a\x00\x04\x00\x000\x00!\x00\x00'
2025-01-03 23:28:22.016 DEBUG (MainThread) [zigpy.zcl] [0xCF49:2:0xfc00] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl<0x1D>(frame_type=<FrameType.CLUSTER_COMMAND: 1>, is_manufacturer_specific=True, direction=<Direction.Server_to_Client: 1>, disable_default_response=1, reserved=0, *is_cluster=True, *is_general=False), manufacturer=4107, tsn=154, command_id=0, *direction=<Direction.Server_to_Client: 1>)
[...]

PR #5002, opened 2026-05-14
Emit Philips SHORT_PRESS immediately, decouple from multi-press queue

### LINKED ISSUE #5001: Hue Dimmer v2 (RWL022): ~1 second delay on every press, blocks responsive dimming
Pressing any of the four buttons on a Hue Dimmer Switch v2 (RWL022) takes close to a full second before anything reacts. With no immediate feedback, users often assume the first press didn't register. For smooth brightness control specifically, the lag makes it nearly impossible to implement a clean, reactive ramp: by the time the lamp starts responding, the button has already been released.

This is the same root cause as the closed/stale #3668, opening a fresh ticket because the responsiveness regression is still very much there.

The clearest path forward came from @TheJulianJES in the old thread:

> We should revert the click at the beginning to the previous behavior IMO. I think SHORT_PRESS and SHORT_RELEASE fire at the exact same time now, only SHORT_RELEASE should be delayed for the multi events. […] If you also use multi events, you just have to use SHORT_RELEASE instead of SHORT_PRESS then. PRs welcome, as always!

Tagging @TheJulianJES  and @fholzer for visibility since they were involved in #2340 and the original discussion.

Setup:
- Home Assistant Core 2026.5.0
- Device: Hue Dimmer Switch v2 (RWL022)

mguaylam:
## Proposed change

Emit \`SHORT_PRESS\` immediately when the Philips remote firmware sends \`press_type=0\`, instead of routing it through \`ButtonPressQueue\`'s 300ms threshold. \`SHORT_RELEASE\` continues to be queued so multi-press detection keeps working.

This restores instant button response on Philips remotes where the previous ~1s perceived delay made the device feel unresponsive — particularly bad for continuous-dimming use cases.

The direction was proposed by @TheJulianJES in #3668:

> We should revert the click at the beginning to the previous behavior IMO. I think SHORT_PRESS and SHORT_RELEASE fire at the exact same time now, only SHORT_RELEASE should be delayed for the multi events. […] If you also use multi events, you just have to use SHORT_RELEASE instead of SHORT_PRESS then.

## Additional information

Fixes #5001.

**Behavior change**: during multi-presses (fast double/triple/etc.), \`SHORT_PRESS\` is now emitted on each individual press in addition to the multi-press event. Automations relying on "exactly one SHORT_PRESS per sequence" should listen to \`SHORT_RELEASE\` instead.

Tested on Hue Dimmer Switch v2 (RWL022) hardware. The difference is dramatic: the previous ~1s perceived lag is gone — button response is now genuinely instant. Multi-press detection (double / triple / quadruple / quintuple) verified to still fire correctly alongside the per-press SHORT_PRESS events.

## Device diagnostics

N/A — modifies behavior of an existing quirk; no new device added.

## Checklist

- [x] \`pre-commit\` checks pass / the code has been formatted using Black


mguaylam:
Here's a demo : 
image

mguaylam:
I’m guessing it’s a mild breaking change?

PR #5031, opened 2026-05-25
Fix Schneider Electric thermostat sensor selection

DamKast:
## Proposed change

The schneider electric thermostat (EKO0725x and EKO30465) has some manufacturer specific cluster attributes that allows selecting between internal or external (air, floor) sensors. The current attribute defined in the quirk is  using the wrong type, which results in failure when trying to change the source sensor.

## Additional information

According to specs found in https://github.com/zigpy/zha-device-handlers/pull/3473, the `LocalTemperatureSourceSelect` is using the uint8 type and not enum8 which is used currently.
image

## Checklist


TheJulianJES (commented):
Thanks. We‘ll need to set the `zcl_type` on the attribute instead then. The enum is just the user-ish-facing part then. See this example: https://github.com/zigpy/zha-device-handlers/blob/1c7c9238486c5e994fce88f155ff45d0eccb62a3/zhaquirks/sonoff/mini_zbrbs.py#L42

PR #5067, opened 2026-06-08
Add overridable `reset_after` property to motion reset cluster

TheJulianJES:
## Proposed change

This changes two Aqara motion sensor quirks, which have a configurable number entity in ZHA for setting their motion timeout, to not depend on special ZHA logic to set the `reset_s` attribute in the cluster classes, so zha-quirks knows when to reset the timer.

Instead, `reset_after` is now a property that subclasses can override to derive the reset interval from a device attribute. Then, the attribute cache is only accessed when it's needed, not at startup, like with the ZHA approach. Otherwise, `reset_s` remains as a static attribute for all other quirks that use this, or as the default when nothing is in the attribute cache.

## Additional information

Implements the changes suggested in (and kinda requires the PR):
- https://github.com/zigpy/zha/pull/782#discussion_r3344900453

Couple of side notes:
- We should make sure that the attribute is read when pairing by ZHA (or automatically done when moved to quirks v2)
- We should likely also make sure that the static/default `reset_s` matches what the sensor has as the default (30s?).
- This should also be tested with a real device.
- We should really migrate the [last 7 remaining quirks](https://github.com/search?q=repo%3Azigpy%2Fzha-device-handlers%20%22%20attributes%20%3D%20%7B%22&type=code) that still use the old attribute definitions. (I thought we already did this? 😅)
- We should add proper test coverage for this "auto-reset motion cluster". I don't think we have any, as this was added years ago.

## Device diagnostics

TODO (maybe)

## Checklist

### DIFF
--- a/tests/test_xiaomi.py
+++ b/tests/test_xiaomi.py
@@ -1673,6 +1673,31 @@ async def test_xiaomi_p1_t1_motion_sensor(
 
 
 @pytest.mark.parametrize(
+    "quirk, default_reset_s",
+    (
+        (zhaquirks.xiaomi.aqara.motion_ac02.LumiMotionAC02, 30),
+        (zhaquirks.xiaomi.aqara.motion_agl04.LumiLumiMotionAgl04, 60),
+    ),
+)
+async def test_aqara_motion_reset_after_detection_interval(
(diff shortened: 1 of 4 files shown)


TheJulianJES:
@zigpy-review-bot Fix the conflict.

PR #5094, opened 2026-06-16
Add Tuya sensor quirk for _TZE284_hodyryli

cablespaghetti:
## Proposed change

Add support for "_TZE284_hodyryli", "TS0601" (and 8se38w3c which seems to have the exact same requirements) using inspiration from support for this device in https://github.com/0leg7/zigbee-herdsman-converters/blob/master/src/devices/tuya.ts.

## Additional information

Works on my home assistant install
image

## Checklist

### DIFF
--- a/tests/test_tuya_sensor.py
+++ b/tests/test_tuya_sensor.py
@@ -208,7 +208,9 @@ async def test_handle_get_data_probe_sensor(zigpy_device_from_v2_quirk, model, m
     assert ep.temperature.get("measured_value") == 2530  # 253 * 10
     assert ep.humidity.get("measured_value") == 7100  # 71 * 100
     assert ep.power.get("battery_percentage_remaining") == 100  # state 1 → 100
-    assert ep.tuya_manufacturer.get("temperature_probe") == 245  # raw value, ÷10 for display
+    assert (
+        ep.tuya_manufacturer.get("temperature_probe") == 245
+    )  # raw value, ÷10 for display
 
     # Battery state 0 maps to 0, not 50 (unlike _TZE200_upagmta9)
     bat_empty = b"\x09\xe0\x02\x0b\x33\x03\x02\x00\x04\x00\x00\x00\x00"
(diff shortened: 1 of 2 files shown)


cablespaghetti:
Duplicates #4520 

Hopefully addresses the issues in that PR but starts fresh...

Fixes #4489

lenucksi:
Since this is the freshest and mergeable attempt for this device, one important piece is still missing: the LCD **clock**.

The ZT08 display only accepts the MCU time sync if a `DP 17 = false` write follows ~500 ms later. The generic `NoManufTimeTuyaMCUCluster` alone is ignored by the firmware, so the clock stays at `00:xx`. This was pinned down on the zigbee2mqtt side (Koenkk/zigbee2mqtt#29627, Koenkk/zigbee-herdsman-converters#12180) and applies to the ZHA path the same way.

Working implementation, verified on a live device (clock stays correct across restarts and HA reboots):

```python
TUYA_SET_DATA = 0x00


class HodyryliTimeSyncCluster(NoManufTimeTuyaMCUCluster):
    """Time-sync cluster with DP17 LCD flush.

    After mcuSyncTime the firmware silently ignores the sync unless a dp17=false
    write follows ~500 ms later. Without it the LCD stays at 00:xx.
    See: Koenkk/zigbee2mqtt#29627
    """

    def handle_set_time_request(self, payload):
        result = super().handle_set_time_request(payload)
        self.create_catching_task(self._dp17_flush())
        return result

    async def _dp17_flush(self):
        await asyncio.sleep(0.5)
        dpd = TuyaDatapointData(17, TuyaData(False))
        cmd = TuyaCommand(
            status=0,
            tsn=self.endpoint.device.application.get_sequence(),
            datapoints=[dpd],
        )
        await TuyaNewManufCluster.command(self, TUYA_SET_DATA, cmd, expect_reply=False)
```

and then `.add_to_registry(replacement_cluster=HodyryliTimeSyncCluster)` instead of `NoManufTimeTuyaMCUCluster` (needs `import asyncio`, `TuyaCommand`, `TuyaData`, `TuyaDatapointData`, `TuyaNewManufCluster`).

FYI: #4520 has the `set_time_request` plumbing but not this flush either, so folding this in would make this PR the first one that actually drives the clock.

Two smaller notes from a device that has been running a local variant of this quirk for a while:

* the battery mapping `{0: 20, 1: 50, 2: 100}` (or `{0: 10, ...}`) is a judgement call — the hardware only reports three discrete states and no voltage, so please just pick one deliberately rather than leaving it undocumented;
* the internal humidity/battery also arrive on the standard `0x0405` / `0x0001` clusters, so keeping the DP mapping is fine but worth a comment to avoid confusion about which value wins.

Happy to open a small follow-up PR against this branch with the clock-sync part, or to leave the code here if you would rather fold it in yourself.

Issue #3972, opened 2025-03-21
[Device Support Request]  Zigbee WZ-SPI RGBW LED Strip SPI Controller


THK-artjom:
### Problem description

Device can be found here:
https://www.iskydance.com/index.php?c=product_show&a=index&id=2433

![Image](https://github.com/user-attachments/assets/c41cd7ef-6b08-4c61-ab7a-ebf94209cfd5)

### Solution description

Currently the device is not support. It would be great if someone could help me integrating it.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

![Image](https://github.com/user-attachments/assets/2c6bcb1f-b956-451e-9c75-ce1e2d7a557d)

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4417,
    "maximum_buffer_size": 66,
    "maximum_incoming_transfer_size": 66,
    "server_mask": 10752,
    "maximum_outgoing_transfer_size": 66,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZE204_8fffc3kb",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

### Diagnostic information

<details><summary>Diagnostic information</summary>
I found another request here at the git with some information:
https://github.com/Koenkk/zigbee2mqtt/issues/23991

Available DataPoints:
```json
{"1":"Switch Led",
"2":"Work Mode",
"3":"Bright Value",
"4":"Temp Value",
"5":"Colour Data",
"7":"Countdown",
"8":"Music Data",
"46":"Light Length",
"47":"Light Pixel",
"51":"Dreamlight Scene Mode",
"52":"Dreamlightmic Music Data",
[...]

Issue #3442, opened 2024-10-18
[Device Support Request] Skydance WZ-SPI RGBW LED Strip Controller


kim366:
### Problem description

The device connects to Home Assistant correctly. Firmware and diagnostic information is shown. It is, however, not possible to control the color/brightness/pattern of the LED strip.

Product Page: https://iskydance.com/index.php?c=product_show&a=index&id=2439

### Solution description

Control LED strips in the same way as other LED strip controllers already supported by ZHA.

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4417,
    "maximum_buffer_size": 66,
    "maximum_incoming_transfer_size": 66,
    "server_mask": 10752,
    "maximum_outgoing_transfer_size": 66,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZE204_8fffc3kb",
  "model": "TS0601",
  "class": "zigpy.device.Device"
}
```

</details>

dav972:
We can't know the dates for the updates? We can't know the dates for the updates?

diahellfire:
+1

gleb1983:
+1

wawan822:
+1

Ernesto1957:
+1

falconius:
+1

wawan822:
+1

djsamlb:
+1

KrolikROJER:
+1

KrolikROJER:
i can add support device, but settings work for:
* bright
* state: on/off
"that's all, "(c)

create `wz_spi.js` and put in external converters

```
const fz = require('zigbee-herdsman-converters/converters/fromZigbee');
[...]

PR #5095, opened 2026-06-17
Add Gledopto GL-SPI-206P (TS0601 _TZE284_gt5al3bl) RGB light quirk

### LINKED ISSUE #4610: [Device Support Request] _TZE284_gt5al3bl (Gledopto GL-SPI-206P)
### Problem description

Tuya TS0601-based LED strip controller, produced by [Gledopto](https://gledopto.com/h-pd-101.html).

### Solution description

Support for this device exists in [Zigbee2MQTT](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/gledopto.ts#L271), so I used this as a start to write a quirk.

However, I cannot get the color part to work, because I need to send a raw payload to DP 61. On/off and level control seem to work.

Quirk V2 lacks some examples on RGB/Lights, so I have no clue how to continue.

### Screenshots/Video

None

### Diagnostics information

None

### Device signature

```json
{
  "node_descriptor": {
    "logical_type": 1,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4417,
    "maximum_buffer_size": 66,
    "server_mask": 10752,
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0051",
      "input_clusters": [
        "0x0000",
        "0x0004",
        "0x0005",
        "0xed00",
        "0xef00"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZE284_gt5al3bl",
  "model": "TS0601",
}
```

</details>

### Logs

None

### Custom quirk

```python

"""Tuya SPI Pixel Controller GL-SPI-206P quirk."""

from zigpy.quirks.v2 import EntityType
import zigpy.types as t
from zigpy.zcl.clusters.general import LevelControl
from zigpy.zcl.clusters.lighting import Color

from zhaquirks.tuya import TuyaLocalCluster, NoManufacturerCluster, TuyaData
    TuyaLevelControl,
    TuyaDatapointData,
    TUYA_MCU_COMMAND,
)
from zhaquirks.tuya.builder import TuyaQuirkBuilder
from zigpy.profiles import zha
from zhaquirks.tuya.mcu import DPToAttributeMapping

from homeassistant.util.color import color_xy_to_hs
from typing import Any


class WorkMode(t.enum8):
    """Work mode enum."""

    White = 0x00
    Colour = 0x01
    Scene = 0x02
    Music = 0x03


class ChipType(t.enum8):
    """Chip type enum."""

    WS2801 = 0x00
    LPD6803 = 0x01
    LPD8803 = 0x02
    WS2811 = 0x03
    TM1814B = 0x04
    TM1934A = 0x05
    SK6812 = 0x06
    SK9822 = 0x07
    UCS8904B = 0x08
    WS2805 = 0x09


class TuyaColor(Color, TuyaLocalCluster):
    """Tuya MCU Level cluster for color device."""

    _CONSTANT_ATTRIBUTES = {
        Color.AttributeDefs.color_mode.id: Color.ColorMode.Hue_and_saturation
    }

    class AttributeDefs(Color.AttributeDefs):
        """Cluster attributes."""

    def _xy_to_hsv(self, x: int, y: int) -> tuple[int, int, int]:
        """Convert CIE xy to HSV for Tuya."""
        # Normalize x, y (ZCL sends as uint16 where 65535 = 1.0)
        x_norm = x / 65535
        y_norm = y / 65535

        # Use Home Assistant's color utility to convert xy to HS
        hue_deg, sat_percent = color_xy_to_hs(x_norm, y_norm)

        # Convert to Tuya format (0-360 for hue, 0-1000 for saturation)
        hue = max(0, min(360, round(hue_deg)))
        sat = max(0, min(1000, round(sat_percent * 10)))
        val = 1000  # Fixed at 1000 as per TypeScript implementation

        return (hue, sat, val)

    def _send_dp61_color(self, hue: int, sat: int, val: int = 1000):
        """Send DP 61 raw payload with HSV data."""
        # Build the 11-byte payload as per TypeScript implementation
        payload = bytes([
            0x00, 0x01, 0x01, 0x14, 0x00,  # Header
            (hue >> 8) & 0xFF, hue & 0xFF,  # Hue (16-bit big-endian)
            (sat >> 8) & 0xFF, sat & 0xFF,  # Saturation (16-bit)
            (val >> 8) & 0xFF, val & 0xFF,  # Value (16-bit)
        ])

        self.debug(
            "Sending DP 61 color: H=%d S=%d V=%d, payload=%s",
            hue, sat, val, payload.hex()
        )

        # Send via Tuya MCU command bus with the raw payload as attribute value
        cluster_data = TuyaDatapointData(
            dp=61,
            data=TuyaData(raw=payload)
        )
        self.endpoint.device.command_bus.listener_event(
            TUYA_MCU_COMMAND,
            cluster_data,
        )

    async def command(
        self,
        command_id: foundation.GeneralCommand | int | t.uint8_t,
        *args,
        manufacturer: int | t.uint16_t | None = None,
        expect_reply: bool = True,
        tsn: int | t.uint8_t | None = None,
        **kwargs: Any,
    ):
        """Override the default Cluster command."""
        self.debug(
            "Tuya Cluster Command: %x, args=%s, kwargs=%s",
            command_id,
            args,
            kwargs,
        )

        if command_id == Color.ServerCommandDefs.move_to_color.id:
            x = kwargs.get("color_x")
            y = kwargs.get("color_y")

            hue, sat, val = self._xy_to_hsv(x, y)
            self._send_dp61_color(hue, sat, val)
            return foundation.GENERAL_COMMANDS[
                foundation.GeneralCommand.Default_Response
            ].schema(command_id=command_id, status=foundation.Status.SUCCESS)


class TuyaColorNM(NoManufacturerCluster, TuyaColor):
    """Tuya Color cluster for SPI Pixel Controller."""


class TuyaLevelControlNM(NoManufacturerCluster, TuyaLevelControl):
    """Tuya OnOff cluster with NoManufacturerID."""


(
    TuyaQuirkBuilder("_TZE284_gt5al3bl", "TS0601")
    # DP 1: On/Off (works)
    .tuya_onoff(dp_id=1)

    # DP 2: Work mode (not tested)
    .tuya_enum(
        dp_id=2,
        attribute_name="work_mode",
        enum_class=WorkMode,
        translation_key="work_mode",
        fallback_name="Work mode",
    )

    # DP 3: Brightness/Level control (works)
    .tuya_dp(
        dp_id=3,
        ep_attribute=TuyaLevelControlNM.ep_attribute,
        attribute_name=LevelControl.AttributeDefs.current_level.name
    )
    .adds(TuyaLevelControlNM)


    # DP 61: Color control (does not work, needs raw payload support)
    .tuya_dp_multi(
        dp_id=61,
        attribute_mapping=[
            DPToAttributeMapping(
                ep_attribute=TuyaColorNM.ep_attribute,
                attribute_name=Color.AttributeDefs.current_hue.name
            ),
        ])
    .adds(TuyaColorNM)

    # DP 53: Light pixel number setting (works)
    .tuya_number(
        dp_id=53,
        attribute_name="lightpixel_number_set",
        type=t.uint16_t,
        min_value=1,
        max_value=1000,
        step=1,
        translation_key="lightpixel_number_set",
        fallback_name="Pixel count",
    )
    # DP 102: Chip type (works)
    .tuya_enum(
        dp_id=102,
        attribute_name="chip_type",
        enum_class=ChipType,
        translation_key="chip_type",
        fallback_name="Chip type",
    )
    .replaces_endpoint(1, device_type=zha.DeviceType.COLOR_DIMMABLE_LIGHT)
    .skip_configuration()
    .add_to_registry()
)
```

</details>

### Additional information

None

strepto42:
## Proposed change

Adds a quirk for the **Gledopto GL-SPI-206P** addressable (SPI) RGB(W) LED
controller — a Tuya-MCU (`0xEF00`) device that reports as model `TS0601` and
exposes all functionality through Tuya datapoints. Covers manufacturers
`_TZE284_gt5al3bl` and `_TZE204_8fffc3kb`.

`TuyaQuirkBuilder` has no light helper, so the quirk builds a full Home
Assistant light from local `OnOff`/`LevelControl`/`Color` clusters mapped onto
the datapoints, and adds the device's config and effect controls.

**Entities**

- **Light** — on/off (DP 1), brightness (DP 3), colour via the XY colour
  picker (DP 61), colour temperature (DP 4)
- **Config** — work mode (DP 2), countdown (DP 7), pixel count (DP 53), LED
  colour order (DP 101), LED chip type (DP 102), do-not-disturb (DP 103)
- **Effects** — scene select (16 presets, DP 51), music-reactive mode select
  (6 modes) + music sensitivity number (DP 52)

**Implementation notes**

- All datapoint encodings (HSV colour, scene/music raw payloads, scaling) were
  taken from `zigbee-herdsman-converters` `src/devices/gledopto.ts`.
- The `zha` light platform implements only XY and colour-temp colour modes (no
  Hue/Saturation), so the `Color` cluster advertises
  `XY_attributes | Color_temperature` and converts HA's CIE-xy into the
  device's native HSV datapoint.
- The MCU is slow to send its application-level response; with zigpy's
  per-device concurrency of 1, `expect_reply=True` writes stall the request
  queue, so all command paths and the effects cluster send datapoints
  fire-and-forget (`expect_reply=False`). Work mode (DP 2) is only re-sent when
  it changes.
- The device does not report colour/level state back, so light state is
  command-optimistic.

## Additional information

Fixes #4610.

The light (on/off, brightness, XY colour, colour temperature) has been verified
on hardware, as have scenes and music.

## Device diagnostics

[zha-01KG1S34902H50EV2T4W19WJ5Z-_TZE284_gt5al3bl TS0601-dc3b6a1264bd9dbd2e24302390b9f5b0.json](https://github.com/user-attachments/files/29069213/zha-01KG1S34902H50EV2T4W19WJ5Z-_TZE284_gt5al3bl.TS0601-dc3b6a1264bd9dbd2e24302390b9f5b0.json)

## Checklist

Related: #3442, #3972

### DIFF
--- a/tests/test_tuya_light.py
+++ b/tests/test_tuya_light.py
@@ -1,0 +1,313 @@
+"""Tests for the Tuya TS0601 SPI/addressable LED quirk (Gledopto GL-SPI-206P)."""
+
+from unittest import mock
+
+from zigpy.quirks.v2 import CustomDeviceV2
+from zigpy.zcl import foundation
+from zigpy.zcl.clusters.general import LevelControl, OnOff
+from zigpy.zcl.clusters.lighting import Color
+
+from tests.common import ClusterListener, wait_for_zigpy_tasks
+import zhaquirks
+from zhaquirks.tuya import TuyaCommand, TuyaData, TuyaDatapointData
+from zhaquirks.tuya.mcu import TuyaMCUCluster
+from zhaquirks.tuya.ts0601_light import (
+    TuyaColorControl,
+    TuyaSpiLevelControl,
+    TuyaSpiOnOff,
+    TuyaWorkMode,
+    decode_hue,
+    decode_saturation,
+    dp_to_level,
+    dp_to_mireds,
+    encode_color,
+    level_to_dp,
+    mireds_to_dp,
+    xy_to_hs,
+)
+
+zhaquirks.setup()
+
+MANUFACTURER = "_TZE284_gt5al3bl"
+MODEL = "TS0601"
+
+
+def test_encode_color_exact_bytes():
+    """The DP 61 payload must match the Z2M byte layout exactly."""
+    # pure red: hue 0, full saturation -> h=0, s=1000, v=1000
+    assert encode_color(0, 254) == bytes.fromhex("0001011400000003e803e8")
+    # max hue (254 -> 360 deg = 0x0168), full saturation
+    assert encode_color(254, 254) == bytes.fromhex("0001011400016803e803e8")
+    # zero saturation (h=0, s=0, v=1000)
+    assert encode_color(0, 0) == bytes.fromhex("00010114000000000003e8")
+
+
+def test_color_roundtrip():
+    """Encoding then decoding hue/saturation returns the originals."""
+    for hue, sat in ((0, 0), (127, 200), (254, 254), (60, 90)):
+        raw = encode_color(hue, sat)
+        assert decode_hue(raw) == hue
+        assert decode_saturation(raw) == sat
+
+
+def test_decode_short_payload_is_safe():
+    """Malformed/short payloads decode to 0 rather than raising."""
+    assert decode_hue(b"\x00\x01") == 0
+    assert decode_saturation(b"") == 0
+
+
+def test_level_scaling():
+    """ZCL level 0-254 maps to Tuya brightness 10-1000."""
+    assert level_to_dp(0) == 10
+    assert level_to_dp(254) == 1000
+    assert dp_to_level(1000) == 254
+    assert dp_to_level(10) == 1
+
+
+def test_color_temp_scaling():
+    """Tuya color temp 0=warm/1000=cold maps to ZCL mireds (inverse)."""
+    assert mireds_to_dp(500) == 0
+    assert mireds_to_dp(153) == 1000
+    assert dp_to_mireds(0) == 500
+    assert dp_to_mireds(1000) == 153
+
+
+async def test_gl_spi_206p_clusters(zigpy_device_from_v2_quirk):
+    """The quirk builds a color light with the expected clusters."""
+    quirked = zigpy_device_from_v2_quirk(MANUFACTURER, MODEL)
+    assert isinstance(quirked, CustomDeviceV2)
+
+    ep = quirked.endpoints[1]
+    assert isinstance(ep.on_off, OnOff)
+    assert isinstance(ep.level, LevelControl)
+    assert isinstance(ep.light_color, TuyaColorControl)
+    assert isinstance(ep.tuya_manufacturer, TuyaMCUCluster)
+
+    # zha only supports XY + color temp, so advertise those capabilities
+    caps = ep.light_color.get(Color.AttributeDefs.color_capabilities.name)
+    assert caps == (
+        Color.ColorCapabilities.XY_attributes
+        | Color.ColorCapabilities.Color_temperature
+    )
+
+
+def test_xy_to_hs_known_colors():
+    """CIE xy converts to sensible ZCL hue (0-254) for primary colors."""
+    red_hue, _ = xy_to_hs(int(0.69 * 65535), int(0.31 * 65535))
+    assert red_hue <= 5 or red_hue >= 250  # hue wraps around 0
+    green_hue, _ = xy_to_hs(int(0.17 * 65535), int(0.70 * 65535))
+    assert 78 <= green_hue <= 95  # ~120 deg
+    blue_hue, _ = xy_to_hs(int(0.15 * 65535), int(0.06 * 65535))
+    assert 160 <= blue_hue <= 185  # ~240-250 deg
+
+
+async def test_move_to_color_sends_dp61(zigpy_device_from_v2_quirk):
+    """An XY color command (as zha sends it) sets colour mode + DP 61."""
+    quirked = zigpy_device_from_v2_quirk(MANUFACTURER, MODEL)
+    ep = quirked.endpoints[1]
+    color = ep.light_color
+    tuya = ep.tuya_manufacturer
+
+    color_x, color_y = int(0.7 * 65535), int(0.3 * 65535)
+    exp_hue, exp_sat = xy_to_hs(color_x, color_y)
+
+    with mock.patch.object(
+        tuya.endpoint, "request", return_value=foundation.Status.SUCCESS
+    ) as req_mock:
+        # zha invokes the named command with keyword args
+        await color.command(
+            Color.ServerCommandDefs.move_to_color.id,
+            color_x=color_x,
+            color_y=color_y,
+            transition_time=0,
+        )
+        await wait_for_zigpy_tasks()
+
+    sent = b"".join(
+        (call.kwargs.get("data") or b"") for call in req_mock.call_args_list
+    )
+
+    # DP 2 (work_mode) set to colour: dp=0x02, enum type=0x04, len=1, value=1
+    assert b"\x02\x04\x00\x01\x01" in sent
+    # DP 61 raw payload: dp=0x3d, raw type=0x00, len=0x0b, then 11 HSV bytes
+    assert b"\x3d\x00\x00\x0b" + encode_color(exp_hue, exp_sat) in sent
+
+    # optimistic state: zha reads current_x/current_y
+    assert color.get(Color.AttributeDefs.current_x.name) == color_x
+    assert color.get(Color.AttributeDefs.current_y.name) == color_y
+    assert color.get(Color.AttributeDefs.color_mode.name) == Color.ColorMode.X_and_Y
+
+    # datapoints are sent fire-and-forget so streamed updates stay responsive
+    assert req_mock.call_args_list
+    assert all(
+        call.kwargs.get("expect_reply") is False for call in req_mock.call_args_list
+    )
+
+
+async def test_work_mode_not_resent_when_unchanged(zigpy_device_from_v2_quirk):
+    """work_mode (DP 2) is only sent when the mode changes, not every update.
+
+    Resending it on every streamed colour update floods the MCU and causes the
+    dropped/laggy commands users see when dragging the colour wheel.
+    """
+    quirked = zigpy_device_from_v2_quirk(MANUFACTURER, MODEL)
+    ep = quirked.endpoints[1]
+    color = ep.light_color
+    tuya = ep.tuya_manufacturer
+
+    work_mode_colour = b"\x02\x04\x00\x01\x01"  # DP 2 enum -> colour
+
+    # first colour command: mode changes (None -> colour), so DP 2 is sent
+    with mock.patch.object(
+        tuya.endpoint, "request", return_value=foundation.Status.SUCCESS
+    ) as req1:
+        await color.command(
+            Color.ServerCommandDefs.move_to_color.id,
+            color_x=30000,
+            color_y=30000,
+            transition_time=0,
+        )
+        await wait_for_zigpy_tasks()
+    sent1 = b"".join((c.kwargs.get("data") or b"") for c in req1.call_args_list)
+    assert work_mode_colour in sent1
+
+    # second colour command: already in colour mode, so DP 2 is NOT resent
+    with mock.patch.object(
+        tuya.endpoint, "request", return_value=foundation.Status.SUCCESS
+    ) as req2:
+        await color.command(
+            Color.ServerCommandDefs.move_to_color.id,
+            color_x=40000,
+            color_y=20000,
+            transition_time=0,
+        )
+        await wait_for_zigpy_tasks()
+    sent2 = b"".join((c.kwargs.get("data") or b"") for c in req2.call_args_list)
+    assert work_mode_colour not in sent2
+    # the colour itself (DP 61) is still sent every time
+    assert b"\x3d\x00\x00\x0b" in sent2
+
+
+async def test_move_to_color_temp_sends_dp4(zigpy_device_from_v2_quirk):
+    """A color-temp command sets white mode (DP 2) and color temp (DP 4)."""
+    quirked = zigpy_device_from_v2_quirk(MANUFACTURER, MODEL)
+    ep = quirked.endpoints[1]
+    color = ep.light_color
+    tuya = ep.tuya_manufacturer
+
+    with mock.patch.object(
+        tuya.endpoint, "request", return_value=foundation.Status.SUCCESS
+    ) as req_mock:
+        # zha invokes move_to_color_temp with keyword args
+        await color.command(
+            Color.ServerCommandDefs.move_to_color_temp.id,
+            color_temp_mireds=326,
+            transition_time=0,
+        )
+        await wait_for_zigpy_tasks()
+
+    sent = b"".join(
+        (call.kwargs.get("data") or b"") for call in req_mock.call_args_list
+    )
+
+    # DP 2 (work_mode) set to white (value 0)
+    assert b"\x02\x04\x00\x01\x00" in sent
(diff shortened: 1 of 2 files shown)


Deniz-KK:
Thanks for this — I have a `_TZE204_8fffc3kb` (Skydance WZ-SPI) and independently ran into exactly the wall you describe. I had not tested your quirk yet, so this is not a confirmation that it works on my unit, but two things from my findings may be useful.

**Your fire-and-forget note is very likely the key.** I built a colour cluster on DP 61 and could see, in the zigpy debug log, a correctly encoded frame going out and the device answering with `set_data_response`, status 0, no error — and the strip simply not changing colour. Payload content was byte-identical to what the RF remote produces for the same colour. I had `expect_reply=True` throughout. Your explanation of the MCU stalling zigpy's request queue fits that symptom exactly, and it is not something the log makes obvious.

**Two datapoint differences on my unit**, in case there are hardware variants:

- **DP 101 / 102 are swapped.** On mine, DP 101 is the chip type (factory default 1 = TM1809) and DP 102 is the LED colour order (mine is BRG, index 4 in the manual's list RGB, RBG, GRB, GBR, BRG, BGR). Your quirk maps 101 as colour order and 102 as chip type.
- **DP 61 payload header.** Yours is `00 01 01 14 00` with an 11-byte payload. Mine sends and receives `00 01 00 14 00` + hue(2) + sat(2) + val(2) + `00`, i.e. 12 bytes — third byte `00` instead of `01`, plus a trailing `00`. Verified in both directions: green (120°) is `0001001400007803e803e800`. Hue is in degrees 0–360, saturation and value in 0–1000.

One more observation: my unit **never reports DP 2 (work mode) back**, not even after a Tuya data query spell, so its enum values cannot be confirmed from the device side. Issue #3972 lists `colour = 0x00, white = 0x01` for the same manufacturer ID, which is the opposite of the mapping I had assumed.

Happy to test your quirk on `_TZE204_8fffc3kb` and report back if that would help move this along.

Deniz-KK:
Testing this on _TZE204_8fffc3kb (Skydance WZ-SPI, 280 px WS2811, fw 0x0000004a), which this quirk also claims. Four findings:

1. expect_reply=False alone is sufficient — the work-mode write isn't part of the fix. Worth stating explicitly since the two changes arrived together. With the controller already in colour mode, _ensure_work_mode short-circuits and only DP 61 goes out (tsn=89 dp=1, tsn=90 dp=61, no DP 2) — and the strip still follows. Before the change, byte-identical frames were acknowledged with set_data_response status 0 and had no effect.

2. Incoming DP 61 is written to attributes ZHA never reads. test_incoming_color_report_updates_hue_sat asserts current_hue/current_saturation, but the cluster advertises XY_attributes | Color_temperature and the command path sets color_mode = X_and_Y. In that mode HA derives colour from current_x/current_y, so the report never reaches the entity. Not hypothetical here: the RF wall panel does send DP 61 back on the _TZE204, so "the device does not report colour state" holds for the _TZE284 but not for both. Writing current_x, current_y and color_mode alongside hue/sat restores it.

3. The DP 61 frame length differs between the variants — decoding survives, encoding doesn't. Yours is an 11-byte frame (header + hue/sat/value) with header 00 01 01 14 00. Mine is 12 bytes with header 00 01 00 14 00 and a trailing byte. Since both headers are 5 bytes, decode_hue/decode_saturation read the right offsets for either; encode_color produces a frame my controller ignores.

4. DP 101/102 are swapped. On _TZE204_8fffc3kb, DP 101 is the chip type and DP 102 the colour order — the reverse of the mapping here.

Happy to attach diagnostics and debug logs.

strepto42:
Thanks — this is exactly the data I was missing, and the retraction in #3442 is appreciated.

Confirmed on my side, against your green frame `0001001400007803e803e800`:

- **Decode is variant-safe, encode is not.** Both headers are 5 bytes, so `decode_hue`/`decode_saturation` read the right offsets and my `len < 11` guard passes your 12-byte frame — it decodes to hue 120°, sat 1000 correctly. `encode_color` emits `0001011400007803e803e8` (11 bytes, third byte `01`), which is the frame your unit ignores.
- **Your four-attribute point is right and I'm adopting it.** The cluster advertises `XY_attributes` and the command path sets `color_mode = X_and_Y`, so HA reads `current_x`/`current_y` and the hue/sat writes go nowhere. I'll maintain `current_x`/`current_y` and `color_mode` alongside. My module docstring asserts the device doesn't report colour state — that's from the herdsman converter having no `fromZigbee` colour parser, not from device data, and your RF-remote reports disprove it for the `_TZE204`. It'll be corrected.
- **`expect_reply=False` alone being sufficient** — noted, I'll stop implying the DP 2 write is part of the fix. The two changes did land together.
- **Root cause of the variant errors:** I took the DP map, the 11-byte header and the DP 101/102 assignment from herdsman `gledopto.ts`, which fingerprints `_TZE204_8fffc3kb` and `_TZE284_gt5al3bl` under one definition. That's an assumption I inherited without device data, and `.applies_to("_TZE204_8fffc3kb")` shouldn't stand unverified. Likely outcome is a separate registration for the `_TZE204` (own encoder, DP 101/102 mapping and chip enum) rather than one shared quirk.

One thing I can't resolve from the comments. Your finding 1 references `_ensure_work_mode` by name and says the strip follows, which reads as my quirk working on your unit; finding 3 says `encode_color` produces a frame your controller ignores. Did you run the quirk unmodified, or with your 12-byte header patched into `encode_color`? If the 11-byte frame does work on the `_TZE204`, the split is unnecessary — so this decides the design.

If the offer still stands, what would help most:

1. **Device diagnostics JSON** for the `_TZE204_8fffc3kb`.
2. **Debug log of an inbound DP 61** triggered from the RF remote (`zigpy: debug`), so I can pin the exact report frame rather than infer it.
3. **A direct A/B**: unmodified `encode_color` (11-byte) vs your 12-byte frame on the same colour — confirms whether the header byte or the trailing byte is what the MCU rejects.
4. **DP 101 chip-type enum and DP 102 colour-order list** as your manual gives them. Mine has no `TM1809` at all, so the enum is variant-specific too, not just the DP number.
5. **work_mode (DP 2):** have you ever seen it reported, in any state? Mine follows herdsman (`white=0, colour=1`); #3972's quirk for the same WZ-SPI has them inverted (`colour=0, white=1`). Since DP 2 isn't needed for colour, I'd rather mark it unverified than guess.

Happy to put a branch up with the four-attribute mapping and a `_TZE204` variant for you to test once the above narrows it down.

Deniz-KK:
First things first: it works. Full colour control over ZHA on _TZE204_8fffc3kb, plus RF-remote colour changes propagating back into Home Assistant. The thing that unlocked it was your expect_reply=False finding — nothing else. I had a byte-correct frame going out and set_data_response coming back with status 0, and no idea why the strip ignored it. That one sentence in your PR description ended a hunt that had already cost me two evenings. Thank you.

Now to your question, and the answer is uncomfortable: I never ran your quirk. What I tested is my own implementation, written before I found #5095, using the 12-byte frame I had captured from my device. From your PR I took only the expect_reply insight. _ensure_work_mode appeared in my notes because I had read your code, not run it. That was misleading, and I'm sorry — it sent you looking for a contradiction that isn't there.

Finding 3 is therefore withdrawn. I never A/B tested the 11-byte frame; I compared frames and inferred a difference mattered. Please don't split the quirk on the strength of that claim.

A third source, which I only found today, makes that withdrawal firmer and settles the DP question independently: github.com/JirkaDeLucina/zigbee2mqtt-wz-spi, a standalone Z2M external converter written specifically for this fingerprint, not derived from the shared gledopto.ts definition.

It has 101 = chip_type and 102 = light_bead_sequence, matching my device and contradicting the definition you inherited. Two independent sources for that swap, so DP 101/102 stands regardless of how the frame question resolves.

It also sends the 11-byte header. Written for _TZE204_8fffc3kb, evidently in use — so the MCU most likely tolerates both frame lengths and my finding 3 was noise.

And it builds DP 51 constructively instead of shipping fixed blobs: [0x01, sceneID, effectByte, speed, speed, flagsByte, 0x00, 0x00, brightnessPercent] followed by hue/saturation triplets. I checked that layout against the 16 fixed payloads in your PR and they resolve completely — remainder always divisible by 3, all hues ≤ 360, all saturations ≤ 100, and the colours match the names (Ice Land Blue = 193/180/181/196°, Late Autumn = 25/34/44/20/12°). With that, all 44 factory scenes plus speed and brightness are reachable rather than 16 opaque presets. I've implemented scenes on that basis and they work. Two bytes I could not pin down: the effect byte (only 1 and 10 observed, the manual claims 16 variations) and the flags byte (0xE0 and 0x60 observed, meaning unclear) — I left both fixed rather than guess. Licence there is CC BY-NC 4.0, so the format is usable as knowledge, the code isn't.

To the points I can answer from measurement:

DP 61 outbound, 12 bytes on my unit: 00 01 00 14 00 + hue(2) + sat(2) + val(2) + 00, hue in degrees 0–360, saturation and value 0–1000.

DP 61 inbound arrives spontaneously when the colour is changed on the RF remote, roughly every 1–2 seconds while dragging. That is device data, captured with zigpy.zcl: debug — it disproves the "device doesn't report colour state" assumption for the _TZE204, as you suspected.

DP 101 / 102 from my manual. Chip types: TM1803, TM1809 (factory default here), TM1829, TLS3001, GW6205, MBI6120, TM1814B(RGBW), SK6812(RGBW), UCS8904B(RGBW), LPD6803, LPD8803, WS2801, P9813, SK9822, TM1914A, GS8206, UCS2904, SM16804, SM16825, SM16714(RGBW), UCS5603, UCS2603, SM16714D, UCS7604(RGBW), UCS7804(RGBW). Colour order: RGB, RBG, GRB, GBR, BRG, BGR, then the RGBW variants (RGBW, RBGW, GRBW, GBRW, BRGW, BGRW, WRGB, WRBG, WGRB, WGBR, WBRG, WBGR); mine is BRG. So the enum is variant-specific too, not just the DP number.

DP 2: never reported, in any state. Not on a Tuya data query spell, not after reconfigure, not in any of the four modes, not after a power cycle. Writes appear to be accepted — behaviour changes — but nothing comes back, so the enum cannot be confirmed from the device side. Marking it unverified is the right call.

Your four-attribute conclusion matches what I ended up doing, and it's the one addition I'd push for before merge: mapping DP 61 to four attributes and maintaining current_x/current_y is what makes the RF remote drive HA state. Without it the remote is invisible to Home Assistant, which is the whole point for anyone syncing the strip to other lights.

Thanks again — from me and from Claude, which did the log archaeology on this and, in hindsight, should have searched for the manufacturer ID on day one rather than reconstructing frames from debug output for two days. Your PR is the reason this device is usable at all.

strepto42:
All good, thanks for the update! It took me several tries over several months to get this far with it; AI assistance drastically improved over that time, as did my skill with guiding it, and learning lessons from previously failed attempts to get a satisfactory implementation going. The ultimate moral of my story is not to buy the same type of crappy quirky controller again, but in the meantime, it's nice if this can help others! Yay open source!

Deniz-KK:
Yeah, its really incredible. Im not a dev and im not a LED or HA professional. Im just a IT guy who is interested in smart home and bought a zigbee led controller from ali express and asked claude for help 😜
But your work was the key to success, thanks again!

Deniz-KK:
btw. did you know that you can install github on your HA - after that connect the repository with claude code... thats so nice!

strepto42:
I did not! That is quite handy, easier than going through custom quirks and HACS! Thanks!

PR #5099, opened 2026-06-18
Fix `TuyaQuirkBuilder` parameter type annotations

TheJulianJES:
## Summary

Follow-up to the pre-commit hook autoupdate (#3849), where a Copilot review noticed that `tuya_formaldehyde`'s `converter` parameter is annotated `float` despite taking a callable. Investigating that surfaced a whole cluster of the same class of bug in `TuyaQuirkBuilder`: methods annotate a cluster-**class** parameter as the **instance** type while passing/using the class.

This corrects all of them in `zhaquirks/tuya/builder/__init__.py` (17 annotations):

- 12 `*_cfg: TuyaLocalCluster = <ClusterClass>` → `type[TuyaLocalCluster]` (illuminance, co2, ec, form, pm25, metering, onoff, cover, rh, soil, temp, voc) — used as classes (`cfg.ep_attribute`, passed to `.adds()`).
- `replacement_cluster: TuyaMCUCluster` → `type[TuyaMCUCluster]`.
- `power_cfg` in `_tuya_battery` and `tuya_battery` → `type[PowerConfiguration]` (`| None` where defaulted).
- `ias_cfg: TuyaLocalCluster` in `tuya_ias` → `type[TuyaLocalCluster]`.
- `converter: float` → `Callable[[Any], Any] | None`, matching the other converter methods in the file.

**Annotation-only — no runtime behavior change.**

## Why these weren't caught

These are all suppressed by the global `disable_error_code` list in `pyproject.toml`, so mypy never reports them: the default-value mismatches surface under `assignment` and the call-site mismatches under `arg-type` — both disabled.

Verified by re-enabling each code:
- `--enable-error-code assignment`: the 14 default-value lines clear.
- `--enable-error-code arg-type`: the 4 `ias_cfg`/`power_cfg` call-site errors clear.

Two unrelated pre-existing issues are deliberately left out (they need more than an annotation tweak): the `inspect.currentframe()` `FrameType | None` locals (211-212) and `battery_type` (238).

## Test plan

- `mypy 2.1.0` (default repo config) — clean.
- `mypy 2.1.0 --enable-error-code assignment` / `--enable-error-code arg-type` — target lines no longer error.
- `ruff check` / `ruff format` — clean.
- `pytest -k tuya` — 1160 passed.

### DIFF
--- a/zhaquirks/tuya/builder/__init__.py
+++ b/zhaquirks/tuya/builder/__init__.py
@@ -216,7 +216,7 @@ class TuyaQuirkBuilder(QuirkBuilder):
     def _tuya_battery(
         self,
         dp_id: int,
-        power_cfg: PowerConfiguration,
+        power_cfg: type[PowerConfiguration],
         scale: float,
         endpoint_id: int = 1,
     ) -> Self:
@@ -234,7 +234,7 @@ class TuyaQuirkBuilder(QuirkBuilder):
     def tuya_battery(
         self,
         dp_id: int,
-        power_cfg: PowerConfiguration | None = None,
+        power_cfg: type[PowerConfiguration] | None = None,
         battery_type: BatterySize | None = BatterySize.AA,
         battery_qty: int | None = 2,
         battery_voltage: int | None = None,
@@ -399,7 +399,7 @@ class TuyaQuirkBuilder(QuirkBuilder):
     def tuya_ias(
         self,
         dp_id: int,
-        ias_cfg: TuyaLocalCluster,
+        ias_cfg: type[TuyaLocalCluster],
         converter: Callable[[Any], Any] | None = None,
         endpoint_id: int = 1,
     ) -> Self:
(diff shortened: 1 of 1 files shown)


TheJulianJES:
AI comment:

Good catch — folded those in (9110e8c4). `_tuya_battery`/`tuya_battery` `power_cfg` → `type[PowerConfiguration]`, and `tuya_ias` `ias_cfg` → `type[TuyaLocalCluster]`. They're the same class-as-instance bug; I'd missed them because my verification only re-enabled `assignment`, and these surface under `arg-type` (at the call sites: `ias_cfg=TuyaIasContact` etc.).

Verified with `--enable-error-code arg-type`: 0 errors now (was 4). The only remaining issues under the stricter codes are the two unrelated categories left out by design — the `inspect.currentframe()` `FrameType | None` locals (211-212) and `battery_type` (238), which need more than an annotation tweak.

PR #5110, opened 2026-06-22
Add Sonoff dual-channel water valve (SWV-ZF2E and variants)

yezi289:
## Checklist

Related: #4340


yezi289:
@zigpy-review-bot Thanks for pointing this out.

I reviewed zigpy/zha#845 and agree that this appears to be an upstream diagnostics tooling issue: `ZhaJsonEncoder` currently cannot serialize `bytes` values in quirk entity info.

In this PR, the `bytes` value in `command_args` is intentional because it represents the raw payload required by the SONOFF irrigation/valve command. Converting it to an integer or another JSON-safe type inside the quirk could change the actual ZCL command payload or make the command metadata no longer match what is sent to the device.

My preference is to keep the quirk payload unchanged and address the JSON serialization issue upstream in zigpy/zha#845. Once that fix is available, I can rerun the diagnostics import/snapshot tests for this PR.

Please let me know if you would prefer a temporary JSON-safe metadata workaround in this PR.

PR #5112, opened 2026-06-22
Add quirk for Bituo SPM01X

abmantis:
## Proposed change

The device is single-phase but was adding 3phase entities. This PR
removes those.
It also showed the "energy reset" command as a switch entity. This PR
replaces it with a button entity.
This also removes an "opening" binary_sensor which comes from the on_off
cluster.

Without quirk:
image

## Additional information

Official Zigbee reference: https://docs.bituo-technik.com/device-api/spm-sdm-zigbee-cluster-reference

## Device diagnostics

[BITUO TECHNIK SPM01X.json](https://github.com/user-attachments/files/29225977/BITUO.TECHNIK.SPM01X.json)

## Checklist


zhaquirks/bituo/spm01.py
```diff
@@ -0,0 +1,63 @@
...
+        """Ignore reports/restores for the phantom Phase B/C attributes."""
+        attr_id = attrid.id if isinstance(attrid, ZCLAttributeDef) else attrid
+        if attr_id in self.PHANTOM_PHASE_BC_ATTRIBUTES:
+            return
+        super()._update_attribute(attrid, value)
+
+
+(
+    QuirkBuilder("BITUO TECHNIK", "SPM01X")
+    .replaces(SPM01ElectricalMeasurement)
```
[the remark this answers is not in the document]
abmantis: Added to the PR

pyproject.toml
```diff
@@ -92,7 +92,7 @@ norecursedirs = ".git testing_config"
 
 [tool.codespell]
 skip = "Contributors.md,uv.lock"
-ignore-words-list = "hass, dout, potentiels, checkin"
+ignore-words-list = "hass, dout, potentiels, checkin, technik"
```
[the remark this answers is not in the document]
abmantis: Addressed

tests/test_bituo.py
```diff
@@ -0,0 +1,43 @@
...
+        assert electrical_cluster.get(attr_id) is None
+
+    # Phase A attributes are still cached normally.
+    electrical_cluster.update_attribute(
+        ElectricalMeasurement.AttributeDefs.active_power.id, 100
+    )
+    assert (
+        electrical_cluster.get(ElectricalMeasurement.AttributeDefs.active_power.id)
+        == 100
+    )
```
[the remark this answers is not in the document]
puddly: ZHA already has a form of snapshot tests to ensure entity metadata for real devices does not change unexpectedly.

zhaquirks/bituo/spm01.py
```diff
@@ -0,0 +1,63 @@
...
+from zigpy.zcl.clusters.general import OnOff
+from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
+from zigpy.zcl.foundation import ZCLAttributeDef
+
+
+class SPM01ElectricalMeasurement(CustomCluster, ElectricalMeasurement):
+    """ElectricalMeasurement cluster for the single-phase SPM01X.
+
+    The SPM01X is a single-phase meter, but its firmware reports the Phase B/C
+    measurement attributes with placeholder values (0 or -1). That populates
```
TheJulianJES: Does the device send `0` and `-1`, or `0xFFFF` (as used in the test)? Or everything of that?

Because `0xFFFF` would be somewhat correct, as that's the `non_value` for the unsigned attributes with an unsigned 16 bit integer. But even if, that would still cause the entities to be created, just marked as "unknown" state. (`0x8000` + `0x80` for other attributes)

I guess we should still adjust the docstring and/or the test to match what the device actually does.

zhaquirks/bituo/spm01.py
```diff
@@ -0,0 +1,63 @@
...
+    # for this cluster (the switch, and the "opening" binary sensor)
+    # and expose a button instead.
+    .prevent_default_entity_creation(
+        endpoint_id=1,
+        cluster_id=OnOff.cluster_id,
+        # Avoid preventing the quirk-defined reset button from being created
+        function=lambda entity: (
+            getattr(entity, "translation_key", None) != "reset_energy"
+        ),
+    )
```
TheJulianJES, line 54: Ok, this kind of sucks 😅. Something like https://github.com/zigpy/zha/pull/763 may improve this a bit, so that `prevent_default_entity_creation` truly only removes "default ZCL entities". But I think we should also add a `platform` kwarg to that.

(And with https://github.com/zigpy/zha/pull/762, I think we can also generally improve how we target default ZCL entities for removal (maybe directly reference the class?). Otherwise, maybe we should just have some sort of `key: str` to target.)

zhaquirks/bituo/spm01.py
```diff
@@ -0,0 +1,63 @@
...
+    .command_button(
+        OnOff.ServerCommandDefs.on.name,
+        OnOff.cluster_id,
+        entity_type=EntityType.CONFIG,
+        initially_disabled=True,
```
TheJulianJES, line 59: I think we could even enable this button by default – I think that's what we do for other quirks.
Side note: Other quirks use `reset_summation_delivered` as the translation key ("Reset summation delivered"). This should also be fine though.

zhaquirks/bituo/spm01.py
```diff
@@ -0,0 +1,63 @@
...
+            ElectricalMeasurement.AttributeDefs.power_factor_ph_b.id,
+            ElectricalMeasurement.AttributeDefs.power_factor_ph_c.id,
+        }
+    )
+
+    def _update_attribute(self, attrid, value):
+        """Ignore reports/restores for the phantom Phase B/C attributes."""
+        attr_id = attrid.id if isinstance(attrid, ZCLAttributeDef) else attrid
+        if attr_id in self.PHANTOM_PHASE_BC_ATTRIBUTES:
+            return
```
TheJulianJES: I guess this is fine but:
- This has the downside of not fixing older devices, unless they're re-interviewed once via the "Reconfigure" button. We don't have a good way of migrating or removing anything in ZCL cache at the moment. Calling `update_attribute` (with `value=None`) in `__init__` does not work btw.

So just wondered whether we should instead:
1. Remove the entities (and just keep the attribute cache as-is)
    - This would be the most "modern / quirks v2"-like approach – not modifying underlying ZCL and just the (Z)HA entity representation.
2. Override `is_attribute_unsupported`?
    - This would have the benefit of retroactively "fixing" previously paired devices automatically – ZHA just doesn't create the sensor entities then.
    - We would be able to keep that same list of attributes.
    - This would be somewhat similar to how `_CONSTANT_ATTRIBUTES` is implemented (in zigpy itself).

zhaquirks/bituo/spm01.py
TheJulianJES: Ok, looking at the screenshots, it seems to be all three values.

TheJulianJES (commented):
Nice! I think this is one of the first devices where a bunch of attributes aren't supported 😅. I'd kind of prefer just overriding `is_attribute_unsupported` for now, instead of `_update_attribute`.

zhaquirks/bituo/spm01.py
abmantis: yes, this is very ugly atm! :laughing: [#763](https://github.com/zigpy/zha/pull/763) would indeed sort it out.

zhaquirks/bituo/spm01.py
abmantis: I think this button also resets summation_received (I haven't tested it since my device is only on the "delivering" side).
I would suggest keeping it disabled by default, since it is not an entity that has a low chance of being needed/used.

abmantis: Side-note: it would be nice if we eventually replace the "summation" attr nomenclature with "energy"/"gas"/"water"/etc to make those a bit more user friendly

zhaquirks/bituo/spm01.py
abmantis: Changed to overriding `is_attribute_unsupported`

TheJulianJES (approved):
Updated to work with the recent quirks/ZHA refactor (https://github.com/zigpy/zha-device-handlers/pull/5113).
The refactor is part of HA Core 2026.7.0 (beta) and newer, so this quirk won’t work when installed as a custom quirk on previous HA versions.

TheJulianJES:
Thanks!

PR #5113, opened 2026-06-22
Move the quirks API out of zigpy

puddly:
## Proposed change

This is the quirks third of https://github.com/zigpy/zha/pull/762, which also depends on https://github.com/zigpy/zigpy/pull/1846.

This PR moves the entire quirks API out of zigpy and into zha-device-handlers. It implements it on top of the new ZHA device matching API, which gives zha-device-handlers full access to ZHA entities as primitives: complex devices can be implemented directly in this repo as self-contained modules.

One major mechanical TODO is to move all of the zigpy quirks tests here, to maintain test coverage.

## Additional information

N/A

## Device diagnostics

N/A

## Checklist


zhaquirks/builder/device.py
```diff
@@ -0,0 +1,188 @@
...
+import zigpy.zcl
+
+from zhaquirks.builder.discovery import discover_quirks_v2_entities
+from zhaquirks.builder.metadata import QuirkDefinition
+
+if TYPE_CHECKING:
+    from zha.application.gateway import Gateway
+
+
+def _entity_targets_cluster(
```
dmulcahey: would it make sense to make this a method on PlatformEntity itself?

zhaquirks/philips/hue_light.py
```diff
@@ -144,20 +143,6 @@
     .add_to_registry()
 )
 
-(
```
dmulcahey, line 147: was this intentional? I may be missing something...

puddly: The entire quirk definition was duplicated like 80 lines up 😅

PR #5171, opened 2026-07-12
Add APIs for groups, binding, reporting, and writing attributes

### LINKED ISSUE #708: Potential issue with events in bellows for IKEA Bilresa 2 button remote?
Not sure where to bring this up and happy to move the conversation somewhere else if desired. Or close it if there already discussion happening amongst the devs.

Anyway, I submitted the [PR](https://github.com/zigpy/zha-device-handlers/pull/4612) for the quirk v2 for IKEA Bilresa 2 button remote and I started a [post on the community forums](https://community.home-assistant.io/t/ikea-bilresa-2-button-remote-pairs-with-zha/968513) about it working with ZHA. A number of commenters in the post have mentioned that they are not seeing any events from the remote before or after using the quirk. I was rereading through the post and noticed that all of the affected users seem to be using coordinators with EFR32MG24 and EFR32MG21 based radios, including the ZBT-2.

I am personally using a Sonoff Dongle P with 20211217 firmware and don't have a second coordinator, let alone one with EFR32MG24 or EFR32MG21 radios to test this out.

I am wondering if there is something in bellows that is blocking or not generating the zha_event on the event bus for the Bilresa 2 button remote (and possibly other devices). The data set from the post is obviously very small, but the coordinator radio stood out as a common theme.

puddly:
## Proposed change

This PR adds a few new quirks v2 APIs (zigpy and ZHA PRs will be linked later):
- `subscribes_to_multicast_group`: Fixes https://github.com/zigpy/bellows/issues/708 by adding the coordinator to the required groups, provided by a quirk. I've added a few default implementations (there aren't many devices that need this)
- `binds_cluster`: binds to a cluster. This replaces the current manufacturer-specific binding done in ZHA.
- `configures_reporting`: similar to `.sensor(..., reporting_config=...)` but without requiring an entity to be created.
- `writes_attributes`: writes an attribute during initialization. This replaces a few `binds()` overrides and further reduces the amount of Python runtime code.

`writes_attributes` allowed the Opple remote quirk to be rewritten from V1 to V2.

## Checklist

### DIFF
--- a/tests/test_quirks_v2_builder.py
+++ b/tests/test_quirks_v2_builder.py
@@ -27,7 +27,11 @@ from zigpy.zdo.types import LogicalType, NodeDescriptor
 
 from zhaquirks.builder import QuirkBuilder
 from zhaquirks.builder.device import QuirkV2Device
-from zhaquirks.builder.metadata import QuirkDefinition, recursive_freeze
+from zhaquirks.builder.metadata import (
+    QuirkDefinition,
+    ReportingConfig,
+    recursive_freeze,
+)
 from zhaquirks.clusters import CustomCluster
 from zhaquirks.const import COMMAND, COMMAND_ON, SHORT_PRESS, TURN_ON
 from zhaquirks.device import CustomZigpyDevice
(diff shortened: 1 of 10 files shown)


TheJulianJES (commented):
Nice! I had started with the binding + reporting as well 😄: https://github.com/zigpy/zha-device-handlers/pull/5128. Some differences for binding/reporting behavior (separate cluster configs, thus multiple virtual entities with different unique ID behavior in my PR).

We can go with whatever PR you want. I guess a default cluster config (like here) makes more sense?
(nit: Splitting up the PR between APIs and modifying existing quirks might be nice when everything's done.)

PR #5179, opened 2026-07-15
Support Aeotec Pico Switch

puddly:
## Proposed change

This PR adds support for the Aeotec Pico Switch.

## Additional information

The Pico Switch Duo is effectively identical but requires better support for duplicate entities so I've left it out of this PR.

## Device diagnostics

[zha-01KTWBERD1KGPHZRX3YA4VC7FS-Aeotec Pico Switch-28d93814024f7c14ba06c8737b5a0cd9.json](https://github.com/user-attachments/files/30030490/zha-01KTWBERD1KGPHZRX3YA4VC7FS-Aeotec.Pico.Switch-28d93814024f7c14ba06c8737b5a0cd9.json)

## Checklist


zhaquirks/aeotec/pico_switch.py
```diff
@@ -0,0 +1,128 @@
...
+            type=AeotecControls,
+            zcl_type=DataTypeId.enum8,
+            access="rw",
+            manufacturer_code=AEOTEC_MANUFACTURER_ID,
+        )
+
+
+(
+    QuirkBuilder("AEOTEC", "ZGA002")
+    .friendly_name(model="Pico Switch", manufacturer="Aeotec")
```
TheJulianJES: nit: I think I prefer following the same manufacturer/model order we use for matching quirks:
```suggestion
    .friendly_name(manufacturer="Aeotec", model="Pico Switch")
```

zhaquirks/aeotec/pico_switch.py
```diff
@@ -0,0 +1,128 @@
...
+    name: str = "Aeotec Switch Type Configuration"
+    ep_attribute: str = "aeotec_switch_type_config"
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        switch_type: Final = ZCLAttributeDef(
+            id=0x0000,
+            type=AeotecSwitchType,
+            zcl_type=DataTypeId.enum8,
```
TheJulianJES: Why do we need to set `zcl_type` to `enum8` here? `AeotecSwitchType` already subclasses `t.enum8`, so that should work fine and be "detected" by zigpy? But we also do this for a NodOn switch apparently: https://github.com/zigpy/zha-device-handlers/blob/899e26dc856c88a5f50ee30bb69c89ee2d81fa0a/zhaquirks/nodon/switch.py#L30-L31

zhaquirks/aeotec/pico_switch.py
```diff
@@ -0,0 +1,128 @@
...
+    .replaces(AeotecSwitchTypeConfigCluster, endpoint_id=2)
+    .replaces(AeotecSwitchTypeConfigCluster, endpoint_id=3)
+    .enum(
+        attribute_name=AeotecSwitchTypeConfigCluster.AttributeDefs.switch_type.name,
+        enum_class=AeotecSwitchType,
+        cluster_id=AeotecSwitchTypeConfigCluster.cluster_id,
+        endpoint_id=2,
+        entity_type=EntityType.CONFIG,
+        translation_key="s1_switch_type",
+        fallback_name="S1 switch type",
```
TheJulianJES: note: I'm not sure we want/need to do that here, but we could technically use `translation_placeholders` here. See this as an example: [Ubisys quirk](https://github.com/zigpy/zha-device-handlers/blob/899e26dc856c88a5f50ee30bb69c89ee2d81fa0a/zhaquirks/ubisys/control_c4.py#L67-L69) + [HA `strings.json` entry](https://github.com/home-assistant/core/blob/f37df79107e57922cd320f00087ec75659d809b2/homeassistant/components/zha/strings.json#L1279-L1281).

We'll have to "manually" add that to HA `strings.json` when doing the quirks bump though, if we even do this. I think the S1/S2 being part of the translation may also be fine.

zhaquirks/aeotec/pico_switch.py
```diff
@@ -0,0 +1,128 @@
...
+    QuirkBuilder("AEOTEC", "ZGA002")
+    .friendly_name(model="Pico Switch", manufacturer="Aeotec")
+    .replaces(AeotecSwitchTypeConfigCluster, endpoint_id=2)
+    .replaces(AeotecSwitchTypeConfigCluster, endpoint_id=3)
+    .enum(
+        attribute_name=AeotecSwitchTypeConfigCluster.AttributeDefs.switch_type.name,
+        enum_class=AeotecSwitchType,
+        cluster_id=AeotecSwitchTypeConfigCluster.cluster_id,
+        endpoint_id=2,
+        entity_type=EntityType.CONFIG,
```
TheJulianJES: nit: Last note, we're really inconsistent on when we provide default kwargs 😅. It's fine to keep it explicitly as `CONFIG` though – some other quirks already do this. (Reasoning behind `CONFIG` default is that most v2 quirks just expose config entities only.)

puddly:
Should be cleaned up now!

TheJulianJES (approved):

PR #5200, opened 2026-07-24
Add network LED switch for Sonoff S60ZBTPF smart plug

FMeinicke:
## Proposed change

Adds a binary switch for the network LED attribute in the Sonoff/eWeLink cluster (0xFC11). Turning this off, turns off the blue status indicator LED on the plug.

## Checklist

### DIFF
--- a/zhaquirks/sonoff/s60zbtpf.py
+++ b/zhaquirks/sonoff/s60zbtpf.py
@@ -20,6 +20,7 @@ from zigpy.zcl import foundation
 from zigpy.zcl.clusters.general import OnOff
 from zigpy.zcl.clusters.homeautomation import ElectricalMeasurement
 from zigpy.zcl.clusters.smartenergy import Metering
+from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 
 from zhaquirks.builder import QuirkBuilder
 from zhaquirks.clusters import CustomCluster
@@ -48,6 +49,23 @@ class SonoffS60OnOff(CustomCluster, OnOff):
             )
 
         super()._update_attribute(attrid, value)
+
+
+class SonoffEwelinkCluster(CustomCluster):
+    """Sonoff/eWeLink cluster."""
+
+    cluster_id = 0xFC11
+    name = "SONOFF eWeLink cluster"
+    ep_attribute = "ewelink"
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        network_led = ZCLAttributeDef(
+            id=0x0001,
+            type=t.Bool,
+            manufacturer_code=None,
+        )
 
 
 class SonoffS60ElectricalMeasurement(CustomCluster, ElectricalMeasurement):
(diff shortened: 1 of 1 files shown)


zhaquirks/sonoff/s60zbtpf.py
```diff
@@ -71,6 +90,17 @@ def _update_attribute(
     .applies_to("SONOFF", "S60ZBTPG")
     .replaces(SonoffS60OnOff)
     .replaces(SonoffS60ElectricalMeasurement)
+    .replaces(SonoffEwelinkCluster, endpoint_id=1)
+    .switch(
+        attribute_name=SonoffEwelinkCluster.AttributeDefs.network_led.name,
+        cluster_id=SonoffEwelinkCluster.cluster_id,
+        endpoint_id=1,
+        off_value=t.Bool.false,
+        on_value=t.Bool.true,
```
TheJulianJES: This seems unnecessary

zhaquirks/sonoff/s60zbtpf.py
```diff
@@ -71,6 +90,17 @@ def _update_attribute(
...
+        attribute_initialized_from_cache=False,
```
TheJulianJES: Any reason why you set this to `False`?

zhaquirks/sonoff/s60zbtpf.py
```diff
@@ -44,6 +45,24 @@ def _update_attribute(
...
+    ep_attribute = "ewelink"
+
+    class AttributeDefs(BaseAttributeDefs):
+        """Attribute definitions."""
+
+        network_led = ZCLAttributeDef(
+            id=0x0001,
+            type=t.Bool,
+            manufacturer_code=None,
+            is_manufacturer_specific=True,
```
TheJulianJES: `manufacturer_code=None` takes priority over `is_manufacturer_specific`. And `is_manufacturer_specific=False` would've been the same as `manufacturer_code=None`. So remove the bottom line:
```suggestion
```

zhaquirks/sonoff/s60zbtpf.py
FMeinicke: True. Removing.

zhaquirks/sonoff/s60zbtpf.py
FMeinicke: I had an issue where the switch would show up in HA but would be *unavailable*. Setting `attribute_initialized_from_cache=False` was one thing I tried to force zha to read the current value from the device upon initialization. I'm not sure anymore now that this was actually the fix.

I removed this line and tested again locally with my devices and it seems this setting was unnecessary. Removing.

zhaquirks/sonoff/s60zbtpf.py
FMeinicke: Thanks for the explanation. I might have switched that up in my head, sorry. 😅

FMeinicke:
Applied your suggestions and rebased on dev.

Thanks, @TheJulianJES.

Stono:
Hey, any chance we can get this merged please?  I just raised a PR to fix the same problem (admittedly didn't check for dupes first)

PR #5201, opened 2026-07-24
Add SONOFF TP-WGZBA thermostat support

viys:
## Proposed change

Add support for the SONOFF TP-WGZBA thermostat.

The quirk replaces the device thermostat and manufacturer-specific private cluster, exposes the device controls and diagnostics as ZHA entities, and encodes the device-specific weekly schedule and temporary-mode commands. It also includes behavior tests for entity creation, system-mode handling, schedule parsing/encoding, and write failure handling.

## Additional information

The device uses manufacturer code `0x1286` and private cluster `0xFC11`.

The imports include compatibility fallbacks for older Home Assistant/ZHA releases. The primary target is the current quirk builder API, while the fallback was verified with Home Assistant 2026.3.3.

Full-suite testing completed with two unrelated pre-existing Tuya time-payload failures:

- `tests/test_tuya.py::test_moes[MoesHY368_Type1]`
- `tests/test_tuya_mcu.py::test_tuya_mcu_set_time[TuyaDoubleSwitchDimmer]`

## Device diagnostics

Diagnostic dump provided by the reporter:

[zha-01KY9HWXPG21KE9QBDGQ6YABRW-SONOFF TP-WGZBA-98d70e73f61912a31a5ea9db77861c38.json](https://github.com/user-attachments/files/30344018/zha-01KY9HWXPG21KE9QBDGQ6YABRW-SONOFF.TP-WGZBA-98d70e73f61912a31a5ea9db77861c38.json)

The dump was collected from Home Assistant 2026.3.3 with ZHA 1.0.2. It reports `quirk_applied: true`, `issues: []`, and the expected thermostat endpoint `0x0201` plus private cluster `0xFC11`. The original JSON should be attached to this section through the GitHub PR page.

## Checklist

### DIFF
--- a/tests/test_sonoff_tp_wgzba.py
+++ b/tests/test_sonoff_tp_wgzba.py
@@ -1,7 +1,9 @@
 """Tests for the SONOFF TP-WGZBA thermostat quirk."""
 
+from types import SimpleNamespace
 from unittest import mock
 
+import pytest
 from zha.quirks import DEVICE_REGISTRY
 from zigpy.zcl import ClusterType, foundation
 from zigpy.zcl.clusters.hvac import Thermostat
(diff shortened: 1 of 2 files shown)


viys:
@zigpy-review-bot Thanks for pointing this out. I reviewed zigpy/zha#845 and agree that this is an upstream diagnostics tooling issue: ZhaJsonEncoder currently cannot serialize bytes in quirk entity info.

The bytes value in command_args is intentional because it is the raw payload required by the TP-WGZBA temporary-mode command. Changing it to an integer would risk changing the actual ZCL command payload. I propose keeping the quirk payload unchanged and addressing the ZhaJsonEncoder fix upstream in zigpy/zha#845. Once that fix is available, I will rerun the diagnostics import/snapshot tests for this PR.

Please let me know if you would prefer a temporary JSON-safe metadata workaround in this PR.

Issue #1623, opened 2022-06-15
[Device Support Request] Schneider Electric / LK Wiser 550D6001 four-button Battery device


mrwee:
**Is your feature request related to a problem? Please describe.**
The device is recognized by zha and I can make one (1) out of four devices work in hass. The rest does not generate any events.

**Describe the solution you'd like**
Full support for this device

<details>
```
<!-- Device signature can be acquired by clicking on the "Zigbee Device Signature" button in the device settings view -->
<summary>Device signature</summary>
```

```yaml
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=4190, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=11264, maximum_outgoing_transfer_size=82, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "21": {
      "profile_id": 260,
      "device_type": "0x0104",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0020",
        "0xff17"
      ],
      "out_clusters": [
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0019",
        "0x0102"
      ]
    },
    "22": {
      "profile_id": 260,
      "device_type": "0x0104",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0xff17"
      ],
      "out_clusters": [
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
        "0x0102"
      ]
    },
    "23": {
      "profile_id": 260,
      "device_type": "0x0104",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0xff17"
      ],
      "out_clusters": [
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0x0008",
[...]

PR #5220, opened 2026-07-30
Add quirk for Schneider Electric FLS/AIRLINK/4 wireless switch

Crocmagnon:
## Proposed change

Add a v2 quirk for the battery-powered Schneider Electric FLS/AIRLINK/4 wireless wall switch (Odace connectable wireless switch).

The device exposes two rockers (endpoint 22 = left, endpoint 21 = right), each sending on/off (OnOff) on a short press and level move/stop (LevelControl) on a long press/release. The quirk:

- adds device automation triggers for both rockers, and
- exposes a single battery entity by hiding the redundant ones on the secondary endpoints (22/23/24), keeping endpoint 21.

Purely declarative quirk (no custom cluster or logic), so no dedicated
test is required per the contributing guidelines.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Device diagnostics

[without quirk](https://github.com/user-attachments/files/30538206/zha-01KN1ZMJ05EV56PWVPE42Q7WVK-Schneider.Electric.FLS_AIRLINK_4-8ffff7ea5c2e8a1402f44abee430e95d.json)

[with quirk](https://github.com/user-attachments/files/30538211/zha-01KN1ZMJ05EV56PWVPE42Q7WVK-Schneider.Electric.FLS_AIRLINK_4-8ffff7ea5c2e8a1402f44abee430e95d.1.json)

## Checklist

Related: #1623

### DIFF
--- a/zhaquirks/schneiderelectric/wireless_switch.py
+++ b/zhaquirks/schneiderelectric/wireless_switch.py
@@ -1,0 +1,100 @@
+"""Quirk for the Schneider Electric FLS/AIRLINK/4 wireless wall switch."""
+
+from zigpy.zcl.clusters.general import LevelControl, OnOff, PowerConfiguration
+
+from zhaquirks.builder import QuirkBuilder
+from zhaquirks.const import (
+    CLUSTER_ID,
+    COMMAND,
+    COMMAND_MOVE,
+    COMMAND_MOVE_ON_OFF,
+    COMMAND_OFF,
+    COMMAND_ON,
+    COMMAND_STOP,
+    DIM_DOWN,
+    DIM_UP,
+    ENDPOINT_ID,
+    LEFT,
+    LONG_RELEASE,
+    RIGHT,
+    TURN_OFF,
+    TURN_ON,
+)
+from zhaquirks.schneiderelectric import SE_MANUF_NAME
+
+# Battery-powered remote with two rockers. Each rocker sends on/off on a short
+# press (OnOff) and level move/stop on a long press/release (LevelControl).
+LEFT_EP = 22
+RIGHT_EP = 21
+
+(
+    QuirkBuilder(SE_MANUF_NAME, "FLS/AIRLINK/4")
+    .device_automation_triggers(
+        {
+            # Left rocker (endpoint 22)
+            (TURN_ON, LEFT): {
+                COMMAND: COMMAND_ON,
+                CLUSTER_ID: OnOff.cluster_id,
+                ENDPOINT_ID: LEFT_EP,
+            },
+            (TURN_OFF, LEFT): {
+                COMMAND: COMMAND_OFF,
+                CLUSTER_ID: OnOff.cluster_id,
+                ENDPOINT_ID: LEFT_EP,
+            },
+            (DIM_UP, LEFT): {
+                COMMAND: COMMAND_MOVE_ON_OFF,
+                CLUSTER_ID: LevelControl.cluster_id,
+                ENDPOINT_ID: LEFT_EP,
+            },
+            (DIM_DOWN, LEFT): {
+                COMMAND: COMMAND_MOVE,
+                CLUSTER_ID: LevelControl.cluster_id,
+                ENDPOINT_ID: LEFT_EP,
+            },
+            (LONG_RELEASE, LEFT): {
+                COMMAND: COMMAND_STOP,
+                CLUSTER_ID: LevelControl.cluster_id,
+                ENDPOINT_ID: LEFT_EP,
+            },
+            # Right rocker (endpoint 21)
+            (TURN_ON, RIGHT): {
+                COMMAND: COMMAND_ON,
+                CLUSTER_ID: OnOff.cluster_id,
+                ENDPOINT_ID: RIGHT_EP,
+            },
+            (TURN_OFF, RIGHT): {
+                COMMAND: COMMAND_OFF,
+                CLUSTER_ID: OnOff.cluster_id,
+                ENDPOINT_ID: RIGHT_EP,
+            },
+            (DIM_UP, RIGHT): {
+                COMMAND: COMMAND_MOVE_ON_OFF,
+                CLUSTER_ID: LevelControl.cluster_id,
+                ENDPOINT_ID: RIGHT_EP,
+            },
+            (DIM_DOWN, RIGHT): {
+                COMMAND: COMMAND_MOVE,
+                CLUSTER_ID: LevelControl.cluster_id,
+                ENDPOINT_ID: RIGHT_EP,
+            },
+            (LONG_RELEASE, RIGHT): {
+                COMMAND: COMMAND_STOP,
+                CLUSTER_ID: LevelControl.cluster_id,
+                ENDPOINT_ID: RIGHT_EP,
+            },
+        }
+    )
+    # Expose a single battery entity: keep endpoint 21 (the main endpoint, which
+    # also carries poll control and OTA) and hide the redundant ones.
+    .prevent_default_entity_creation(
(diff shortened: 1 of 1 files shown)


KevinFabre:
Hi @Crocmagnon — I have the same device and went through a fairly extensive capture campaign on it. Your endpoint mapping matches mine exactly (21 = right, 22 = left), so that's independently confirmed on a second unit. Happy to fold my findings into your PR rather than open a competing one — let me know what works best for you.

One issue I think is blocking: the quirk doesn't replace `SESwitchConfiguration` (0xFF17), so there's no way to see or change `se_switch_actions` from HA.

On my unit, endpoints 22/23/24 were `NotUsed` out of the box. Only endpoint 21 emitted anything, so the five "left" triggers were dead and there was no way to find out why: without the quirk the cluster shows up as unknown and its attributes aren't listed, so the panel can't read it either. Adding the cluster plus a `switch_actions` select is what let me activate the second rocker.

Two more things I found while capturing every `se_switch_actions` value:

1. The shutter modes need `SEWindowCovering` replacing the **client** window covering cluster (same as `shutters.py` does). Without it ZHA  silently drops `se_stop_or_step_lift_percentage` (0x80) and a short press in `SchneiderShutter` mode looks like it emits nothing at all. With it, the command carries `direction` 0/1 — the only mode that tells the two buttons of a rocker apart on a short press.

2. `ToggleLight` and `ToggleDimmer` are annotated "No observable function?" in `schneiderelectric/__init__.py`, but on this device they do work: both buttons send `toggle` on a short press, and a long press sends `off` (ToggleLight) or a level move + `stop` (ToggleDimmer). `AlternateLight`, `AlternateDimmer` and `Scene` really are silent.

Full trigger table covering all 14 modes, replayed against 52 captured `zha_event`s. Also, your battery dedup is a good catch — I hadn't noticed the power cluster was on all four endpoints.

image

Crocmagnon:
I suggest we start with a simple quirk and build upon it later?

KevinFabre:
Hi @Crocmagnon, happy to work with whatever you prefer here. This would be my first contribution to this repo, so I don't know the conventions yet, and I'll defer to you and the maintainers on process.

Starting simple and building on it sounds right to me. Two things I'd suggest keeping in this first iteration, and the rest can definitely wait.

**1. The `switch_actions` select**

My unit had endpoints 22, 23 and 24 set to `NotUsed`, so only the right rocker emitted anything, and the five "left" triggers sat in the UI without ever firing. I can't tell you whether that's the factory default (yours clearly wasn't in that state), but a device can be in it, and right now there's no way to find out. Without the quirk, 0xFF17 shows up as an unknown cluster, so the ZHA panel can't list its attributes either. Your own diagnostics show the symptom: 0xFF17 has an empty attribute list on all four endpoints, in both the with-quirk and without-quirk dumps.

Replacing `SESwitchConfiguration` on endpoints 21 and 22, plus one `.enum()` each, makes the state visible and fixable. It's about 10 lines, and it's what let me activate my second rocker.

**2. Trigger naming**

HA automations store the device trigger's `type` and `subtype`, so renaming them after a release silently breaks whatever users have built in the meantime. Worth settling once rather than after the fact.

Your current scheme, per rocker:

```python
(TURN_ON, LEFT):      {COMMAND: COMMAND_ON,           ENDPOINT_ID: 22}
(TURN_OFF, LEFT):     {COMMAND: COMMAND_OFF,          ENDPOINT_ID: 22}
(DIM_UP, LEFT):       {COMMAND: COMMAND_MOVE_ON_OFF,  ENDPOINT_ID: 22}
(DIM_DOWN, LEFT):     {COMMAND: COMMAND_MOVE,         ENDPOINT_ID: 22}
(LONG_RELEASE, LEFT): {COMMAND: COMMAND_STOP,         ENDPOINT_ID: 22}
```

The alternative I've been using keys on the button numbers that are printed on the device itself. On mine the layout is:

```
3   1
4   2
```

so endpoint 21 carries buttons 1 and 2, endpoint 22 carries 3 and 4:

```python
(SHORT_PRESS, BUTTON_3):  {COMMAND: COMMAND_ON,          ENDPOINT_ID: 22}
(SHORT_PRESS, BUTTON_4):  {COMMAND: COMMAND_OFF,         ENDPOINT_ID: 22}
(LONG_PRESS, BUTTON_3):   {COMMAND: COMMAND_MOVE_ON_OFF, ENDPOINT_ID: 22}
(LONG_PRESS, BUTTON_4):   {COMMAND: COMMAND_MOVE,        ENDPOINT_ID: 22}
(LONG_RELEASE, LEFT):     {COMMAND: COMMAND_STOP,        ENDPOINT_ID: 22}
```

The practical difference is what a user picks in the automation UI: "Turn on, Left" versus "Remote button short press, Button 3", where Button 3 is the number they can read on the wall. It also puts a press duration in the `type` slot, which is what most quirks in this repo do.

Both schemes extend to the other switch actions without renaming anything, so this isn't a correctness issue. Yours would grow like this:

```python
(OPEN, LEFT):   {COMMAND: "up_open",    CLUSTER_ID: 258, ENDPOINT_ID: 22}
(CLOSE, LEFT):  {COMMAND: "down_close", CLUSTER_ID: 258, ENDPOINT_ID: 22}
(STOP, LEFT):   {COMMAND: "stop",       CLUSTER_ID: 258, ENDPOINT_ID: 22}
(TOGGLE, LEFT): {COMMAND: COMMAND_TOGGLE, CLUSTER_ID: 6, ENDPOINT_ID: 22}
```

The one spot that needs more granularity either way is `SchneiderShutter`. Its short press sends the vendor command `se_stop_or_step_lift_percentage` carrying `direction` 0 or 1, so it does distinguish the two buttons of a rocker, and a single `LEFT` subtype can't express that:

```python
(STOP, BUTTON_3): {COMMAND: "se_stop_or_step_lift_percentage",
                   CLUSTER_ID: 258, ENDPOINT_ID: 22,
                   PARAMS: {"direction": 0}}
(STOP, BUTTON_4): {COMMAND: "se_stop_or_step_lift_percentage",
                   CLUSTER_ID: 258, ENDPOINT_ID: 22,
                   PARAMS: {"direction": 1}}
```

I have no strong attachment to my scheme, I just think it's better to pick one now than to change it later.

Everything else is genuinely follow-up material: full coverage of the 14 `se_switch_actions` values, and the client-side `SEWindowCovering` the shutter modes need for ZHA to decode 0x80 at all. Happy to send you a patch for the two points above, or to open a second PR once yours is in, whichever suits you.

Crocmagnon:
Pardon me if I’m wrong, but these two comments scream LLM. Please disclose it when addressing me.
Trigger naming is easier with « top left » rather than « button 3 » because the number isn’t visible on the outside of the device. It may be visible inside but I don’t have access to the device anymore (installed in another home) so I can’t check.
Switch action can be changed by changing the device mode with a combination of button presses (see documentation), so it’s not mandatory to start with.

KevinFabre:
You're right, and thanks for saying it plainly. Yes, I use AI to speed up my work, and that includes those comments. I'll state it up front from now on.

For context, since it probably explains why they read the way they did: I installed these switches on my Zigbee network, discovered custom quirks because of them, and worked with Claude to understand how quirks are built and to put together a full one for this device. My goal was to contribute, not to lecture you on your own PR.

On the button numbers, you're right,  the numbers are invisible once the switch is on the wall. "Top left" is the better label. Please ignore that suggestion.

On switch actions, fair enough. If the mode can be changed with a button combination then it isn't blocking, and it can come later.

One thing I'd still stand behind: everything I reported came off my own device. I set each `se_switch_actions` value in turn and captured the `zha_event` output for every button and gesture, which is where the shutter mode findings came from. Happy to share the raw captures if they're ever useful. Otherwise I'll wait for your PR to land and propose the rest on top of it.

PR #5131, opened 2026-06-29
Add SONOFF SNZB-04PR2 tamper quirk

phipz:
## Proposed change

Add a dedicated ZHA quirk for the SONOFF SNZB-04PR2 contact sensor.

The SNZB-04PR2 is a different device from the existing SNZB-04P and reports tamper on Sonoff manufacturer-specific cluster `0xFC11`, attribute `0x2000`.

The official SONOFF documentation links a ZHA script from FAQ point 5: https://help.sonoff.tech/docs/SNZB-04PR2

In local testing, the SONOFF-provided script exposed the tamper entity, but live tamper changes did not reliably update the ZHA binary sensor.

Compared with the SONOFF-provided ZHA script, this PR:

- Defines tamper attribute `0x2000` on cluster `0xFC11` as `types.uint8_t` instead of `types.Bool`.
- Removes `is_manufacturer_specific=True`, because the device reports `manufacturer_code=None`.
- Keeps cluster `0xFC11` as the Sonoff manufacturer-specific cluster.
- Handles live `Report Attributes` updates for `0x2000` so the ZHA tamper binary sensor updates immediately.

Debug logs from the device show that the SNZB-04PR2 reports tamper as a normal ZCL `Report Attributes` frame on the manufacturer-specific cluster:

```text
cluster_id=0xFC11
attribute_id=0x2000
type=uint8_t
manufacturer_code=None
value=0/1
```

## Additional information

This is intentionally added as a separate SNZB-04PR2 quirk rather than modifying `snzb04p.py`, because the PR2 is a different model and uses a different tamper reporting path.

The important distinction observed in the logs is:

```text
manufacturer-specific cluster: yes
manufacturer-specific attribute report: no
```

Example live report for tamper active:

```text
[0xA89E:1:0xfc11] Received ZCL frame: '18 a4 0a 00 20 20 01'
Decoded ZCL frame header:
FrameControl<0x18>(..., is_manufacturer_specific=0, ...)
Decoded ZCL frame:
SonoffContactCluster:Report_Attributes(
  attribute_reports=[
    Attribute(attrid=0x2000, value=TypeValue(type=uint8_t, value=1))
  ]
)
Emitting event attribute_report:
attribute_name='tamper',
attribute_id=8192,
manufacturer_code=None,
raw_value=1,
value=1
```

Example live report for tamper clear:

```text
[0xA89E:1:0xfc11] Received ZCL frame: '18 a5 0a 00 20 20 00'
Decoded ZCL frame:
[...]

PR #5225, opened 2026-08-02
Add SNZB-04PR2 tamper quirk

slackticus:
## Summary
- add a ZHA quirk for SONOFF SNZB-04PR2
- expose the rear tamper switch as a diagnostic binary sensor
- cover the PR2 standard-ZCL tamper attribute in a regression test

## Root cause
The SNZB-04PR2 reports tamper state on manufacturer cluster `0xFC11`, attribute `0x2000`, using a standard ZCL report frame. ZHA did not have a PR2 quirk, so it did not expose that attribute as an entity.

## Validation
- `uv run pytest tests/test_sonoff.py -q`
- `uv run ruff check zhaquirks/sonoff/snzb04p.py tests/test_sonoff.py`
- Live ZHA verification on an SNZB-04PR2: press changed tamper `on` to `off`; release restored `on`.

### DIFF
--- a/tests/test_sonoff.py
+++ b/tests/test_sonoff.py
@@ -19,6 +19,21 @@ from zhaquirks.sonoff.zbm5 import (
 zhaquirks.setup()
 
 
+async def test_snzb04pr2_tamper_attribute_is_standard_zcl(
+    zigpy_device_from_v2_quirk,
+):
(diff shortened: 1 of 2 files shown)


slackticus:
Closing as a duplicate of #5131, which already implements the complete SNZB-04PR2 tamper-report handling and live update path. I validated #5131’s approach on a live device.

Issue #5116, opened 2026-06-23
[BUG] Aqara FP300 custom quirk battery issue


rich5741:
### Bug description

I have a different battery issue than the already identified one while using the FP300 with the latest (as of 6/21/2026) custom quirk and latest (as of 6/21/2026) firmware.  Unsure how to pull the version numbers of quirk to confirm, but I did reinstall 6/21 to check.  Firmware on device page is 0x0000412a but unsure about accuracy given its a custom quirk.  Custom quirk from: https://github.com/zigpy/zha-device-handlers/pull/4504

My battery will bounce between zero and 100, as in the below picture.  this happens regardless of whether the device is idle or active (its a living room device so inactive during the day, active starting around 5-6pm.

See attached.

<img width="600" height="1305" alt="Image" src="https://github.com/user-attachments/assets/0cb64a94-229c-4ad2-8316-b5ce7d3ca8f3" />

### Steps to reproduce

View Battery stats

### Expected behavior

Consistent battery reading

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

<img width="600" height="1305" alt="Image" src="https://github.com/user-attachments/assets/32d5b1d9-2517-424f-b1dc-de61383209f3" />

</details>

### Diagnostics information

[zha-FP300-diagnostics-anonymized.json](https://github.com/user-attachments/files/29253917/zha-FP300-diagnostics-anonymized.json)

Anonymized file here.  If full file needed, I'm open to providing it privately.

### Device signature

<details><summary>Device signature</summary>

```json

{
  "node_descriptor": {
    "logical_type": 2,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 128,
    "manufacturer_code": 4447,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 1036,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 1036,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0107",
      "input_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0012",
        "0x0400",
        "0x0402",
        "0x0405",
        "0xfcc0",
        "0xfcf0",
        "0xfcf1"
      ],
      "output_clusters": [
        "0x000a",
[...]

PR #5259, opened 2026-08-17
Add Aqara FP300 presence sensor quirk (lumi.sensor_occupy.agl8)

### LINKED ISSUE #4487: [Device Support Request] Aqara FP300 (lumi.sensor_occupy.agl8)

wiltodelta:
Takes over #4504, which has had no code updates since March 2026 while the device support request #4487 stays open. All quirk code in this branch comes from @cemizm's PR branch; on top of it sits one commit fixing the battery path investigated in #5116. If @cemizm prefers to resume his PR, I will gladly close this one in favor of it.

**The battery defect** (entity toggling 0% to 100% every report cycle, #5116), measured on a live FP300 over the ZHA websocket (`zha/devices/clusters/attributes/value`):

| Source | Attribute | Raw value |
|---|---|---|
| Standard PowerConfiguration (0x0001) | 0x0020 battery_voltage | 3 |
| Standard PowerConfiguration (0x0001) | 0x0021 battery_percentage_remaining | 200 (100%) |
| Aqara TLV 0x00F7, tag 0x17 | battery voltage | 320 |

Three fixes, verified in production on HA 2026.8.2:

1. TLV tag 0x17 arrives in units of 10 mV (320 = 3.2 V), not millivolts. Read as mV it clamps to `MIN_VOLTS_MV` and derives 0%. Now rescaled when below 1000 and sanity checked against a 2.0 to 4.0 V window.
2. `battery_voltage` (0x0020) is defined in units of 100 mV; the quirk wrote volts (`/ 1000`), showing 0.32 V on a full battery. Restored the parent `XiaomiPowerConfiguration` scale (`/ 100`).
3. The device also pushes a stuck-at-full percentage and 3 V through the standard power cluster, fighting the TLV-derived values. Standard battery attribute updates are now dropped; only quirk-derived values enter the cache (same idea as @nokside's quirk linked from #5116).

Tests updated to the corrected expectations plus new cases: a TLV scaling table (306/320/2950/2750) and standard reports not overriding TLV values. All 28 tests in `tests/test_motion_agl8.py` pass with zigpy 1.2.x. The battery entity has been stable across two HA restarts since deployment.

Fixes #4487

Related: #5116


wiltodelta:
Data point on the tag 0x17 scale, since @nokside reports his unit always sends millivolts (see his comment on the fork PR): my FP300 runs firmware `0x0000412a` and sends `320` in tag 0x17 while simultaneously reporting 3.0 V (0x0020) and 100% (0x0021) through the standard power cluster, so on this firmware the tag is in units of 10 mV. The `306` fixture in the original tests has the same shape. The `voltage_mv < 1000` guard in `battery_reported()` keeps both scales working: mV values pass through unchanged, 10 mV values are rescaled, and anything outside a 2.0 to 4.0 V window is dropped.

parcou:
@wiltodelta This question is coming from a newbie.

If fixes are verified in production on HA 2026.8.2, do we still need to use the quirk? Asking because I am new to this. Or does the 2026.8.2 fix work in conjunction with the quirk installed?

Thx

wiltodelta:
@parcou Home Assistant 2026.8.2 does not include this quirk. "Verified in production on HA 2026.8.2" means I ran the file from this PR as a custom quirk on that version.

HA 2026.8.2 ships zha-quirks 2.2.0, and that release has no FP300 file. You still need the custom quirk until this PR merges and a zha-quirks release is picked up by Home Assistant.

If you already dropped motion_agl8.py into custom_quirks_path, keep it. After an official release includes this file, delete the custom copy so the same signature is not registered twice.

parcou:
> @parcou Home Assistant 2026.8.2 does not include this quirk. "Verified in production on HA 2026.8.2" means I ran the file from this PR as a custom quirk on that version.
> 
> HA 2026.8.2 ships zha-quirks 2.2.0, and that release has no FP300 file. You still need the custom quirk until this PR merges and a zha-quirks release is picked up by Home Assistant.
> 
> If you already dropped motion_agl8.py into custom_quirks_path, keep it. After an official release includes this file, delete the custom copy so the same signature is not registered twice.

Thank you for this detailed reply. Yes, motion_agl8.py was working great so far. I will stand down for now. 👍🏽

rich5741:
Some findings from Claude-assisted troubleshooting on https://github.com/zigpy/zha-device-handlers/pull/4504#issuecomment-5574122042

Thanks @wiltodelta — I originally had the oscillating 0 to 100 swing, but in the course of troubleshooting for the issue above, I reset all my devices and switched firmware to thread and back.  That alone fixed the issue, I never implemented the fix you outlined here.  I did notice the devices did not unpair from HA easily - I would remove them from ZHA, but they tended to repair on their own, even after a reset.  I found that I needed to keep the devices awake (via button presses) while unpairing from ZHA.  Then, after a reset I would let it sit on zigbee for a few minutes, make sure it did not repair with ZHA, the proceed with switching the firmware to thread and back to Zigbee.  My hypothesis is that the stickiness means the device never unpairs, and its something in the true un-pair/re-pair procedure that actually fixes the issue/

I'm not a dev, so in case its helpful, here's a Claude generated output about the voltage swing issue:

**Environment:** Aqara FP300 (`lumi.sensor_occupy.agl8`), firmware `0x0000412a` (build 6542), Home Assistant 2026.9.0, ZHA, 4 units, running the #4504 quirk with an added lifeline read purely for logging.

The FP300 intermittently reports `battery_voltage_mV = 0` in its lifeline TLV. Captured on two devices, three times in one evening:

```
20:05:37  [0x2D16]  FP300 lifeline read OK, battery_voltage_mV=0
20:37:12  [0x2D16]  FP300 lifeline read OK, battery_voltage_mV=0
21:01:43  [0x143C]  FP300 lifeline read OK, battery_voltage_mV=0
```

Entity history for the first, with good reads either side:

```
20:05:01  ->  96.0     (2994 mV)
20:05:37  ->   0.0     (0 mV)
20:07:07  ->  91.5     (2987 mV)
```

Without your guard, `_update_battery_percentage(0)` clamps to the minimum-voltage floor and emits 0 %; the next good read snaps it back. **This is the mechanism behind the 0↔100 battery oscillation**, which I originally reported in #5116 without knowing the cause. Every other documented case I could find is *stuck-high* (the 30 V / 0.3 V scaling discussion) — I couldn't find the low-side variant reported anywhere with the raw value captured.

Your `if not 2000 <= voltage_mv <= 4000: return` discards these rather than clamping them, which is the correct handling.

One case it wouldn't cover: another unit returned a lifeline whose TLV parsed to an empty dict (`{}`) — the read succeeded, but no recognised tags came back. It's intermittent; the same device returned a valid payload later the same evening. Might be worth also guarding against a `_parse_aqara_attributes` result with no voltage key.

To be clear, I have **not** deployed this patch — my units aren't currently oscillating, so I'd have no way to validate it. I'm reporting the defect rather than the fix. Happy to supply full logs or diagnostics if useful.

wiltodelta:
@rich5741 Thank you, this is the first published capture of the low-side variant with the raw value. Your logs come from the #4504 quirk, but the parse path is the same base class, so the capture applies here.

The 0 mV case is covered twice in this PR. `battery_reported()` drops any value outside 2000 to 4000 mV after the rescale, and `test_aqara_fp300_battery_tlv_out_of_window` parametrizes `tlv_voltage=0` (alongside 50, 100, 4500) and asserts zero power cluster updates, so the cache keeps the last good reading instead of clamping to 0%.

The empty parse needs no guard, and it is now pinned by a test. The Xiaomi dispatch checks for key membership: `battery_reported()` is only called `if BATTERY_VOLTAGE_MV in attributes`, so a TLV that parses to `{}` never reaches the power cluster and the last good voltage stays cached. Per the base parser loop, a truly empty dict means the payload was empty or a single stray null byte; unknown tags still get `0xff01-N` keys. I added `test_aqara_fp300_battery_tlv_without_voltage_tag_is_noop` covering all three shapes: it seeds a good reading, pushes each payload, and asserts nothing changes.

Your re-pair observation is a useful data point for #5116: a stale pairing can apparently produce the same symptom, while the guard covers it regardless of cause. If you still have the raw payload bytes of the empty parse, I would like to see whether it was an empty report or the stray null byte.

nokside:
Just to clarify the reference to my quirk:

> Standard battery attribute updates are now dropped; only quirk-derived values enter the cache (same idea as @nokside's quirk linked from https://github.com/zigpy/zha-device-handlers/issues/5116).
That is not really what my workaround was doing.

My workaround was only about filtering invalid standard `PowerConfiguration` battery reports. On older firmware, my FP300 periodically reported `0` through the standard `0x0020` / `0x0021` battery attributes, exactly once every three hours. Those invalid standard reports could then replace the battery state that had been derived from the Aqara-specific `0x00F7` data.

I also checked how Aqara handles `0xFCC0 / 0x00F7` on the gateway side. I did not find any extra scale conversion for item `0x17`. The voltage formulas I found on the gateway side also work with millivolt values.

It is also important to consider how ZHA processes these reports, because the raw data still goes through Xiaomi report parsing, FP300 mapping, and then power cluster/cache handling before it becomes the final HA battery value.

So without raw `0x00F7` payload bytes and the full handling path for the `313/320/0` cases, especially since I did not find this conversion in Aqara’s own gateway-side handling, I would not treat `0x00F7` itself as the confirmed source of the scale or zero-value issue.

rich5741:
@TheJulianJES for visibility on this PR. I think he was the ZHA developer working on getting the quirk officially into ZHA

rich5741:
> @rich5741 Thank you, this is the first published capture of the low-side variant with the raw value. Your logs come from the #4504 quirk, but the parse path is the same base class, so the capture applies here.
> 
> The 0 mV case is covered twice in this PR. `battery_reported()` drops any value outside 2000 to 4000 mV after the rescale, and `test_aqara_fp300_battery_tlv_out_of_window` parametrizes `tlv_voltage=0` (alongside 50, 100, 4500) and asserts zero power cluster updates, so the cache keeps the last good reading instead of clamping to 0%.
> 
> The empty parse needs no guard, and it is now pinned by a test. The Xiaomi dispatch checks for key membership: `battery_reported()` is only called `if BATTERY_VOLTAGE_MV in attributes`, so a TLV that parses to `{}` never reaches the power cluster and the last good voltage stays cached. Per the base parser loop, a truly empty dict means the payload was empty or a single stray null byte; unknown tags still get `0xff01-N` keys. I added `test_aqara_fp300_battery_tlv_without_voltage_tag_is_noop` covering all three shapes: it seeds a good reading, pushes each payload, and asserts nothing changes.
> 
> Your re-pair observation is a useful data point for #5116: a stale pairing can apparently produce the same symptom, while the guard covers it regardless of cause. If you still have the raw payload bytes of the empty parse, I would like to see whether it was an empty report or the stray null byte.

Unfortunately I do not, I never captured the raw payload

PR #5294, opened 2026-09-01
Migrate Aqara pet feeder ACN001 to quirks v2 and add onboard scheduling

iamjoshk:
## Proposed change

This PR updates the Aqara Pet Feeder ACN001 to a quirk V2. There are several improvements to the existing quirk.
The quirk v2:

+ Sets the feeder's time to the user's local timezone instead of UTC during the interview and configuration (similar to Z2M).
  + Makes it easier to set the onboard schedule
  + Makes the LED Indicator switch slightly more useful
+ Adds the ability to set the onboard schedule for feedings independent of Home Assistant.
+ Adds a sensor that reports the most recently set onboard schedule.
+ Allows for removing the Aqara Feeder entities in ZHA, for which I will create a separate PR in the ZHA repo.
+ Updates the existing `FeedingSource` class to include `Schedule` and renames `Remote` to `HomeAssistant`.
+ Is compatible with devices that have previously been paired in ZHA.
+ Generates events for attribute reports and updates (enables things like the onboard schedule sensor above).

## Additional information

Updating the onboard schedule can be done using the `zha.set_zigbee_cluster_attribute` like this:
```yaml
data:
  ieee: 54:EF:44:10:11:11:11:11
  cluster_type: in
  endpoint_id: 1
  cluster_id: 64704
  attribute: schedule
  value: >-
    [{"days":"everyday","hour":6,"minute":0,"portions":3},{"days":"everyday","hour":11,"minute":0,"portions":2},{"days":"everyday","hour":20,"minute":0,"portions":2}]
action: zha.set_zigbee_cluster_attribute
```
When the schedule has been set successfully, it will update the `Schedule` sensor. For example:
image

This quirk v2 is also prepared for changes required in ZHA mentioned here:
https://github.com/zigpy/zha/issues/705
PR for ZHA "legacy" quirks: https://github.com/zigpy/zha/pull/883

After removing the entities from ZHA, when a feeding has been provided from the onboard schedule the `Last feeding source` sensor will display `Schedule` instead of `undefined_0x00`:
image

## Checklist


mtwo (approved):

mtwo:
It would be incredible if this could land in ZHA. I currently use an HA automation to trigger the feeder, but I'm always worried about potential flakiness in my Zigbee network, an HA issue, or a power outage causing a missed feeding.

TheJulianJES (commented):
We should ideally try to keep the refactor and new features in separate quirks PRs

zhaquirks/xiaomi/aqara/feeder_acn001.py
```diff
@@ -262,77 +558,124 @@ async def write_attributes(
...
+        fallback_name="Error detected",
+    )
+    .sensor(
+        attribute_name=OppleCluster.AttributeDefs.schedule.name,
+        cluster_id=OppleCluster.cluster_id,
+        entity_type=EntityType.DIAGNOSTIC,
+        unique_id_suffix="64704-schedule",
+        translation_key="schedule",
+        fallback_name="Schedule",
+    )
```
TheJulianJES, line 670: I don't think HA Core allows us to have a sensor with JSON text. We can add that "fake" attribute, so it can be read/written via the cluster dialog (and service/action calls to set attributes) but I don't think we're allowed to expose such a schedule sensor entity as pure JSON.

TheJulianJES: I think in the future we may have a separate "states/valeus" tab on that "Manage Zigbee device" page, somewhat similar to what Z-Wave has, where we can expose certain settings that do not fit nicely into HA entities, but still make them easier to change, compared to just raw ZCL/custom attributes.

iamjoshk:
@TheJulianJES do you want me to open a separate PR for just the v2 migration? That should be relatively easy. That would also enable https://github.com/zigpy/zha/pull/883.

Then I can work on refactoring this separately, even in a new PR to keep it cleaner. There is a lot to address. :sweat_smile:

TheJulianJES:
I think a separate PR makes sense, yes. Thanks for working on this!

iamjoshk:
v2 quirk PR: https://github.com/zigpy/zha-device-handlers/pull/5296

zhaquirks/xiaomi/aqara/feeder_acn001.py
iamjoshk: That would be cool. I can remove this. I just liked the idea of having the schedule available. In the meantime, a trigger-based template sensor helper can be created to parse the event data to track the schedule.

Issue #5260, opened 2026-08-18
[BUG] TS0002 _TZ3000_tonrapsk: both gangs switch together with Switch_2G_GPP, works correctly in Z2M


corrin:
Zemismart ZB811 (KS811-ZB-21-L) 2 gang wall switch, `_TZ3000_tonrapsk` / `TS0002`,
app_version 82.

Turning either gang off in ZHA turns both off. Independent control is impossible from
HA, though the physical buttons on the faceplate work fine. `Switch_2G_GPP` is applied
and its signature matches the device exactly, so this isn't an unsupported device.

I don't have a root cause or a fix, but I do have a clean A/B on the same physical unit,
so here's what I tried.

## What I tried

I have two of these, bought together, same firmware.

The first one went on ZHA driving an outside light and a stairwell light. It couldn't
switch them separately. It also had a worse problem where sending `off` would lock the
MCU (both relays dropped, faceplate LEDs went out, buttons unresponsive, mains cycle at
the fusebox to recover), and a custom quirk didn't help with that. I gave up and replaced
it with a SONOFF ZBM5-3C-120 in the same spot on the same ZHA network, which has been
fine ever since, so it isn't position or mesh.

The second one went on Zigbee2MQTT 2.13.0 (ember adapter, SkyConnect) and worked
correctly for a week.

Today I moved that second unit onto ZHA to check whether the first had just been faulty.
It reproduced the ganging immediately: either gang off turns both off. I moved it
straight back to Z2M and independent control returned, same afternoon, same firmware.

So the ganging follows the stack, not the unit. The MCU lockups I only ever saw on the
first unit and never on the second, so I think that one was faulty on top of this. I'd
ignore the lockups, I've only mentioned them for completeness. The bug I'm reporting is
the ganging.

## Same device, both stacks

|  | ZHA 2.1.0 | Z2M 2.13.0 |
| --- | --- | --- |
| independent gangs | no, either off turns both off | yes |
| power_on_behavior | one entity, device wide, from ep1 | two, `_l1` and `_l2` |
| countdown | not exposed | two, `_l1` and `_l2` |
| switch_type, indicator_mode, backlight | partial | all exposed |

Both coordinators are EFR32MG21 running EmberZNet, so I don't think it's the radio.

## Signature

```json
{
  "manufacturer": "_TZ3000_tonrapsk",
  "model": "TS0002",
  "quirk_applied": true,
[...]

Issue #4102, opened 2025-05-31
[Device Support Request] _TZ3000_c7xsiexw 2 gang switch


ausfas:
### Problem description

I'm trying to override the default ZHA quirk for a Tuya 2-gang switch (_TZ3000_c7xsiexw, model: TS0002), so I can add support for dp_id 105 (backlight control). Here's the issue:

What I want to achieve

Stop the device from being picked up by the default Switch_2G_GPP quirk
Register both switch endpoints properly (dp_id 1, dp_id 2)
Expose dp_id 105 as a select entity in Home Assistant for backlight on/off

### Solution description

What I’ve tried
Created a v2-style quirk using CustomDeviceV2 and TuyaZHADevice — not picked up

Created a v1-style quirk using CustomDevice and TuyaMCUCluster — picked up and working for switches

Successfully controlled the cluster using zha_toolkit.attr_write
```
service: zha_toolkit.attr_write
data:
  ieee: "entity"
  endpoint: 1
  cluster: 0x0006
  attribute: 0x5000
  attr_type: 0x30
  attr_val:   0 # or 1

```

Tried exposing dp_id 105 as a select entity via TuyaDPConverter, but no select entity appears

```
# ts0002_c7xsiexw_manual.py

from zigpy.profiles import zha
from zigpy.quirks import CustomDevice
from zigpy.zcl.clusters.general import Basic, Identify, Groups, Scenes, OnOff, Ota, Time
from zhaquirks.tuya import TuyaZBDevice
from zhaquirks.tuya.mcu import TuyaMCUCluster, TuyaOnOffType, TuyaDPConverter


class BacklightDP(TuyaDPConverter):
    """Handle backlight enum (dp_id 105). 0 = off, 1 = normal"""
    values = {
        0: "off",
        1: "normal",
    }
    reverse_values = {v: k for k, v in values.items()}

    def deserialize(self, value):
        return self.values.get(value, f"unknown({value})")

    def serialize(self, value):
        return self.reverse_values.get(value)


class TS0002_C7XS(CustomDevice):
    signature = {
        "models_info": [("_TZ3000_c7xsiexw", "TS0002")],
        "endpoints": {
            1: {
                "profile_id": zha.PROFILE_ID,
                "device_type": zha.DeviceType.ON_OFF_LIGHT,
                "input_clusters": [
                    Basic.cluster_id,
                    Identify.cluster_id,
                    Groups.cluster_id,
                    Scenes.cluster_id,
[...]

PR #5305, opened 2026-09-07
Add Tuya TS0002/TS0003 relay separation commissioning quirks

dfgentag:
## Proposed change

Add device-specific commissioning quirks for two Tuya wall switches that require a proprietary initialization sequence before their relay endpoints operate independently:

- `_TZ3000_c7xsiexw` / `TS0002` — 2-gang
- `_TZ3000_iol4bl2y` / `TS0003` — 3-gang

Both devices use manufacturer code `0x1141` (4417) and were observed with application version `0x52`.

Without this commissioning sequence, standard ZCL On/Off commands sent to an individual endpoint actuate all relays together. Physical button reports, however, are correctly associated with their respective endpoints.

## Commissioning sequence

Packet captures of Tuya/Zigbee commissioning identified the following sequence:

1. Send a Basic cluster Read Attributes request to APS destination endpoint `0xFF`.
2. Write Basic attribute `0xFFDE` as `uint8` with value `0x0D`.
3. Wait approximately 2.55 seconds.
4. Send Basic cluster-specific command `0xF0` to endpoint 1.

After this sequence, standard ZCL On/Off commands independently control:

- endpoints 1 and 2 on the TS0002
- endpoints 1, 2 and 3 on the TS0003

The quirk is intentionally restricted to the two manufacturer/model combinations that have been physically tested.

### TS0002 — `_TZ3000_c7xsiexw`

Controlled testing showed:

- Default ZHA behavior: both relays actuate together.
- `0xFFDE` write only: relays remain coupled.
- `0xFFDE` + delay + `0xF0`: relays remain coupled.
- Endpoint `0xFF` Basic read + `0xFFDE` + delay + `0xF0`: relays operate independently.

The independent relay configuration also survived Home Assistant restarts and a physical mains power cycle.

### TS0003 — `_TZ3000_iol4bl2y`

The same coupled-relay behavior was reproduced on the 3-gang version.

After re-pairing with the commissioning quirk enabled, endpoints 1, 2 and 3 controlled their corresponding relays independently.

A Zigbee packet capture of this commissioning confirmed the same sequence. The observed interval between the `0xFFDE` write and the `0xF0` command was approximately 2.6 seconds, consistent with the 2.55 second delay plus processing/transmission time.

## Tests

The tests cover both manufacturer/model combinations and verify the exact commissioning requests, payloads, destination endpoints and delay.

Local regression testing:

`39 passed`

CI also passes on the supported Python versions, including pre-commit and coverage checks.

Related to #4102.

Related: #5260


dfgentag:
A note for maintainers regarding the low-level `device.request()` calls in this quirk:

The raw ZCL requests are intentional. I initially attempted to reproduce the required behavior using only the normal Basic cluster operations, but the critical first request must be transmitted to APS destination endpoint `0xFF`.

Packet captures from the Tuya gateway showed the following exact commissioning sequence:

1. Basic Read Attributes request → APS destination endpoint `0xFF`
2. Write Basic attribute `0xFFDE = 0x0D`
3. ~2.55 s delay
4. Basic cluster-specific command `0xF0` → endpoint 1

I tested the individual parts of the sequence on the physical devices. The relays remain coupled when the initial endpoint-`0xFF` request is omitted. Only the complete sequence causes the relay endpoints to operate independently.

The behavior was reproduced and packet-captured on both `_TZ3000_c7xsiexw / TS0002` and `_TZ3000_iol4bl2y / TS0003`.

I used the low-level request path because the standard cluster API normally targets the cluster endpoint and does not appear to provide a way to reproduce this APS destination endpoint `0xFF` request exactly.

I'm happy to implement this through a higher-level zigpy/ZCL API instead if there is a preferred way to send this specific commissioning request.

PR #5308, opened 2026-09-07
Bump zigpy to 2.2.0 and update dependencies

TheJulianJES:
## Proposed change

With this PR, zha-quirks requires zigpy 2.2.0 or later. This is required for:
- https://github.com/zigpy/zha-device-handlers/pull/5307
  it requires: https://github.com/zigpy/zigpy/pull/1885, part of [zigpy 2.2.0](https://github.com/zigpy/zigpy/releases/tag/2.2.0)

## Additional information

Other changes in the zigpy bump do not affect anything.


pyproject.toml
```diff
@@ -11,7 +11,7 @@ authors = [{ name = "David F. Mulcahey", email = "david.mulcahey@icloud.com" }]
 readme = "README.md"
 license = { text = "Apache-2.0" }
 requires-python = ">=3.12"
-dependencies = ["zigpy>=2.0.0", "zha>=2.0.0"]
+dependencies = ["zigpy>=2.2.0", "zha>=2.0.0"]
```
TheJulianJES, line 14: ZHA doesn't need to be bumped as the zigpy requirement also requires ZHA 2.2.2 then anyway (transitively): ZHA itself pins the exact zigpy version.

But if we were to bump the explicit requirement for ZHA, we'd have to change the `starting_version` in ZHA's `pyproject.toml` (for the circular dependency in tests, as we have the issue with dependabot not resolving the correct version from tags there). For now, this is just left alone. This is fine as is.

Issue #3354, opened 2024-09-08
[Device Support Request] TS0004 _TZ3000_imaccztn 4 channel dry contact relay


lpt2007:
### Problem description

First problem:
Device shows as light.
Second problem:
If I switch any of four relays to on, automatically goes almost immediately to off state

### Solution description

Device shows as switch and relay stay on when I turn it on.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

https://github.com/user-attachments/assets/604711a0-4176-4962-ae4b-c26bfe4210f0

</details>

### Device signature

<details><summary>Device signature</summary>

```json
{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4417,
    "maximum_buffer_size": 66,
    "maximum_incoming_transfer_size": 66,
    "server_mask": 10752,
    "maximum_outgoing_transfer_size": 66,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0100",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0006",
        "0xe000",
        "0xe001"
      ],
      "output_clusters": [
        "0x000a",
        "0x0019"
      ]
    },
    "2": {
      "profile_id": "0x0104",
      "device_type": "0x0100",
      "input_clusters": [
        "0x0004",
        "0x0005",
        "0x0006",
        "0xe001"
      ],
      "output_clusters": []
    },
    "3": {
      "profile_id": "0x0104",
      "device_type": "0x0100",
      "input_clusters": [
        "0x0004",
        "0x0005",
        "0x0006",
        "0xe001"
      ],
      "output_clusters": []
    },
    "4": {
      "profile_id": "0x0104",
      "device_type": "0x0100",
      "input_clusters": [
        "0x0004",
        "0x0005",
        "0x0006",
        "0xe001"
      ],
      "output_clusters": []
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0061",
      "input_clusters": [],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "_TZ3000_imaccztn",
  "model": "TS0004",
[...]

PR #5327, opened 2026-09-12
Add Salus SQ610RF thermostat and KL08RF wiring centre quirks

AlexanderYW:
> **⚠️ This is a DRAFT PR — opened to share progress and to ask for help, not to be merged as-is.**
> Everything below is honest about what is verified on hardware and what is not. One feature
> (setpoint writing) is blocked on a capture I cannot make myself: it needs someone with a
> **Salus UGE600 or UGE800 gateway** and a sniffer. 
> **Other contributions are equally welcome**: review, code, or just confirmation on a white-label device.

Disclaimer: The code is heavily AI assisted due to my lack of knowledge in this area.

## Proposed change

Adds v2 `QuirkBuilder` quirks for two devices in the Salus / Computime **iT600RF** heating
family (manufacturer code `0x1078`), both of which are currently **unusable on ZHA** - not
degraded, unusable:

| Device | Also sold as | Behaviour without a quirk |
|---|---|---|
| **SQ610RF / SQ610RFNH** Quantum thermostat | - | joins, then re-associates forever (52 attempts observed); LED never stops blinking |
| **KL08RF** 8-zone wiring centre | Cosmo **CFK8L / CFKL8** | joins, then leaves within seconds, in a loop |

Neither device implements standard `hvacThermostat` (`0x0201`) in any useful way. The real
protocol is manufacturer clusters `0xFC00` / `0xFC09`, reverse-engineered here from decrypted
over-the-air captures of a **native Salus installation** (CO10RF coordinator + KL08RF + six
SQ610RF thermostats), captured alongside the same devices failing on ZHA, including a full
network teardown and rebuild.

### What the quirks do

**SQ610RF / SQ610RFNH**

- **Commissioning fix.** The device sends `fc00` command `0x10` and waits for a
  manufacturer-specific application reply. Stock ZHA only APS-acks, so the device re-associates
  in a loop. The quirk answers with `0x81`, exactly as a native controller does, and it
  commissions cleanly.
- **Identity gate.** After commissioning it reads Basic `ModelIdentifier` from what it believes
  is its controller and leaves unless the answer is a Salus controller string. The quirk answers
  `SAL6DC1`.
- **OTA gate.** It sends one OTA `Query Next Image Request` and stalls on ZHA's real OTA-server
  answer; a native controller replies with a plain ZCL Default Response / Invalid Field. The
  quirk does the same.
- **Entities.** Temperature, Setpoint (read-only) and Heating, parsed from the `fc00` `0x12`
  operational report, plus device-clock diagnostics and a "set clock to HA time" button.

**KL08RF**

- **Join fix**, verified over 3+ hours with zero NWK Leaves. The exact failure in both existing
  public reports of this device.
- **Correct node-slot assignment** (see below). It builds no entities: the wiring centre
  transmits nothing operational and drives its zone relays locally from the demand bit in the
  reports it receives. It is on the network as a router and as the bind target for thermostats,
  which is all it can be. The thermostat quirk points zone binds at the real hardware when
  one is present.

### The core finding: `fc00 0x81` byte 0 is a node **slot**

Documenting this here because it is not written down anywhere else and it will bite the next
person:

```
device -> controller   0x10   [contact] ...       contact 0xff = first contact
controller -> device   0x81   [SLOT] 00 [counter x2 LE] [session x2] 03 0d 01 ff ff ff ff 30 00
                              ^^^^^^
```

Byte 0 is the slot the controller **assigns**. Wiring centres take `0x01`–`0x09` (the control-box
number, that a thermostat's on-screen `01..09` picker selects and what the KL08RF shows on its
zone LEDs); thermostats take `0x2b`+ in pairing order. Proven by watching a factory-reset
coordinator rebuild a whole network from zero: the wiring centre took `01` and lit exactly one
zone LED at zone 1, six thermostats took `2b 2c 2d 2e 2f 30`, and all eleven subsequent zone
binds were accepted (`0x88` byte 0 = `00`).

Give a wiring centre a thermostat's slot and it joins, stays, looks perfectly healthy and
silently refuses every zone bind with `0x88 01ff…`. That is what the slot constant in the KL08RF
quirk exists to prevent.

### What's missing and how you can help

**Writing a setpoint does not work, and I could not find a way to make it work.** The thermostat
is the setpoint *master*: in a native Salus system nothing ever pushes a setpoint down over the
air, so the command was never captured. Three rounds of probing came back empty:

- standard Thermostat cluster `0x0201` is a **stub** `OccupiedHeatingSetpoint` write →
  Unsupported Attribute, `SetpointRaiseLower` → Unsupported Cluster
- `fc00` attribute writes `0x02`–`0x07`, `0x10`–`0x12`, `0x20` → all Unsupported
- `fc00` commands `0x02 / 0x03 / 0x04 / 0x08 / 0x0c / 0x12 / 0x13`, correctly framed as
  Server→Client (on-air ZCL FCF `0x1d`, mfr `0x1078`) → all silently ignored, setpoint never moved

Remote control demonstrably exists and the **UGE600** gateway does it from the it600 app, but that path is the gateway's **local HTTP API**, off Zigbee entirely. What nobody has captured is **what the gateway puts on the air** when a setpoint changes.

#### 🙏 If you own a UGE600 or UGE800, this one capture finishes the feature

Any 802.15.4 sniffer works (CC2531, nRF52840, or a Sonoff ZBDongle-E flashed with sniffer
firmware) plus Wireshark:

1. **Find your Salus channel.** Sniff 11–26 briefly and look for frames from OUI `00:1e:5e:09`.
   Don't assume a gateway picks its channel at first setup.
2. **Get the network key.** Put any Salus device into pairing mode while sniffing. The join sends
   a Transport Key encrypted with the default TC link key `ZigBeeAlliance09`, so Wireshark derives
   the network key on its own, just make sure
   `"5A:69:67:42:65:65:41:6C:6C:69:61:6E:63:65:30:39","Normal","default TC link"` is in your
   `zigbee_pc_keys` UAT.
3. **Capture while you change a setpoint** from the it600 app. Leave a minute either side.
4. **Find the frame.** Filter `zbee_aps.cluster==0xfc00` and look for **gateway → thermostat** at
   the moment of the change. What I need is its manufacturer-specific command id
   (`zbee_zcl.cs.cmd.id`) and its payload (`data.data`) or just drop the pcap here.

That command id and payload are the entire missing piece; with them the read-only Setpoint sensor
becomes a real climate entity.

**Also very welcome, if you have the hardware:** confirmation on a **Cosmo CFK8L / CFKL8**, or on a
plain **SQ610RF** (non-NH) both should match unchanged, neither has been confirmed by anyone.

**And ordinary contributions are welcome too**, review comments, restructuring to match how the
maintainers want these files laid out, test coverage, or picking up any of the open items above.
I'd rather this land as a shared thing than as my thing.

⚠️ **If you go probing yourself:** `fc00` command `0x40` is the thermostat's reset-and-leave. It
will kick the device off your network. The whole high band (`0x20`, `0x80`, `0xe0`) is untested and neighbours it don't sweep blindly.

## Additional information

- **Not a breaking change**, new quirk modules only, no changes to existing quirks or shared code.
  I've left them at the top level for now; happy to move them into whatever vendor directory the
  maintainers prefer (these report manufacturer `SALUS` with manufacturer code `0x1078`/Computime,
  so either home is arguable).
- **Known-unverified:** the KL08RF slot-assignment path is implemented and byte-verified offline
  against captured native behaviour, but has not been exercised end-to-end on ZHA, because forcing
  it requires a factory reset of the wiring centre, which takes a whole heating system offline.
  Flagging it rather than quietly shipping it.
- **Explicitly out of scope, because the hardware doesn't do it:** HVAC mode and schedules are
  100% local to the thermostat, never transmitted, not readable, not writable; the `fc09`
  operating-mode attribute only ever reports one value in practice. Battery is not exposed over
  Zigbee at all, there is no PowerConfiguration cluster and the 18-byte report is fully accounted
  for without one. Reports arrive on a fixed ~615 s heartbeat, plus immediately on a setpoint
  change at the device.
- **Implementation notes for reviewers** (each of these cost real debugging time):
  - Salus points its two device classes in **opposite ZCL directions**, the thermostat sends FCF
    `0x1d` (Server→Client), the wiring centre `0x15` (Client→Server). zigpy picks its command table
    from that bit, and an id missing from the table it picks comes back as raw bytes with no
    `.data`, silently. Hence mfr commands declared on **both** `ClientCommandDefs` and
    `ServerCommandDefs`, with extraction through a `payload_bytes()` helper.
  - Replies must be Server→Client with default-response disabled (on-air FCF `0x1d`). zigpy's
    high-level `reply()` can't produce that from a server cluster, so those frames are built by
    hand. ZHA's action services can't do it either, which voided two earlier rounds of probing.
  - Attribute-read overrides must return zigpy types (`t.CharacterString`, not `str`); on zigpy 2.x
    a bare `str` throws during serialization and the only symptom is the device quietly leaving.
- **Developed against** zha 2.2.1 / zha-quirks 2.2.1 / zigpy 2.1.0.
- **Prior art**, both of which document the KL08RF failure without a fix:
  [Koenkk/zigbee-herdsman-converters#3354](https://github.com/Koenkk/zigbee-herdsman-converters/issues/3354),
  [Koenkk/zigbee2mqtt#15208](https://github.com/Koenkk/zigbee2mqtt/issues/15208).

## Device diagnostics

- **SQ610RFNH** attaching diagnostics from the device running under the quirk, plus the
  pre-quirk capture where it interviews as `unk_manufacturer unk_model`, since the difference is
  the point.
- **KL08RF** diagnostics to follow; the device is in a live heating system and I want to pull a
  clean download rather than a mid-join one.

## Checklist

**On the boxes, since this is a draft:**

- *Tested:* SQ610RF is verified on real hardware (commissioning, identity gate, OTA gate, entity
  population). KL08RF join is verified on real hardware. The KL08RF slot-assignment path is **not**
  hardware-tested see Additional information.
- *pre-commit / Black:* not yet run against this repo's config; will do before taking the PR out of
  draft.
- *Tests:* there is a standalone validation script with 51 assertions covering device matching,
  cluster replacement, slot assignment, the `0x12` report parser and the exact on-air bytes of
  every reply each one corresponding to something that actually broke or that a capture
  confirmed. It is **not** yet ported to this repo's pytest suite; I'd like guidance on how much of
  it maintainers want in-tree.
- *Diagnostics:* SQ610RFNH attached, KL08RF pending as noted above.

Devices i'm aware of that these quirks would benefit
| Device | Tested |
|--------|--------|
| KL08RF(iT600RF) | [x] |
| CFK8L | [] - white-label of KL08RF |
| SQ610RF | [] |
| SQ610RFNH | [x] - white-label of SQ610RF |

- NH in SQ610RFNH is neotherm

References: [#1524](https://github.com/Koenkk/zigbee-herdsman-converters/issues/1524), [#3354](https://github.com/Koenkk/zigbee-herdsman-converters/issues/3354), [#15208](https://github.com/Koenkk/zigbee2mqtt/issues/15208)

### DIFF
--- a/zhaquirks/salus/kl08rf.py
+++ b/zhaquirks/salus/kl08rf.py
@@ -163,10 +163,7 @@ from typing import Final
 import zigpy.types as t
 from zigpy.zcl import foundation
 from zigpy.zcl.clusters.general import Basic, Ota
-from zigpy.zcl.foundation import (
-    BaseCommandDefs,
-    ZCLCommandDef,
-)
+from zigpy.zcl.foundation import BaseCommandDefs, ZCLCommandDef
 
 from zhaquirks.builder import QuirkBuilder
 from zhaquirks.clusters import CustomCluster
--- a/zhaquirks/salus/kl08rf.py
+++ b/zhaquirks/salus/kl08rf.py
@@ -1,0 +1,529 @@
+"""ZHA quirk for the Salus KL08RF 8-zone wiring centre (Computime, mfr 0x1078).
+
+Built with the zha-device-handlers **v2 QuirkBuilder** pattern, as a companion
+to `salus_sq610rf.py` (the SQ610RFNH thermostat quirk, confirmed working on
+ZHA 2026-07-04).
+
+STATUS: VERIFIED ON HARDWARE 2026-09-10/11. The KL08RF joins ZHA and stays --
+3+ hours, zero NWK Leaves -- which is the failure both z2m reports hit. Its
+interview matched the predicted signature exactly. The one thing that did NOT
+work, and why, is decoded under THE SLOT ASSIGNMENT below; the fix is in this
+file and is the single most important thing to understand about it.
+
+What the native captures establish (facts)
+------------------------------------------
+
+* The KL08RF (native short 0xb15f, IEEE 00:1e:5e:09:09:03:dc:67, Computime
+  OUI) is a mains-powered ZIGBEE ROUTER -- one of only two NWK Link Status
+  originators on the native net (the other is the CO10RF coordinator).
+* Its wiring-centre service lives on **endpoint 8**: thermostats address
+  their fc00 0x12 operational reports and the 0x18/0x14 zone-bind requests to
+  wc:ep8, and the KL08RF answers 0x88/0x84 from ep8.
+* It runs its OWN gateway-facing fc00 session: a 0x10 -> 0x81 handshake
+  followed by command **0x37** (payload `110000`), once per ~90 min -- the
+  same Computime commissioning/keep-alive grammar the SQ610RF uses
+  (reply = 0x80 | low-nibble(request), ZCL FCF 0x1d, mfr 0x1078).
+* The 0x81 reply body is `[slot] 00 [counter x2 LE] [session const x2] 03 0d
+  01 ff ff ff ff 30 00`. Byte 0 is the device's ASSIGNED SLOT, not a session
+  token -- see THE SLOT ASSIGNMENT below. Bytes 2-5 are a counter plus a
+  per-session constant and are NOT validated by devices.
+* Heat demand / relays are HARDWIRED, never on-air: driving a zone to 28 C
+  produced NO Zigbee demand command from the KL08RF (heat-demand capture).
+  The KL08RF switches its zone relays from the demand bit it receives in the
+  thermostats' 0x12 reports; it transmits nothing operational upstream.
+
+=> There is NOTHING operational to read from the KL08RF over Zigbee, so this
+quirk builds NO entities (ZHA still provides LQI/RSSI). Its whole job is:
+keep the device joined by answering the Computime handshake, and LOG every
+proprietary frame it sends for future decoding.
+
+Confirmed by two independent ZHA-family interviews (z2m #3354 + #15208)
+----------------------------------------------------------------------
+This exact device shows up in two zigbee2mqtt reports -- as the Salus KL08RF
+(Koenkk/zigbee-herdsman-converters#3354, 2021) and as its white-label the
+Cosmo CFK8L / "CFKL8" 8-Zonen-Funk-Regelklemmleiste (Koenkk/zigbee2mqtt#15208,
+2022). Both are DIFFERENT physical units than ours; the same reporter linked
+them and called the CFK8L "a whitelabel of the Salus KL08RF". They confirm the
+failure mode this quirk fixes and pin down the join signature.
+
+WHITE-LABELS: the same hardware is sold as at least Salus KL08RF and Cosmo
+CFK8L (CFKL8). All report identically over Zigbee (see below), so this one
+quirk covers every rebrand -- there is no model string to tell them apart
+anyway.
+
+Full join signature (authoritative -- from the #15208 herdsman DB entry, whose
+interview completed further than #3354's):
+* node type ROUTER (mains); manufId 4216 = 0x1078 (Computime);
+  powerSource "Unknown"; zclVersion 0; hwVersion 1 (#3354 saw hwVersion 197 --
+  a firmware/label difference the quirk does not depend on).
+* Basic manufacturerName "Computime" (NOT "SALUS") and NO ModelIdentifier
+  (both reports log the model as 'undefined'). Our own SQ610RF thermostats
+  report "SALUS"; this device's firmware still reports the raw Computime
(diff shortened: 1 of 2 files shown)


hmlendea:
Does this work in a system that also uses an RX10RF?

AlexanderYW:
It doesn't work yet. I need more sniffed logs to finish it, I just shared the progress hoping someone would provide either the remaining features sniffed or contribute them as code

PR #5344, opened 2026-09-15
Align attribute method overrides with new zigpy typed responses

puddly:
## Proposed change

A few quirks modify low-level attribute reading/writing/configuring methods. These aren't _technically_ part of the quirks API and can change, as we are doing with https://github.com/zigpy/zigpy/pull/1778. This PR updates them to use the new zigpy response types.

## Checklist


puddly:
As a test, I've migrated the Eurotronic thermostat to use direct entity replacement instead of conforming to a fake ZCL: https://github.com/puddly/zha-device-handlers/blob/755cd1737f028c6b20f986e8fe657faa6b7974c8/zhaquirks/eurotronic/__init__.py.

It looks pretty okay to me.

PR #4881, opened 2026-03-30
Improve Nimly lock operation event decoding

fredrik-lindseth:
## Proposed change

This updates decoding of the Onesti/Nimly last-operation event (Door Lock attribute `0x0100`).

* Add source value `0x05` as `unattributed`. It has been reported on NimlyCodePRO and NimlyPRO24 devices.
* Add NimlyShared to the quirk match list.
* Replace the current hex-string slicing with byte shifts and masks. The field extraction is equivalent; the new form makes the bitmap layout explicit.
* Use `unknown` for source and action bytes that do not have a mapping.
* Cover the known mappings and fallback values with parametrized tests.

The event is a 32-bit value with the user slot in bits 0–15, action in bits 16–23 and source in bits 24–31.

## Scope and open question

This PR changes how known values are decoded. It does not address the separate report that operation events are not delivered to the entities in some installations (#5235).

The branch currently renames source `0x0A` from `self` to `auto`, based on local hardware testing. My attached diagnostic also contains `0x0A` with an unlock action, and Zigbee2MQTT uses a different label. I am happy to retain `self` until that value has been confirmed more broadly, since changing the exposed state may affect automations.

## Device diagnostics

Nimly Touch Pro (NimlyPRO) diagnostics are attached in a comment below. The `0x05` reports are linked above.

## Checklist

* [x] The changes are tested and work correctly
* [x] `pre-commit` checks pass / the code has been formatted using Black
* [x] Tests cover known mappings and unknown values

PR #5345, opened 2026-09-16
Add support for NimlyCodePRO in QuirkBuilder

vinnyspb:
## Proposed change
Nimly Code Pro lock support added.

## Additional information
N/A

## Device diagnostics
N/A

## Checklist


fredrik-lindseth:
Related: #4881 also adds NimlyCodePRO (and NimlyShared) to this model list, plus the 0x05 source decoding a Code Pro needs before the action sensors say anything sensible for Zigbee commands, auto-relock and the interior keypad (it reports all three as 0x05 with user 0). Whichever PR lands second gets a small conflict in the model list, so they may be worth reviewing together. And if you get a chance to run the #4881 branch on your Code Pro, a second device confirming the decoding would be welcome.

Issue #5038, opened 2026-05-27
[Device Support Request] Aqara Dimmer Switch H2 EU (decoupled mode - lumi.switch.agl011)


FreddieDev:
### Problem description

<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/94b5ae56-6b69-4da2-9cd3-86db48ae5bb1" />

While the lumi.switch.agl011 does appear to be supported, the decoupled mode cannot be changed and it's not firing events for all 3 buttons.

PRESS - Event fires okay
DOWN - No events fired
UP - No events fired

There's also buttons for multi-press and lock relay which report errors that the attributes can't be set when I try to use them.

I have switched to Zigbee2MQTT for now, and that seems to offer full support.

### Solution description

Full decoupled mode events logging in HA so I can link them to a smart bulb.

### Diagnostics information

Not accessible at this time. Please let me know if this is also required given I have attached the full signature

### Device signature

<details><summary>Device signature</summary>

I extracted this using the ZHA Toolkit. 

```json

{
  "ieee": "54:ef:44:10:01:1c:76:c1",
  "nwk": "0xfced",
  "model": "lumi.switch.agl011",
  "manufacturer": "Aqara",
  "manufacturer_id": "0x4447",
  "endpoints": [
    {
      "id": 21,
      "device_type": "0x0000",
      "profile": "0x0104",
      "in_clusters": {
        "0x000c": {
          "cluster_id": "0x000c",
          "title": "PowerMeasurementCluster",
          "name": "analog_input",
          "attributes": {
            "0x0051": {
              "attribute_id": "0x0051",
              "attribute_name": "out_of_service",
              "value_type": [
                "0x10",
                "Bool",
                "Discrete"
              ],
              "access": "READ|REPORT",
              "access_acl": 5,
              "attribute_value": 0
            },
            "0x0055": {
              "attribute_id": "0x0055",
              "attribute_name": "present_value",
              "value_type": [
                "0x39",
                "Single",
                "Analog"
              ],
              "access": "READ|REPORT",
              "access_acl": 5,
              "attribute_value": 0.0
            },
            "0x006f": {
              "attribute_id": "0x006f",
              "attribute_name": "status_flags",
              "value_type": [
                "0x18",
                "bitmap8",
                "Discrete"
              ],
              "access": "READ|REPORT",
              "access_acl": 5,
              "attribute_value": 0
            }
          },
          "commands_received": {},
          "commands_generated": {}
[...]

Issue #4346, opened 2025-09-15
[Device Support Request] Aqara Dimmer Switch H2 EU -- support some kind of sensor to determine which way the knob is being turned


pizzaisdavid:
### Problem description

I want to be able to program an automation that when I am turning the knob in one direction, that the brightness of my smart light goes up, and the other way, the brightness goes down.

### Solution description

For example, something like "is being turned left", "is being turned right". Or "is brightness being changed".

https://www.zigbee2mqtt.io/devices/KD-R01D.html

> brightness: To control the brightness publish a message to topic zigbee2mqtt/FRIENDLY_NAME/set with payload {"brightness": VALUE} where VALUE is a number between 0 and 254. To read the brightness send a message to zigbee2mqtt/FRIENDLY_NAME/get with payload {"brightness": ""}.

### Screenshots/Video

<details><summary>Screenshots/Video</summary>

[Paste/upload your media here]

</details>

### Diagnostics information

No response

### Device signature

<details><summary>Device signature</summary>

```json

[Paste the device signature here]

```

</details>

### Logs

<details><summary>Logs</summary>

```python

[Paste the logs here]

```

</details>

### Custom quirk

<details><summary>Custom quirk</summary>

```python

[Paste your custom quirk here]

```

</details>

pizzaisdavid:
Possible alternative, from https://www.zigbee2mqtt.io/devices/KD-R01D.html

> Action (enum)
> 
> Triggered action (e.g. a button click). Value can be found in the published state on the action property. It's not possible to read (/get) or write (/set) this value. The possible values are: start_rotating, rotation, stop_rotating, hold, single, double, release.

If using start_rotating, then I would need someway to get the current angle / brightness.

I can also imagine that "start_rotating" would not constantly update as I am continuously rotating a knob.

PR #5162, opened 2026-07-09
Add multi-click mode and button press events to Aqara Dimmer Switch H2 EU

DidntRead:
…2 EU (lumi.switch.agl011)

## Proposed change
Add support for multi click mode on Aqara Dimmer H2 EU allowing custom automations.

## Additional information
Copied from Z2MQTT and tested on my own H2 Dimmer running latest firmware(4120)

## Device diagnostics
[zha-01KR51KGAPFMXGWB2ZB520KHD1-Aqara lumi.switch.agl011-726c7dc73ae5d33d205260beec7711a2.json](https://github.com/user-attachments/files/29860381/zha-01KR51KGAPFMXGWB2ZB520KHD1-Aqara.lumi.switch.agl011-726c7dc73ae5d33d205260beec7711a2.json)

## Checklist
- [x] The changes are tested and work correctly
- [x] `pre-commit` checks pass / the code has been formatted using Black
- [x] Tests have been added to verify that the new code works
- [x] Device diagnostics data has been attached

Related: #5038

PR #5346, opened 2026-09-16
Add knob events and voltage fix to Aqara Dimmer Switch H2 EU quirk

exi:
## Proposed change

The lumi.switch.agl011 quirk only exposed configuration attributes, so the knob produced no events in ZHA and could not be used as a controller in decoupled mode.

Knob presses:
- Replace MultistateInput on endpoint 1 to turn present_value reports into zha_events: 0=hold, 1=single, 2=double, 255=release.
- Add device automation triggers for short/double/long press and release.
- Expose the multi_click attribute (0x0286) as a select. It must be set to Multi for hold, double and release to be reported.

Knob rotation:
- Rotation is reported on the manufacturer cluster (0xFCC0) of endpoint 71, not endpoint 1. Add that endpoint and a cluster that parses the reports.
- 0x023A carries the action (1=start_rotating, 2=rotation, 3=stop_rotating), with bit 0x80 set while the knob is held. It is sent last in the report frame, so the angle (0x022E), angle speed (0x0230), time (0x0231), percent speed (0x0232) and percent (0x0233) attributes are buffered and sent as arguments of the zha_event fired on the action.
- Add device automation triggers for the three rotation actions.

Mapping is based on the Zigbee2MQTT lumi converter and checked against captured frames from a real device.

Other changes:
- The device reports rms_voltage in 0.1 V, which showed up as ~2325 V in HA. Force ac_voltage_divisor to 10.
- Limit the rotation sensitivity number (0x0234) to 180-720 degrees and show it as a slider. The value is the knob rotation that maps to 0-100 %: on a real device, 270 gave percent = angle / 2.7 and 720 gave percent = angle / 7.2. Zigbee2MQTT only offers 180 (high), 360 (medium) and 720 (low), but values in between work as well. The entity stays a number with the same unique_id, so existing entities are kept.

## Device Diagnostics

[zha-01J7Y2HJHV0A3N7BS8N5HDE7GV-Aqara lumi.switch.agl011-1636ea08686b41cbfc6bc9247e0508d5.json](https://github.com/user-attachments/files/32310141/zha-01J7Y2HJHV0A3N7BS8N5HDE7GV-Aqara.lumi.switch.agl011-1636ea08686b41cbfc6bc9247e0508d5.json)

## Checklist

- [X] Device diagnostics data has been attached

Related: #5038
Related: #4346


exi:
@zigpy-review-bot you did not leave any comments, what changes are needed here?

TheJulianJES:
It's more of a label for us to triage all PRs but I'll have the internal review posted here.

exi:
Thanks for the review. I pushed two commits: b618378 (voltage) and f12e8b9 (the other points). Filtered ZHA logs for this device, from before and after the voltage change, are attached in the next comment.

## Voltage

The divisor doesn't scale anything twice on this model: its Xiaomi attribute report sends voltage in **0.01 V**, not 0.1 V. The attribute report is also the only source, because the device rejects direct `rms_voltage` reads.

Direct read of `rms_voltage` (1285), answered with `UNSUPPORTED_ATTRIBUTE` (`05 05 86`):

```
[0xD620:1:0x0b04] Sending request: Read_Attributes(attribute_ids=[1285, 2309, 2565, 768])
[0xD620:1:0x0b04] Received ZCL frame: '18 08 01 05 05 86 05 09 86 05 0a 86 00 03 86'
```

Attribute report from this device. Sub-attribute 150 is `23245.7`, and the shared `* 0.1` handler stores `2324.57`:

```
54:ef:44:10:01:24:f8:25 - Xiaomi attribute report. attribute_id: [247] value: [{... 'power': 0.24606062471866608, 'voltage': 23245.705078125, ...}]
AttributeUpdatedEvent(... device_ieee='54:ef:44:10:01:24:f8:25', ... attribute_name='rms_voltage', ... value=2324.5705078125)
```

Before this PR, HA therefore showed ~2325 V. With `ac_voltage_divisor = 10` it shows 232.46 V. For comparison, another Aqara plug on the same network reports in 0.1 V, which is why I didn't change the shared handler:

```
54:ef:44:10:00:4e:c2:53 - Xiaomi attribute report. ... 'voltage': 2280.0 ...
AttributeUpdatedEvent(... device_ieee='54:ef:44:10:00:4e:c2:53', ... attribute_name='rms_voltage', ... value=228.0)
```

While testing, I found a second problem. zigpy stores the `UNSUPPORTED_ATTRIBUTE` status from the direct read, and ZHA doesn't create a sensor for an unsupported attribute. So after a restart, the voltage sensor sometimes disappeared ("This entity is no longer being provided by the zha integration"), depending on whether a failed read or a report value was stored last. b618378 bases the cluster on the Xiaomi `LocalDataCluster` variant (`zhaquirks.xiaomi.ElectricalMeasurementCluster`), so reads are answered locally and voltage is never marked unsupported. After the change, no request is sent to the device and the sensor came back right away:

```
[0xD620:1:0x0b04] reading attributes for LocalDataCluster: attributes=[1537, 1285, 2309, 2565, 768] manufacturer=None
AttributeReadEvent(... device_ieee='54:ef:44:10:01:24:f8:25', ... attribute_name='rms_voltage', ... raw_value=2343.622265625, value=2343)
```

## Other points (f12e8b9)

- **Stale rotation buffer:** the buffer is now cleared before the action lookup, and there is a regression test for an unknown action followed by a valid one.
- **Rotation names:** switched to `started_rotating` / `continued_rotating` / `stopped_rotating` with `rotary_knob` as the subtype.
- **Button constants and payload:** events use the `COMMAND_*` constants, and the payload now has `endpoint_id`, `press_type`, `attr_id` and `value`, like the other Xiaomi buttons.
- **Rotation attributes:** the six attributes are now declared on `RotationCluster` with their types and `manufacturer_code=0x115F`. A new test sends an encoded manufacturer-specific report frame through `packet_received`, so decoding is covered as well.
- **Button endpoints:** `MultistateInput` is now replaced with `.replace_cluster_occurrences(..., replace_client_instances=False)`, so the extra button endpoints in the #5038 signature are covered, and no cluster is added where the device doesn't have one. There's a test for that.
- **Reusing `opple_remote.MultistateInputCluster`:** I kept a separate cluster. The shared one would rename the events to `1_single` etc. It also has no-op `configure_reporting`, which I'd rather not bring in without testing it on the device.
- **`is_manufacturer_specific`:** I left it on the existing `OppleCluster` attributes to keep the diff small. The new rotation attributes use `manufacturer_code`.
- **Rotation sensitivity range:** the value is the knob rotation in degrees that covers 0–100 %. On the device, 270 gave `percent = angle / 2.7` and 720 gave `percent = angle / 7.2`, so values between the Z2M presets work. I kept 180–720 because that range is what has been tested.
- **`.adds_endpoint(71)`:** agreed. It's kept for the test fixture.
- **Related issues:** this also answers #4346, since the sign of `action_rotation_angle` gives the turn direction.
- **#5162:** a maintainer can decide which PR to take. I'm happy to fold anything from #5162 into this one.

exi:
ZHA event log before and after the voltage fix:


[lumi.switch.agl011-zha-log-before-voltage-fix.log](https://github.com/user-attachments/files/32339692/lumi.switch.agl011-zha-log-before-voltage-fix.log)

[lumi.switch.agl011-zha-log-after-voltage-fix.log](https://github.com/user-attachments/files/32339691/lumi.switch.agl011-zha-log-after-voltage-fix.log)

exi:
Pushed 079a6ef after capturing knob gestures with ZHA debug logging on my device (firmware `current_file_version=4120`, i.e. 0x1018). Two findings that the attached diagnostics don't show, filtered logs for both captures are in the next comment.

### Rotation with the knob held down is reported on endpoint 72

Turning while pressing the knob produces no `MultistateInput` report at all — the device swallows the press — and the rotation frames arrive on **endpoint 72** instead of 71, with the documented 0x80 pressed bit never set. Turning right while pressed:

```
08:04:10.872 [0xD620:72:0xfcc0] Report_Attributes(... 0x022E=0.0,    0x0231=0,   0x0233=0.0,  0x023A=0)
08:04:11.211 [0xD620:72:0xfcc0] Report_Attributes(... 0x022E=144.0,  0x0231=100, 0x0233=40.0, 0x023A=2)
08:04:11.747 [0xD620:72:0xfcc0] Report_Attributes(... 0x022E=144.0,  0x0231=100, 0x0233=40.0, 0x023A=3)
```

The same gesture turning left gives 0x022E=-48.0 then -156.0, again on endpoint 72. A plain turn, in the second capture, uses endpoint 71:

```
08:12:24.905 [0xD620:71:0xfcc0] RotationCluster:Report_Attributes(... 0x022E=0.0,   0x023A=0)
08:12:25.373 [0xD620:71:0xfcc0] RotationCluster:Report_Attributes(... 0x022E=72.0,  0x023A=2)
08:12:26.088 [0xD620:71:0xfcc0] RotationCluster:Report_Attributes(... 0x022E=108.0, 0x023A=3)
```

So the endpoint carries the pressed state on this firmware. The quirk now adds endpoint 72 with a `PressedRotationCluster` that emits the same events with `action_rotation_button_state: pressed`, and still honours the 0x80 bit if another firmware sets it.

### Rotation starts with action 0, not 1

Both captures start every rotation with `0x023A=0`, in a frame whose values are all zero, followed by 2 while turning and 3 at the end. Z2M's `lumiKnobRotation` documents 1 for `start_rotating` and I never see it, so without mapping 0 there is no `started_rotating` event at all. Both values now map to `started_rotating`.

As a sanity check on the angle/percent relation: 144° gave 40.0 % with rotation sensitivity at 360 (144/360), and 72° gave 20.0 %.

Note that endpoint 72 is not in the diagnostics attached here, nor in the dump on #5162, so it may well be firmware-specific — worth keeping in mind if someone tests a different firmware. Both endpoints are added by the quirk, so a device that only has 71 is unaffected.

exi:
Captures for a normal turn of the know and turns while know was pressed.

[lumi.switch.agl011-zha-log-plain-turns.log](https://github.com/user-attachments/files/32368200/lumi.switch.agl011-zha-log-plain-turns.log)

[lumi.switch.agl011-zha-log-pressed-turns.log](https://github.com/user-attachments/files/32368204/lumi.switch.agl011-zha-log-pressed-turns.log)

exi:
New Diagnostics with fix applied:

[Aqara lumi.switch.agl011-1636ea08686b41cbfc6bc9247e0508d5.json](https://github.com/user-attachments/files/32369533/Aqara.lumi.switch.agl011-1636ea08686b41cbfc6bc9247e0508d5.json)

Correction to my previous comment: endpoint 72 **is** present in the diagnostics
already attached to this PR — I checked the dump itself rather than trusting the
endpoint list in the review above. Its `original_signature` has both 71 and 72
with cluster `0xfcc0`, and so does the new dump attached here. What was new in my
packet capture is only what endpoint 72 is *used for* (rotation while the knob is
held), not the endpoint itself. I can't check the dump on #5162, but since 71 and
72 appear together in both of mine, they're probably both there as well.

The new dump is taken with 079a6ef applied and shows:

- **Endpoint 71** — `opple_cluster` with the rotation attributes decoded by name:
  `rotation_action: 3`, `rotation_angle: 60.0`, `rotation_percent: 16.67`,
  `rotation_time: 100`.
- **Endpoint 72** — same cluster and attribute names, from a turn with the knob
  held: `rotation_action: 3`, `rotation_angle: -240.0`, `rotation_percent: -66.67`.
  In the older dump this endpoint is plain `manufacturer_specific` with no
  attributes, since the quirk didn't cover it.
- **Voltage** — `rms_voltage` is now a normal cached value instead of
  `unsupported: true`, and the voltage sensor has a value again. In the older dump
  the attribute is `unsupported` and `ElectricalMeasurementRMSVoltage` has
  `native_value: null`, which is the disappearing sensor described above.
  `ac_voltage_divisor` is 10 and `ac_power_divisor` 10 from the local cluster.
- **Rotation sensitivity** — now `rotation_sensitivity`, shown as a slider in
  degrees (180–720), reading 360 here.

Both dumps are from firmware `0x1018` (`current_file_version: 4120`).

exi:
@TheJulianJES I hope it's ok that I extended the scope a bit. I discovered the meaning of endpoint 72 in the meantime

Issue #4338, opened 2025-09-10
[Device Support Request] Add quirk for Sonoff ZBMINIL2


rvveber:
### Problem description

It should be the same as ZBMINIR2 - which already got added.
https://github.com/zigpy/zha-device-handlers/pull/3428

Only difference is that it doesn't act as router (has no neutral requirement) and does not allow detached mode! 
It is not possible to switch this device without also switching the relais!
(For anyone coming across this post.. i sent them back for this reason)

### Diagnostics information
I use conbee3 as coordinator.

```
{
  "home_assistant": {
    "installation_type": "Home Assistant OS",
    "version": "2025.9.1",
    "dev": false,
    "hassio": true,
    "virtualenv": false,
    "python_version": "3.13.7",
    "docker": true,
    "arch": "x86_64",
    "timezone": "Europe/Berlin",
    "os_name": "Linux",
    "os_version": "6.12.43-haos",
    "container_arch": "amd64",
    "supervisor": "2025.09.0",
    "host_os": "Home Assistant OS 16.2",
    "docker_version": "28.3.3",
    "chassis": "vm",
    "run_as_root": true
  },
  "custom_components": {
    "localtuya": {
      "documentation": "https://github.com/rospogrigio/localtuya/",
      "version": "5.2.3",
      "requirements": []
    },
    "spotcast": {
      "documentation": "https://github.com/fondberg/spotcast",
      "version": "v6.0.0-a15",
      "requirements": [
        "spotipy>=2.23.0",
        "spotifyaio>=0.8.8",
        "RapidFuzz>=3.10.1"
      ]
    },
    "hacs": {
      "documentation": "https://hacs.xyz/docs/use/",
      "version": "2.0.5",
      "requirements": [
        "aiogithubapi>=22.10.1"
      ]
    },
    "sleep_as_android": {
      "documentation": "https://github.com/IATkachenko/HA-SleepAsAndroid/wiki",
      "version": "2.3.2",
      "requirements": [
        "pyhaversion"
      ]
    }
  },
  "integration_manifest": {
    "domain": "zha",
    "name": "Zigbee Home Automation",
    "after_dependencies": [
      "hassio",
      "onboarding",
      "usb"
    ],
    "codeowners": [
      "dmulcahey",
      "adminiuga",
      "puddly",
      "TheJulianJES"
    ],
    "config_flow": true,
    "dependencies": [
      "file_upload",
      "homeassistant_hardware"
    ],
[...]

PR #5349, opened 2026-09-17
Add `KeepAlive` cluster to ZBMINIL2

puddly:
## Proposed change

The ZBMINIL2 reads attributes from the coordinator's `KeepAlive` cluster but does not itself advertise supporting this cluster. https://github.com/zigpy/zigpy/pull/1888 is one solution to this problem but the more immediate one is just patching the device with a quirk.

## Additional information

Might address https://github.com/home-assistant/core/issues/181626

## Checklist

Related: #4338


puddly:
@1775939187 I think this is a bug with the ZBMINIL2 firmware. It's using `KeepAlive` without advertising support for this cluster.

Wackymax:
The device-specific cluster addition in this PR matches an isolated zigpy 2.2.0 replay: an incoming Keep-Alive Read Attributes request is dropped with the original ZBMINIL2 endpoint descriptor, and adding the missing `0x0025` client cluster allows the coordinator to form the matching response.

There is a separate APS-layer detail for this fix: zigpy 2.2.0 does not set `TransmitOptions.APS_Encryption` for Keep-Alive responses, while [ZCL revision 8 §3.18.4](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf) requires the response to be APS-encrypted. I noted the proposed narrow change on [zigpy/zigpy#1888](https://github.com/zigpy/zigpy/pull/1888#issuecomment-5817075719). The quirk resolves dispatch; the Zigpy change would ensure the response is sent with the required security flag.

